PR #4, opened 2023-08-30
features/multi-denom-deposits

baktun14:
- Added support for USDC payments for deployments
- Added caching for market data
- Update akashjs@0.4.11


api/src/providers/apiNodeProvider.ts
```diff
@@ -339,7 +339,9 @@ export async function getDeployment(owner: string, dseq: string) {
     include: [{ model: ProviderAttribute }]
   });
 
-  const aktPrice = getAktMarketData()?.price;
+  const marketData = await getMarketData();
```
Redm4x, line 342: Isn't this doing a query to coingecko every time?

deploy-web/src/components/newDeploymentWizard/PrerequisiteList.tsx
```diff
@@ -86,7 +86,7 @@ export const PrerequisiteList: React.FunctionComponent<Props> = ({ onClose, onCo
               </ListItemIcon>
               <ListItemText
                 primary="Wallet Balance"
-                secondary="The balance of the wallet needs to be of at least 5 AKT. If you do not have 5 AKT, you will need to specify an authorized depositor."
+                secondary="The balance of the wallet needs to be of at least 5 AKT or USDC. If you do not have 5 AKT, you will need to specify an authorized depositor."
```
Redm4x, line 89:
```suggestion
                secondary="The balance of the wallet needs to be of at least 5 AKT or USDC. If you do not have 5 AKT or USDC, you will need to specify an authorized depositor."
```

api/src/providers/apiNodeProvider.ts
baktun14: Yes indeed I'll add caching

PR #33, opened 2023-10-16
feat: Connect to Leap Wallet support

hsnice16:
Hello, team!
I am Himanshu from Leap's dev team.

I have raised this PR to add support for Connect Wallet with Leap Browser Extension.
Please review this PR, check whether everything works fine, and let me know if any other changes are required.


deploy-web/src/types/gloabl.d.ts
baktun14, line 1: We have a file called `global.ts` in the same folder with the `Window` interface, I suggest they be merged. Also, the file name has a typo :).

deploy-web/src/components/layout/WelcomeModal.tsx
```diff
@@ -1,5 +1,5 @@
 import { Button, Dialog, DialogContent, DialogActions, Typography, Box, Paper, useTheme } from "@mui/material";
-import { useKeplr } from "@src/context/KeplrWalletProvider";
+import { useKeplr } from "@src/context/WalletProvider";
```
Redm4x, line 2:
```suggestion
```
Should be `useWallet`, but it's not used anymore so could just remove the import

deploy-web/src/context/WalletProvider/WalletProvider.tsx
```diff
@@ -85,29 +93,43 @@ export const KeplrWalletProvider = ({ children }) => {
...
+        window.addEventListener("leap_keystorechange", () => onKeystoreChange(Wallets.LEAP));
 
         return () => {
           isMounted.current = false;
 
-          console.log("useKeplr on unmount");
-
-          window.removeEventListener("keplr_keystorechange", onKeystoreChange);
+          window.removeEventListener("keplr_keystorechange", () => onKeystoreChange(Wallets.KEPLR));
+          window.removeEventListener("leap_keystorechange", () => onKeystoreChange(Wallets.LEAP));
```
baktun14, line 132: remove/add listeners must use the same reference to properly work, so we could just set the functions per wallet in respective variables just before adding the listeners.

baktun14, line 132:
```suggestion
          window.removeEventListener("keplr_keystorechange", keplrOnKeystoreChange);
          window.removeEventListener("leap_keystorechange", leapOnKeystoreChange);
```

hsnice16: Have made the changes.

deploy-web/src/context/WalletProvider/WalletProvider.tsx
```diff
@@ -85,29 +93,43 @@ export const KeplrWalletProvider = ({ children }) => {
...
+          };
+
           loadWallet();
         } else {
           setIsWalletLoaded(true);
         }
 
-        window.addEventListener("keplr_keystorechange", onKeystoreChange);
+        window.addEventListener("keplr_keystorechange", () => onKeystoreChange(Wallets.KEPLR));
+        window.addEventListener("leap_keystorechange", () => onKeystoreChange(Wallets.LEAP));
```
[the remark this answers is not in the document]
hsnice16: Though, I made the changes, but, Was it not working properly? because it should work, as we can use arrow functions in event listeners. And, since, these are two different events then there should be no issue in removing listeners as well.

deploy-web/src/components/layout/WelcomeModal.tsx
hsnice16: I have made the changes.

deploy-web/src/types/gloabl.d.ts
hsnice16: I've updated the `global.ts` file only.

Redm4x:
Thank you @hsnice16 for the contribution! Will get that in the next release 🚀

deploy-web/src/context/WalletProvider/WalletProvider.tsx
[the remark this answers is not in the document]
Redm4x: The events were added correctly, but the `removeEventListener` was not removing them since they were not the same arrow functions as the ones added. This explains the issue: https://www.semicolonandsons.com/code_diary/javascript/anonymous-functions-do-not-work-with-remove-event-listener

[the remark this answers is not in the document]
hsnice16: Oh! Nice. Makes sense, we have not given the same function reference.
Thanks for this.

PR #86, opened 2024-01-23
Refactor & improve public API

Redm4x:
Improve public api
- Added swagger UI documentation
- Standardized endpoint urls (ex: `/getMainnetVersion => /version/mainnet`)
- Switched to [honojs](https://hono.dev/) instead of express for a better dev experience (better typing + facilitate openapi definition)
- Added versioning to url and file structure to enable backward compatible upgrades (`/v1/endpoint`)
- Added eslint and improved typing across the board


api/src/caching/helpers.ts
```diff
@@ -3,16 +3,16 @@ import MemoryCacheEngine from "./memoryCacheEngine";
...
 
 const cacheEngine = new MemoryCacheEngine();
-const pendingRequests: { [key: string]: Promise<any> } = {};
+const pendingRequests: { [key: string]: Promise<unknown> } = {};
 
-interface CachedObject {
+interface CachedObject<T> {
   date: Date;
-  data: any;
+  data: T;
```
baktun14, line 10: good idea

baktun14:
Rename providers into services?

api/src/routes/v1/deployments/byOwnerDseq.ts
```diff
@@ -0,0 +1,91 @@
...
+const route = createRoute({
+  method: "get",
+  path: "/deployment/{owner}/{dseq}",
+  summary: "Get deployment details",
+  tags: ["Deployments"],
+  request: {
+    params: z.object({
+      owner: z.string().openapi({
+        description: "Owner's Address",
+        example: "akash13265twfqejnma6cc93rw5dxk4cldyz2zyy8cdm"
```
baktun14: constant?

api/src/utils/pricing.ts
```diff
@@ -1,13 +1,12 @@
...
 
-
 export function getAkashPricing(cpu: number, memory: number, storage: number) {
   // Pricing obtained from https://github.com/akash-network/helm-charts/blob/main/charts/akash-provider/scripts/price_script_generic.sh
   const targetMemory = 1.25; // USD/GB-month
   const targetHDEphemeral = 0.08; // USD/GB-month
-  const targetHDPersHDD = 0.1; // USD/GB-month
-  const targetHDPersSSD = 0.12; // USD/GB-month
-  const targetHDPersNVME = 0.14; // USD/GB-month
+  // const targetHDPersHDD = 0.1; // USD/GB-month
```
baktun14, line 7: It wasn't used?

deploy-web/src/utils/apiUtils.ts
```diff
@@ -58,10 +58,10 @@ export class ApiUrlService {
     return `${apiEndpoint}/cosmos/feegrant/v1beta1/allowances/${address}`;
   }
   static dashboardData() {
-    return `${BASE_API_URL}/dashboardData`;
+    return `${BASE_API_URL}/dashboard-data`;
```
baktun14, line 61: Should we version apiUtils as well? I see that you added the version prefix in the BASE_API_URL variable but it applies to all the routes. If we add or change routes between versions, it might break these routes or I guess we can change the apiUtils to the new versions of the routes.

baktun14:
Does this new version of the API support the old routes? We should for a grace period of time because some clients use the API actively in their app and will break if we release without support.

api/src/utils/pricing.ts
Redm4x: We only compare ephemeral storage pricing, no persistent storage for now

Redm4x:
> Does this new version of the API support the old routes? We should for a grace period of time because some clients use the API actively in their app and will break if we release without support.

deploy-web/src/utils/apiUtils.ts
Redm4x: We should not end up calling different versions of the api in the deploy tool. If we publish a new version of the api, we would update `apiUtils.ts` to use the latest version everywhere.

Redm4x:
> Does this new version of the API support the old routes? We should for a grace period of time because some clients use the API actively in their app and will break if we release without support.

Good point, I added `legacyRouter.ts` with redirects so that it doesn't break existing integrations. We should still contact those clients and point them to the new documentation so that they can switch to using `v1`

baktun14 (approved):

PR #95, opened 2024-02-09
Features/implement new status endpoint

Redm4x:
- Support for the new provider's status endpoint (grpc)
  - Keep track of nodes and cpu/gpu info (vendor/model)
  - Fallback to old tracking if provider isn't updated yet (based on the `/version` endpoint)
- Improved performance for provider graphs


indexer/src/providers/statusEndpointHandlers/grpc.ts
```diff
@@ -0,0 +1,332 @@
...
+    } catch (err) {
+      reject(err);
+    }
+  });
+}
+
+const protosetBuffer = fs.readFileSync("./src/proto/akash/providerServiceDescriptor.bin");
+const descriptorSet = loadFileDescriptorSetFromBuffer(protosetBuffer);
+const packageDef = grpc.loadPackageDefinition(descriptorSet);
+const clientInsecureCreds = grpc.credentials.createInsecure();
```
baktun14, line 14: Move those constants at the top of the file?

indexer/src/providers/statusEndpointHandlers/grpc.ts
```diff
@@ -0,0 +1,332 @@
...
+      {
+        cpu: 0,
+        gpu: 0,
+        memory: 0,
+        storage: 0
+      }
+    );
+  const checkDate = toUTC(new Date());
+
+  console.time("updateData");
```
baktun14: Is it fast?

indexer/src/providers/statusEndpointHandlers/grpc.ts
```diff
@@ -0,0 +1,332 @@
...
+      {
+        cpu: 0,
+        gpu: 0,
+        memory: 0,
+        storage: 0
+      }
+    );
+}
+
+type NewStatusResponseType = {
```
baktun14: I don't think we have one, but we should have a folder with all the types.

Redm4x:
Added a flag to disable the new status endpoint for now, will deploy like this to get performance improvements asap and will enable grpc later on.

indexer/src/providers/statusEndpointHandlers/grpc.ts
Redm4x: Moved the type to a new `types` folder, with a TODO to remove it once we have grpc client generation

baktun14 (approved):

PR #112, opened 2024-02-27
Fix provider status tracking bugs

Redm4x:
- Add timeout to grpc status calls (10s, same as rest calls)
- Cap available resources to 0 when `allocated` is larger than `allocatable` to prevent negative numbers. This can happen due to [provider overcommit](https://github.com/akash-network/docs/blob/master/operator/provider/README.md#cluster-resources-overcommit).
- Fix a discrepancy between provider capacity tiles and their corresponding graphs. This was caused by an incorrect handling of provider snapshots when status endpoint is not accessible.
- Fetch status of recently online providers first (so that they come back online sooner after a temporary outage)

### DIFF
--- a/indexer/src/providers/providerStatusProvider.ts
+++ b/indexer/src/providers/providerStatusProvider.ts
@@ -4,12 +4,12 @@ import semver from "semver";
 import { Provider, ProviderSnapshotNode, ProviderSnapshotNodeCPU, ProviderSnapshotNodeGPU } from "@shared/dbSchemas/akash";
 import { asyncify, eachLimit } from "async";
 import { ProviderSnapshot } from "@src/../../shared/dbSchemas/akash/providerSnapshot";
-import { fetchAndSaveProviderStats as grpcFetchAndSaveProviderStats } from "./statusEndpointHandlers/grpc";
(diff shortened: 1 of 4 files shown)


indexer/src/providers/providerStatusProvider.ts
```diff
@@ -28,103 +32,160 @@ export async function syncProvidersInfo() {
...
+        );
+
+        await ProviderSnapshotNodeGPU.bulkCreate(
+          node.gpus.map((gpuInfo) => ({
+            snapshotNodeId: providerSnapshotNode.id,
+            vendor: gpuInfo.vendor,
+            name: gpuInfo.name,
+            modelId: gpuInfo.modelId,
+            interface: gpuInfo.interface,
+            memorySize: gpuInfo.memorySize // TODO: Change type to bytes?
```
baktun14: is it supposed to be bytes?

baktun14 (approved):

indexer/src/providers/providerStatusProvider.ts
Redm4x: It's fine to keep it as a string I think. Only use for a numerical type would be to query GPUs with X vram "or more", but for now it's simpler like this. I removed the todo, we can always change the type in the futur if needed.

PR #117, opened 2024-03-03
bugfixes/fix-stats-deployments

baktun14:
- Fix address deployments

### DIFF
--- a/stats-web/package-lock.json
+++ b/stats-web/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "stats-web",
-  "version": "0.15.0",
+  "version": "0.19.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "stats-web",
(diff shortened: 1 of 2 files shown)


Redm4x (approved):

tgrecojs:
is there a staging url where the changes from [collab-rebrand](https://github.com/akash-network/cloudmos/tree/collab-rebrand) can be viewed? i'm really interested in checking out the fix #117 but it seems as though there are a handful of fixes (all within the collab-rebrand branch) that are intended to go live together at some time in the future. am i right to assume this? 

just my 2 cents here as an outside contributor - i haven't inspected the other changes from that branch but i think there's a case to be made for the deployments fix to get pushed out the door asap. granted that the fix has been confirmed, i don't see the benefit in having the live UI render broken features for any longer than necessary 🤷

Redm4x:
@tgrecojs This bugfix is now live in v0.19.1 of the stats app

Issue #93, opened 2024-01-30
GPU model display in deployment detail (Lease)


baktun14:
- With the indexer, we need to track the bid that was selected for the lease and extract from the bid resource the gpu model and attributes
- Implement a new api endpoint to retrieve the data
- Display the GPU model in the deployment detail

brewsterdrinkwater:
March 12th, 2024

- This is going out this week.

Redm4x:
This was deployed last week as planned. The gpu model is now visible in the deployment detail page.

![Image](https://github.com/akash-network/cloudmos/assets/2829180/ff1bd43e-9355-4d68-bdef-cd069061d2ae)

PR #127, opened 2024-03-13
Show gpu model in deployment detail page

Redm4x:
- Fetch bid info on deployment detail page and show gpu models (#93)
image
- Fix Kube version in provider page on refresh


deploy-web/src/pages/providers/[owner]/index.tsx
```diff
@@ -60,7 +58,7 @@ const ProviderDetailPage: React.FunctionComponent<Props> = ({ owner, _provider }
     isLoading: isLoadingStatus,
     refetch: getProviderStatus
   } = useProviderStatus(provider?.hostUri, {
-    enabled: false,
+    enabled: true,
```
Comment (baktun14) [inline deploy-web/src/pages/providers/[owner]/index.tsx:61]: It was not working properly? I don't remember exactly why I made explicity query only once before. It just made me think, at some point the `/status` rest endpoint won't exist anymore, so we'll probably have to migrate to the grpc in the frontend as well.

baktun14 (approved):

deploy-web/src/pages/providers/[owner]/index.tsx
Comment (Redm4x) [reply deploy-web/src/pages/providers/[owner]/index.tsx:61]: It was working when coming to the provider page from another page, but it was not loading when opening/refreshing directly on the provider page. I think you may have put it to false so that it doesnt query the endpoint during server rendering, didn't look too far into it

Comment (baktun14) [reply deploy-web/src/pages/providers/[owner]/index.tsx:61]: I don't think it was for ssr, but more to not spam the provider endpoint since the status doesn't really change.

Comment (Redm4x) [reply deploy-web/src/pages/providers/[owner]/index.tsx:61]: Got it, yeah for sure we could change react-query options to keep the cache longer for `/version`. I'll let it like this for now, but we will need to refactor the provider pages after the rebrand. This merging of 3 sources into the same provider object is causing weird issues. Also just realized a similar problem happens with the "Your active leases" count.

PR #129, opened 2024-03-19
Improve api typing

Redm4x:
- Add missing types in Api project
- Enable `noImplicitAny` in tsconfig
- Remove unused proposal code (deploy-web)


api/src/db/dbConnection.ts
```diff
@@ -6,7 +6,7 @@ import { chainModels, getChainModels, userModels } from "@shared/dbSchemas";
 import { Template, TemplateFavorite, UserAddressName, UserSetting } from "@shared/dbSchemas/user";
 import { chainDefinitions } from "@shared/chainDefinitions";
 
-const csMap = {
+const csMap: { [key: string]: string } = {
```
ygrishajev: is it needed to be this flexible? if you skip this type it's implicitly typed as object with specified keys and values. It seems that what really needed here is typed env

api/src/routes/v1/graphData.ts
```diff
@@ -1,29 +1,13 @@
...
+import { AuthorizedGraphDataNames, getGraphData, isValidGraphDataName } from "@src/services/db/statsService";
 
 const route = createRoute({
   method: "get",
   path: "/graph-data/{dataName}",
   tags: ["Analytics"],
   request: {
     params: z.object({
-      dataName: z.string().openapi({ example: "dailyUAktSpent", enum: authorizedDataNames })
+      dataName: z.string().openapi({ example: "dailyUAktSpent", enum: Array.from(AuthorizedGraphDataNames) })
```
ygrishajev: `Array.from` is already a hack here needed to overcome `as const` cast. To me it would be the same in terms of code being clean to do:
```typescript
AuthorizedGraphDataNames as unknown as string[]
```

However I'd just created a dedicated type in the source instead of deriving type from an array. Yes, some duplication but ts would enforce to follow and it works consistently without hacks

```typescript
type AuthorizedGraphDataName = (
  "dailyUAktSpent" |
  "dailyUUsdcSpent" |
  "dailyUUsdSpent" |
  "dailyLeaseCount" |
  "totalUAktSpent" |
  "totalUUsdcSpent" |
  "totalUUsdSpent" |
  "activeLeaseCount" |
  "totalLeaseCount" |
  "activeCPU" |
  "activeGPU" |
  "activeMemory" |
  "activeStorage"
  )

export const authorizedGraphDataNames: AuthorizedGraphDataName[] = [
  "dailyUAktSpent",
  "dailyUUsdcSpent",
  "dailyUUsdSpent",
  "dailyLeaseCount",
  "totalUAktSpent",
  "totalUUsdcSpent",
  "totalUUsdSpent",
  "activeLeaseCount",
  "totalLeaseCount",
  "activeCPU",
  "activeGPU",
  "activeMemory",
  "activeStorage"
];
```

api/src/services/external/apiNodeService.ts
```diff
@@ -272,18 +279,18 @@ export async function getProposal(id: number) {
...
-      // const proposerData = await proposerResponse.json();
-      // const validatorFromDb = await Validator.findOne({ where: { accountAddress: proposerData.result.proposer } });
-      // proposer = {
-      //   address: proposer,
-      //   moniker: validatorFromDb?.moniker,
-      //   operatorAddress: validatorFromDb?.operatorAddress,
-      //   avatarUrl: validatorFromDb?.keybaseAvatarUrl
-      // };
-    }
+    // const proposer = null; // TODO: Fix
```
ygrishajev, line 282: is there a reason we can't remove this code? I know it's here only because of prettier so just asking

ygrishajev (approved):

api/src/services/external/apiNodeService.ts
Redm4x: We have kept it in hope of fixing it one day. There is an issue with proposals created during genesis. I created #132 to keep track of it.

api/src/db/dbConnection.ts
Redm4x: I removed the type and added a typeguard instead. But yeah ideally we would have env var typing & validation.

api/src/services/db/statsService.ts
```diff
@@ -75,7 +75,22 @@ export const getDashboardData = async () => {
   };
 };
 
-export const AuthorizedGraphDataNames = [
+type AuthorizedGraphDataName =
```
ygrishajev, line 78: ❤️

baktun14 (approved):

PR #134, opened 2024-03-21
Features/handle storage class validation

Redm4x:
- Update akashjs and remove temporary patch
- Call akashjs validation method and show errors as CustomValidationError


deploy-web/src/utils/deploymentData/helpers.ts
```diff
@@ -57,12 +57,12 @@ export function parseSizeStr(str: string) {
 
 type NetworkType = "beta2" | "beta3";
 
-function isString(value: unknown): value is string {
-  return typeof value === "object" && value !== null && value.constructor === String;
+function isValidString(value: unknown): value is string {
```
baktun14, line 60: We could replace with [lodash isString](https://www.geeksforgeeks.org/lodash-_-isstring-method/)

PR #137, opened 2024-03-28
Features/rent gpu query

baktun14:
- Update Akashjs to 0.5.11
- Added support for multiple GPU model with ram + interface in SDL Builder
- Added new API endpoint that fetches the possible GPU models from the [provider config file](https://raw.githubusercontent.com/akash-network/provider-configs/main/devices/pcie/gpus.json)
- Added query param logic to pre-populate a specific model + vram + interface in the rent gpu form
  - Example query: `?vendor=nvidia&gpu=h100&vram=80Gi&interface=sxm`

image


deploy-web/src/components/sdl/RentGpusForm.tsx
```diff
@@ -80,6 +81,34 @@ export const RentGpusForm: React.FunctionComponent<Props> = ({}) => {
...
   }, []);
 
+  useEffect(() => {
+    if (router.query.vendor && router.query.gpu && gpuModels && !isQueryInit) {
+      // Example query: ?vendor=nvidia&gpu=h100&vram=80Gi&interface=sxm
+      const vendorQuery = router.query.vendor as string;
+      const gpuQuery = router.query.gpu as string;
+      const gpuModel = gpuModels.find(x => x.name === vendorQuery)?.models.find(x => x.name === gpuQuery);
+
+      console.log("GPU model", gpuModel);
```
baktun14: :scissors:

api/src/routers/internalRouter.ts
```diff
@@ -232,3 +236,49 @@ internalRouter.get("leases-duration/:owner", async (c) => {
...
+      models: []
+    };
+
+    // Loop over models
+    for (const [modelKey, modelValue] of Object.entries(vendorValue.devices)) {
+      const _modelValue = modelValue as {
+        name: string;
+        memory_size: string;
+        interface: string;
+      };
```
Redm4x, line 261: Can we type the entire response instead of using `object` and casting here? With a type like this:
```
type ProviderConfigGpusType = {
  [key: string]: {
    name: string;
    devices: {
      [key: string]: {
        name: string;
        memory_size: string;
        interface: string;
      };
    };
  };
};

```

deploy-web/src/utils/sdl/sdlGenerator.ts
```diff
@@ -188,6 +213,8 @@ export const generateSdl = (services: Service[], region?: string) => {
     }
   });
 
+  console.log(result);
```
baktun14: :scissors:

Redm4x: ✂️ ?

api/src/routers/internalRouter.ts
baktun14: Good idea yea

Redm4x (approved):

PR #143, opened 2024-04-01
Bugfixes/fix provider gpus UI

Redm4x:
## Display gpu models from feature discovery instead of capabilities attributes
- Return gpu models from feature discovery in providers endpoints
- Use those model instead of capabilities in provider list/detail pages
- Added missing indexes to improve performance

```
CREATE INDEX IF NOT EXISTS provider_snapshot_node_snapshot_id
    ON public."providerSnapshotNode" USING btree
    ("snapshotId" ASC NULLS LAST)
    WITH (deduplicate_items=True)
    TABLESPACE pg_default;
	
CREATE INDEX IF NOT EXISTS provider_snapshot_node_cpu_snapshot_node_id
    ON public."providerSnapshotNodeCPU" USING btree
    ("snapshotNodeId" ASC NULLS LAST)
    WITH (deduplicate_items=True)
    TABLESPACE pg_default;
	
CREATE INDEX IF NOT EXISTS provider_snapshot_node_gpu_snapshot_node_id
    ON public."providerSnapshotNodeGPU" USING btree
    ("snapshotNodeId" ASC NULLS LAST)
    WITH (deduplicate_items=True)
    TABLESPACE pg_default;
```

### DIFF
--- a/api/src/services/db/providerStatusService.ts
+++ b/api/src/services/db/providerStatusService.ts
@@ -42,8 +42,7 @@ export async function getNetworkCapacity() {
 }
 
 export const getProviderList = async () => {
-  // Fetch provider list with their attributes & auditors
-  const providers = await Provider.findAll({
+  const providersWithAttributesAndAuditors = await Provider.findAll({
     where: {
       deletedHeight: null
     },
@@ -58,8 +57,7 @@ export const getProviderList = async () => {
     ]
   });
 
-  // Fetch latest snapshot nodes for each provider
-  const providerNodes = await Provider.findAll({
+  const providerWithNodes = await Provider.findAll({
     attributes: ["owner"],
     where: {
       deletedHeight: null
@@ -82,14 +80,14 @@ export const getProviderList = async () => {
     ]
   });
 
-  const filteredProviders = providers.filter((value, index, self) => self.map((x) => x.hostUri).lastIndexOf(value.hostUri) === index);
+  const distinctProviders = providersWithAttributesAndAuditors.filter((value, index, self) => self.map((x) => x.hostUri).lastIndexOf(value.hostUri) === index);
   const providerAttributeSchemaQuery = getProviderAttributesSchema();
   const auditorsQuery = getAuditors();
 
   const [auditors, providerAttributeSchema] = await Promise.all([auditorsQuery, providerAttributeSchemaQuery]);
 
-  return filteredProviders.map((x) => {
-    const nodes = providerNodes.find((p) => p.owner === x.owner)?.lastSnapshot?.nodes;
+  return distinctProviders.map((x) => {
+    const nodes = providerWithNodes.find((p) => p.owner === x.owner)?.lastSnapshot?.nodes;
     return mapProviderToList(x, providerAttributeSchema, auditors, nodes);
   });
 };
--- a/api/src/services/db/providerStatusService.ts
+++ b/api/src/services/db/providerStatusService.ts
@@ -88,9 +88,10 @@ export const getProviderList = async () => {
 
   const [auditors, providerAttributeSchema] = await Promise.all([auditorsQuery, providerAttributeSchemaQuery]);
 
-  return filteredProviders.map((x) =>
-    mapProviderToList(x, providerAttributeSchema, auditors, providerNodes.find((p) => p.owner === x.owner)?.lastSnapshot?.nodes)
-  );
+  return filteredProviders.map((x) => {
+    const nodes = providerNodes.find((p) => p.owner === x.owner)?.lastSnapshot?.nodes;
+    return mapProviderToList(x, providerAttributeSchema, auditors, nodes);
+  });
 };
 
 export const getProviderDetail = async (address: string): Promise<ProviderDetail> => {
@@ -122,7 +123,7 @@ export const getProviderDetail = async (address: string): Promise<ProviderDetail
     }
   });
 
-  const lastOnlineSnapshot = await ProviderSnapshot.findOne({
+  const lastSnapshot = await ProviderSnapshot.findOne({
     where: {
       id: provider.lastSnapshotId
     },
@@ -141,7 +142,7 @@ export const getProviderDetail = async (address: string): Promise<ProviderDetail
   const [auditors, providerAttributeSchema] = await Promise.all([auditorsQuery, providerAttributeSchemaQuery]);
 
   return {
-    ...mapProviderToList(provider, providerAttributeSchema, auditors, lastOnlineSnapshot?.nodes),
+    ...mapProviderToList(provider, providerAttributeSchema, auditors, lastSnapshot?.nodes),
     uptime: uptimeSnapshots.map((ps) => ({
       id: ps.id,
       isOnline: ps.isOnline,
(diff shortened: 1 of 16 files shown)


api/src/services/db/providerStatusService.ts
```diff
@@ -42,6 +42,7 @@ export async function getNetworkCapacity() {
 }
 
 export const getProviderList = async () => {
+  // Fetch provider list with their attributes & auditors
```
ygrishajev: imho this kind of comments don't help much. It's already obvious that `providers` is a providers list. If it's essential to emphasise it's got attributes and auditors (is it?) it can be named accordingly. e.g. `providersWithAttrsAndAuditors`. It would also help understanding code better further if it's important.

api/src/utils/map/provider.ts
```diff
@@ -1,10 +1,19 @@
...
+  providerAttributeSchema: ProviderAttributesSchema,
+  auditors: Array<Auditor>,
+  nodes?: ProviderSnapshotNode[]
+): ProviderList => {
   const isValidVersion = provider.cosmosSdkVersion ? semver.gte(provider.cosmosSdkVersion, "v0.45.9") : false;
   const name = provider.isOnline ? new URL(provider.hostUri).hostname : null;
+  const gpuModels = (nodes || [])
+    .flatMap((x) => x.gpus)
+    .map((x) => ({ vendor: x.vendor, model: x.name, ram: x.memorySize, interface: x.interface }))
+    .filter((x, i, arr) => arr.findIndex((o) => x.vendor === o.vendor && x.model === o.model && x.ram === o.ram && x.interface === o.interface) === i);
```
ygrishajev: this one is hard to understand, ideally I'd imagine this should be extracted into a named function with some vars that would better explain the conditions.

api/src/utils/map/provider.ts
```diff
@@ -83,6 +90,14 @@ export const mapProviderToList = (provider: Provider, providerAttributeSchema: P
   } as ProviderList;
 };
 
+function getDistinctGpuModelsFromNodes(nodes: ProviderSnapshotNode[]) {
+  const gpuModels = nodes.flatMap((x) => x.gpus).map((x) => ({ vendor: x.vendor, model: x.name, ram: x.memorySize, interface: x.interface }));
+  const distinctGpuModels = gpuModels.filter(
+    (x, i, arr) => arr.findIndex((o) => x.vendor === o.vendor && x.model === o.model && x.ram === o.ram && x.interface === o.interface) === i
```
ygrishajev: Could you pls explain what this means?
I understand that `findIndex` matching same vendor, model, etc. Why should it also be of the same index?

ygrishajev: ok, I see, it's to remove duplicates

ygrishajev: It's really hard to understand this. I'd suggest if we could add a dedicated function for this.

```typescript
type Matcher<T> = (a: T, b: T) => boolean;

function createFilterUnique<T>(matcher: Matcher<T> = (a, b) => a === b): (value: T, index: number, array: T[]) => boolean {
  return (value, index, array) => {
    return array.findIndex((other) => matcher(value, other)) === index;
  };
}
```

in this case the code above would be:

```typescript
const distinctGpuModels = gpuModels.filter(
    createFilterUnique(x, o) => x.vendor === o.vendor && x.model === o.model && x.ram === o.ram && x.interface === o.interface))
);
```

and example [below](https://github.com/akash-network/cloudmos/pull/143/files#diff-11422e0f72849b5cbe75e9e428d718b2798c564d11f7e80f91f72f03a27030dcR29) would be

```typescript
.filter(createFilterUnique())
```

ygrishajev: can be tested [here](https://www.typescriptlang.org/play?#code/C4TwDgpgBAsghsAxgCwgJwDwBUB8UC8UAFHAFxRYA0UARuVgJQF40D2rANhHAHYDcAKAEAzAK49EwAJaseURGm7AIAMSkdlaAKo8pAR1ERsOIgFsEKdOXhJUmXAWJxqNJvjxwC+Qq-JEAbnAchvTUUjwAJhAAHuQ8oqY06NRwaGhwIPQA2gC6bizsXLxQAN4CUFCKwKJocgFBhmGRMSlpGfml5RWVENW1UKnpIAB0wuERAJLN0URErMB2Hea26PXBENTzi27eUOMxghUAvoJHQoiyAM7AUMLsjlkA5HCP1I80r1CPiJ-vvy85UbqTREBRKVTA9A6fSGIgMBjnK6cCDDDisADmRDurARAguPGut3YAEYHiUiaxyO9Uo8oEdqOTsVSaDS6QyKcy4AAvWn00ocr4stDE3mAsYaVZghAQiXaXQGCCzZy0DpwUb3bw+dU43H4y7I1EYrEkhhAA)

Redm4x: It's to keep only the first occurence of each gpus since there are duplicates in the array. If the current index `i` matches `findIndex` that means it's the first occurence so it's kept, otherwise it is filtered out as duplicate.

ygrishajev: yea, I figured after asking the question :) check out a suggestion pls

Redm4x: mb your other replies weren't showing when I answered. For sure that would be easier to understand, I'll add it! 👍

Redm4x: I added the `createFilterUnique` helper with relevant unit tests

deploy-web/src/components/providers/ProviderListRow.tsx
```diff
@@ -62,7 +62,7 @@ export const ProviderListRow: React.FunctionComponent<Props> = ({ provider }) =>
   const _totalStorage = provider.isOnline
     ? bytesToShrink(provider.availableStats.storage + provider.pendingStats.storage + provider.activeStats.storage)
     : null;
-  const gpuModels = provider.hardwareGpuModels.map(gpu => gpu.substring(gpu.lastIndexOf(" ") + 1, gpu.length));
+  const gpuModels = provider.gpuModels.map(x => x.model).filter((x, i, arr) => arr.indexOf(x) === i);
```
ygrishajev: This could also utilise that new filter function. What do you think?
```suggestion
  const gpuModels = provider.gpuModels.map(x => x.model).filter(createFilterUnique());
```

deploy-web/src/components/providers/ProviderSpecs.tsx
```diff
@@ -20,12 +18,16 @@ const useStyles = makeStyles()(theme => ({
...
 
-export const ProviderSpecs: React.FunctionComponent<Props> = ({ provider, providerAttributesSchema }) => {
+export const ProviderSpecs: React.FunctionComponent<Props> = ({ provider }) => {
   const { classes } = useStyles();
-  const router = useRouter();
+
+  const gpuModels =
+    provider?.gpuModels
+      ?.map(x => x.model + " " + x.ram)
+      .filter((x, i, arr) => arr.indexOf(x) === i)
```
ygrishajev:
```suggestion
      .filter(createFilterUnique())
```

deploy-web/src/components/providers/ProviderListRow.tsx
Redm4x: Yeah, I added it to the deploy-web project too. No unit tests for now, but we could add them once the big refactor is done (rebrand).

ygrishajev (approved):

Issue #152, opened 2024-04-04
Improve uptime checks logic


Redm4x:
## Improve uptime checks logic when there is an outage
When there is a temporary outage, the provider is immediatly flagged as offline and disappears from stats pages (capacity/gpu pricing) for at least 15min until next check.

We could implement a exponential backoff algorithm like this:
```
- Continue with the 15-minute checks as you are currently doing. Don't reduce.
- If a check fails, do not immediately conclude that the provider/GPU is unavailable. Instead, initiate a retry with exponential backoff, capped at a 32-minute window. This process would involve re-attempting the query according to the following sequence:
- If the initial 15-minute check fails for a specific single provider at a time, then:
    - Retry after 2 minutes. If this attempt fails, retry in 4 minutes; if that fails, retry in 8 minutes; following a failure, retry in 16 minutes. If all attempts fail, perform a final retry.
```

Also need to investigate if we can query the ingress subdomain as a new metric and maybe show that in UI as the uptime %.

Potential implementation details:
- Keep a `nextCheckDate` and `failedCheckCount` on the providers table and update them after every check.
- Change the uptime check task to run more often, but look at `nextCheckDate` instead of checking every providers. That would also have the benefit of better distributing provider checks. We will also stop checking
- Keep providers included in the stats as long as the final retry is not reached to prevent rapid fluctuation for short outages.
- After a specific retry count (ex: 32min) we can flag provider as offline and remove from stats, but we should still continue checking from time to time in case it comes back online (1h+ interval).
- For provider pages and bids, we could keep the provider visible, but add an indicator for a potential outage.
- Need to adjust uptime bars in provider page as well as uptime percentages (30d,7d,1d). We can no longer assume that every uptime checks are spaced 15 minutes apart.
- Add a `lastSuccessulSnapshotId` on the provider table to simplify stats queries.

Redm4x:
Feedback from providers about the current system: https://discord.com/channels/747885925232672829/1111749073322660034/1227151601186377798

PR #167, opened 2024-04-23
feat: Add grace period to uptime check

Redm4x:
## Improvements to the provider uptime tracking system #152 

- Added a grace period before treating a provider as completely offline. This will reflect in the following places:
  - `/internal/gpu-prices` endpoint used by the GPU Pricing page
  - `/internal/gpu` & `/internal/provider-versions` endpoints used for internal tracking
  - Network capacity tiles + historical graph (shown on https://stats.akash.network/)
  - Provider detail page in Cloudmos / Console
  
  The default grace period is **3h**, but this can be changed with the `ProviderUptimeGracePeriodMinutes` env var in the api.
- When a provider fails an uptime check, start retrying every minutes and slowly increase the interval based on the following table:

| Downtime Duration | Retry Frequency |
| - | - |
| < 15m | 1m |
| 15m - 1h | 5m |
| 1h - 6h | 15m |
| 6h - 24h | 30m |
| 24h - 7d | 1h |
| 7d+ | 24h |
- Merge the uptime bars when they occurs in the same 15m window (during retries) so that they each represent the same interval. Added an orange bar when a provider is up and down in the same 15m window.
image
- Uptime percentage (1d,7d,30d) will now take into account the interval between each check. That way if a provider fails a check, but the retry is successful after 1m then it will only count for 1 minute of downtime.

### DIFF
--- a/deploy-web/src/pages/providers/[owner]/index.tsx
+++ b/deploy-web/src/pages/providers/[owner]/index.tsx
@@ -1,4 +1,4 @@
-import { useState, useEffect } from "react";
+import { useState, useEffect, useMemo } from "react";
 import { Typography, Box, Paper, useTheme, CircularProgress, Alert } from "@mui/material";
 import { useAllLeases } from "@src/queries/useLeaseQuery";
 import Layout from "@src/components/layout/Layout";
@@ -18,6 +18,7 @@ import { CustomNextSeo } from "@src/components/shared/CustomNextSeo";
 import { UrlService } from "@src/utils/urlUtils";
 import { getNetworkBaseApiUrl } from "@src/utils/constants";
 import axios from "axios";
+import { differenceInMinutes, sub } from "date-fns";
 
 const NetworkCapacity = dynamic(() => import("../../../components/providers/NetworkCapacity"), {
   ssr: false
@@ -87,6 +88,33 @@ const ProviderDetailPage: React.FunctionComponent<Props> = ({ owner, _provider }
     getProviderStatus();
   };
 
+  function groupUptimeChecksByPeriod(uptimeChecks: { isOnline: boolean; checkDate: string }[] = []) {
+    const groupedSnapshots: { checkDate: Date; checks: boolean[] }[] = [];
+
+    const sortedUptimeChecks = uptimeChecks.toSorted((a, b) => new Date(a.checkDate).getTime() - new Date(b.checkDate).getTime());
+
+    for (const snapshot of sortedUptimeChecks) {
+      const recentGroup = groupedSnapshots.find(x => differenceInMinutes(new Date(snapshot.checkDate), x.checkDate) < 15);
+
+      if (recentGroup) {
+        recentGroup.checks.push(snapshot.isOnline);
+      } else {
+        groupedSnapshots.push({
+          checkDate: new Date(snapshot.checkDate),
+          checks: [snapshot.isOnline]
+        });
+      }
+    }
+
+    return groupedSnapshots.map(x => ({
+      date: x.checkDate,
+      status: x.checks.every(x => x) ? "online" : x.checks.every(x => !x) ? "offline" : "partial"
+    }));
+  }
+
+  const uptimePeriods = useMemo(() => groupUptimeChecksByPeriod(provider?.uptime || []), [provider?.uptime]);
+  const wasRecentlyOnline = provider && (provider.isOnline || (provider.lastCheckDate && new Date(provider.lastCheckDate) >= sub(new Date(), { hours: 24 })));
+
   return (
     <Layout isLoading={isLoading}>
       <CustomNextSeo title={`Provider detail ${provider?.name || provider?.owner}`} url={`https://deploy.cloudmos.io${UrlService.providerDetail(owner)}`} />
@@ -98,7 +126,7 @@ const ProviderDetailPage: React.FunctionComponent<Props> = ({ owner, _provider }
           </Box>
         )}
 
-        {provider && !provider.isOnline && !isLoading && (
+        {provider && !wasRecentlyOnline && !isLoading && (
           <Alert
             variant="outlined"
             severity="warning"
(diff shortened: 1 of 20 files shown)


deploy-web/src/pages/providers/[owner]/index.tsx
```diff
@@ -87,6 +88,33 @@ const ProviderDetailPage: React.FunctionComponent<Props> = ({ owner, _provider }
     getProviderStatus();
   };
 
+  function groupUptimeChecksByPeriod(uptimeChecks: { isOnline: boolean; checkDate: string }[] = []) {
+    const groupedSnapshots: { checkDate: Date; checks: boolean[] }[] = [];
+
+    uptimeChecks.toSorted((a, b) => new Date(a.checkDate).getTime() - new Date(b.checkDate).getTime());
```
Comment (ygrishajev) [inline deploy-web/src/pages/providers/[owner]/index.tsx]: Didn't know about this method :) after looking into the doc and testing it seems like it's not mutating the original array. So seems like this usage might not be intended

image

deploy-web/src/pages/providers/[owner]/index.tsx
```diff
@@ -18,6 +18,7 @@ import { CustomNextSeo } from "@src/components/shared/CustomNextSeo";
 import { UrlService } from "@src/utils/urlUtils";
 import { getNetworkBaseApiUrl } from "@src/utils/constants";
 import axios from "axios";
+import { differenceInMinutes, sub } from "date-fns";
```
Comment (ygrishajev) [inline deploy-web/src/pages/providers/[owner]/index.tsx:21]: date fns supports tree shaking but to make use of it I think functions need to be imported one by one
```suggestion
import differenceInMinutes from "date-fns/differenceInMinutes";
import sub from "date-fns/sub";
```

indexer/src/tasks/providerUptimeTracker.ts
```diff
@@ -1,66 +1,49 @@
...
 import { sequelize } from "@src/db/dbConnection";
-import { toUTC } from "@src/shared/utils/date";
-import { add } from "date-fns";
-import { Op, QueryTypes } from "sequelize";
+import { QueryTypes } from "sequelize";
 
 export async function updateProviderUptime() {
   console.log("Updating provider uptimes.");
-  console.time("getAllProviders");
+  console.time("updateProviderUptimes");
```
ygrishajev, line 8: how much sense does it make to have 2 logs?

deploy-web/src/pages/providers/[owner]/index.tsx
Comment (Redm4x) [reply deploy-web/src/pages/providers/[owner]/index.tsx]: Good catch, I changed the `sort` for `toSorted` to make use of the new immutable array methods which I prefer, but forgot to add the new variable.  😄 There's some other cool ones like `toSpliced` and `toReversed`

indexer/src/tasks/providerUptimeTracker.ts
```diff
@@ -1,66 +1,49 @@
...
-  console.timeEnd("getAllProviders");
+    );
 
-  console.time("updateProviderUptime");
-
-  for (const provider of providers) {
-    const uptime1d = provider.total1d > 0 ? provider.online1d / provider.total1d : 0;
-    const uptime7d = provider.total7d > 0 ? provider.online7d / provider.total7d : 0;
-    const uptime30d = provider.total30d > 0 ? provider.online30d / provider.total30d : 0;
+    const uptime1d = Math.max(0, 1 - offline_seconds_1d / (24 * 60 * 60));
```
ygrishajev: i don't mind this numbers. just dropping it here https://www.npmjs.com/package/time-constants - used to use it. a lib is quite simple but improves readability quite a bit

indexer/src/tasks/providerUptimeTracker.ts
ygrishajev: oh, I see, this is is a timer :) discard my comment

deploy-web/src/pages/providers/[owner]/index.tsx
Comment (ygrishajev) [reply deploy-web/src/pages/providers/[owner]/index.tsx]: nice, I like that too!

deploy-web/src/pages/providers/[owner]/index.tsx
```diff
@@ -87,6 +88,33 @@ const ProviderDetailPage: React.FunctionComponent<Props> = ({ owner, _provider }
...
+      }
+    }
+
+    return groupedSnapshots.map(x => ({
+      date: x.checkDate,
+      status: x.checks.every(x => x) ? "online" : x.checks.every(x => !x) ? "offline" : "partial"
+    }));
+  }
+
+  const uptimePeriods = groupUptimeChecksByPeriod(provider?.uptime || []);
```
Comment (baktun14) [inline deploy-web/src/pages/providers/[owner]/index.tsx]: I think this method could be extracted as a hook with `useMemo` since it's called on every render and does a decent amount of calculations.

deploy-web/src/pages/providers/[owner]/index.tsx
Comment (Redm4x) [reply deploy-web/src/pages/providers/[owner]/index.tsx:21]: I just did some tests and it looks like the unused functions are not included in the final bundle even when using the `import { sub } from "date-fns";` syntax.

Turns out Nextjs has a [compile step](https://vercel.com/blog/how-we-optimized-package-imports-in-next-js) that automatically converts `import { sub } from "date-fns"` to `import sub from "date-fns/sub"`. This feature is configurable, but is enabled by default for the most common libraries including date-fns.

Comment (ygrishajev) [reply deploy-web/src/pages/providers/[owner]/index.tsx:21]: Oh, that's cool. I mean I'm not hard on this. If this behaviour is not consistent across other envs I'd still use unified way. Again not a blocker from me

Comment (Redm4x) [reply deploy-web/src/pages/providers/[owner]/index.tsx:21]: One thing we could do is set the optimisation config explicitly in the `next.config.js` file to at least not depend on an unknown library list. Let's revisit this after the rebrand when the codebase will be more stable (lots of new libraries being added).

indexer/src/tasks/providerUptimeTracker.ts
Redm4x: Good idea, I found out that `date-fns` has similar constants so I used those instead. Way more readable 👍

baktun14 (approved):

PR #182, opened 2024-05-06
feat(api): use generated code by version from akash-api

ygrishajev:
Also remove relevant local barrel index files.

refs akash-network/support#184

### DIFF
--- a/deploy-web/src/utils/proto/index.ts
+++ b/deploy-web/src/utils/proto/index.ts
@@ -1,9 +1,11 @@
-import * as v1beta3 from "./v1beta3";
-import * as v1beta4 from "./v1beta4";
+import * as v1beta3 from "@akashnetwork/akash-api/v1beta3";
+import * as v1beta4 from "@akashnetwork/akash-api/v1beta4";
 import { mainnetId, testnetId, sandboxId } from "../constants";
+import omit from "lodash/omit";
 
-const mainnetTypes = { ...v1beta3, ...v1beta4 };
-const sandboxTypes = { ...v1beta3, ...v1beta4 };
+const commonTypes = { ...omit(v1beta3, "GPU"), ...v1beta4 };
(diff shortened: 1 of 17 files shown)


deploy-web/src/utils/proto/index.ts
```diff
@@ -1,9 +1,11 @@
-import * as v1beta3 from "./v1beta3";
-import * as v1beta4 from "./v1beta4";
+import * as v1beta3 from "@akashnetwork/akash-api/v1beta3";
+import * as v1beta4 from "@akashnetwork/akash-api/v1beta4";
 import { mainnetId, testnetId, sandboxId } from "../constants";
+import omit from "lodash/omit";
 
-const mainnetTypes = { ...v1beta3, ...v1beta4 };
-const sandboxTypes = { ...v1beta3, ...v1beta4 };
+const commonTypes = { ...omit(v1beta3, "GPU"), ...v1beta4 };
```
baktun14: Why are we omitting GPU again?

baktun14 (approved):

deploy-web/src/utils/proto/index.ts
ygrishajev: technically because it wasn't [there](https://github.com/akash-network/cloudmos/blob/bc026912cacc8062a9d230a2000d7b33df6113df/deploy-web/src/utils/proto/v1beta3.ts) before. Product wise, I'd ask @Redm4x

Redm4x: Yeah we are not using it in the `deploy-web` project. We can keep it ommited, but I don't think it's that bad if it gets included here anyway.

Redm4x (approved):

deploy-web/src/utils/proto/index.ts
ygrishajev: @Redm4x I can add it back, should I?

PR #184, opened 2024-05-06
Feature/sdl

ygrishajev:
### Changes:
- **Validation Movement**: Shifts validation responsibilities from Cloudmos to Akashjs.
- **Integration**: Updates code to address interface changes
- **Private Registries**: Enables support for private registries

**Note**:
- This is just a part of all the validation. More is to come.
- Requires https://github.com/akash-network/akashjs/pull/82 to be valid and merged

### DIFF
--- a/deploy-web/src/utils/deploymentData/index.ts
+++ b/deploy-web/src/utils/deploymentData/index.ts
@@ -1,22 +1,5 @@
 import * as v1beta3 from "./v1beta3";
-import { mainnetId, testnetId, sandboxId } from "../constants";
 export * from "./helpers";
 
-export let deploymentData;
+export const deploymentData = v1beta3;
 export let selectedNetworkId: string;
-
-export function initDeploymentData() {
-  selectedNetworkId = localStorage.getItem("selectedNetworkId");
-
-  switch (selectedNetworkId) {
-    case mainnetId:
-    case testnetId:
-    case sandboxId:
-      deploymentData = v1beta3;
-      break;
-
-    default:
-      deploymentData = v1beta3;
-      break;
-  }
-}
--- a/deploy-web/src/utils/deploymentData/v1beta2.ts
+++ b/deploy-web/src/utils/deploymentData/v1beta2.ts
@@ -1,38 +1,18 @@
-import { CustomValidationError, ManifestVersion, getCurrentHeight, Manifest, DeploymentGroups, getSdl, parseSizeStr } from "./helpers";
+import { CustomValidationError, DeploymentGroups, getCurrentHeight, getSdl, Manifest, ManifestVersion, parseSizeStr } from "./helpers";
 import { defaultInitialDeposit } from "../constants";
 import { stringToBoolean } from "../stringUtils";
 import path from "path";
 import yaml from "js-yaml";
 import { SDL } from "@akashnetwork/akashjs/build/sdl";
+import { getSelectedNetwork } from "@src/hooks/useSelectedNetwork";
+import { NetworkId } from "@akashnetwork/akashjs/build/types/network";
 
-const endpointNameValidationRegex = /^[a-z]+[-_\da-z]+$/;
-const endpointKindIP = "ip";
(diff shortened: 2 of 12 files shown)


deploy-web/src/utils/deploymentData/v1beta2.ts
ygrishajev, line 1: I see this file is not used anymore. Let's remove it?

deploy-web/src/utils/deploymentData/index.ts
```diff
@@ -1,22 +1,5 @@
 import * as v1beta3 from "./v1beta3";
-import { mainnetId, testnetId, sandboxId } from "../constants";
 export * from "./helpers";
 
-export let deploymentData;
+export const deploymentData = v1beta3;
 export let selectedNetworkId: string;
-
-export function initDeploymentData() {
```
baktun14, line 8: This was used to support different versions on different networks at the same time. Most of the time it's not used, but when new features are released on the sandbox for a few weeks, it's useful to have it.

deploy-web/src/utils/deploymentData/v1beta3.ts
```diff
@@ -1,107 +1,30 @@
...
+import { NetworkId } from "@akashnetwork/akashjs/build/types/network";
 
 export const endpointNameValidationRegex = /^[a-z]+[-_\da-z]+$/;
-export const endpointKindIP = "ip";
 
-function validate(yamlStr: string, yamlJson) {
+function validate(yamlStr: string, yamlJson, networkId: NetworkId) {
   try {
-    SDL.validate(yamlStr);
+    // TODO: use result of this in client code instead of just using validation
```
baktun14, line 13: It's only called at one place in `NewDeploymentData`, maybe we can replace it there.

baktun14: Actually, `getSdl` is called multiple times `DeploymentGroups` `Manifest` `ManifestVersion`, perhaps there's a better to handle this.

deploy-web/src/utils/deploymentData/v1beta2.ts
baktun14: Sure we can clean it up

deploy-web/src/utils/deploymentData/v1beta3.ts
ygrishajev: yes, definitely. I've just left it for the upcoming changes. When I move out all the validation it would be easier to streamline code duplication

deploy-web/src/utils/deploymentData/index.ts
ygrishajev: you mean, you manually update the code?

baktun14: Yes... I will revert it back to what it was.

ygrishajev: this is a funky way to do stuff. types and refs are lost this way I IDEs which can cause untraceable errors. there should be no dynamic exports

ygrishajev: smth like this would do much better

```typescript
import * as v1beta3 from "./v1beta3";
import { mainnetId, testnetId, sandboxId } from "../constants";
import { getSelectedNetwork } from "@src/hooks/useSelectedNetwork";
import { NetworkId } from "@akashnetwork/akashjs/build/types/network";
export * from "./helpers";

const NETWORK_SDL: Record<NetworkId, typeof v1beta3> = {
  [mainnetId]: v1beta3,
  [testnetId]: v1beta3,
  [sandboxId]: v1beta3
};

export let selectedNetwork = getSelectedNetwork();
export let selectedNetworkId = selectedNetwork.id;
export const deploymentData = NETWORK_SDL[selectedNetworkId];
```

baktun14: Yea that works and looks better 👍

PR #204, opened 2024-05-17
feat(setting): implement basic local storage data exporter/importer

ygrishajev:
refs akash-network/cloudmos#56

### DIFF
--- a/api/package-lock.json
+++ b/api/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "cloudmos-api",
-  "version": "2.18.3",
+  "version": "2.18.4",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "cloudmos-api",
-      "version": "2.18.3",
+      "version": "2.18.4",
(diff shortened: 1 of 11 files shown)


Redm4x:
Importing data will overwrite any existing data of the same key right? In my implementation of the [Import from Cloudmos](https://github.com/akash-network/cloudmos/pull/206/files) feature I ended up only importing new keys and merging existing wallets for certs. In this case users will most likely only import data if they have no current data (ex: got wiped), but I'm wondering if we should put a warning that existing data may be overwritten. Wdyt?

ygrishajev:
That was my concern as well. I'm not sure we need to try guessing what user wants. Having a warning is ideal imo too

ygrishajev:
@Redm4x @baktun14 to address what we discussed with @Redm4x above added a confirmation popup on data import attempt.
Also added a handy popup context/hook with promised based `confirm` method. This can be extended to other methods too. This approach reduces boilerplate code/duplication quite a bit.

Redm4x (approved):

PR #205, opened 2024-05-17
fix(deploy-web): change export page to an authorization popup

Redm4x:
The previous page was expected to be loaded in an iframe, but this approach ended up failing because of [Storage Partitioning](https://developers.google.com/privacy-sandbox/3pcd/storage-partitioning) in chrome. This PR changes the export page to instead be loaded as a popup and send back the data to Console via `window.opener`. This also has the benefit of asking for the user permission before sending the data reducing the security risk.


deploy-web/src/pages/standalone/localstorage-export.tsx
```diff
@@ -1,30 +1,101 @@
...
+    }
   }
 
   return (
-    <div>
-      <p>Exporting local storage data to {validExternalDomain}</p>
-    </div>
+    <>
+      <Head>
+        <title>Cloudmos Import</title>
```
baktun14, line 54: Export? Not a big deal :D

baktun14: But I guess when we open it from Console it makes more sense to be Import

baktun14 (approved):

PR #211, opened 2024-05-22
feat(sdl): persist sdl builder form to localstorage

ygrishajev:
I tried to use [this package](https://github.com/tiaanduplessis/react-hook-form-persist/blob/master/tests/index.test.tsx) for this issues, however it doesn't work nicely with `defaultValues`. So for now I copy/pasted that one and going to make a PR to that package as it generally looks handy.

### DIFF
--- a/deploy-web/src/hooks/useFormPersist.tsx
+++ b/deploy-web/src/hooks/useFormPersist.tsx
@@ -1,0 +1,93 @@
+import { useEffect } from "react";
+import { SetFieldValue } from "react-hook-form";
+
+export interface FormPersistConfig {
+  storage?: Storage;
+  watch: (names?: string | string[]) => any;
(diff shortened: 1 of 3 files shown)


deploy-web/src/hooks/useFormPersist.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const watchedValues = watch();
+
+  const getStorage = () => storage || window.sessionStorage;
+
+  const clearStorage = () => getStorage().removeItem(name);
+
+  useEffect(() => {
+    const str = getStorage().getItem(name);
+    const parsed = str ? JSON.parse(str) : defaultValues;
+    console.log("DEBUG parsed", parsed);
```
baktun14: ✂️ ?

baktun14 (approved):

PR #213, opened 2024-05-23
fix(deploy-web): fixed error handling for transaction page

baktun14:
- Handle transaction not found when indexer has picked it up yet.
- Fix error pages on stats app

### DIFF
--- a/apps/api/src/services/db/transactionsService.ts
+++ b/apps/api/src/services/db/transactionsService.ts
@@ -42,7 +42,7 @@ export async function getTransactions(limit: number) {
   }));
 }
 
-export async function getTransaction(hash: string): Promise<ApiTransactionResponse> {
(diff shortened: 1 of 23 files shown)


Redm4x (approved):

apps/api/src/services/db/transactionsService.ts
```diff
@@ -62,7 +63,9 @@ export async function getTransaction(hash: string) {
     ]
   });
 
-  if (!tx) return null;
+  if (!tx) {
+    return null;
```
ygrishajev, line 67: how come is this not failing a type check? `ApiTransactionResponse` can't be null

baktun14: not sure, but I'll modify the return type with `| null`

PR #217, opened 2024-06-03
fix: ensure deploy web type check is passing

ygrishajev:


apps/deploy-web/src/utils/TransactionMessageData.ts
```diff
@@ -221,7 +221,7 @@ export class TransactionMessageData {
           ],
           expiration: expiration
             ? {
-                seconds: longify(Math.floor(expiration.getTime() / 1_000)),
+                seconds: longify(Math.floor(expiration.getTime() / 1_000)) as unknown as Long,
```
baktun14, line 224: Similar fix in the metamask pr
https://github.com/akash-network/cloudmos/pull/216/files#diff-fe2e9c31ce613d1bb35616c66aaa537a33c848cf39b9a6a3d359bca6e207eb86R223

baktun14 (approved):

apps/deploy-web/src/utils/TransactionMessageData.ts
ygrishajev: hehe, yea - that's a fix )))

PR #231, opened 2024-06-07
fix: ensure apps build consistently in docker and locally

ygrishajev:
This PR ensures that all the core dependencies like `react` and `next` are aligned across all monorepo apps/packages. Additionally, it updates the Docker files accordingly. With this PR, all builds and types get consistency across environments and build the same way in both Docker and locally within the context of the monorepo.

**NOTE**: This is 1 of 3 PR which should be merged together. It should be reviewed 1st and merged last after https://github.com/akash-network/cloudmos/pull/232


apps/deploy-web/src/utils/localStorage.ts
```diff
@@ -1,36 +1,26 @@
...
+  const currentVersion: string = publicRuntimeConfig.version;
+  const version = getVersion();
+  const hasPreviousVersion = version && neq(currentVersion, version);
 
-  // Only apply migrations if there was a previous version
-  if (latestUpdatedVersion && neq(currentVersion, latestUpdatedVersion)) {
+  if (hasPreviousVersion) {
     Object.keys(migrations).forEach(version => {
-      if (gt(version, latestUpdatedVersion)) {
+      if (gt(version, version)) {
```
baktun14: maybe not the best to both have the same name :)

apps/landing/package.json
```diff
@@ -18,30 +18,30 @@
...
-    "next-seo": "^5.5.0",
+    "next-pwa": "^5.6.0",
+    "next-seo": "^6.1.0",
     "nextjs-google-analytics": "^2.0.0",
-    "notistack": "^2.0.5",
+    "notistack": "2.0.8",
     "nprogress": "^0.2.0",
-    "react": "^17.0.2",
-    "react-dom": "^17.0.2",
+    "react": "^18.3.1",
```
baktun14: I think we should have the version without `^` so we don't have surprise updates.

Also, there was a new version that kind of introduced a weird warning about `fetchPriority` props on a DOM element. [People are saying that in `18.2.0` is fine](https://github.com/vercel/next.js/issues/65161)

apps/deploy-web/src/utils/localStorage.ts
ygrishajev: right

apps/landing/package.json
ygrishajev: ok, I'll try that. versions are locked anyway. but if it'll fix that warning it makes sense

ygrishajev: done

apps/deploy-web/src/utils/localStorage.ts
ygrishajev: done

ygrishajev:
I'll rebase the other 2 PRs once we're done with this one. I can also go the opposite way and merge this one and rebase the rest against main

baktun14 (approved):

Redm4x (approved):

PR #232, opened 2024-06-07
style(dx): centralize code style tools 

ygrishajev:
This PR centralizes the configurations of ESLint and Prettier under a shared package `dev-config`. All the apps and packages within the monorepo now use these centralized configurations. Additionally, the root monorepo combines them to ensure consistency. This approach allows both linting and formatting to be used globally and within each app/package. Another benefit is that this `dev-config` package can be published to npm for use in other projects.
 
**NOTE**: This is 1 of 3 PR which should be merged together. It should be reviewed 2nd after https://github.com/akash-network/cloudmos/pull/231 and merged 2nd after https://github.com/akash-network/cloudmos/pull/233. Never mind the failing checks: it's due to actual fixes done in a dedicated PR.


packages/dev-config/.prettierrc.js
```diff
@@ -0,0 +1,17 @@
...
+  jsxSingleQuote: false,
+  bracketSameLine: false,
+  trailingComma: "none",
+  requirePragma: false,
+  insertPragma: false,
+  singleQuote: false,
+  arrowParens: "avoid",
+  endOfLine: "crlf",
+  htmlWhitespaceSensitivity: "strict",
+  plugins: ["prettier-plugin-tailwindcss"]
```
Redm4x, line 16: Are the [overrides for tailwindConfig](https://github.com/akash-network/cloudmos/pull/218/commits/1937dcc56c7408b27d03a674d1b8bffa61b3a713#diff-663ade211b3a1552162de21c4031fcd16be99407aae5ceecbb491a2efc43d5d2) no longer needed here?

ygrishajev: They are in the [root .prettierrc.js](https://github.com/akash-network/cloudmos/pull/232/files#diff-7312296367dd7c725b9beb500d7dab7bbeeb2aa03a1010a3fb2935f826d03fa6R5) as package one is more general

Redm4x (approved):

PR #236, opened 2024-06-11
chore: setup prod and dev docker with compose

ygrishajev:
This PR introduces several enhancements to the Docker setup:

- **Staged Builds in Dockerfiles**: Ensures all Dockerfiles use multi-stage builds, allowing services to be built for both development and production using the same file.
- **Alpine-Based Images**: Standardizes all images to use Alpine as the base image to reduce the overall size.
- **Compose Files for Different Environments**:
  - **Build**: Adds a Docker Compose file for building the images to verify they won't break in CI.
  - **Production**: Adds a Docker Compose file for running services in production mode to ensure they won't break once deployed.
  - **Development**: Adds a Docker Compose file for development mode, including hot-reload and other development tools.
- **Root Package.json Updates**: Adds relevant commands to the root `package.json` to improve the developer experience (DX).

With these changes, we can ensure that any developer can test and run the project in any environment, whether it's for development, continuous integration, or production, providing a consistent and reliable setup across all stages.


baktun14:
We need to test actually deploying in prod with this config

apps/stats-web/next.config.js
```diff
@@ -1,4 +1,5 @@
 const { version } = require("./package.json");
+const path = require("path");
```
Redm4x: Is it needed?

docker/Dockerfile.api
```diff
@@ -7,28 +13,20 @@ COPY /packages /app/packages
 COPY package.json /app
 COPY package-lock.json /app
 
-RUN apk add --no-cache python3 make g++ libc6-compat postgresql-dev
+RUN npm install
```
Redm4x, line 16: Is there a reason for using `npm install` instead of `npm ci` in the docker files?

apps/stats-web/next.config.js
ygrishajev: nope, forgot to remove

ygrishajev:
After some thinking I removed the dep on `.env.local` for some services in docker compose file. Added a `.env.sandbox.docker-compose-dev` to git having all that's needed for staging

docker/Dockerfile.stats-web
```diff
@@ -1,61 +1,46 @@
...
-FROM node:20 AS runner
+FROM base AS production
+
 WORKDIR /app
 
 ENV NODE_ENV production
-ENV NEXT_TELEMETRY_DISABLED 1
-
-RUN addgroup --system --gid 1001 nodejs
-RUN adduser --system --uid 1001 nextjs
```
Redm4x: Those user/group comes from the [example dockerfile](https://github.com/vercel/next.js/blob/canary/examples/with-docker/Dockerfile). I think so that nextjs run as a non-root user. Do you think it is not needed? If so I guess we should also remove the `--chown=nextjs:nodejs` from the `COPY` commands bellow

ygrishajev: This is needed for security. If someone potentially hacks into there's less harm that could be done. My reasoning was that it might not be needed for a ui only service as it has no db access. However this can change. Perhaps it's better to get this back. Wdyt?

ygrishajev: anyway, decided to roll this back and add this to other images too

ygrishajev:
@baktun14 @baktun14 improved this further today:

- ensured all apps are build the same way and this allowed to use same docker files for apps with the same stack: nodejs and nextjs
- added health checks to so that apps wait to start till migration is done
- ensured apps are served in non-privileged mode for security implications

ygrishajev:
@baktun14 @Redm4x another thing is I noticed that console is served on port 3001 https://github.com/akash-network/cloudmos/blob/main/apps/deploy-web/deploy.yml#L15 which basically means that Nginx is unused. So I removed the config and relevant steps in the Dockerfile. Let me know if I'm missing anything.

README.md
```diff
@@ -123,6 +123,21 @@ Note: you may pass any `docker compose` related arguments to the above commands.
...
+### Using Turbo Repo
+Another way to run apps in dev mode is using turbo repo setup. Some available commands are:
+```shell
+npm run console:dev # run console ui in dev mode with dependencies
+npm run stats:dev # run stats ui in dev mode with dependencies
+npm run api:dev # run api in dev mode with dependencies
+npm run indexer:dev # run indexer in dev mode with dependencies
+```
+
+Note the above commands still depend on docker to ru postgres database. If you need to run them without db you can use the following commands:
```
baktun14: **run** postgres database

Redm4x (approved):

Issue #230, opened 2024-06-06
Shared packages


baktun14:
Praetor brand will be dissolved into Console to unify the Akash experience and we want to share the same styling with Console by creating a library of components (tailwind/shadcn)

- [x] Create the shared component library under a folder called “ui” in the packages folder
- [x] #250
- [x] Refactor deploy-web to use the config and components
- [x] Add documentation to import components
- [x] Extract re-used custom components from stats + console
- [x] Fix the form components
- ~Add storybook to be able to view and develop shared components~

dominikusbrian:
Will there be opportunity to contribute to this effort as open-source developer ? Contributing to this would be helpful in understading and gaining more experience for building a frontend for projects that are integrating with Akash network.

anilmurty:
@dominikusbrian - yes - the cloudmos/ console code is open source and so is the praetor code

baktun14:
Update June 18 2024:
- Extracted all the shadcn components from deploy-web and stats-web into packages/ui and replaced everywhere we use them.
- I will continue to extract more custom re-used components in another branch

baktun14:
Update June 25 2024:

- Currently working in another branch to extract more shared components

baktun14:
Update July 2nd 2024:
- Extracted more custom components #256 
- I will continue working on this, mainly the form components for the provider console and deploy console

baktun14:
Update:

- Currently working on refactoring the form components

baktun14:
Update:

- I had issues with the schema library for form validation, but figured it out this morning. I will create my pr this week to finish this task and set the standard for form schema validation and components.

baktun14:
Update:
- I figured out all the edge cases for using zod as a schema validation for forms and created my pr https://github.com/akash-network/console/pull/283
- This pr sets the standard for form validation in all the clients
- I will close this issue once this pr is merged.

PR #239, opened 2024-06-14
feat: extract UI components shadcn

baktun14:
- Extract from stats + console the shadcn components + some re-used components in the ui package
- I will continue extracting custom components in another branch


baktun14:
#230

PR #246, opened 2024-06-20
feat: improve provider leases graph

baktun14:
- Improve query to only fetch the data for the lifespan of the provider and not the network
- Improve endpoint for active gpu to only return data from gpu launch
- Update nivo lib version that fixed local errors
- Fixed some styling

### **GPU Leased**
Before:
image
After:
image

---

### **Provider active lease (Europlots)**
Before:
image

After:
image


baktun14:
I will run `npm run lint` once [this pr](https://github.com/akash-network/cloudmos/pull/244) is merged to avoid conflicts

Redm4x (approved):

PR #248, opened 2024-06-25
feat(deployment): implement plain linux deployment page

ygrishajev:
This is an initial implementation.
Upcoming changes:
1) more distros
2) keys generation

[Current DEMO](https://drive.google.com/file/d/1BhHNdGoFExFNUtBkFKOEAlrBn6dXgbrv/view?usp=share_link)


apps/deploy-web/src/components/sdl/SimpleServiceFormControl.tsx
```diff
@@ -219,59 +229,116 @@ export const SimpleServiceFormControl: React.FunctionComponent<Props> = ({
                         control={control}
                         name={`services.${serviceIndex}.image`}
                         rules={{
-                          required: "Docker image name is required.",
-                          validate: value => {
-                            const hasValidChars = /^[a-z0-9\-_/:.]+$/.test(value);
```
baktun14, line 224: We could use the `register` function on the input component if we want to keep the validation instead of the `Controller` component.

apps/deploy-web/src/utils/sdl/transformCustomSdlFields.ts
```diff
@@ -0,0 +1,94 @@
...
+
+export const transformCustomSdlFields = (services: Service[], options?: TransformOptions) => {
+  const pipeline = [addSshPubKey, ensureServiceCount];
+
+  if (options?.withSSH) {
+    pipeline.push(ensureSSHExpose);
+    pipeline.push(mapImage);
+  }
+
+  const transform = flowRight(pipeline);
```
baktun14: It executes each functions sequentially from right to left? Never used it

baktun14 (approved):

baktun14:
Maybe we can find a better name for the route other than `/plain-linux` like `/deploy-linux`, thoughts @anilmurty ?

anilmurty:
hey @ygrishajev - this is pretty much what we wanted - the SSH instructions at the end are 😍 . 

Just a couple comments:

1. I'd suggest a tweak to the copy of the instructions (mainly avoiding calling it a "VM" directly) as follows:
Screenshot 2024-06-25 at 3 01 19 PM

#### Instructions to SSH into the deployed Linux container:
  - Open a command terminal on your machine and copy this into it:
        `ssh root@provider.akash.pub -p 31463 -i ~/.ssh/id_rsa`
  - Replace `~/.ssh/id_rsa` with the path to the private key (stored on your local machine) corresponding to the public key you provided earlier and run the command

2. Agree with @baktun14 's comment re. route.

apps/deploy-web/src/utils/sdl/transformCustomSdlFields.ts
ygrishajev: yea, actually I intended to use simple `flow` but in this case it doesn't matter much

apps/deploy-web/src/components/sdl/SimpleServiceFormControl.tsx
ygrishajev: this is removed by mistake

PR #255, opened 2024-07-01
Feature/sdl

ygrishajev:
This PR adds a context for the SDL builder to share some state that is currently passed through the component tree.
ATM it only exposes values for controlling components visibility to add SSH exposure functionality. However much more can be refactored this way (e.g. some form methods).

The main purpose for now was for us to be able to do stuff like in `apps/deploy-web/src/pages/deploy-linux/index.tsx` or `apps/deploy-web/src/pages/new-deployment/index.tsx` and deeper in the components have access to context to control visibility of certain components.

With the described above SSH is added to main SDL builder.


apps/deploy-web/src/context/SdlBuilderProvider/SdlBuilderProvider.tsx
```diff
@@ -0,0 +1,57 @@
...
+  );
+
+  return (
+    <SdlContext.Provider value={context}>
+      <NewDeploymentContainer />
+    </SdlContext.Provider>
+  );
+};
+
+export const useSdl = () => {
```
baktun14: I think `useSdlBuilder` is a better name for the hook

apps/deploy-web/src/pages/deploy-linux/index.tsx
```diff
@@ -1,5 +1,5 @@
-import { NewDeploymentContainer } from "@src/components/new-deployment/NewDeploymentContainer";
+import { SdlBuilderProvider } from "@src/context/SdlBuilderProvider/SdlBuilderProvider";
 
-export default function NewDeploymentPage() {
-  return <NewDeploymentContainer imageSource="ssh-vms" ssh />;
+export default function DeployLinuxPage() {
+  return <SdlBuilderProvider imageSource="ssh-vms" hiddenComponents={["command", "service-count", "ssh-toggle", "yml-editor", "yml-uploader"]} />;
```
baktun14: I'm not sure I like using context providers at the page level, because for instance we need to declare the provider on each page that would use one of the sub components of the SdlBuilder, which is currently 3 pages (new-deployment, deploy-linux, sdl-builder).

I think a better approach would be to use jotai and custom hook to set/access the global state.

ygrishajev: Yea, I know and I'm going to add this to other pages. This whole thing is already overly complicated. Context is an in built thing that imo fits this perfectly as it's not really needed globally but within the context of a certain page (every page would have own context). Unlike state managers are needed for global state.

baktun14: I think using the Context api is fine. I always use it at the layout level to avoid having to repeat the provider at multiple places, especially if the component is hardcoded. Perhaps we could just put `{children}` and let the page call it's own children.

apps/deploy-web/src/context/SdlBuilderProvider/SdlBuilderProvider.tsx
```diff
@@ -0,0 +1,57 @@
...
+    (component: ComponentNames) => {
+      setHiddenComponents(prev => {
+        const next = new Set(prev);
+        next.has(component) ? next.delete(component) : next.add(component);
+        return next;
+      });
+    },
+    [setHiddenComponents]
+  );
+  const hasCmp = useCallback((component: ComponentNames) => !hiddenComponents.has(component), [hiddenComponents]);
```
baktun14: `hasComponent` would be more explicit.

apps/deploy-web/src/pages/deploy-linux/index.tsx
ygrishajev: Agreed

apps/deploy-web/src/context/SdlBuilderProvider/SdlBuilderProvider.tsx
ygrishajev: Ok, makes sense

apps/deploy-web/src/pages/deploy-linux/index.tsx
ygrishajev: I didn't do that just coz haven't thought of other pages. After Anil's comment in slack I realised I'd need to do that

baktun14: Makes sense!

apps/deploy-web/src/context/SdlBuilderProvider/SdlBuilderProvider.tsx
```diff
@@ -38,20 +53,27 @@ export const SdlBuilderProvider: FCWithChildren<SdlBuilderProviderProps> = ({ im
...
   const context = React.useContext(SdlContext);
   if (!context) {
-    throw new Error("usePopup must be used within a PopupProvider");
+    throw new Error("useSdlBuilder must be used within a SdlContext");
   }
   return context;
 };
+
+export const withSdlBuilder = (options: SdlBuilderProviderProps = {}) =>
+  function wrapWithSdlBiolder<P extends JSX.IntrinsicAttributes>(Component: React.ComponentType<P>): FC<P> | React.ComponentType<P> {
```
baktun14:
```suggestion
  function wrapWithSdlBuilder<P extends JSX.IntrinsicAttributes>(Component: React.ComponentType<P>): FC<P> | React.ComponentType<P> {
```

Redm4x (approved):

baktun14 (approved):

PR #256, opened 2024-07-01
feat: extract custom components

baktun14:
- Extracted snackbar + popup + errorfallback + address components
- Added readme for ui package


apps/deploy-web/src/context/PopupProvider/PopupProvider.tsx
ygrishajev, line 1: this can be extracted along Popup. This is a lib like thing and reduces bilerplate code when using popups quite a bit

baktun14: Oh right good point.

Issue #247, opened 2024-06-24
[Console 2.0] Fiat Payments


baktun14:
### **Overview and Problem Statement**

One of the most, if not the most, requested feature of the Akash Console is to be able to pay with fiat, aka credit cards. The system will be credit based (pre-payment), which will easily allow for free trial through free credits, like all the other cloud platforms currently do.

The goal is to keep the current deployment UX the same and only abstract the wallet interactions with a managed wallet server side.

This is a high level specification of the order and how we will implement the system.

- [x] API to interact with the managed wallet
  - [x] Class to manage wallets (Create wallet derived from main seed, Authz from main wallet to user wallet, sign transactions)
  - ~~[ ] Class to manage deployments (send manifest to provider, interact with provider for logs/shell, write tests, need to think about network upgrades)~~
  - [x] Generic signing tx endpoint, instead of creating specific endpoints for each of the transactions from the frontend
  - [x] User service to assign wallet/deployments (Update the db schema, write tests)
  - [x] Structure the payment API module/router
- [x] Credit manager; scheduled tasks to manage user credits
  - ~~[ ] Auto-refill deployment deposits~~
  - [x] Auto-refill authz for the user managed wallets
  - [x] Implement Stripe integration
  - ~~[ ] Auto-refill account credits with credit card~~
  - [x] Improve logging
  - [x] Free trial
- [x] Frontend
  - [x] Deployment detail, and everywhere we display the balance, modify to hide balance of the deployment since it's handled for the user
- [ ] Sign in with Github
- [ ] Email notifications
  - [ ] Credit purchase confirmation
- [ ] Billing section
  - [ ] Credit transaction history

### **Db Schema changes**

**v1.0**

- user_payment
  - user_id? null when anonymous
  - AddressIndex (hdpath)
  - Address
  - StripeCustomerId
  - CreditAmount

- user_certificate
  - cert
  - cert_key (Encrypted)
 
- UserDeployment
  - Name
  - Dseq
  - SDL (Encrypted)

- CreditActivity (For integrity and debug + billing)
  - Add/Remove credit
  - Authz Tx hash
  - StripePaymentId

ygrishajev:
At the moment, initial wallet creation is implemented on the backend, allowing for the creation of a wallet for a user and granting it a defined trial spending limit. The next steps involve using that wallet to create deployments. The following steps are currently in progress:
	1.	A generic sign endpoint is being added so we can sign transactions from the console UI as usual but on our backend (~1-2 days).
[...]

PR #265, opened 2024-07-10
feat(billing): implement sign-tx endpoint

ygrishajev:
as a part of this:
- refactor some common deps within the billing module
- add custom exceptions and global error handler

refs #247


apps/api/src/billing/routes/sign-tx/sign-tx.router.ts
```diff
@@ -0,0 +1,65 @@
...
+  messages: z
+    .array(
+      z.object({
+        typeUrl: z.string(),
+        value: z.object({})
+      })
+    )
+    .min(1)
+    .openapi({}),
+  fee: z.object({
```
baktun14: I think we should handle the fee in the api to avoid getting it called with huge fees from potential attackers.

apps/api/src/billing/services/managed-user-wallet/managed-user-wallet.service.ts
```diff
@@ -62,42 +64,46 @@ export class WalletService {
...
         this.logger.debug({ event: "SPENDING_ALREADY_AUTHORIZED", address: options.address });
+
+        const revokeMessage = this.rpcMessageService.getRevokeAllowanceMsg({
+          granter: await this.masterWalletService.getFirstAddress(),
+          grantee: options.address
+        });
+
+        console.log("DEBUG revokeMessage", JSON.stringify(revokeMessage, null, 2));
+        const fee = await this.estimateFee([revokeMessage], this.config.TRIAL_ALLOWANCE_DENOM);
+        await this.masterSigningClientService.signAndBroadcast([revokeMessage], fee);
```
baktun14: We revoke the allowance if it already exists?

apps/deploy-web/src/components/new-deployment/ManifestEdit.tsx
```diff
@@ -61,6 +62,7 @@ export const ManifestEdit: React.FunctionComponent<Props> = ({ editedManifest, s
   const sdlBuilderRef = useRef<SdlBuilderRefType>(null);
   const { minDeposit } = useChainParam();
   const { hasComponent } = useSdlBuilder();
+  const wallet = useManagedWallet();
```
baktun14:
```suggestion
  const managedWallet = useManagedWallet();
```

apps/deploy-web/src/components/new-deployment/ManifestEdit.tsx
```diff
@@ -163,7 +165,11 @@ export const ManifestEdit: React.FunctionComponent<Props> = ({ editedManifest, s
       if (!valid) return;
     }
 
-    setIsCheckingPrerequisites(true);
+    if (wallet) {
+      await handleCreateClick(defaultInitialDeposit, "akash1fq8fhssjs0w8x9ysggrm4r8x26522elefm737l");
```
baktun14: The hardcoded address is for testing?

apps/deploy-web/src/components/user/UserSettingsForm.tsx
```diff
@@ -83,98 +83,100 @@ export const UserSettingsForm: React.FunctionComponent = () => {
...
-          <FormPaper>
-            <form onSubmit={handleSubmit(onSubmit)}>
-              <LabelValue label="Email" value={user.email} />
-              <LabelValue
-                label="Username"
-                value={
-                  <>
+      {user?.username && user?.bio && (
+        <UserProfileLayout page="settings" username={user?.username} bio={user?.bio}>
+          {isLoading || !user ? (
```
baktun14: I guess we don't need to check user if the user has a username.
```suggestion
          {isLoading ? (
```

apps/deploy-web/src/context/WalletProvider/WalletProvider.tsx
```diff
@@ -162,24 +181,40 @@ export const WalletProvider = ({ children }) => {
...
+    let txResult: DeliverTxResponse;
 
-      const txResult = await broadcast(txRaw);
+    try {
+      if (user && managedWallet) {
+        const messages = msgs.map(m => ({ ...m, value: Buffer.from(customRegistry.encode(m)).toString("base64") }));
 
-      setIsBroadcastingTx(false);
+        enqueueTxSnackbar();
+        const { data } = await axios.post("http://localhost:3080/v1/tx", {
```
baktun14: We can store the url in the `apiUtils.ts` file

apps/deploy-web/src/pages/_app.tsx
```diff
@@ -1,20 +1,26 @@
+import "nprogress/nprogress.css";
+CertificateProvider;
```
baktun14: :scissors:?

apps/deploy-web/src/queries/useAnonymousUserQuery.ts
```diff
@@ -0,0 +1,43 @@
...
+  stripeCustomerId?: string;
+  bio?: string;
+  subscribedToNewsletter: boolean;
+  youtubeUsername?: string;
+  twitterUsername?: string;
+  githubUsername?: string;
+}
+
+export const anonymousUsersHttp = axios.create({
+  baseURL: `${BASE_API_URL}/v1/anonymous-users`
```
baktun14: `apiUtils.ts`

apps/deploy-web/src/queries/useAnonymousUserQuery.ts
```diff
@@ -0,0 +1,43 @@
...
+const createAnonymousUser = async () => (await anonymousUsersHttp.post<ApiUserOutput>("")).data;
+const getAnonymousUser = async (id: string) => (await anonymousUsersHttp.get<ApiUserOutput>(id)).data;
+
+let userAsPromised: Promise<ApiUserOutput>;
+const findOrCreateAnonymousUser = async (id?: string) => {
+  userAsPromised = userAsPromised || (id ? getAnonymousUser(id) : createAnonymousUser());
+  return await userAsPromised;
+};
+
+export function useAnonymousUserQuery(id?: string, options?: { enabled?: boolean }) {
```
baktun14, line 20: So since this is not called anywhere, it won't affect the state of the console?

apps/deploy-web/src/utils/TransactionMessageData.ts
```diff
@@ -221,7 +221,7 @@ export class TransactionMessageData {
           ],
           expiration: expiration
             ? {
-                seconds: Long.fromInt(Math.floor(expiration.getTime() / 1_000)) as unknown as Long,
+                seconds: BigInt(Math.floor(expiration.getTime() / 1_000)),
```
baktun14, line 224: I think it was fine with `Long.fromInt`, any reason for the change?

apps/deploy-web/src/utils/constants.ts
```diff
@@ -127,7 +127,7 @@ function getProviderProxyWsUrl() {
 export let selectedNetworkId = "";
 
 // 0.5AKT aka 500000uakt
-export const defaultInitialDeposit = 500000;
+export const defaultInitialDeposit = 5000000;
```
baktun14: It's supposed to be 0.5 AKT, any reason for the change?

apps/api/src/billing/routes/sign-tx/sign-tx.router.ts
ygrishajev: yes, missed to remove this from schema. fees are calculated on the backend

ygrishajev: oh, this is an old comment. yes, this is now removed

apps/api/src/billing/services/managed-user-wallet/managed-user-wallet.service.ts
ygrishajev: so this is more for testing purposes. however when there is an allowance it's not possible to add another one. would you suggest a better way? For not the case is that while testing address indexes repeat and allowances are granted repeatedly. However I can imagine we might need to increase allowance when refilling the wallet.

apps/deploy-web/src/components/new-deployment/ManifestEdit.tsx
ygrishajev: yes, should be in env vars

apps/deploy-web/src/queries/useAnonymousUserQuery.ts
ygrishajev: not sure I follow

apps/deploy-web/src/utils/TransactionMessageData.ts
ygrishajev: yes, update package with different types

apps/deploy-web/src/utils/constants.ts
ygrishajev: wasn't working for me this way. let's take a look together into this

apps/api/src/billing/routes/sign-tx/sign-tx.router.ts
baktun14: Sounds good

apps/deploy-web/src/queries/useAnonymousUserQuery.ts
baktun14: When we merge into main and deploy, it won't be called anywhere right?

apps/deploy-web/src/utils/TransactionMessageData.ts
baktun14: Ah ok

apps/deploy-web/src/utils/constants.ts
baktun14: Hmm, maybe the sandbox doesn't have the same params. In any case, we have a `useDepositParams` hook that fetchs the deposit min from the on-chain params, so this variable is only used as the default value and shouldn't matter.

apps/api/src/billing/services/managed-user-wallet/managed-user-wallet.service.ts
baktun14: I guess if we can't have more than one allowance, it makes sense to remove it before we add another one to refresh the allowance.

apps/deploy-web/src/queries/useAnonymousUserQuery.ts
ygrishajev: it would actually. It's not hidden behind any flags

baktun14: Is this what we want? Since the feature is not production ready

ygrishajev: nope, however this is WIP. Even the current chunk of work is not really mergeable. I need to test more, make sure it doesn't affect anything. I do not plan to add new features before merge but I want to test more and adjust if needed. Will disable newly added functionality for sure. thanks for the heads up!

apps/deploy-web/src/utils/constants.ts
ygrishajev: ok, let's take a look at this. doesn't work for me 

``` ERROR (799): Error: Query failed with (6): rpc error: code = Unknown desc = failed to execute message; message index: 1: Deposit too low - 500000 < 5000000uakt: Deposit invalid [akash-network/akash-api@v0.0.67/go/node/deployment/v1beta3/params.go:54] With gas wanted: '0' and gas used: '91641' : unknown request```

baktun14: Perhaps the sandbox param hasn't been updated. We can probably remove this variable and use the `useDepositParams` everywhere we need this value to fix.

ygrishajev: ok, thanks for the hint, [this works](https://github.com/akash-network/console/pull/265/commits/8ed3422ec6a1c417c9f01b2249565ff3f46a5d18)

apps/deploy-web/src/components/user/UserSettingsForm.tsx
ygrishajev: ok, to handle cases like this added a container which would provide user guaranteed. 
https://github.com/akash-network/console/pull/265/commits/5487667ac93d0acc80c1fcc57113b52390ebeb27

baktun14 (approved):

PR #267, opened 2024-07-11
Features/track persistent storage

Redm4x:
## Track provider persistent storage
- Store persistent storage information in the new `ProviderSnapshotStorage` table
- Added columns for persistent storage to `ProviderSnapshot` and renamed the old ones to ephemeral.
- Removed all stats columns from the `Provider` table. We recently added `lastSnapshotId` & `lastSuccessfulSnapshotId` on the `Provider` table and it is easy to get the stats from there.

⚠️ I did not implement persistent storage tracking for the REST status endpoint since almost all active providers have already migrated to GRPC.

### DIFF
--- a/apps/indexer/src/providers/statusEndpointHandlers/grpc.ts
+++ b/apps/indexer/src/providers/statusEndpointHandlers/grpc.ts
@@ -1,4 +1,4 @@
-import { NodeResources } from "@akashnetwork/akash-api/akash/inventory/v1";
+import { NodeResources, Storage } from "@akashnetwork/akash-api/akash/inventory/v1";
 import { ResourcesMetric, Status } from "@akashnetwork/akash-api/akash/provider/v1";
 import { ProviderRPCClient } from "@akashnetwork/akash-api/akash/provider/v1/grpc-js";
 import { Empty } from "@akashnetwork/akash-api/google/protobuf";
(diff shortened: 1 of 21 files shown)


apps/indexer/src/providers/statusEndpointHandlers/grpc.ts
```diff
@@ -22,32 +22,43 @@ export async function fetchProviderStatusFromGRPC(provider: Provider, timeout: n
         cpu: prev.cpu + next.cpu,
         gpu: prev.gpu + next.gpu,
         memory: prev.memory + next.memory,
-        storage: prev.storage + next.storage
+        ephemeralStorage: prev.ephemeralStorage + next.storage,
+        persistentStorage: prev.persistentStorage
```
baktun14, line 26: `prev.persistentStorage + next.persistentStorage`?

baktun14: also `ephemeralStorage: prev.ephemeralStorage + next.ephemeralStorage ,`?

packages/database/dbSchemas/akash/providerSnapshotStorage.ts
```diff
@@ -0,0 +1,17 @@
+import { DataTypes } from "sequelize";
+import { Column, Default, Model, PrimaryKey, Table } from "sequelize-typescript";
+
+import { Required } from "../decorators/requiredDecorator";
+
+@Table({
+  modelName: "providerSnapshotStorage",
```
baktun14, line 7: should we start naming the table name with _ `provider_snapshot_storage`?

apps/indexer/src/providers/statusEndpointHandlers/grpc.ts
Redm4x: The array being reduced only has a `storage` property which is ephemeral storage. It is reduced into a `ephemeralStorage` property and `persistentStorage` stays the same as the initial value which is `0`.

packages/database/dbSchemas/akash/providerSnapshotStorage.ts
Redm4x: I would prefer changing naming convention as we move tables to drizzle or do a single PR for renaming everything. Instead of having mixed convention inside our sequelize schemas. wdyt?

apps/indexer/src/providers/statusEndpointHandlers/grpc.ts
baktun14: Ah ok got it

packages/database/dbSchemas/akash/providerSnapshotStorage.ts
baktun14: Ok yea makes sense

baktun14 (approved):

PR #268, opened 2024-07-12
Wallet connect & sign arbitrary

jigar-arc10:
- Wallet Connect
- cosmos-kit integration
- merge current praetor functionality to new wallet connect
- store jwt for process API calls
- public key storage for encrypting things in future


apps/provider-console/src/context/SettingsProvider/SettingsProviderContext.tsx
```diff
@@ -0,0 +1,307 @@
...
+const defaultSettings: Settings = {
+  apiEndpoint: "",
+  rpcEndpoint: "",
+  isCustomNode: false,
+  nodes: [],
+  selectedNode: null,
+  customNode: null
+};
+
+export const SettingsProvider: FC<{ children: ReactNode }> = ({ children }) => {
```
baktun14, line 54: I'm not sure you guys need all the logic for custom nodes and list of nodes, but I guess it works without a ui to change it.

apps/provider-console/src/types/deployment.ts
```diff
@@ -0,0 +1,335 @@
+export interface DeploymentDetail {
```
baktun14, line 1: We should extract all the types into a separate package

baktun14 (approved):
There's a lot of code duplication with deploy Console. We should try and extract what can be re-used between both apps into the ui package or for type a new types package.

jigar-arc10:
> There's a lot of code duplication with deploy Console. We should try and extract what can be re-used between both apps into the ui package or for type a new types package.

Let's plan the next phase for extraction and shared packages

PR #275, opened 2024-07-24
chore(deploy-web): update Google Analytics configuration 

ghost:
add ga config to track flow from web to console

### DIFF
--- a/apps/deploy-web/src/components/layout/CustomGoogleAnalytics.tsx
+++ b/apps/deploy-web/src/components/layout/CustomGoogleAnalytics.tsx
@@ -16,7 +16,7 @@ export default function GoogleAnalytics() {
 
   return (
     <>
-      {!isProd && (
+      {isProd && (
         <>
           <GAnalytics trackPageViews />
           <GAnalytics trackPageViews gaMeasurementId="G-LFRGN2J2RV" />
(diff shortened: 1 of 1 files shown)


apps/deploy-web/src/components/layout/CustomGoogleAnalytics.tsx
```diff
@@ -14,5 +14,14 @@ export default function GoogleAnalytics() {
     });
   });
 
-  return <>{isProd && <GAnalytics trackPageViews />}</>;
+  return (
+    <>
+      {isProd && (
+        <>
+          <GAnalytics trackPageViews />
+          <GAnalytics trackPageViews gaMeasurementId="G-LFRGN2J2RV" />
```
baktun14, line 22: Which one is this? The main website? We should get this from an env variable

HoomanDigital: yes from main website, we have to use same ga on both: Akash website and Akash console.. we can also keep it without env in this case..
WhatsApp Image 2024-07-24 at 8 57 47 PM

baktun14: Ok but there's going to be 2 GA scripts?

HoomanDigital: Yes, for the time being. We will remove the existing one once the second one starts functioning well as per our needs.

Redm4x (approved):

baktun14 (approved):

PR #281, opened 2024-07-29
fix: move leapwallet CSS library import to component-specific usage

ghost:
## Description

The import of the Leap Wallet CSS library was previously located at the top level of the component, which inadvertently affected the container class on the main console website. This container class is crucial for maintaining the layout and ensuring that the div elements do not extend beyond a specified width.

## Changes Made
- Moved the import of the Leap Wallet CSS library from the top of the component file to the component-specific context.

- This change addresses layout issues by isolating the CSS impact to only where it is needed, preventing unintended interference with the container class and preserving the intended design constraints.

## Benefits
- Improved Layout Consistency: Ensures that the container class on the main console website functions as intended without being affected by external styles.
- Scoped CSS: Reduces potential conflicts by importing the CSS library only in the context where it is actually used.

### Before
image

### After
image

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -1,6 +1,6 @@
 {
   "name": "console-api",
-  "version": "2.20.0",
+  "version": "2.20.1",
   "description": "Api providing data to the deploy tool",
   "repository": {
     "type": "git",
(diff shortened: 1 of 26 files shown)


baktun14:
Good find, thanks for fixing.

PR #282, opened 2024-07-29
feat(user): implement anonymous user registration

ygrishajev:
refs #247


apps/deploy-web/src/pages/api/auth/[...auth0].ts
```diff
@@ -21,6 +22,15 @@ export default handleAuth({
...
             const user_metadata = session.user["https://console.akash.network/user_metadata"];
+            const headers = new AxiosHeaders({
+              Authorization: `Bearer ${session.accessToken}`
+            });
+
+            const anonymousId = req.headers["x-anonymous-user-id"];
+
+            if (anonymousId) {
+              headers.set("X-ANONYMOUS-USER-ID", anonymousId);
+            }
```
Comment (baktun14) [inline apps/deploy-web/src/pages/api/auth/[...auth0].ts]: Should they be both the same format?

baktun14 (approved):
Good stuff!

apps/deploy-web/src/pages/api/auth/[...auth0].ts
Comment (ygrishajev) [reply apps/deploy-web/src/pages/api/auth/[...auth0].ts]: It's lowercased somewhere along the way anyway. But prob makes sense to make it lowercase everywhere to avoid confusion

PR #289, opened 2024-08-06
Feature/billing

ygrishajev:


apps/api/src/billing/services/master-signing-client/master-signing-client.service.ts
```diff
@@ -119,6 +123,15 @@ export class MasterSigningClientService {
       const hashes = [...responses.map(hash => toHex(hash.hash)), lastDelivery.transactionHash];
 
       return await Promise.all(hashes.map(hash => client.getTx(hash)));
+    } catch (error) {
+      if (error.message.includes("account sequence mismatch")) {
```
baktun14, line 91: Does this happen a lot?

apps/deploy-web/src/context/WalletProvider/WalletProvider.tsx
```diff
@@ -175,40 +182,46 @@ export const WalletProvider = ({ children }) => {
...
-        setIsBroadcastingTx(false);
+        setLoadingState(undefined);
       }
 
       if (txResult.code !== 0) {
         throw new Error(txResult.rawLog);
       }
 
-      showTransactionSnackbar("Transaction success!", "", txResult.transactionHash, "success");
+      if (!managedWallet?.isWalletConnected) {
```
baktun14: This feels like a weird condition. If managedWallet is truthy, there will never be a wallet connected. Looking at the hook, we set it to if the managed wallet is thruthy in a ternary for the wallet value. Perhaps we could just check if the userWallet is defined or has an address.

baktun14:
There's currently no way to exit a trial and go back to self custody wallet. We could add the same (...) button with just one option to opt out of the trial and a confirmation?
image

apps/api/src/billing/services/master-signing-client/master-signing-client.service.ts
ygrishajev: it happened once to me

ygrishajev:
> There's currently no way to exit a trial and go back to self custody wallet. We could add the same (...) button with just one option to opt out of the trial and a confirmation? image

looks like a product decision generally. Let's discuss this with @anilmurty
in the meanwhile I don't really see any reason to add extra into this PR, the feature is incomplete anyway

anilmurty:
Funn

> > There's currently no way to exit a trial and go back to self custody wallet. We could add the same (...) button with just one option to opt out of the trial and a confirmation? image
> 
> looks like a product decision generally. Let's discuss this with @anilmurty in the meanwhile I don't really see any reason to add extra into this PR, the feature is incomplete anyway

nice coincidence. I ran into that just this morning - was trying to show someone a demo of the "Start Trial" button and couldn't get back into that state after starting a trial. let's discuss for sure

baktun14 (approved):

PR #291, opened 2024-08-06
Add Metamask Support

schnetzlerjoe:
This PR adds Metamask support to Akash Console allowing MM signing and deployment on Akash.


apps/deploy-web/src/context/CustomChainProvider/CustomChainProvider.tsx
```diff
@@ -13,6 +13,16 @@ import { akash, akashSandbox, akashTestnet, assetLists } from "@src/chains";
 import { useSelectedNetwork } from "@src/hooks/useSelectedNetwork";
 import { customRegistry } from "@src/utils/customRegistry";
 
+import "@interchain-ui/react/styles";
+import "@interchain-ui/react/globalStyles";
```
baktun14: This is already imported at the top of the file

apps/deploy-web/src/context/CustomChainProvider/CustomChainProvider.tsx
```diff
@@ -13,6 +13,16 @@ import { akash, akashSandbox, akashTestnet, assetLists } from "@src/chains";
...
+
+let mm: any[] = [];
+if (typeof window !== 'undefined') {
+  import("@cosmos-kit/cosmos-extension-metamask").then((module) => {
+    mm = module.wallets;
+  });
+}
```
baktun14, line 21: I think a cleaner way would be to use next dynamic import, what do you think?
```
const metamask = dynamic(() => import("@cosmos-kit/cosmos-extension-metamask").wallets, {
  ssr: false
});
```
(I haven't tested this code)

package.json
Redm4x, line 1: Some things have been removed and patch-package added, are those changes needed or a rebase issue?

apps/deploy-web/src/context/CustomChainProvider/CustomChainProvider.tsx
schnetzlerjoe: @baktun14 Tried that but it fails. Could not get it to work . This solution is definitely more stable @baktun14

package.json
schnetzlerjoe: @Redm4x Nope! I fixed it

apps/deploy-web/src/context/CustomChainProvider/CustomChainProvider.tsx
schnetzlerjoe: Fixed! @baktun14

baktun14:
There are way too many changes to the package-lock.json and the package needs to be installed in the console app. I'll make a pr with the changes...

PR #292, opened 2024-08-07
feat(error): add sentry reporter integrated with otl

ygrishajev:
refs #247

### DIFF
--- a/apps/api/src/core/config/env.config.ts
+++ b/apps/api/src/core/config/env.config.ts
@@ -7,7 +7,12 @@ const envSchema = z.object({
   // TODO: make required once billing is in prod
   POSTGRES_DB_URI: z.string().optional(),
   POSTGRES_MAX_CONNECTIONS: z.number({ coerce: true }).optional().default(20),
-  DRIZZLE_MIGRATIONS_FOLDER: z.string().optional().default("./drizzle")
+  DRIZZLE_MIGRATIONS_FOLDER: z.string().optional().default("./drizzle"),
+  DEPLOYMENT_ENV: z.string().optional().default("production"),
+  SENTRY_TRACES_RATE: z.number({ coerce: true }).optional().default(0.01),
+  SENTRY_ENABLED: z.enum(["true", "false"]).optional().default("false"),
+  SENTRY_SERVER_NAME: z.string().optional(),
+  SENTRY_DSN: z.string().optional()
 });
 
 export const envConfig = envSchema.parse(process.env);
--- a/apps/api/src/core/providers/sentry.provider.ts
+++ b/apps/api/src/core/providers/sentry.provider.ts
@@ -1,0 +1,27 @@
+import * as GlobalSentry from "@sentry/node";
+import { container, inject } from "tsyringe";
+
+import { config } from "@src/core/config";
+import packageJson from "../../../package.json";
+
+export const sentryOptions = {
+  dsn: config.SENTRY_DSN,
+  environment: config.DEPLOYMENT_ENV || config.NODE_ENV,
+  tracesSampleRate: config.SENTRY_TRACES_RATE,
+  release: packageJson.version,
+  enabled: config.SENTRY_ENABLED === "true"
+};
+
+GlobalSentry.init({
+  ...sentryOptions,
+  serverName: config.SENTRY_SERVER_NAME,
+  integrations: [new GlobalSentry.Integrations.Http({ tracing: true })]
+});
+
+const SENTRY = "SENTRY";
+
+container.register(SENTRY, { useValue: GlobalSentry });
+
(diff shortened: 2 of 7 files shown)


apps/api/src/core/services/sentry-event/sentry-event.service.ts
```diff
@@ -0,0 +1,92 @@
...
+type ErrorType = Error & {
+  status?: number;
+  statusCode?: number;
+  toJSON?: () => Record<string, any>;
+};
+type ObjectErrorType = Record<string, any>;
+export type ErrorCatchType = ErrorType | ObjectErrorType | string | unknown;
+
+@singleton()
+export class SentryEventService {
```
ygrishajev, line 15: quite often it's not an Error instance that is thrown but a string, or an object. This class makes sure a reasonable event is reported to sentry in any case.

apps/api/src/core/config/env.config.ts
```diff
@@ -7,7 +7,12 @@ const envSchema = z.object({
   // TODO: make required once billing is in prod
   POSTGRES_DB_URI: z.string().optional(),
   POSTGRES_MAX_CONNECTIONS: z.number({ coerce: true }).optional().default(20),
-  DRIZZLE_MIGRATIONS_FOLDER: z.string().optional().default("./drizzle")
+  DRIZZLE_MIGRATIONS_FOLDER: z.string().optional().default("./drizzle"),
+  DEPLOYMENT_ENV: z.string().optional().default("production"),
+  SENTRY_TRACES_RATE: z.number({ coerce: true }).optional().default(0.01),
+  SENTRY_ENABLED: z.enum(["true", "false"]).optional().default("false"),
+  SENTRY_SERVER_NAME: z.string().optional(),
+  SENTRY_DSN: z.string().optional()
```
ygrishajev, line 15: these var names need to be updated before next deploy, I'll add new names that before merging.

apps/api/src/core/providers/sentry.provider.ts
```diff
@@ -0,0 +1,28 @@
...
+import { config } from "@src/core/config";
+import packageJson from "../../../package.json";
+
+export const sentryOptions = {
+  dsn: config.SENTRY_DSN,
+  environment: config.DEPLOYMENT_ENV || config.NODE_ENV,
+  tracesSampleRate: config.SENTRY_TRACES_RATE,
+  release: packageJson.version,
+  enabled: config.SENTRY_ENABLED === "true",
+  debug: true
```
Redm4x: Do we keep `debug: true`?

apps/api/src/core/services/hono-error-handler/hono-error-handler.service.ts
```diff
@@ -25,6 +32,36 @@ export class HonoErrorHandlerService {
...
+  private async reportError<E extends Env = any>(error: Error, c: Context<E>): Promise<void> {
+    const id = this.sentry.captureEvent(await this.getSentryEvent(error, c));
+    this.logger.info({ event: "SENTRY_EVENT_REPORTED", id });
+  }
+
+  private async getSentryEvent<E extends Env = any>(error: Error, c: Context<E>): Promise<Event> {
+    const event = this.sentry.addRequestDataToEvent(this.sentryEventService.toEvent(error), {
+      method: c.req.method,
+      url: c.req.url,
+      headers: Object.fromEntries(c.req.raw.headers),
```
Redm4x: Should we strip out sensitive headers like the `x-anonymous-user-id`?

apps/api/src/core/providers/sentry.provider.ts
ygrishajev: nope :)

apps/api/src/core/services/hono-error-handler/hono-error-handler.service.ts
ygrishajev: I'd leave it there on purpose actually. should help with debugging. do u see any issues with this?

Redm4x: Well it's kind of a username/password so logging it in sentry increases the risk. Do we have any other ID that we could use to identify the anonymous user without being able to login with it? Or maybe replace it by a hashed version?

ygrishajev: imho we shouldn't go crazy trying to protect anonymous user accounts. I don't see any potential risks there. however, I can filter that out just in case. tracing info should be sufficient.

ygrishajev: discussed that it still makes sense to omit. fixed

Redm4x (approved):

PR #293, opened 2024-08-07
fix(auth): properly authenticate new endpoints

ygrishajev:
refs #247


apps/api/src/auth/services/auth.interceptor.ts
```diff
@@ -43,4 +50,12 @@ export class AuthInterceptor implements HonoInterceptor {
       return await next();
     };
   }
+
+  private async getValidUserId(bearer: string, c: Context) {
```
ygrishajev, line 54: I'd put this method into `AuthTokenService` but I don't want that one to depend on http context. And seems like this method needs that in any way. Any thoughts @Redm4x ?

apps/api/webpack.dev.js
```diff
@@ -29,7 +29,7 @@ module.exports = {
       }
     ]
   },
-  plugins: [new NodemonPlugin(), new webpack.IgnorePlugin({ resourceRegExp: /^pg-native$/ })],
+  plugins: [new NodemonPlugin({ watch: [".env.local"] }), new webpack.IgnorePlugin({ resourceRegExp: /^pg-native$/ })],
```
baktun14: Wasn't that not working?

baktun14 (approved):

apps/api/webpack.dev.js
ygrishajev: Yes, need to add more dirs here for it to work properly I think

PR #295, opened 2024-08-12
feat(setting): only show setting page for a connected user wallet

ygrishajev:
refs #247


baktun14 (changes requested):
There's also the `NodeStatusBar` that leads to the settings page. Since we're hiding the settings page entirely, we could just hide the node status in the sidebar.

ygrishajev:
> There's also the `NodeStatusBar` that leads to the settings page. Since we're hiding the settings page entirely, we could just hide the node status in the sidebar.

done https://github.com/akash-network/console/pull/295/files#diff-5dd4aa7a833b73576cfe9e9019748d150105bbb832dae13a062e6c9bb68e8f83R190

baktun14 (approved):

PR #296, opened 2024-08-12
feat(network): stick to a certain network for managed wallets

ygrishajev:
Also refactor network store, so it can serve both inside and outside react context as a single source of truth.

refs #247


apps/deploy-web/src/store/networkStore.ts
```diff
@@ -66,8 +69,42 @@ export const initiateNetworkData = async () => {
   );
 };
 
-const selectedNetwork = atom<Network>(networks[0]);
+const selectedNetworkId = atomWithStorage<Network["id"]>(
```
baktun14, line 72: Didn't know about this util, pretty cool!

apps/deploy-web/src/store/networkStore.ts
```diff
@@ -66,8 +69,42 @@ export const initiateNetworkData = async () => {
...
+  "selectedNetworkId",
+  defaultNetworkId,
```
baktun14, line 74: So this sets the default value in the localStorage on first execution if not set?

apps/deploy-web/src/store/networkStore.ts
```diff
@@ -66,8 +69,42 @@ export const initiateNetworkData = async () => {
...
+    set(selectedNetworkId, next.id);
+  }
+);
 
 export default {
-  selectedNetwork
+  selectedNetworkId,
+  selectedNetwork,
+  getSelectedNetwork: () => store.get(selectedNetwork),
+  useSelectedNetwork: () => useAtom(selectedNetwork)[0]
```
baktun14, line 115: Should we could return the atom instead of just the value so we can set as well?

apps/deploy-web/src/store/networkStore.ts
ygrishajev: it is not, but it wouldn't matter if we didn't have direct local storage accesses. Just realise that. Will add some logic so it does set it till that task is done.

apps/deploy-web/src/store/networkStore.ts
ygrishajev: there is an atom above that can be used. This ones if for backward compatibility with the prev implementation. I can add this detail to the task

apps/deploy-web/src/store/networkStore.ts
ygrishajev: ok, here it is https://github.com/akash-network/console/pull/296/files#diff-9b9e612ba8a4e17b81a175b35a8ce752159500f007a1566958bc9cbabc1240cdR86

apps/deploy-web/src/store/networkStore.ts
ygrishajev: this is actually pretty common pattern. I've seen implementations with every state management lib

baktun14 (approved):

PR #306, opened 2024-08-14
Feature/billing

ygrishajev:


apps/deploy-web/src/utils/walletUtils.ts
```diff
@@ -7,12 +7,7 @@ export type LocalWalletDataType = {
   name: string;
   selected: boolean;
   isManaged?: boolean;
-};
-
-export const useStorageWallets = () => {
-  const wallets = getStorageWallets();
-
-  return { wallets };
+  userId?: string;
```
baktun14, line 10: Maybe I missed it, where do we set this?

apps/deploy-web/src/hooks/useManagedWallet.ts
```diff
@@ -36,6 +36,12 @@ export const useManagedWallet = () => {
     setSelectedNetworkId(NEXT_PUBLIC_MANAGED_WALLET_NETWORK_ID);
   });
 
+  useEffect(() => {
+    if (user?.id && !user.userId) {
```
baktun14, line 40: This is for anonymous user account?

baktun14 (approved):

apps/deploy-web/src/utils/walletUtils.ts
ygrishajev: when managed wallet is fetched from the backend it has a userId and is persisted as is to the local storage

apps/deploy-web/src/hooks/useManagedWallet.ts
ygrishajev: yes

PR #308, opened 2024-08-14
feat(console): revoke all authz

baktun14:
- Added revoke all + checkboxes to select which authorization to revoke
- Added pagination for authorizations given, both deployment deposit and fee allowance
- #302 

image


Redm4x:
I feel like it's more common to have selection checkboxes on the left, wdyt?

baktun14:
> I feel like it's more common to have selection checkboxes on the left, wdyt?

We have them on the right for deployment list as well, so shouldn't be an issue. My reasoning was to have all the actions near each other on the right. I'll merge as is and we can adjust if users complain.

Redm4x (approved):

PR #309, opened 2024-08-15
Feature/env

ygrishajev:


apps/api/env/.env.sample
```diff
@@ -0,0 +1,43 @@
+# Secrets
+AkashDatabaseCS=
+AkashlyticsGithubPAT=
+ANONYMOUS_USER_TOKEN_SECRET=
+CosmosDatabaseCS=
+HealthChecks_SyncAKTMarketData=
+JunoDatabaseCS=
+MASTER_WALLET_MNEMONIC=
+OsmosisDatabaseCS=
```
baktun14:
```suggestion
HealthChecks_SyncAKTMarketData=
MASTER_WALLET_MNEMONIC=
```

apps/api/env/.env.sample
```diff
@@ -0,0 +1,43 @@
...
+POSTGRES_DB_URI=
+SecretToken=
+SentryDSN=
+SENTRY_DSN=
+StargazeDatabaseCS=
+StripeSecretKey=
```
baktun14:
```suggestion
SENTRY_DSN=
StripeSecretKey=
```

baktun14 (changes requested):

.dockerignore
```diff
@@ -2,8 +2,30 @@
...
+apps/landing/.env.*
+
+apps/provider-console/.env
+apps/provider-console/.env.*
+
+apps/provider-proxy/.env
+apps/provider-proxy/.env.*
+
+apps/stats-web/.env
+apps/stats-web/.env.*
```
Redm4x, line 27: Could we do this instead to include all apps and prevent an oversight when creating new apps?
```
apps/*/.env
apps/*/.env.*
```

Redm4x: Same for .gitignore if it makes sense

ygrishajev: I need these to be ignored for all apps except for api now. Eventually when all apps are loading envs in the same way this can be patterned

Redm4x (approved):

PR #316, opened 2024-08-16
Fix uptime checks for some providers

Redm4x:
This PR changes the grpc `client_idle_timeout_ms` config to 5min to prevent an uptime check bug with some providers:
- [provider.nmfakash.uk](https://console.akash.network/providers/akash16jlteppn4tu58848l7q6efhstr5t00rhl03zve?network=mainnet)
- [provider.akash-palmito.org](https://console.akash.network/providers/akash15ksejj7g4su7ljufsg0a8eglvkje94z8qsh68a)

The issue causes uptime checks to fail at first and then succeed at the first retry. This affects the uptime percentage as it counts as multiple 1min downtime period.

While the exact root cause of this issue has not yet been identified, this prevents it for now. If we find a better fix later on we could revert it back to the default value.

### DIFF
--- a/apps/indexer/src/providers/statusEndpointHandlers/grpc.ts
+++ b/apps/indexer/src/providers/statusEndpointHandlers/grpc.ts
@@ -1,4 +1,4 @@
-import { NodeResources, Storage } from "@akashnetwork/akash-api/akash/inventory/v1";
+import { NodeResources } from "@akashnetwork/akash-api/akash/inventory/v1";
 import { ResourcesMetric, Status } from "@akashnetwork/akash-api/akash/provider/v1";
 import { ProviderRPCClient } from "@akashnetwork/akash-api/akash/provider/v1/grpc-js";
 import { Empty } from "@akashnetwork/akash-api/google/protobuf";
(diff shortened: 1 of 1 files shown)


baktun14 (approved):

apps/indexer/src/providers/statusEndpointHandlers/grpc.ts
```diff
@@ -9,6 +9,7 @@ import { promisify } from "util";
 import { parseDecimalKubernetesString, parseSizeStr } from "@src/shared/utils/files";
 import { FakeInsecureCredentials } from "./fake-insecure-credentials";
 import { ProviderStatusInfo } from "./types";
+import { minutesToMilliseconds } from "date-fns";
```
ygrishajev: Date fns supports tree shaking as I remember if u import the function directly

ygrishajev (approved):

Redm4x:
Changed it a bit to set the keepalive interval. Instead of closing the connection after 5min it will send pings to make sure the connection is still active.

[grpc.keepalive_time_ms](https://grpc.github.io/grpc/core/group__grpc__arg__keys.html#gabeeccb441a671122c75384e062b1b91b)

PR #318, opened 2024-08-19
feat(user): transfer a wallet from anonymous user to a logged in one

ygrishajev:
The scenario this PR is addressing:

1) there is a registered user that has no managed wallet
2) the one goes to the console without being logged it
3) starts trial as an anonymous users (and say, creates a deployment)
4) logs in

In this case the anonymous user managed wallet (and the deployment along with it) is transferred to the given already registered user.

Also this PR adds timestamps to user and it's wallet which are really useful when debugging.


apps/api/src/services/db/userDataService.ts
```diff
@@ -142,6 +150,24 @@ export async function getSettingsOrInit({ anonymousUserId, userId, wantedUsernam
   ]);
 }
 
+async function tryToTransferWallet(prevUserId: string, nextUserId: string) {
+  if (process.env.BILLING_ENABLED !== "true") {
+    return;
+  }
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const { UserWalletRepository } = require("@src/billing/repositories/user-wallet/user-wallet.repository");
```
baktun14, line 158: I guess we can add a TODO to refactor this file and make it like the services you built with DI

baktun14 (approved):

apps/api/src/services/db/userDataService.ts
ygrishajev: https://github.com/akash-network/console/issues/322

PR #336, opened 2024-08-23
fix(console): refactor useAllowance hook

baktun14:
- There was a circular reference between the WalletProvider and fee allowances creating a bug where the value of fee granter was undefined in the WalletProvider
- Merged the useAllowance in the WalletProvider because they're intrinsically related and there's no more circular reference


ygrishajev (commented):
I understand the issue here but I can't say I like the solution. `WalletProvider` is already quite complex and large. I'd think on a better way to keep these two separate. In the worst case we can provide wallet as an argument to `useAllowance`.

baktun14:
> I understand the issue here but I can't say I like the solution. `WalletProvider` is already quite complex and large. I'd think on a better way to keep these two separate. In the worst case we can provide wallet as an argument to `useAllowance`.

I can try to refactor it back to being separated. I personally think it makes sense to have them in the same file, even though it's becoming a bit big.

baktun14:
@ygrishajev I refactored it back to 2 files but with params to the hook

Redm4x (approved):

PR #343, opened 2024-08-30
feat(billing): handle tx errors

### LINKED ISSUE #340: Improved managed wallets error handling
- The errors happening in the api return a generic 500, but we should handle return better error messages based on known potential errors.
- Example: When user is out of credits
image
image

ygrishajev:
So far only added 1 error to handle as the range of possible errors we'd want to show to a user with a managed wallet is limited compared to the ones with custodial, but we can add more as we go.

### DIFF
--- a/apps/api/src/billing/routes/sign-and-broadcast-tx/sign-and-broadcast-tx.router.ts
+++ b/apps/api/src/billing/routes/sign-and-broadcast-tx/sign-and-broadcast-tx.router.ts
@@ -11,7 +11,13 @@ export const SignTxRequestInputSchema = z.object({
     messages: z
       .array(
         z.object({
-          typeUrl: z.string(),
+          typeUrl: z.enum([
+            "/akash.deployment.v1beta3.MsgCreateDeployment",
+            "/akash.cert.v1beta3.MsgCreateCertificate",
+            "/akash.market.v1beta4.MsgCreateLease",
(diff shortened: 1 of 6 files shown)


apps/api/src/billing/routes/sign-and-broadcast-tx/sign-and-broadcast-tx.router.ts
```diff
@@ -11,7 +11,13 @@ export const SignTxRequestInputSchema = z.object({
     messages: z
       .array(
         z.object({
-          typeUrl: z.string(),
+          typeUrl: z.enum([
+            "/akash.deployment.v1beta3.MsgCreateDeployment",
```
baktun14, line 15: Not an issue, but we'll need to be aware when there's new types or network upgrades. I think zod throw an error if we send a value outside of the enum.

baktun14 (approved):

apps/api/src/billing/routes/sign-and-broadcast-tx/sign-and-broadcast-tx.router.ts
ygrishajev: Exactly, I think we discussed this. It's unwanted to misuse an api. So I'd prefer any other values are rejected. If there's an upgrade we'll do it as well

Issue #298, opened 2024-08-12
Improve stats app charts


baktun14:
It's currently only possible to zoom for a specific period of time for all the charts on stats.akash.network. We would need to change chart library to another one to support this type of custom zooming.

Suggestions:
https://www.tradingview.com/lightweight-charts/

forbesus:
Hello @baktun14 I will handle this task

ShinySyntax:
Hello, @baktun14 
May i take this issue?

baktun14:
Hey @forbesus and @ShinySyntax , thanks again for your interesting in contributing. @forbesus I will assign this to you if you're willing to give this issue a shot.

forbesus:
Okay, I can do this @baktun14

PR #347, opened 2024-09-02
Feature to improve stats app charts

forbesus:
refs #298


forbesus:
hello @baktun14 
I have done #298 would you check this pr?

baktun14:
Hey @forbesus I just tested and it works great! I have a few modifications requirements:
- Would it possible to remove the TV logo?
- When you switch between 7d 1M ALL, the zoom works but it only includes the data of that range so when you zoom out in 7d/1M, it doesn't really work. There's some logic that slices the data coming from props `rangedData`
- Would it possible to have less padding by default at the top and bottom of the chart?
image
- Would it be possible to have the same unit as the current value in the right axis labels? 
image
image
- Would it be possible to decrease the line width a tiny bit?

Thanks for your contribution!

apps/stats-web/package.json
```diff
@@ -24,7 +24,9 @@
     "geist": "^1.3.0",
     "iconoir-react": "^7.3.0",
     "jotai": "^2.5.1",
+    "lightweight-charts": "^4.2.0",
     "lucide-react": "^0.292.0",
+    "moment": "^2.30.1",
```
baktun14: Please remove moment and use `data-fns`, that's the library we use in other projects for time based functions.

apps/stats-web/src/components/graph/Graph.tsx
```diff
@@ -25,74 +30,140 @@ const Graph: React.FunctionComponent<IGraphProps> = ({ rangedData, snapshotMetad
...
-              x: _snapshot.date,
-              y: roundDecimal(snapshotMetadata.unitFn(_snapshot.value).value)
-            }))
-            .sort(function (a, b) {
-              return Number(new Date(a.x)) - Number(new Date(b.x));
-            })
+    ? rangedData.map(_snapshot => (
+      {
+        time: moment.utc(_snapshot.date).format('YYYY-MM-DD'),
+        value: _snapshot.value
```
baktun14: There was a unitFn called in the previous code `y: roundDecimal(snapshotMetadata.unitFn(_snapshot.value).value)` which is probably why the values don't match.

baktun14 (changes requested):

forbesus:
Hello @baktun14 i fixed and completed
Would you check again?

baktun14:
@forbesus Great, I will check today.

baktun14:
@forbesus  There's a bug when changing the time range, it duplicates the graph. I'm investigating it.
image

forbesus:
Hello @baktun14 I checked it but I have a question
Did you check it after removing browser cache?

baktun14:
> Hello @baktun14 I checked it but I have a question Did you check it after removing browser cache?

It doesn't change the behavior for me

forbesus:
Hello @baktun14 
I fixed that problem 
Would you check again?

baktun14:
@forbesus It works great!

baktun14 (approved):

baktun14:
Great work @forbesus ! I fixed a resize issue and theme issue, but otherwise all is good to merge.

forbesus:
Thank you @baktun14

PR #348, opened 2024-09-03
Improve validation and error handling

Redm4x:
## Improved dseq validation
Improved dseq validation by making sure the string is a positive integer. The previous check was using `parseInt`, but some input that parses successfully are not valid dseq like negative numbers or numbers with leading/trailing spaces. I tried doing the validation with zod, but zod's coersion uses `Number()` which has the same pitfalls as `parseInt()`. 

Endpoints affected:
- `/v1/deployment/{owner}/{dseq}`
- `/internal/leases-duration/{owner}`

## Fixed body parsing during sentry error reporting
The [getSentryEvent](https://github.com/akash-network/console/blob/3430a089629e40019b90fa712d668279b9774982/apps/api/src/core/services/hono-error-handler/hono-error-handler.service.ts#L46-L68) method in [hono-error-handler.service.ts](https://github.com/akash-network/console/blob/3430a089629e40019b90fa712d668279b9774982/apps/api/src/core/services/hono-error-handler/hono-error-handler.service.ts) was calling `c.req.json()` no matter the request's `content-type`. That was failing when an exception was thrown inside an endpoint with no body. I added a `getSentryEventRequestBody` function that parses the body differently based on the `content-type` header.

Error that was previously thrown: `SyntaxError: Unexpected end of JSON input`

I also wrapped the `reportError()` call into a try/catch so that if the sentry reporting fail, the `{"error":"InternalServerError"}` response is still returned.

## No longer report an error when an ibc denom is not found
In the address endpoint we use `coinToAsset()` to looks if some additional information (symbol, logo, description, etc) is available in `@chain-registry/assets` for a specific asset. If an this information was not found we previously reported it as an error in sentry. The goal was to be notified of a potential update needed to the `@chain-registry/assets`. Making sure that the assets have all that additional information is out of scope as there is way too many denoms out there and even the `@chain-registry/assets` package does not include them all. On top of that this information was used inside the block explorer pages which we are no longer promoting. As such I removed the sentry reporting for this. I left a log so that we can always see which coin we are missing in the future if needed.

Endpoint affected: `/v1/addresses/{address}`

### DIFF
--- a/apps/api/src/core/services/hono-error-handler/hono-error-handler.service.ts
+++ b/apps/api/src/core/services/hono-error-handler/hono-error-handler.service.ts
@@ -39,8 +39,12 @@ export class HonoErrorHandlerService {
   }
 
   private async reportError<E extends Env = any>(error: Error, c: Context<E>): Promise<void> {
-    const id = this.sentry.captureEvent(await this.getSentryEvent(error, c));
-    this.logger.info({ event: "SENTRY_EVENT_REPORTED", id });
+    try {
+      const id = this.sentry.captureEvent(await this.getSentryEvent(error, c));
+      this.logger.info({ event: "SENTRY_EVENT_REPORTED", id });
+    } catch (e) {
+      this.logger.error(e);
+    }
   }
 
   private async getSentryEvent<E extends Env = any>(error: Error, c: Context<E>): Promise<Event> {
@@ -48,7 +52,7 @@ export class HonoErrorHandlerService {
       method: c.req.method,
       url: c.req.url,
       headers: omit(Object.fromEntries(c.req.raw.headers), ["x-anonymous-user-id"]),
-      body: await c.req.json()
+      body: await this.getSentryEventRequestBody(c)
     });
     const currentSpan = trace.getSpan(context.active());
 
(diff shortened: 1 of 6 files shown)


apps/api/src/core/services/hono-error-handler/hono-error-handler.service.ts
```diff
@@ -33,7 +33,11 @@ export class HonoErrorHandlerService {
       return c.json({ error: "BadRequestError", data: error.errors }, { status: 400 });
     }
 
-    await this.reportError(error, c);
+    try {
```
ygrishajev: what would you think of handling this in the `reportError` method itself. Seems like it's more relevant there and this method is cleaner

apps/api/src/utils/coin.ts
```diff
@@ -1,48 +1,51 @@
...
+      symbol: "AKT",
+      logoUrl: "https://console.akash.network/images/akash-logo.svg",
+      amount: parseFloat(coin.amount)
+    };
+  } else {
+    const akashChain = asset_lists.find(c => c.chain_name === "akash");
+    const ibcAsset = akashChain.assets.find(a => a.base === coin.denom);
+
+    if (!ibcAsset) {
+      console.log(`Unknown asset ${coin.denom}`);
```
ygrishajev: do we need to act on this one somehow?

Redm4x: Nothing to act on. I left a log so that if in the future we change our mind and want to see which one we are missing we can always filter the logs to find them.

baktun14: Also, perhaps log with the new pino logger?

baktun14 (approved):

ygrishajev (approved):

PR #349, opened 2024-09-04
Update react query

ShinySyntax:


apps/deploy-web/package.json
```diff
@@ -49,6 +49,7 @@
     "@sentry/nextjs": "^7.80.0",
     "@stripe/react-stripe-js": "^1.12.0",
     "@stripe/stripe-js": "^1.41.0",
+    "@tanstack/react-query": "^5.53.2",
```
ygrishajev, line 52: it seems like this version is the latest. Any reason why it's not used?
```suggestion
    "@tanstack/react-query": "^5.54.1",
```

apps/deploy-web/src/context/SettingsProvider/SettingsProviderContext.tsx
```diff
@@ -21,6 +21,31 @@ export type BlockchainNode = {
   id: string;
 };
 
+export type Settings = {
```
ygrishajev, line 24: smth's wrong here with the changes in this file in general. code is duplicated

PR #359, opened 2024-09-09
feat(console): balance authz deployments

baktun14:
- Refactored how we fetch the balances into a single hook
- Improved the display in "Your Account" panel for managed wallets
- Added a modal to explain the escrow accounts/deployments for managed wallets
- Added a warning in the create deployment confirmation when insufficient funds

image
image
image
image


apps/deploy-web/src/hooks/useWalletBalance.ts
```diff
@@ -5,26 +5,93 @@ import { TX_FEE_BUFFER } from "@src/config/tx.config";
...
+        .filter(d => d.authorization.spend_limit.denom === UAKT_DENOM)
+        .reduce((acc, d) => acc + parseFloat(d.authorization.spend_limit.amount), 0);
+      const totalGrantsUUSDC = balances.deploymentGrants.grants
+        .filter(d => d.authorization.spend_limit.denom === usdcIbcDenom)
+        .reduce((acc, d) => acc + parseFloat(d.authorization.spend_limit.amount), 0);
+
+      setWalletBalance({
+        totalUsd: aktUsdValue + totalUsdcValue + totalDeploymentEscrowUSD + totalDeploymentGrantsUSD,
+        balanceUAKT: balances.balanceUAKT + totalGrantsUAKT,
+        balanceUUSDC: balances.balanceUUSDC + totalGrantsUUSDC,
```
ygrishajev, line 65: UUSDC? what does this mean?

apps/deploy-web/src/hooks/useWalletBalance.ts
```diff
@@ -5,26 +5,93 @@ import { TX_FEE_BUFFER } from "@src/config/tx.config";
 import { useChainParam } from "@src/context/ChainParamProvider";
 import { usePricing } from "@src/context/PricingProvider";
 import { useWallet } from "@src/context/WalletProvider";
+import { useBalances } from "@src/queries/useBalancesQuery";
 import { udenomToDenom } from "@src/utils/mathHelpers";
 import { uaktToAKT } from "@src/utils/priceUtils";
 import { useUsdcDenom } from "./useDenom";
 
-export const useTotalWalletBalance = () => {
+export type TotalWalletBalance = {
```
ygrishajev: almost all of the props have `total` in it. Would it make sense to remove it and simplify naming? Not insisting, I might be missing the point here

apps/deploy-web/src/hooks/useWalletBalance.ts
```diff
@@ -5,26 +5,93 @@ import { TX_FEE_BUFFER } from "@src/config/tx.config";
...
+  totalDeploymentEscrowUAKT: number;
+  totalDeploymentEscrowUUSDC: number;
+  totalDeploymentEscrowUSD: number;
+  totalDeploymentGrantsUAKT: number;
+  totalDeploymentGrantsUUSDC: number;
+  totalDeploymentGrantsUSD: number;
+};
+
+export type TotalWalletBalanceReturnType = {
+  isLoadingBalances: boolean;
```
ygrishajev: it's kinda clear from the context it's loading balance. Would this make more sense?
```suggestion
  isLoading: boolean;
```

apps/deploy-web/src/hooks/useWalletBalance.ts
```diff
@@ -5,26 +5,93 @@ import { TX_FEE_BUFFER } from "@src/config/tx.config";
...
+  fetchBalances: () => void;
```
ygrishajev: it's fetching by default, right? I kinda feel like original terminology is better on the interface
```suggestion
  refetch: () => void;
```

apps/deploy-web/src/hooks/useWalletBalance.ts
```diff
@@ -5,26 +5,93 @@ import { TX_FEE_BUFFER } from "@src/config/tx.config";
...
+  walletBalance: TotalWalletBalance | null;
```
ygrishajev:
```suggestion
  balance: TotalWalletBalance | null;
```

apps/deploy-web/src/hooks/useWalletBalance.ts
ygrishajev: I'd generally say it's important to consider this is exposed in the context of the hook that is named well. So it's clear it'a all balances. So more laconic and clear names to me would look like: `usd`, `uakt`, `escrowUakt`, `grantsUakt`, perhaps total makes more sense to me in case when it's combining grants and escrow etc.

apps/deploy-web/src/components/deployments/DeploymentDepositModal.tsx
```diff
@@ -167,8 +168,8 @@ export const DeploymentDepositModal: React.FunctionComponent<Props> = ({ handleC
     setError("");
     clearErrors();
     const deposit = denomToUdenom(amount);
-    const uaktBalance = walletBalances?.uakt || 0;
-    const usdcBalance = walletBalances?.usdc || 0;
+    const uaktBalance = walletBalance?.balanceUAKT || 0;
```
ygrishajev, line 171: should we fall back to default in the hook?

apps/deploy-web/src/hooks/useWalletBalance.ts
```diff
@@ -5,26 +5,93 @@ import { TX_FEE_BUFFER } from "@src/config/tx.config";
...
+};
+
+export const useTotalWalletBalance = (): TotalWalletBalanceReturnType => {
```
ygrishajev: it does show more then totals?
```suggestion
export const useWalletBalance = (): TotalWalletBalanceReturnType => {
```

apps/deploy-web/src/components/shared/ManagedEscrowFaqModal.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { ArrowRight } from "iconoir-react";
+
+import { useChainParam } from "@src/context/ChainParamProvider";
+import { useTotalWalletBalance } from "@src/hooks/useWalletBalance";
+
+export type Props = {
+  onClose: () => void;
+};
+
+export const ManagedEscrowFaqModal: React.FunctionComponent<Props> = ({ onClose }) => {
```
ygrishajev: this totally works. I just would like to point out that using the content and same popup container (portal pattern) in the root of the website is a generally better approach in terms of re-usablility. If we do all the modals like this it would have several downsides:

- it's more boilerplate code, you need to re-declare popup everywhere, imagine to refactor such thing if needed
- your can't make a component to self manage it's state (using it in a tooltip is a great example), meaning more boilerplate code on the client code

apps/deploy-web/src/queries/useBalancesQuery.ts
```diff
@@ -1,92 +1,67 @@
...
-    redelegationsPromise,
-    unbondingsPromise
-  ]);
+  const [balanceResponse, authzBalanceResponse, activeDeploymentsResponse] = await Promise.all([balancePromise, authzBalancePromise, activeDeploymentsPromise]);
 
-  // Balance
-  const balanceData = balanceResponse.data;
-  const balance = balanceData.balances.some(b => b.denom === UAKT_DENOM)
-    ? parseFloat(balanceData.balances.find(b => b.denom === UAKT_DENOM)?.amount || "0")
+  // Authz Grants
```
ygrishajev, line 25: are these comments really helpful?
```suggestion
```

apps/deploy-web/src/queries/useBalancesQuery.ts
```diff
@@ -1,92 +1,67 @@
...
-    unbondings
+    balanceUAKT,
+    balanceUUSDC,
+    deploymentEscrowUAKT,
+    deploymentEscrowUUSDC,
+    deploymentGrantsUAKT,
+    deploymentGrantsUUSDC,
+    activeDeployments,
+    deploymentGrants
   } as Balances;
```
ygrishajev: do you think it's possible to avoid this type cast?

ygrishajev (commented):
lgtm, although I've got quite some comments. It's obviously up to you which ones to apply and I'm happy to discuss if needed

apps/deploy-web/src/hooks/useWalletBalance.ts
baktun14: It's the same as `UAKT` but for `USDC` so `1USDC` = `1_000_000UUSDC`. It's how it's stored on chain.

apps/deploy-web/src/hooks/useWalletBalance.ts
baktun14: Sure yea makes sense.

apps/deploy-web/src/queries/useBalancesQuery.ts
baktun14: Yea sure it's not necessary.

apps/deploy-web/src/queries/useBalancesQuery.ts
baktun14: They certainly don't hurt haha.

apps/deploy-web/src/components/shared/ManagedEscrowFaqModal.tsx
baktun14: Yea I understand your point. I'll create a context for it.

apps/deploy-web/src/hooks/useWalletBalance.ts
ygrishajev: Ah, thx )

apps/deploy-web/src/components/shared/ManagedEscrowFaqModal.tsx
ygrishajev: There is a popup context but you will need to do some work for it I think. I'd suggest we keep this as is as it works. We may create an issue, perhaps leave a todo. My comment aims at aligning approach for the future.

apps/deploy-web/src/components/deployments/DeploymentDepositModal.tsx
baktun14: This shouldn't happen. I added a check to disable the button to submit, but I prefer to return null for the wallet balance object as we check in some places to render stuff or not.

apps/deploy-web/src/hooks/useWalletBalance.ts
baktun14: Sure I'll change the names where it makes sense. Some are totals and some are just balances.

apps/deploy-web/src/hooks/useWalletBalance.ts
baktun14: I think for the ones who are named with totals make sense, we summarize all the deployment escrows, all the grants, etc.
image

ygrishajev (approved):

PR #360, opened 2024-09-10
feat(deployment): ensure usd values in deployments for managed wallets

ygrishajev:
image

### DIFF
--- a/apps/api/src/billing/routes/sign-and-broadcast-tx/sign-and-broadcast-tx.router.ts
+++ b/apps/api/src/billing/routes/sign-and-broadcast-tx/sign-and-broadcast-tx.router.ts
@@ -16,7 +16,8 @@ export const SignTxRequestInputSchema = z.object({
             "/akash.cert.v1beta3.MsgCreateCertificate",
             "/akash.market.v1beta4.MsgCreateLease",
             "/akash.deployment.v1beta3.MsgUpdateDeployment",
-            "/akash.deployment.v1beta3.MsgCloseDeployment"
+            "/akash.deployment.v1beta3.MsgCloseDeployment",
+            "/akash.deployment.v1beta3.MsgDepositDeployment"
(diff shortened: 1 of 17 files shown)


apps/api/src/billing/routes/sign-and-broadcast-tx/sign-and-broadcast-tx.router.ts
```diff
@@ -16,7 +16,8 @@ export const SignTxRequestInputSchema = z.object({
             "/akash.cert.v1beta3.MsgCreateCertificate",
             "/akash.market.v1beta4.MsgCreateLease",
             "/akash.deployment.v1beta3.MsgUpdateDeployment",
-            "/akash.deployment.v1beta3.MsgCloseDeployment"
+            "/akash.deployment.v1beta3.MsgCloseDeployment",
+            "/akash.deployment.v1beta3.MsgDepositDeployment"
```
baktun14, line 20: We'll re-add it to enable deposit for managed wallets, so might as well keep it?

baktun14 (commented):
Perhaps we can have the same modifications in the deployment detail. Wdyt?

Actually, they're already shown in usd, but we could remove the tooltips that show in USDC.
image

apps/api/src/billing/routes/sign-and-broadcast-tx/sign-and-broadcast-tx.router.ts
ygrishajev: I'm not sure I follow

baktun14: Oh my bad I must've been tired, I read it like you removed `MsgDepositDeployment`. Ignore my comment!

baktun14 (approved):

PR #362, opened 2024-09-12
refactor(network): retrieves api versions and urls based on network store

ygrishajev:
Network service was bothering me a bit and I went for refactoring. The root of it is attempt to depend on a single source of truth which is a network store that already holds a selected network with all its configurations.


apps/deploy-web/env/.env.sample
```diff
@@ -20,4 +21,8 @@ NEXT_PUBLIC_BASE_API_TESTNET_URL=
 NEXT_PUBLIC_API_BASE_URL=
 NEXT_PUBLIC_SENTRY_DSN=
 NEXT_PUBLIC_SENTRY_SERVER_NAME=
-NEXT_PUBLIC_GA_MEASUREMENT_ID=
\ No newline at end of file
+NEXT_PUBLIC_GA_MEASUREMENT_ID=
+
+BASE_API_MAINNET_URL=
+BASE_API_TESTNET_URL=
+BASE_API_SANDBOX_URL=
```
ygrishajev, line 28: in docker compose env server and browser hosts are different

apps/deploy-web/src/store/networkStore.ts
```diff
@@ -2,70 +2,25 @@ import axios from "axios";
...
-  }
-];
+export const networks: Network[] = cloneDeep(INITIAL_NETWORKS_CONFIG);
 
-/**
- * Get the actual versions and metadata of the available networks
- */
-export const initiateNetworkData = async () => {
-  networks = await Promise.all(
+export const initiateNetworkVersions = async () => {
```
ygrishajev, line 14: ideally this shouldn't happen. We could have a dedicated components/components blocking app rendering during init. I understand that it happens in SettingsProvider, however it's opaque and one should really dig in to understand this.

apps/deploy-web/src/types/network.ts
```diff
@@ -11,4 +13,6 @@ export type Network = {
   rpcEndpoint?: string;
   version: string | null;
   enabled: boolean;
+  apiVersion: ApiVersion;
+  marketApiVersion: ApiVersion;
```
ygrishajev, line 17: moved these here as it's really a network config, and we have selected network definition logic based on configs already.

apps/deploy-web/src/utils/apiUtils.ts
```diff
@@ -66,82 +66,62 @@ export class ApiUrlService {
...
-  static mainnetVersion() {
-    return `${browserEnvConfig.NEXT_PUBLIC_API_BASE_URL}/v1/version/mainnet`;
-  }
-  static testnetVersion() {
-    return `${browserEnvConfig.NEXT_PUBLIC_API_BASE_URL}/v1/version/testnet`;
-  }
-  static sandboxVersion() {
-    return `${browserEnvConfig.NEXT_PUBLIC_API_BASE_URL}/v1/version/sandbox`;
+
+  static get baseApiUrl() {
```
ygrishajev, line 120: fixes a bug introduced by me earlier - api url should vary based on the network for the above endpoints

apps/deploy-web/src/store/networkStore.ts
baktun14: Yea we could, but the version isn't really mandatory for the app the to render. I think we only show it in the `NetworkSelect` component.

baktun14 (approved):

apps/deploy-web/src/store/networkStore.ts
ygrishajev: Well it's used in many api urls

PR #369, opened 2024-09-17
Feature/network store

ygrishajev:


apps/stats-web/package.json
```diff
@@ -43,6 +43,7 @@
     "zod": "^3.22.4"
   },
   "devDependencies": {
+    "@akashnetwork/akashjs": "@akashnetwork/akashjs",
```
baktun14, line 46: What's the reason why this was added to the stats app?

baktun14 (approved):

apps/stats-web/package.json
ygrishajev: this is in progress, but it's a part of unifying configs. THere's same config as in deploy-web with network store. It depends on types from akashjs.

baktun14: Ah ok, it's dev dependency so not a big deal.

PR #378, opened 2024-09-20
feat(deployment): ensure user can re-send manifest

ygrishajev:
This PR splits create lease main function into 2: creating a lease and sending a manifest. The second happens if the first is successful and a state is updated accordingly. If it's not user can retry manually.

This doesn't still solve an issue if user tries to navigate to such deployment from the deployment list. In this case deployment isn't functional and it's not obvious why.

### DIFF
--- a/apps/deploy-web/src/components/new-deployment/CreateLease.tsx
+++ b/apps/deploy-web/src/components/new-deployment/CreateLease.tsx
@@ -1,5 +1,5 @@
 "use client";
-import React, { useEffect, useState } from "react";
+import React, { useCallback, useEffect, useMemo, useState } from "react";
 import {
   Alert,
   Button,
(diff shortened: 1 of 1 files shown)


apps/deploy-web/src/components/new-deployment/CreateLease.tsx
```diff
@@ -159,48 +205,10 @@ export const CreateLease: React.FunctionComponent<Props> = ({ dseq }) => {
         category: "deployments",
         label: "Create lease"
       });
-    } catch (error) {
-      // Rejected transaction
+      setIsLeaseCreated(true);
```
baktun14, line 206: Out of curiousity, why don't we just call `sendManifest` here instead of using the `useWhen`?

baktun14 (approved):

apps/deploy-web/src/components/new-deployment/CreateLease.tsx
ygrishajev: You may be right) I'll check this once again

ygrishajev: so yea, you're right, changed that

Issue #226, opened 2024-06-04
[Console 2.0] CI/ CD style deployments through Console


anilmurty:
## Summary
This issue tracks the work to implement the prototype presented by the @hoomanhq team in https://github.com/orgs/akash-network/discussions/562 & https://github.com/orgs/akash-network/discussions/522, into Console. This will allow people to be able to connect a github repo, having it automatically build and redeploy and essentially have a vercel like experience with Akash/ Console.

### Suggested Design

- Remove "Hello World" Tile and add tile called "Build & Deploy". Description text "Build & deploy your own React/ Node app from code located in Github"
- Clicking it has options to specify the following things (note that for private repositories, GH OAuth via GH app will be implemented, with user actions via the UI)
   - Github Repo location
   - Branch (default "Master")
   - Private Repo checkbox which will ask for creds, if checked
      - Username
      - Password
   - NodeJS version: Text box or Dropdown
   - SSH Public Key: Text Box
   - Generate SSH key (will generate pub + private key pair, populate pub key and promp user to download and save private key)
   - SSH instructions

For a Private Repo: there are additional steps to authorize installation of the Github App - we ideally want to do some Akash branding around this
   
![Screenshot 2024-06-04 at 2 40 53 PM](https://github.com/akash-network/cloudmos/assets/19495789/c23b6c3f-27a6-4ec1-ab74-220a1631adcf)
When the deployment is done - is it possible to point back to the repo that this is linked to, in the lease details page?

anilmurty:
@hoomanhq - per our discussion, I'm going to assign this to you

hiroyukikumazawa:
I want to participate to this task too.

anilmurty:
hey @hiroyukikumazawa - this feature will be implemented by @hoomandigital and team as part of https://github.com/orgs/akash-network/discussions/606 -- that said, if you are interested in regular contributions, I'll tag you on some that we may be interested in bulding going forward + I'd encourage you to come to sig-clients (cc @brewsterdrinkwater - can add you to the meeting invite)

HoomanDgtl:
We are progressing well on this initiative and are currently testing the work internally. The development is nearly complete, and we plan to merge it into the main site within the next two weeks. Thank you.

PR #381, opened 2024-09-21
feature: Enable Deployment from Git Repositories to Akash Network

HoomanDgtl:
This is a PR that addresses the new feature addition, enabling deployment directly from GitHub, GitLab, or Bitbucket onto the Akash Network.

Issue: https://github.com/akash-network/console/issues/226


apps/deploy-web/src/components/deployments/DeploymentDetail.tsx
```diff
@@ -95,6 +107,25 @@ export function DeploymentDetail({ dseq }: React.PropsWithChildren<{ dseq: strin
...
+        const version = await deploymentData.getManifestVersion(yamlVersion);
+
+        setDeploymentVersion(version);
+      } else {
+        setShowOutsideDeploymentMessage(true);
+      }
+    };
+
+    init();
+  }, [deployment]);
```
ygrishajev: I'm not sure I understand. This code is copied from `ManifestUpdate` into here, so now it runs twice in two components? Is it intended?

apps/deploy-web/src/components/deployments/ManifestUpdate.tsx
```diff
@@ -22,21 +22,39 @@ import { deploymentData } from "@src/utils/deploymentData";
...
 type Props = {
   deployment: DeploymentDto;
   leases: LeaseDto[];
   closeManifestEditor: () => void;
+  remoteDeploy: boolean;
+  showOutsideDeploymentMessage: boolean;
+  editedManifest: string;
+  deploymentVersion: string | null;
+  setDeploymentVersion: Dispatch<React.SetStateAction<string | null>>;
+  setEditedManifest: Dispatch<React.SetStateAction<string>>;
```
ygrishajev: After thinking a bit further I'm not sure it's a good idea to share state dispatchers. The outer interface on the component should basically do 2 way data binding: either accept input or callbacks for events. Otherwise it's confusing and couples components quite a bit.
Naming matters quite a bit in terms how this component would be developing in the future. I'll put an example for 1 of the props:

```suggestion
  onManifestChange: (value: string) => void;
```

so than on the client it would look like:

```tsx
<ManifestUpdate
   onManifestChange={setEditedManifest}
   {...props}
/>
```

apps/deploy-web/src/components/deployments/DeploymentDetail.tsx
ygrishajev: After inspecting the code a bit more it seems to me that this code was only moved to get access to `editedManifest` to get some other relevant data. Is it so?

If it is, this change seems to be a bit of overhead as you could just expose `onManifestChange` prop as suggested in the comment below.

apps/deploy-web/src/components/new-deployment/ManifestEdit.tsx
```diff
@@ -119,6 +123,13 @@ export const ManifestEdit: React.FunctionComponent<Props> = ({ editedManifest, s
     }
   };
 
+  useEffect(() => {
```
ygrishajev: let's do it the same way as it's done in the code above on L92 with `useWhen`, also let's move these two hooks close together

apps/deploy-web/src/components/new-deployment/ManifestEdit.tsx
```diff
@@ -47,15 +48,18 @@ type Props = {
...
 
-export const ManifestEdit: React.FunctionComponent<Props> = ({ editedManifest, setEditedManifest, onTemplateSelected, selectedTemplate }) => {
+export const ManifestEdit: React.FunctionComponent<Props> = ({ editedManifest, setEditedManifest, onTemplateSelected, selectedTemplate, github }) => {
   const [parsingError, setParsingError] = useState<string | null>(null);
   const [deploymentName, setDeploymentName] = useState("");
   const [isCreatingDeployment, setIsCreatingDeployment] = useState(false);
   const [isDepositingDeployment, setIsDepositingDeployment] = useState(false);
   const [isCheckingPrerequisites, setIsCheckingPrerequisites] = useState(false);
   const [selectedSdlEditMode, setSelectedSdlEditMode] = useAtom(sdlStore.selectedSdlEditMode);
+  const [isRepoDataValidated, setIsRepoDataValidated] = useState(false);
```
ygrishajev: `validated` sounds like data was validated leaving validity question unanswered. `data` may mean many things, I suggest `input` as it's a bit more specific. You're welcome to find even better name if you like.

```suggestion
  const [isRepoInputValid, setIsRepoInputValid] = useState(false);
```

apps/deploy-web/src/components/new-deployment/TemplateList.tsx
```diff
@@ -76,11 +81,17 @@ export const TemplateList: React.FunctionComponent = () => {
 
       <div className="mb-8">
         <div className="grid grid-cols-1 gap-2 md:grid-cols-2 md:gap-4 lg:grid-cols-4">
-          <DeployOptionBox
+          {/* <DeployOptionBox
```
ygrishajev: why commented?

apps/deploy-web/src/components/remote-deploy/helper-functions.ts
ygrishajev, line 1: quite some exports here are configuration constants which are handled differently in this app. Check out:
https://github.com/akash-network/console/tree/main/apps/deploy-web/src/config
https://github.com/akash-network/console/blob/main/doc/apps-configuration.md

ygrishajev: helper functions is a bad name. I realise we have some like this but we do not introduce new ones. Please create dedicated service files grouped by their application/domain. There's a services dir in the root of the app which can host such logic.

apps/deploy-web/src/components/remote-deploy/helper-functions.ts
```diff
@@ -0,0 +1,137 @@
+import { Control, UseFormSetValue } from "react-hook-form";
+import { nanoid } from "nanoid";
+
+import { SdlBuilderFormValuesType, ServiceType } from "@src/types";
+import { Owner } from "@src/types/remotedeploy";
+export const ciCdTemplateId = "akash-network-awesome-akash-automatic-deployment-CICD-template";
+export type ServiceControl = Control<SdlBuilderFormValuesType>;
+export type ServiceSetValue = UseFormSetValue<SdlBuilderFormValuesType>;
+export type OAuth = "github" | "gitlab" | "bitbucket";
+export const PROXY_API_URL_AUTH = "https://proxy-console-github.vercel.app";
```
ygrishajev: unused

apps/deploy-web/src/components/remote-deploy/helper-functions.ts
```diff
@@ -0,0 +1,137 @@
+import { Control, UseFormSetValue } from "react-hook-form";
+import { nanoid } from "nanoid";
+
+import { SdlBuilderFormValuesType, ServiceType } from "@src/types";
+import { Owner } from "@src/types/remotedeploy";
+export const ciCdTemplateId = "akash-network-awesome-akash-automatic-deployment-CICD-template";
```
ygrishajev: config constants should be SCREAMING_CASE

apps/deploy-web/src/components/remote-deploy/helper-functions.ts
```diff
@@ -0,0 +1,137 @@
...
+  BITBUCKET_ACCESS_TOKEN: "BITBUCKET_ACCESS_TOKEN",
+  BITBUCKET_USER: "BITBUCKET_USER",
+  DISABLE_PULL: "DISABLE_PULL",
+  GITHUB_ACCESS_TOKEN: "GITHUB_ACCESS_TOKEN",
+  FRONTEND_FOLDER: "FRONTEND_FOLDER",
+  INSTALL_COMMAND: "INSTALL_COMMAND"
+};
+
+export const REDIRECT_URL = `${process.env.NEXT_PUBLIC_REDIRECT_URI}?step=edit-deployment&type=github`;
+export function appendEnv(key: string, value: string, isSecret: boolean, setValue: ServiceSetValue, services: ServiceType[]) {
```
ygrishajev: too much arguments, please make it a configuration object. Some random [ref I found](https://stackoverflow.com/questions/2244860/when-a-method-has-too-many-parameters)

apps/deploy-web/src/components/remote-deploy/helper-functions.ts
```diff
@@ -0,0 +1,137 @@
...
+        e.isSecret = isSecret;
+
+        return e;
+      }
+      return e;
+    });
+  } else {
+    previousEnv.push({ id: nanoid(), key, value, isSecret });
+  }
+  setValue("services.0.env", previousEnv);
```
ygrishajev: having a form method and it's internal details breaks separation of concerns. This function cannot be used outside the form context although it's declared far from such context.
I think this should just return a value and values assignment should happen on the client.

apps/deploy-web/src/components/remote-deploy/helper-functions.ts
```diff
@@ -0,0 +1,137 @@
...
+}
+
+export function removeEnv(key: string, setValue: ServiceSetValue, services: ServiceType[]) {
+  const previousEnv = services[0]?.env || [];
+  const newEnv = previousEnv.filter(e => e.key !== key);
+  setValue("services.0.env", newEnv);
+}
+
+export const removeInitialUrl = (url?: string) => {
```
ygrishajev: I understand this function gets the last 2 segments of the url provided. However it's going to have different outcomes for different urls obviously. So the name is confusing. What's it intended for?

apps/deploy-web/src/components/remote-deploy/helper-functions.ts
```diff
@@ -0,0 +1,137 @@
...
+  return url?.split("/").slice(-2).join("/");
+};
+
+export interface RepoType {
```
ygrishajev:
```suggestion
export interface Repo {
```

apps/deploy-web/src/components/remote-deploy/helper-functions.ts
```diff
@@ -0,0 +1,137 @@
...
+  name: string;
+  id?: string;
+  default_branch: string;
+  html_url: string;
+  userName?: string;
+  private: boolean;
+  owner?: Owner;
+}
+
+export const isRedeployImage = (yml: string, ciCdYml?: string) => {
```
ygrishajev: CICD is usually used as an acronym of the same case.

```suggestion
export const isRedeployImage = (yml: string, cicdYml?: string) => {
```

apps/deploy-web/src/components/remote-deploy/remote-deploy-api-queries/bit-bucket-queries.ts
ygrishajev, line 1: we have all the queries in the root of the app. Having one query per file and named accordingly.

apps/deploy-web/src/components/remote-deploy/remote-deploy-api-queries/bit-bucket-queries.ts
```diff
@@ -0,0 +1,208 @@
...
+import { IGithubDirectoryItem, PackageJson } from "@src/types/remotedeploy";
+import { BitProfile } from "@src/types/remoteProfile";
+import { BitRepository, BitWorkspace } from "@src/types/remoteRepos";
+import { RouteStep } from "@src/types/route-steps.type";
+import { UrlService } from "@src/utils/urlUtils";
+import { ciCdTemplateId } from "../helper-functions";
+
+const Bitbucket_API_URL = "https://api.bitbucket.org/2.0";
+
+export const handleLoginBit = () => {
```
ygrishajev:
```suggestion
export const loginWithBitBucket = () => {
```

apps/deploy-web/src/components/remote-deploy/remote-deploy-api-queries/bit-bucket-queries.ts
```diff
@@ -0,0 +1,208 @@
...
+import { tokens } from "@src/store/remoteDeployStore";
+import { BitBucketCommit } from "@src/types/remoteCommits";
+import { IGithubDirectoryItem, PackageJson } from "@src/types/remotedeploy";
+import { BitProfile } from "@src/types/remoteProfile";
+import { BitRepository, BitWorkspace } from "@src/types/remoteRepos";
+import { RouteStep } from "@src/types/route-steps.type";
+import { UrlService } from "@src/utils/urlUtils";
+import { ciCdTemplateId } from "../helper-functions";
+
+const Bitbucket_API_URL = "https://api.bitbucket.org/2.0";
```
ygrishajev:
```suggestion
const BITBUCKET_API_URL = "https://api.bitbucket.org/2.0";
```

apps/deploy-web/src/components/remote-deploy/remote-deploy-api-queries/bit-bucket-queries.ts
```diff
@@ -0,0 +1,208 @@
...
+};
+const axiosInstance = axios.create({
+  baseURL: Bitbucket_API_URL,
+  headers: {
+    "Content-Type": "application/json",
+    Accept: "application/json"
+  }
+});
+
+export const useFetchRefreshBitToken = () => {
```
ygrishajev:
```suggestion
export const useBitBucketTokenRefresh = () => {
```

apps/deploy-web/src/components/remote-deploy/remote-deploy-api-queries/bit-bucket-queries.ts
ygrishajev: I think this was already mentioned before by @baktun14. This code has quite some code related to http requests. We extract it into a class based services which expose specific typed methods. You may find some examples [here](https://github.com/akash-network/console/tree/73ff83843947c16a75085b7d7e2ddcead49bc67a/apps/deploy-web/src/services)

apps/deploy-web/src/components/remote-deploy/remote-deploy-api-queries/bit-bucket-queries.ts
```diff
@@ -0,0 +1,208 @@
...
+      return response.data;
+    },
+    onSuccess: data => {
+      setToken({
+        access_token: data.access_token,
+        refresh_token: data.refresh_token,
+        type: "bitbucket"
+      });
+
+      router.replace(
```
ygrishajev: it's highly unexpected to manage routing an a hook that's only supposed to do a query.

apps/deploy-web/src/components/remote-deploy/remote-deploy-api-queries/github-queries.ts
```diff
@@ -0,0 +1,177 @@
...
+import { GitCommit } from "@src/types/remoteCommits";
+import { GithubRepository, IGithubDirectoryItem, PackageJson } from "@src/types/remotedeploy";
+import { GitHubProfile } from "@src/types/remoteProfile";
+import { RouteStep } from "@src/types/route-steps.type";
+import { UrlService } from "@src/utils/urlUtils";
+import { ciCdTemplateId, REDIRECT_URL } from "../helper-functions";
+
+const GITHUB_API_URL = "https://api.github.com";
+
+export const handleLogin = () => {
```
ygrishajev:
```suggestion
export const loginWithGitHub = () => {
```

apps/deploy-web/src/components/remote-deploy/remote-deploy-api-queries/github-queries.ts
```diff
@@ -0,0 +1,177 @@
...
+  window.location.href = process.env.NEXT_PUBLIC_GITHUB_APP_INSTALLATION_URL as string;
+};
+
+export const handleReLogin = () => {
```
ygrishajev:
```suggestion
export const reLoginWithGitHub = () => {
```

apps/deploy-web/src/components/remote-deploy/remote-deploy-api-queries/github-queries.ts
```diff
@@ -0,0 +1,177 @@
...
+          headers: {
+            Authorization: `Bearer ${token?.access_token}`
+          }
+        }
+      );
+      return response.data;
+    },
+    onError: (error: AxiosError<{ message: string }>) => {
+      if (error?.response?.data?.message === "Bad credentials") {
+        console.log("bad credentials");
```
ygrishajev: who is this intended for?

apps/deploy-web/src/components/remote-deploy/remote-deploy-api-queries/gitlab-queries.ts
```diff
@@ -0,0 +1,208 @@
...
+import { tokens } from "@src/store/remoteDeployStore";
+import { GitLabCommit } from "@src/types/remoteCommits";
+import { IGithubDirectoryItem, PackageJson } from "@src/types/remotedeploy";
+import { GitLabProfile } from "@src/types/remoteProfile";
+import { GitlabGroup, GitlabRepo } from "@src/types/remoteRepos";
+import { RouteStep } from "@src/types/route-steps.type";
+import { UrlService } from "@src/utils/urlUtils";
+import { ciCdTemplateId } from "../helper-functions";
+
+export const handleGitLabLogin = () => {
```
ygrishajev:
```suggestion
export const loginWithGitLab = () => {
```

apps/deploy-web/src/components/remote-deploy/update/RemoteDeployUpdate.tsx
```diff
@@ -0,0 +1,129 @@
...
+import GithubBranches from "../github/GithubBranches";
+import GitBranches from "../gitlab/GitlabBranches";
+import { appendEnv, ciCdTemplateId, protectedEnvironmentVariables } from "../helper-functions";
+import Rollback from "./Rollback";
+
+const RemoteDeployUpdate = ({ sdlString, setEditedManifest }: { sdlString: string; setEditedManifest: Dispatch<React.SetStateAction<string | null>> }) => {
+  console.log(sdlString);
+
+  const [token] = useAtom(tokens);
+  const [, setIsInit] = useState(false);
```
ygrishajev: why is such state needed if it's value is unused?

apps/deploy-web/src/components/remote-deploy/update/RemoteDeployUpdate.tsx
```diff
@@ -0,0 +1,129 @@
...
+  const { enqueueSnackbar } = useSnackbar();
+  const [, setError] = useState<string | null>(null);
+  const [isEditingEnv, setIsEditingEnv] = useState<number | boolean | null>(false);
+  const { control, watch, setValue } = useForm<SdlBuilderFormValuesType>({ defaultValues: { services: [defaultService] } });
+  const { fields: services } = useFieldArray({ control, name: "services", keyName: "id" });
+  const { getTemplateById } = useTemplates();
+  const remoteDeployTemplate = getTemplateById(ciCdTemplateId);
+  useEffect(() => {
```
ygrishajev, line 33:
```suggestion

  useEffect(() => {
```

apps/deploy-web/src/components/remote-deploy/update/Rollback.tsx
```diff
@@ -0,0 +1,48 @@
...
+import RollbackModal from "./RollbackModal";
+
+const Rollback = ({ services, control }: { services: ServiceType[]; control: Control<SdlBuilderFormValuesType> }) => {
+  const { data } = useCommits(
+    services?.[0]?.env?.find(e => e.key === protectedEnvironmentVariables.REPO_URL)?.value?.replace("https://github.com/", "") ?? "",
+    services?.[0]?.env?.find(e => e.key === protectedEnvironmentVariables.BRANCH_NAME)?.value ?? ""
+  );
+  const { data: labCommits } = useGitLabCommits(
+    services?.[0]?.env?.find(e => e.key === protectedEnvironmentVariables.GITLAB_PROJECT_ID)?.value,
+    services?.[0]?.env?.find(e => e.key === protectedEnvironmentVariables.BRANCH_NAME)?.value
```
ygrishajev: please extract this and the other ones above into a properly named var.

apps/deploy-web/src/components/remote-deploy/update/Rollback.tsx
```diff
@@ -0,0 +1,48 @@
...
+  );
+  const { data: bitbucketCommits } = useBitBucketCommits(
+    removeInitialUrl(services?.[0]?.env?.find(e => e.key === protectedEnvironmentVariables.REPO_URL)?.value ?? "")
+  );
+
+  let commits: RollBackType[] | null = null;
+
+  if (data && data?.length > 0) {
```
ygrishajev:
```suggestion
  if (data?.length > 0) {
```

apps/deploy-web/src/components/remote-deploy/update/Rollback.tsx
```diff
@@ -0,0 +1,48 @@
...
+  );
+  const { data: labCommits } = useGitLabCommits(
+    services?.[0]?.env?.find(e => e.key === protectedEnvironmentVariables.GITLAB_PROJECT_ID)?.value,
+    services?.[0]?.env?.find(e => e.key === protectedEnvironmentVariables.BRANCH_NAME)?.value
+  );
+  const { data: bitbucketCommits } = useBitBucketCommits(
+    removeInitialUrl(services?.[0]?.env?.find(e => e.key === protectedEnvironmentVariables.REPO_URL)?.value ?? "")
+  );
+
+  let commits: RollBackType[] | null = null;
```
ygrishajev: wrap into `useMemo`

apps/deploy-web/src/components/remote-deploy/update/RollbackModal.tsx
```diff
@@ -0,0 +1,129 @@
...
+import { RollBackType } from "@src/types/remotedeploy";
+import { protectedEnvironmentVariables } from "../helper-functions";
+
+const RollbackModal = ({ data, control }: { data?: RollBackType[] | null; control: Control<SdlBuilderFormValuesType> }) => {
+  const [filteredData, setFilteredData] = useState<RollBackType[]>([]);
+  const [value, setValue] = useState<string>("");
+  const { fields: services } = useFieldArray({ control, name: "services", keyName: "id" });
+  const { append, update } = useFieldArray({ control, name: "services.0.env", keyName: "id" });
+  const currentHash = services[0]?.env?.find(e => e.key === protectedEnvironmentVariables.COMMIT_HASH);
+  useEffect(() => {
```
ygrishajev, line 34:
```suggestion

  useEffect(() => {
```

apps/deploy-web/src/components/remote-deploy/update/RollbackModal.tsx
```diff
@@ -0,0 +1,129 @@
...
+import { GitCommitVertical, GitGraph, Info } from "lucide-react";
+import { nanoid } from "nanoid";
+
+import { SdlBuilderFormValuesType } from "@src/types";
+import { RollBackType } from "@src/types/remotedeploy";
+import { protectedEnvironmentVariables } from "../helper-functions";
+
+const RollbackModal = ({ data, control }: { data?: RollBackType[] | null; control: Control<SdlBuilderFormValuesType> }) => {
+  const [filteredData, setFilteredData] = useState<RollBackType[]>([]);
+  const [value, setValue] = useState<string>("");
```
ygrishajev: which value is this?

apps/deploy-web/src/components/remote-deploy/update/RollbackModal.tsx
```diff
@@ -0,0 +1,129 @@
...
+  TabsTrigger
+} from "@akashnetwork/ui/components";
+import { GitCommitVertical, GitGraph, Info } from "lucide-react";
+import { nanoid } from "nanoid";
+
+import { SdlBuilderFormValuesType } from "@src/types";
+import { RollBackType } from "@src/types/remotedeploy";
+import { protectedEnvironmentVariables } from "../helper-functions";
+
+const RollbackModal = ({ data, control }: { data?: RollBackType[] | null; control: Control<SdlBuilderFormValuesType> }) => {
```
ygrishajev: what is this data?

apps/deploy-web/src/components/sdl/EnvFormModal.tsx
```diff
@@ -30,7 +41,7 @@ export const EnvFormModal: React.FunctionComponent<Props> = ({ control, serviceI
   });
 
   useEffect(() => {
-    if (_envs.length === 0) {
+    if (_envs.length === 0 || (isRemoteDeployEnvHidden && _envs.filter(e => !(e?.key?.trim() in protectedEnvironmentVariables)).length === 0 && !update)) {
```
ygrishajev: pls create some vars here

apps/deploy-web/src/components/sdl/EnvFormModal.tsx
```diff
@@ -80,83 +91,87 @@ export const EnvFormModal: React.FunctionComponent<Props> = ({ control, serviceI
...
-                  ["justify-end"]: envIndex === 0 || !hasSecretOption
-                })}
-              >
-                {envIndex > 0 && (
-                  <Button onClick={() => removeEnv(envIndex)} size="icon" variant="ghost">
-                    <Bin />
-                  </Button>
-                )}
+        {envs
+          ?.filter(e => !isRemoteDeployEnvHidden || !(e?.key?.trim() in protectedEnvironmentVariables))
```
ygrishajev: extract this to a var

apps/deploy-web/src/components/sdl/EnvVarList.tsx
```diff
@@ -1,23 +1,27 @@
...
   serviceIndex?: number;
   children?: ReactNode;
   setIsEditingEnv: Dispatch<SetStateAction<boolean | number>>;
+  isRemoteDeployEnvHidden?: boolean;
 };
 
-export const EnvVarList: React.FunctionComponent<Props> = ({ currentService, setIsEditingEnv, serviceIndex }) => {
+export const EnvVarList: React.FunctionComponent<Props> = ({ currentService, setIsEditingEnv, serviceIndex, isRemoteDeployEnvHidden }) => {
   const { hasComponent } = useSdlBuilder();
+  const currentEnvs = currentService.env?.filter(e => !isRemoteDeployEnvHidden || !(e?.key?.trim() in protectedEnvironmentVariables));
```
ygrishajev: why is it being trimmed everywhere? Can it be trimmed somewhere around the source?

apps/deploy-web/src/components/sdl/EnvVarList.tsx
```diff
@@ -51,8 +55,8 @@ export const EnvVarList: React.FunctionComponent<Props> = ({ currentService, set
         </span>
       </div>
 
-      {(currentService.env?.length || 0) > 0 ? (
-        currentService.env?.map((e, i) => (
+      {(currentEnvs?.length || 0) > 0 ? (
```
ygrishajev: this should be sufficient. unless I'm missing smth
```suggestion
      {currentEnvs?.length ? (
```

apps/deploy-web/src/components/sdl/EnvVarList.tsx
```diff
@@ -51,8 +55,8 @@ export const EnvVarList: React.FunctionComponent<Props> = ({ currentService, set
...
+        currentEnvs?.map((e, i) => (
```
ygrishajev:
```suggestion
        currentEnvs.map((e, i) => (
```

apps/deploy-web/src/hooks/useRemoteDeployFramework.tsx
```diff
@@ -0,0 +1,57 @@
...
+  services,
+  setValue,
+  subFolder
+}: {
+  services: ServiceType[];
+  setValue: ServiceSetValue;
+
+  subFolder?: string;
+}) => {
+  const [data, setData] = useState<PackageJson | null>(null);
```
ygrishajev:
```suggestion
  const [packageJson, setPackageJson] = useState<PackageJson | null>(null);
```

apps/deploy-web/src/hooks/useRemoteDeployFramework.tsx
```diff
@@ -0,0 +1,57 @@
+import { useState } from "react";
+
+import { removeInitialUrl, ServiceSetValue, supportedFrameworks } from "@src/components/remote-deploy/helper-functions";
+import { useBitPackageJson } from "@src/components/remote-deploy/remote-deploy-api-queries/bit-bucket-queries";
+import { usePackageJson } from "@src/components/remote-deploy/remote-deploy-api-queries/github-queries";
+import { useGitlabPackageJson } from "@src/components/remote-deploy/remote-deploy-api-queries/gitlab-queries";
+import { ServiceType } from "@src/types";
+import { PackageJson } from "@src/types/remotedeploy";
+
+const useRemoteDeployFramework = ({
```
ygrishajev, line 10: I think it's better to remove dep on the form here, pass repo url and other values from envs and resolve with a framework. This would make this hook more reusable and maintainable

apps/deploy-web/src/pages/api/bitbucket/authenticate.ts
```diff
@@ -0,0 +1,23 @@
+import { NextApiRequest, NextApiResponse } from "next";
+
+import BitbucketAuth from "@src/services/auth/bitbucket.service";
+
+const NEXT_PUBLIC_BITBUCKET_CLIENT_ID: string = process.env.NEXT_PUBLIC_BITBUCKET_CLIENT_ID as string;
+const BITBUCKET_CLIENT_SECRET: string = process.env.BITBUCKET_CLIENT_SECRET as string;
```
ygrishajev: use server env config

apps/deploy-web/src/pages/api/bitbucket/authenticate.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export default async function exchangeBitBucketCodeForTokensHandler(req: NextApiRequest, res: NextApiResponse): Promise<void> {
+  const { code }: { code: string } = req.body;
+
+  if (!code) {
+    return res.status(400).send("No authorization code provided");
```
ygrishajev: we're trying to unify api responses. to shed some light https://github.com/akash-network/console/blob/main/apps/api/src/core/services/hono-error-handler/hono-error-handler.service.ts

Basically it should replicate `http-errors` serialised error

```suggestion
    return res.status(400).send({ error: 'BadRequestError', message: "No authorization code provided",  });
```

apps/deploy-web/src/services/auth/bitbucket.service.ts
```diff
@@ -0,0 +1,59 @@
...
+      Authorization: `Basic ${Buffer.from(`${this.clientId}:${this.clientSecret}`).toString("base64")}`,
+      "Content-Type": "application/x-www-form-urlencoded"
+    };
+
+    try {
+      const response: AxiosResponse = await axios.post(this.tokenUrl, params.toString(), { headers });
+      const { access_token, refresh_token }: Tokens = response.data;
+      return { access_token, refresh_token };
+    } catch (error) {
+      throw new Error("Failed to exchange authorization code for tokens");
```
ygrishajev: what is the point of overriding an error here?

apps/deploy-web/src/services/auth/bitbucket.service.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    const headers = {
+      Authorization: `Basic ${Buffer.from(`${this.clientId}:${this.clientSecret}`).toString("base64")}`,
+      "Content-Type": "application/x-www-form-urlencoded"
+    };
+
+    try {
+      const response: AxiosResponse = await axios.post(this.tokenUrl, params.toString(), { headers });
+      const { access_token, refresh_token }: Tokens = response.data;
+      return { access_token, refresh_token };
```
ygrishajev: let's rename to camelCase

apps/deploy-web/src/services/auth/github.service.ts
```diff
@@ -0,0 +1,39 @@
...
+        client_secret: this.clientSecret,
+        code: authorizationCode,
+        redirect_uri: this.redirectUri
+      });
+
+      const params = new URLSearchParams(response.data);
+      const access_token = params.get("access_token");
+
+      if (!access_token) {
+        throw new Error("No access token returned from GitHub");
```
ygrishajev: this error wouldn't make it outside this method

apps/deploy-web/src/types/remoteCommits.ts
ygrishajev, line 1: did you have a chance to check open source types?
I found this one that seems to do the job https://github.com/octokit/openapi-types.ts/tree/main

apps/deploy-web/src/types/remoteProfile.ts
```diff
@@ -0,0 +1,115 @@
+export interface BitProfile {
```
ygrishajev, line 4: do you really use all of the fields? perhaps we could remove the unused ones. this way we wouldn't need to maintain them

apps/deploy-web/src/components/deployments/DeploymentDetail.tsx
baktun14: Yea this is strange, please re-use the same code from `ManifestUpdate`.

apps/deploy-web/src/components/new-deployment/NewDeploymentContainer.tsx
```diff
@@ -34,6 +36,20 @@ export const NewDeploymentContainer: FC = () => {
     const queryStep = searchParams?.get("step");
     const _activeStep = getStepIndexByParam(queryStep);
     setActiveStep(_activeStep);
+    const state = searchParams?.get("state");
+    const redeploy = searchParams?.get("redeploy");
+    const code = searchParams?.get("code");
```
baktun14: Why is this checked in 2 different `useEffect`?

apps/deploy-web/src/components/new-deployment/NewDeploymentContainer.tsx
```diff
@@ -49,9 +65,32 @@ export const NewDeploymentContainer: FC = () => {
...
-        router.replace(UrlService.newDeployment({ ...searchParams, step: RouteStep.editDeployment }));
+        if (isRedeployImage(template?.content as string, getTemplateById(ciCdTemplateId)?.deploy)) {
+          router.replace(
+            UrlService.newDeployment({
+              ...searchParams,
+              step: RouteStep.editDeployment,
+              gitProvider: "github"
+            })
+          );
+          setGithub(true);
```
baktun14: We already set it above, why is it set again here? Also, the only thing changing here is the url object
```
UrlService.newDeployment({
              ...searchParams,
              step: RouteStep.editDeployment,
              gitProvider: "github"
            })
```
It would be nice to refactor this logic into just setting the object based on the conditions and only have 1 `router.replace(...)`

apps/deploy-web/src/components/new-deployment/NewDeploymentContainer.tsx
```diff
@@ -12,12 +12,14 @@ import { RouteStep } from "@src/types/route-steps.type";
...
 import { UrlService } from "@src/utils/urlUtils";
 import Layout from "../layout/Layout";
+import { ciCdTemplateId, isRedeployImage } from "../remote-deploy/helper-functions";
 import { CreateLease } from "./CreateLease";
 import { ManifestEdit } from "./ManifestEdit";
 import { CustomizedSteppers } from "./Stepper";
 import { TemplateList } from "./TemplateList";
 
 export const NewDeploymentContainer: FC = () => {
+  const [github, setGithub] = useState<boolean>(false);
```
baktun14: this should be changed to `isGithub` as it's a boolean.

apps/deploy-web/src/components/new-deployment/SdlBuilder.tsx
```diff
@@ -16,173 +16,193 @@ import { defaultService, defaultSshVMService } from "@src/utils/sdl/data";
 import { generateSdl } from "@src/utils/sdl/sdlGenerator";
 import { importSimpleSdl } from "@src/utils/sdl/sdlImport";
 import { transformCustomSdlFields, TransformError } from "@src/utils/sdl/transformCustomSdlFields";
+import RemoteRepositoryDeployManager from "../remote-deploy/RemoteRepositoryDeployManager";
 import { SimpleServiceFormControl } from "../sdl/SimpleServiceFormControl";
 
 interface Props {
   sdlString: string | null;
   setEditedManifest: Dispatch<string>;
+  github?: boolean;
```
baktun14: `isGithub`

apps/deploy-web/src/components/new-deployment/TemplateList.tsx
baktun14: It replaced the hello world template, but this one should be moved in the list of "Staff templates" at the end of the list.

apps/deploy-web/src/components/remote-deploy/helper-functions.ts
baktun14: Agreed.

apps/deploy-web/src/components/remote-deploy/remote-deploy-api-queries/bit-bucket-queries.ts
baktun14: Yes, why was this not refactored as commented before?

apps/deploy-web/src/components/remote-deploy/remote-deploy-api-queries/bit-bucket-queries.ts
baktun14: Yes, this can lead to very unexpected use cases.

apps/deploy-web/src/components/remote-deploy/remote-deploy-api-queries/bit-bucket-queries.ts
```diff
@@ -0,0 +1,208 @@
...
+      );
+    }
+  });
+};
+
+export const useBitUserProfile = () => {
+  const [token] = useAtom(tokens);
+  const { mutate } = useFetchRefreshBitToken();
+  return useQuery({
+    queryKey: ["userProfile", token.access_token],
```
baktun14: We have a `queryKeys` object where we set all the react-query keys, please move all keys over there, as well move all the queries in the queries folder.

apps/deploy-web/src/components/remote-deploy/remote-deploy-api-queries/gitlab-queries.ts
```diff
@@ -0,0 +1,208 @@
...
+
+      return response.data;
+    },
+    onSuccess: data => {
+      setToken({
+        access_token: data.access_token,
+        refresh_token: data.refresh_token,
+        type: "gitlab"
+      });
+      router.replace(
```
baktun14: same comment as above

apps/deploy-web/src/components/remote-deploy/update/RemoteDeployUpdate.tsx
```diff
@@ -0,0 +1,129 @@
...
+import { generateSdl } from "@src/utils/sdl/sdlGenerator";
+import { importSimpleSdl } from "@src/utils/sdl/sdlImport";
+import BitBranches from "../bitbucket/BitBucketBranches";
+import GithubBranches from "../github/GithubBranches";
+import GitBranches from "../gitlab/GitlabBranches";
+import { appendEnv, ciCdTemplateId, protectedEnvironmentVariables } from "../helper-functions";
+import Rollback from "./Rollback";
+
+const RemoteDeployUpdate = ({ sdlString, setEditedManifest }: { sdlString: string; setEditedManifest: Dispatch<React.SetStateAction<string | null>> }) => {
+  console.log(sdlString);
```
baktun14: ✂️

apps/deploy-web/src/components/remote-deploy/update/RemoteDeployUpdate.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+  const [token] = useAtom(tokens);
+  const [, setIsInit] = useState(false);
+  const { enqueueSnackbar } = useSnackbar();
+  const [, setError] = useState<string | null>(null);
```
baktun14: same comment as above

apps/deploy-web/src/components/sdl/EnvFormModal.tsx
```diff
@@ -16,9 +17,19 @@ type Props = {
   control: Control<SdlBuilderFormValuesType | RentGpusFormValuesType, any>;
   hasSecretOption?: boolean;
   children?: ReactNode;
+  isRemoteDeployEnvHidden?: boolean;
+  update?: boolean;
```
baktun14:
```suggestion
  isUpdate?: boolean;
```

apps/deploy-web/src/components/sdl/EnvVarList.tsx
```diff
@@ -1,23 +1,27 @@
...
+  isRemoteDeployEnvHidden?: boolean;
 };
 
-export const EnvVarList: React.FunctionComponent<Props> = ({ currentService, setIsEditingEnv, serviceIndex }) => {
+export const EnvVarList: React.FunctionComponent<Props> = ({ currentService, setIsEditingEnv, serviceIndex, isRemoteDeployEnvHidden }) => {
   const { hasComponent } = useSdlBuilder();
+  const currentEnvs = currentService.env?.filter(e => !isRemoteDeployEnvHidden || !(e?.key?.trim() in protectedEnvironmentVariables));
   return (
-    <FormPaper className="whitespace-break-spaces break-all">
+    <FormPaper className={clsx("whitespace-break-spaces break-all", isRemoteDeployEnvHidden && "!bg-card")}>
```
baktun14: we use  `cn` not `clsx`

baktun14: `import { cn } from "@src/utils/styleUtils";`

apps/deploy-web/src/hooks/useRemoteDeployFramework.tsx
```diff
@@ -0,0 +1,57 @@
...
+      setData(data);
+      const cpus = (Object.keys(data?.dependencies ?? {})?.length / 10 / 2)?.toFixed(1);
+
+      setValue("services.0.profile.cpu", +cpus > 2 ? +cpus : 2);
+    } else {
+      setData(null);
+    }
+  };
+
+  const { isLoading } = usePackageJson(setValueHandler, removeInitialUrl(selected), subFolder);
```
baktun14: move all the hooks and state at the top

apps/deploy-web/src/pages/api/bitbucket/refresh.ts
```diff
@@ -0,0 +1,23 @@
+import { NextApiRequest, NextApiResponse } from "next";
+
+import BitbucketAuth from "@src/services/auth/bitbucket.service";
+
+const NEXT_PUBLIC_BITBUCKET_CLIENT_ID: string = process.env.NEXT_PUBLIC_BITBUCKET_CLIENT_ID as string;
+const BITBUCKET_CLIENT_SECRET: string = process.env.BITBUCKET_CLIENT_SECRET as string;
```
baktun14: Same comment as above, everywhere env vars are used.

baktun14 (changes requested):

apps/deploy-web/src/components/remote-deploy/remote-deploy-api-queries/bit-bucket-queries.ts
Chaitanyarai899: Last comment was actaully about apis folder only, we will fix this one asap as well.

apps/deploy-web/src/types/remoteCommits.ts
dharamveergit: I searched for open source type for gitlab and bitbucket but they are not available.

apps/deploy-web/package.json
```diff
@@ -43,6 +43,7 @@
     "@nivo/core": "^0.87.0",
     "@nivo/line": "^0.87.0",
     "@nivo/pie": "^0.87.0",
+    "@octokit/openapi-types": "^22.2.0",
```
ygrishajev: should be dev dep

apps/deploy-web/src/components/remote-deploy/RemoteRepositoryDeployManager.tsx
```diff
@@ -3,69 +3,82 @@ import { Control, UseFormSetValue } from "react-hook-form";
...
   const [hydrated, setHydrated] = useState(false);
+  const isRepoAndBranchPresent = (env: Array<{ key: string }>) =>
+    env.some(e => e.key === protectedEnvironmentVariables.REPO_URL) && env.some(e => e.key === protectedEnvironmentVariables.BRANCH_NAME);
+
+  const isValid = isRepoAndBranchPresent(services?.[0]?.env || []);
+  const isRepoUrlDefault = (env: ServiceType["env"]) => env?.some(e => e.key === protectedEnvironmentVariables.REPO_URL && e.value === DEFAULT_ENV_IN_YML);
+
+  const shouldResetValue = isRepoUrlDefault(services?.[0]?.env || []);
+
+  const envVarUpdater = new EnvVarUpdater(services);
```
baktun14: I'm not sure we need a class here. Or if we do it should use like this  `useMemo(() => new EnvVarUpdate(services), [services])`

apps/deploy-web/src/components/remote-deploy/RemoteRepositoryDeployManager.tsx
```diff
@@ -3,69 +3,82 @@ import { Control, UseFormSetValue } from "react-hook-form";
...
+
+  const { reLoginWithGithub, loginWithGithub } = new GitHubService();
```
baktun14: same thing here

apps/deploy-web/src/components/remote-deploy/Repos.tsx
```diff
@@ -43,48 +45,47 @@ const Repos = ({
...
   const [filteredRepos, setFilteredRepos] = useState(repos);
   const [currentAccount, setCurrentAccount] = useState<string>("");
   const [directory, setDirectory] = useState<IGithubDirectoryItem[] | null>(null);
   const [open, setOpen] = useState(false);
   const [accounts, setAccounts] = useState<string[]>([]);
-  const rootFolder = "akash-root-folder-repo-path";
-  const currentRepo = services?.[0]?.env?.find(e => e.key === protectedEnvironmentVariables.REPO_URL);
-  const repo = repos?.find(r => r.html_url === currentRepo?.value);
-  const currentFolder = services?.[0]?.env?.find(e => e.key === protectedEnvironmentVariables.FRONTEND_FOLDER);
+  const envVarUpdater = new EnvVarUpdater(services);
```
baktun14: `useMemo`

apps/deploy-web/src/components/remote-deploy/deployment-configurations/RemoteBuildInstallConfig.tsx
```diff
@@ -1,15 +1,18 @@
...
+import { EnvVarUpdater } from "@src/services/remote-deploy/remote-deployment-controller.service";
+import { SdlBuilderFormValuesType, ServiceType } from "@src/types";
 import BoxTextInput from "../BoxTextInput";
-import { appendEnv, protectedEnvironmentVariables, ServiceSetValue } from "../helper-functions";
 
-const RemoteBuildInstallConfig = ({ services, setValue }: { services: ServiceType[]; setValue: ServiceSetValue }) => {
+const RemoteBuildInstallConfig = ({ services, setValue }: { services: ServiceType[]; setValue: UseFormSetValue<SdlBuilderFormValuesType> }) => {
   const [expanded, setExpanded] = useState(false);
   const currentService = services[0];
+  const envVarUpdater = new EnvVarUpdater(services);
```
baktun14: `useMemo`

apps/deploy-web/src/components/remote-deploy/update/RemoteDeployUpdate.tsx
```diff
@@ -16,35 +18,32 @@ import { importSimpleSdl } from "@src/utils/sdl/sdlImport";
...
-
+const RemoteDeployUpdate = ({ sdlString, onManifestChange }: { sdlString: string; onManifestChange: (value: string) => void }) => {
   const [token] = useAtom(tokens);
-  const [, setIsInit] = useState(false);
   const { enqueueSnackbar } = useSnackbar();
-  const [, setError] = useState<string | null>(null);
   const [isEditingEnv, setIsEditingEnv] = useState<number | boolean | null>(false);
   const { control, watch, setValue } = useForm<SdlBuilderFormValuesType>({ defaultValues: { services: [defaultService] } });
   const { fields: services } = useFieldArray({ control, name: "services", keyName: "id" });
+  const envVarUpdater = new EnvVarUpdater(services);
```
baktun14: `useMemo`

PR #383, opened 2024-09-24
Improve error handling in api

Redm4x:
- If a query fail in `cacheResponse` and the `keepCache` param was not used, re-throw the exception instead of catching and returning undefined. This prevents a second error from happening further in the flow because of the undefined value.
- Increased the caching time for `provider-attributes.json` from 30s to 5min. Also enabled the `keepCache` param for both queries in `githubService.ts` so that the old cached version is returned if the query fails.
- Use the new logger in the caching helper
- Added parameter validation in the address deployments endpoint

### DIFF
--- a/apps/api/src/caching/helpers.ts
+++ b/apps/api/src/caching/helpers.ts
@@ -1,7 +1,10 @@
 import * as Sentry from "@sentry/node";
 import { differenceInSeconds } from "date-fns";
 
+import { LoggerService } from "@src/core";
 import MemoryCacheEngine from "./memoryCacheEngine";
+
+const logger = new LoggerService({ context: "Caching" });
 
 export const cacheEngine = new MemoryCacheEngine();
 const pendingRequests: { [key: string]: Promise<unknown> } = {};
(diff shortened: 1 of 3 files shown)


apps/api/src/caching/helpers.ts
```diff
@@ -1,8 +1,11 @@
 import * as Sentry from "@sentry/node";
 import { differenceInSeconds } from "date-fns";
 
+import { LoggerService } from "@src/core";
 import MemoryCacheEngine from "./memoryCacheEngine";
 
+const logger = new LoggerService({ context: "CachingHelpers" });
```
ygrishajev: I'm not too hard on this, but eventually this would look better on logs :)

```suggestion
const logger = new LoggerService({ context: "Caching" });
```

apps/api/src/caching/helpers.ts
```diff
@@ -30,20 +33,24 @@ export const Memoize = (options?: MemoizeOptions) => (target: object, propertyNa
 export async function cacheResponse<T>(seconds: number, key: string, refreshRequest: () => Promise<T>, keepData?: boolean): Promise<T> {
   const duration = seconds * 1000;
   const cachedObject = cacheEngine.getFromCache(key) as CachedObject<T> | undefined;
-  // console.log(`Cache key: ${key}`);
+  // logger.info(`Cache key: ${key}`);
```
ygrishajev: why commented? perhaps not needed? You may also use `logger.debug` and uncomment if you want to avoid it in some envs

apps/api/src/routes/v1/addresses/deployments.ts
```diff
@@ -59,15 +60,28 @@ const route = createRoute({
...
       }
+    },
+    400: {
+      description: "Invalid address"
     }
   }
 });
 
 export default new OpenAPIHono().openapi(route, async c => {
+  if (!isValidBech32Address(c.req.valid("param").address, "akash")) {
```
ygrishajev, line 71: I started using http errors in other places and those are handled by a global error handler. It would also look a bit better generally:

```typescript
import assert from "http-assert";

assert(isValidBech32Address(c.req.valid("param").address, "Akash"), 400, 'Invalid address)
```

apps/api/src/services/external/githubService.ts
```diff
@@ -22,19 +22,25 @@ export function getOctokit() {
 export const getProviderAttributesSchema = async (): Promise<ProviderAttributesSchema> => {
   // Fetching provider attributes schema
   const response = await cacheResponse(
-    30,
+    60 * 5,
```
ygrishajev:
```typescript
import { secondsInMinute } from "date-fns/constants";

5 * secondsInMinute
```

ygrishajev (approved):
all the comments are nice to have from my perspective

apps/api/src/caching/helpers.ts
Redm4x: Yeah I think that was to quickly uncomment when debugging. I'll switch to using `logger.debug`.

apps/api/src/services/external/githubService.ts
Redm4x: I switched to using `date-fns/minutesToSeconds`

baktun14 (approved):

apps/api/src/routes/v1/addresses/deployments.ts
Redm4x: Wont do it in this PR, but yeah we could change all endpoints to using `http-assert` in a separate PR

Issue #330, opened 2024-08-22
Improve stats app resiliency


Redm4x:
# Improve stats app resiliency

## Problem

Most of the [stats app](https://stats.akash.network/) data come from the `/v1/dashboard-data` endpoint. This endpoint poorly handle exceptions so if any error occurs, the entire dashboard breaks. In this specific case, only the "Bonded token" loading failed, but all the tiles are missing.

![image](https://github.com/user-attachments/assets/8b426e17-5143-4f42-9ad3-3ff7927cc614)

## Proposed solution

We need to make sure that the endpoint is resilient and won't break if some of the data loading fail:
- [ ] In the `/v1/dashboard-data` endpoint. Make sure that if anything fail to load, the rest is still returned
- [ ] In the `getChainStats` function. Make sure that if a query fail, the others are still returned
- [ ] In the stats app. Make sure that if a row or tile data is missing from the api response, the others are still rendered

forbesus:
Hello @Redm4x I can handle this task

baktun14:
Hey @forbesus are you still up for this task?

forbesus:
Sure @baktun14

forbesus:
Hello @baktun14 Will you assign this task to me?

baktun14:
@forbesus Do you need help with something?

forbesus:
Can we chat via discord for smoothly communication? @baktun14 

My discord name in community is `analos.xyz`

baktun14:
@forbesus Ok I'll add you

PR #385, opened 2024-09-25
fix(stats): improve stats app resiliency

forbesus:
ref #330

### DIFF
--- a/apps/api/src/routes/v1/dashboardData.ts
+++ b/apps/api/src/routes/v1/dashboardData.ts
@@ -1,10 +1,16 @@
 import { createRoute, OpenAPIHono, z } from "@hono/zod-openapi";
 
+import { LoggerService } from "@src/core/services/logger/logger.service";
 import { getBlocks } from "@src/services/db/blocksService";
 import { getNetworkCapacity } from "@src/services/db/providerStatusService";
 import { getDashboardData, getProviderGraphData } from "@src/services/db/statsService";
 import { getTransactions } from "@src/services/db/transactionsService";
 import { getChainStats } from "@src/services/external/apiNodeService";
+import { createLoggingExecutor } from "@src/utils/logging";
+
+
+const logger = new LoggerService({ context: "Dashboard" });
+const runOrLog = createLoggingExecutor(logger)
 
 const route = createRoute({
   method: "get",
@@ -144,22 +150,30 @@ const route = createRoute({
 });
 
 export default new OpenAPIHono().openapi(route, async c => {
-  const chainStatsQuery = await getChainStats();
-  const dashboardData = await getDashboardData();
-  const networkCapacity = await getNetworkCapacity();
-  const networkCapacityStats = await getProviderGraphData("count");
-  const latestBlocks = await getBlocks(5);
-  const latestTransactions = await getTransactions(5);
-
+  const [{ now, compare }, chainStatsQuery, networkCapacity, networkCapacityStats, latestBlocks, latestTransactions] = await Promise.all([
+    runOrLog(getDashboardData),
+    runOrLog(getChainStats, {
+      bondedTokens: undefined,
+      totalSupply: undefined,
+      communityPool: undefined,
+      inflation: undefined,
+      stakingAPR: undefined
+    }),
+    runOrLog(getNetworkCapacity),
+    runOrLog(() => getProviderGraphData("count")),
+    runOrLog(() => getBlocks(5)),
+    runOrLog(() => getTransactions(5))
+  ]);
   const chainStats = {
-    height: latestBlocks[0].height,
-    transactionCount: latestBlocks[0].totalTransactionCount,
-    ...chainStatsQuery
-  };
+    ...chainStatsQuery,
+    height: latestBlocks && latestBlocks.length > 0 ? latestBlocks[0].height : undefined,
+    transactionCount: latestBlocks && latestBlocks.length > 0 ? latestBlocks[0].totalTransactionCount : undefined,
+  }
 
   return c.json({
     chainStats,
-    ...dashboardData,
+    now,
+    compare,
     networkCapacity,
     networkCapacityStats,
     latestBlocks,
--- a/apps/api/src/services/external/apiNodeService.ts
+++ b/apps/api/src/services/external/apiNodeService.ts
@@ -6,6 +6,7 @@ import fetch from "node-fetch";
 import { Op } from "sequelize";
 
 import { cacheKeys, cacheResponse } from "@src/caching/helpers";
+import { LoggerService } from "@src/core/services/logger/logger.service";
 import { getTransactionByAddress } from "@src/services/db/transactionsService";
 import {
   CosmosGovProposalResponse,
@@ -28,45 +29,87 @@ import { CosmosMintInflationResponse } from "@src/types/rest/cosmosMintInflation
 import { CosmosStakingPoolResponse } from "@src/types/rest/cosmosStakingPoolResponse";
 import { coinToAsset } from "@src/utils/coin";
 import { apiNodeUrl, averageBlockCountInAMonth, betaTypeVersion, betaTypeVersionMarket } from "@src/utils/constants";
+import { createLoggingExecutor } from "@src/utils/logging";
 import { getDeploymentRelatedMessages } from "../db/deploymentService";
 import { getProviderList } from "../db/providerStatusService";
 
 export async function getChainStats() {
+  const logger = new LoggerService({ context: "ApiNode" })
+  const runOrLog = createLoggingExecutor(logger)
+
   const result = await cacheResponse(
     60 * 5, // 5 minutes
     cacheKeys.getChainStats,
     async () => {
-      const bondedTokensQuery = axios.get<CosmosStakingPoolResponse>(`${apiNodeUrl}/cosmos/staking/v1beta1/pool`);
-      const supplyQuery = axios.get<CosmosBankSupplyResponse>(`${apiNodeUrl}/cosmos/bank/v1beta1/supply?pagination.limit=1000`);
-      const communityPoolQuery = axios.get<CosmosDistributionCommunityPoolResponse>(`${apiNodeUrl}/cosmos/distribution/v1beta1/community_pool`);
-      const inflationQuery = axios.get<CosmosMintInflationResponse>(`${apiNodeUrl}/cosmos/mint/v1beta1/inflation`);
-      const distributionQuery = axios.get<CosmosDistributionParamsResponse>(`${apiNodeUrl}/cosmos/distribution/v1beta1/params`);
+      const bondedTokensAsPromised = await runOrLog(async () => {
+        const bondedTokensQuery = await axios.get<CosmosStakingPoolResponse>(
+          `${apiNodeUrl}/cosmos/staking/v1beta1/pool`
+        );
+        return parseInt(bondedTokensQuery.data.pool.bonded_tokens);
+      });
 
-      const [bondedTokensResponse, supplyResponse, communityPoolResponse, inflationResponse, distributionResponse] = await Promise.all([
-        bondedTokensQuery,
-        supplyQuery,
-        communityPoolQuery,
-        inflationQuery,
-        distributionQuery
+      const totalSupplyAsPromised = await runOrLog(async () => {
+        const supplyQuery = await axios.get<CosmosBankSupplyResponse>(
+          `${apiNodeUrl}/cosmos/bank/v1beta1/supply?pagination.limit=1000`
+        );
+        return parseInt(
+          supplyQuery.data.supply.find((x) => x.denom === "uakt")?.amount || "0"
+        );
+      });
+
+      const communityPoolAsPromised = await runOrLog(async () => {
+        const communityPoolQuery = await axios.get<CosmosDistributionCommunityPoolResponse>(
+          `${apiNodeUrl}/cosmos/distribution/v1beta1/community_pool`
+        );
+        return parseFloat(
+          communityPoolQuery.data.pool.find((x) => x.denom === "uakt")?.amount || "0"
+        );
+      });
+
+      const inflationAsPromised = await runOrLog(async () => {
+        const inflationQuery = await axios.get<CosmosMintInflationResponse>(
+          `${apiNodeUrl}/cosmos/mint/v1beta1/inflation`
+        );
+        return parseFloat(inflationQuery.data.inflation || "0");
+      });
+
+      const communityTaxAsPromised = await runOrLog(async () => {
+        const distributionQuery = await axios.get<CosmosDistributionParamsResponse>(
+          `${apiNodeUrl}/cosmos/distribution/v1beta1/params`
+        );
+        return parseFloat(distributionQuery.data.params.community_tax || "0");
+      });
+
+      const [bondedTokens, totalSupply, communityPool, inflation, communityTax] = await Promise.all([
+        bondedTokensAsPromised,
+        totalSupplyAsPromised,
+        communityPoolAsPromised,
+        inflationAsPromised,
+        communityTaxAsPromised
       ]);
 
       return {
-        communityPool: parseFloat(communityPoolResponse.data.pool.find(x => x.denom === "uakt").amount),
-        inflation: parseFloat(inflationResponse.data.inflation),
-        communityTax: parseFloat(distributionResponse.data.params.community_tax),
-        bondedTokens: parseInt(bondedTokensResponse.data.pool.bonded_tokens),
-        totalSupply: parseInt(supplyResponse.data.supply.find(x => x.denom === "uakt").amount)
+        communityPool,
+        inflation,
+        communityTax,
+        bondedTokens,
+        totalSupply,
       };
     },
     true
   );
+
+  let stakingAPR: number | undefined;
+  if (result.bondedTokens && result.bondedTokens > 0 && result.inflation && result.communityTax && result.totalSupply) {
+    stakingAPR = result.inflation * (1 - result.communityTax) * result.totalSupply / result.bondedTokens
+  }
 
   return {
     bondedTokens: result.bondedTokens,
     totalSupply: result.totalSupply,
     communityPool: result.communityPool,
     inflation: result.inflation,
-    stakingAPR: (result.inflation * (1 - result.communityTax)) / (result.bondedTokens / result.totalSupply)
+    stakingAPR,
   };
 }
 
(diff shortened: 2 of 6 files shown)


forbesus:
Hello @baktun14 would you check it?

apps/api/src/services/db/statsService.ts
```diff
@@ -15,66 +15,107 @@ type GraphData = {
...
+        totalUUsdSpent: compareBlockStats.totalUUsdSpent,
+        dailyUUsdSpent: compareBlockStats.totalUUsdSpent - secondCompareBlockStats.totalUUsdSpent,
+        activeCPU: compareBlockStats.activeCPU,
+        activeGPU: compareBlockStats.activeGPU,
+        activeMemory: compareBlockStats.activeMemory,
+        activeStorage: compareBlockStats.activeEphemeralStorage + compareBlockStats.activePersistentStorage
+      }
+    };
+  }
+  catch (error) {
```
Redm4x: I think there is no need to try/catch in here. We do not want to show 0 everywhere as it would cause confusion. If an exception is thrown in `getDashboardData` we should just let it reach the catch block in `dashboardData.ts` and it will be handled there correctly.

apps/api/src/services/external/apiNodeService.ts
```diff
@@ -36,26 +36,68 @@ export async function getChainStats() {
...
-        communityPoolQuery,
-        inflationQuery,
-        distributionQuery
-      ]);
+      // Define default values in case requests fail
+      let bondedTokens = null;
+      let totalSupply = null;
+      let communityPool = null;
+      let inflation = null;
+      let communityTax = null;
```
Redm4x: In the frontend (`Dashboard.tsx`) the tiles are only hidden if those values are `undefined`. Since they default to `null` here that means if some of them fail to load the tiles are still shown with values of 0.

apps/api/src/routes/v1/dashboardData.ts
```diff
@@ -144,22 +144,65 @@ const route = createRoute({
...
+  let networkCapacityStats;
+  let latestBlocks;
+  let latestTransactions;
+
+  try {
+    const chainStatsQuery = await getChainStats();
+    chainStats = { ...chainStatsQuery };
+  } catch (error) {
+    chainStats = {}
+    console.error("Failed to fetch getChainStats:", error);
```
Redm4x: Please use the new logger instead of `console.error`. You can instantiate it at the top of the file like so:
`const logger = new LoggerService({ context: "DashboardDataEndpoint" });`

Redm4x (changes requested):

apps/api/src/routes/v1/dashboardData.ts
Redm4x: It would also be a good idea to send those exceptions to Sentry.
```
import * as Sentry from "@sentry/node";
```
```
Sentry.captureException(error);
```

apps/api/src/services/db/statsService.ts
```diff
@@ -15,66 +15,107 @@ type GraphData = {
...
+    const secondCompareDate = subHours(latestBlockStats.datetime, 48);
+    const secondCompareBlockStats = await Block.findOne({
+      order: [["datetime", "ASC"]],
+      where: {
+        datetime: { [Op.gte]: secondCompareDate }
+      }
+    });
+  
+    return {
+      now: {
```
ygrishajev: I think props `prev`/`next`, `prev`/`current` would better describe the contents. `compare` makes me wonder what that might be, it's also a verb which complicates it more. `now` sounds like a timestamp value, sounds confusing.

apps/api/src/routes/v1/dashboardData.ts
```diff
@@ -144,22 +144,65 @@ const route = createRoute({
...
+  }
+
+  try {
```
ygrishajev: what would you say if we added smth like this:


```typescript
async function runOrLog<T>(cb: () => Promise<T>, defaultValue: T): Promise<T> {
// or async function runOrLog<T>(cb: () => Promise<T>, defaultValue: T): Promise<T> {
  try {
    return await cb();
  } catch (error) {
    console.error("Error occurred:", error);
    // or report to sentry

    return defaultValue;
  }
}

// to later be used like
const { now, compare } = await runOrLog(getDashboardData, {})
const networkCapacityStats = await runOrLog(getProviderGraphData("count"), 0);
```

it would clean the code up quite a bit

apps/api/src/routes/v1/dashboardData.ts
```diff
@@ -144,22 +144,65 @@ const route = createRoute({
...
-  const networkCapacityStats = await getProviderGraphData("count");
-  const latestBlocks = await getBlocks(5);
-  const latestTransactions = await getTransactions(5);
 
-  const chainStats = {
-    height: latestBlocks[0].height,
-    transactionCount: latestBlocks[0].totalTransactionCount,
-    ...chainStatsQuery
-  };
+  let chainStats;
```
ygrishajev: if you stick to this kind of declarations, please type them to avoid possible issues in the future

apps/api/src/services/external/apiNodeService.ts
```diff
@@ -36,26 +36,68 @@ export async function getChainStats() {
...
-      const distributionQuery = axios.get<CosmosDistributionParamsResponse>(`${apiNodeUrl}/cosmos/distribution/v1beta1/params`);
-
-      const [bondedTokensResponse, supplyResponse, communityPoolResponse, inflationResponse, distributionResponse] = await Promise.all([
-        bondedTokensQuery,
-        supplyQuery,
-        communityPoolQuery,
-        inflationQuery,
-        distributionQuery
-      ]);
+      // Define default values in case requests fail
```
ygrishajev: the purpose of default declarations is kinda obvious
```suggestion
```

apps/api/src/services/external/apiNodeService.ts
```diff
@@ -36,26 +36,68 @@ export async function getChainStats() {
...
+      let bondedTokens = null;
```
ygrishajev: these vars should be typed according to what would be assigned further

apps/stats-web/src/app/(home)/DashboardContainer.tsx
```diff
@@ -23,7 +23,7 @@ export const DashboardContainer: React.FunctionComponent = () => {
 
           <div className="mb-8">
             <p className="text-italic text-sm italic text-muted-foreground">
-              Last updated: <FormattedDate value={dashboardData.now.date} /> <FormattedTime value={dashboardData.now.date} />
+              Last updated: {dashboardData?.now?.date !== undefined && <FormattedDate value={dashboardData.now.date} />} {dashboardData?.now?.date !== undefined && <FormattedTime value={dashboardData.now.date} />}
```
ygrishajev:
```suggestion
              Last updated: {!!dashboardData?.now?.date && <FormattedDate value={dashboardData.now.date} />} {!!dashboardData?.now?.date && <FormattedTime value={dashboardData.now.date} />}
```

apps/api/src/services/db/statsService.ts
Redm4x: I agree `prev`/`current` would be better names. This is part of the public api however, maybe we should hold off renaming those if it's not needed in case somebody is using it already? At least until we release a v2 of the api.

ygrishajev: Yea, agreed. This would be a breaking change, should be versioned. I missed these props were here before. Scratch my comment )

ygrishajev: I missed these props were here before. Scratch my comment )

forbesus: I agree with @Redm4x 
But if there is no one who are using `now`/`compare`, I will change

ygrishajev: We can't know for sure. This kind of changes should be versioned

forbesus:
Hello @Redm4x  @ygrishajev 
I finished to fix
Would you check again?

apps/api/src/routes/v1/dashboardData.ts
```diff
@@ -1,11 +1,16 @@
...
+import { getSentry } from "@src/core/providers/sentry.provider";
+import { LoggerService } from "@src/core/services/logger/logger.service";
 import { getBlocks } from "@src/services/db/blocksService";
 import { getNetworkCapacity } from "@src/services/db/providerStatusService";
 import { getDashboardData, getProviderGraphData } from "@src/services/db/statsService";
 import { getTransactions } from "@src/services/db/transactionsService";
 import { getChainStats } from "@src/services/external/apiNodeService";
 
+
+const logger = new LoggerService({ context: "DashboardDataEndpoint" });
```
ygrishajev:
```suggestion
const logger = new LoggerService({ context: "Dashboard" });
```

apps/api/src/services/external/apiNodeService.ts
```diff
@@ -36,26 +36,67 @@ export async function getChainStats() {
...
+      let communityTax;
+
+      try {
+        const bondedTokensQuery = await axios.get<CosmosStakingPoolResponse>(
+          `${apiNodeUrl}/cosmos/staking/v1beta1/pool`
+        );
+        bondedTokens = parseInt(bondedTokensQuery.data.pool.bonded_tokens);
+      } catch (error) {
+        console.error("Failed to fetch bonded tokens:", error);
+      }
```
ygrishajev: this call and below is really the same as in another file. Logger can be a parameter.

```typescript
const createLoggingExecutor = (logger: LoggerService) => async <T>(cb: () => Promise<T>, defaultValue?: T): Promise<T> => {
   try {
     return await cb();
   } catch (error) {
     logger.error({ event: `Failed to fetch ${cb.name}`, error });
     getSentry().captureException(error);
     return defaultValue;
   }
}

const logger = new LoggerService()
const runOrLog = createLoggingExecutor(logger)

const bondedTokens = await runOrLog(() => {
   const bondedTokensQuery = await axios.get<CosmosStakingPoolResponse>(
      `${apiNodeUrl}/cosmos/staking/v1beta1/pool`
   );
   return parseInt(bondedTokensQuery.data.pool.bonded_tokens);
})

// or
const bondedTokens = await runOrLog(
   () => axios.get<CosmosStakingPoolResponse>(`${apiNodeUrl}/cosmos/staking/v1beta1/pool`)
      .then(({ data }) => parseInt(data.pool.bonded_tokens))
)
```

apps/api/src/services/external/apiNodeService.ts
```diff
@@ -66,7 +107,13 @@ export async function getChainStats() {
...
     communityPool: result.communityPool,
     inflation: result.inflation,
-    stakingAPR: (result.inflation * (1 - result.communityTax)) / (result.bondedTokens / result.totalSupply)
+    stakingAPR:
+      result.bondedTokens && result.inflation && result.communityTax && result.totalSupply
+        ? (result.bondedTokens > 0
+          ? (result.inflation * (1 - result.communityTax)) /
+          (result.bondedTokens / result.totalSupply)
+          : null)
+        : undefined,
```
ygrishajev: this nested ternaries are really hard to understand. Could you please organize this into a flow with a couple of descriptive vars?

forbesus:
Hello @ygrishajev I just fixed 
Would you check again?

apps/api/src/services/external/apiNodeService.ts
```diff
@@ -32,41 +34,86 @@ import { getDeploymentRelatedMessages } from "../db/deploymentService";
 import { getProviderList } from "../db/providerStatusService";
 
 export async function getChainStats() {
+  const createLoggingExecutor = (logger: LoggerService) => async <T>(cb: () => Promise<T>, defaultValue?: T): Promise<T> => {
```
ygrishajev: what I meant is to extract this function into a dedicated module/file/service and then use it here and in another file

apps/api/src/services/external/apiNodeService.ts
```diff
@@ -32,41 +34,86 @@ import { getDeploymentRelatedMessages } from "../db/deploymentService";
...
+    }
+  }
+
   return {
     bondedTokens: result.bondedTokens,
     totalSupply: result.totalSupply,
     communityPool: result.communityPool,
     inflation: result.inflation,
-    stakingAPR: (result.inflation * (1 - result.communityTax)) / (result.bondedTokens / result.totalSupply)
+    stakingAPR: stakingAPR,
```
ygrishajev:
```suggestion
    stakingAPR,
```

apps/api/src/services/external/apiNodeService.ts
```diff
@@ -32,41 +34,86 @@ import { getDeploymentRelatedMessages } from "../db/deploymentService";
...
+        inflation,
+        communityTax,
+        bondedTokens,
+        totalSupply,
       };
     },
     true
   );
 
+  let stakingAPR;
```
ygrishajev:
```suggestion
  let stakingAPR: number | undefined;
```

apps/api/src/services/external/apiNodeService.ts
```diff
@@ -32,41 +34,86 @@ import { getDeploymentRelatedMessages } from "../db/deploymentService";
...
+  if (result.bondedTokens && result.inflation && result.communityTax && result.totalSupply) {
```
ygrishajev:
```suggestion
  if (result.bondedTokens && result.bondedTokens > 0 && result.inflation && result.communityTax && result.totalSupply) {
```

apps/api/src/services/external/apiNodeService.ts
```diff
@@ -32,41 +34,86 @@ import { getDeploymentRelatedMessages } from "../db/deploymentService";
...
+    if (result.bondedTokens > 0) {
```
ygrishajev:
```suggestion
```

forbesus:
Hello @ygrishajev  How about now?

apps/api/src/routes/v1/dashboardData.ts
```diff
@@ -144,25 +150,34 @@ const route = createRoute({
...
   const chainStats = {
-    height: latestBlocks[0].height,
-    transactionCount: latestBlocks[0].totalTransactionCount,
-    ...chainStatsQuery
-  };
+    ...chainStatsQuery,
+    height: latestBlocks && latestBlocks.length > 0 ? latestBlocks[0].height : undefined,
+    transactionCount: latestBlocks && latestBlocks.length > 0 ? latestBlocks[0].totalTransactionCount : undefined,
+  }
+  const latestTransactions = await runOrLog(() => getTransactions(5));
```
ygrishajev: what do you think about one last improvement here? Promise.all should speed up this endpoint quite a bit

```typescript
const [{ now, compare }, chainStatsQuery, networkCapacity, networkCapacityStats, latestBlocks, latestTransactions] = await Promise.all([
    runOrLog(getDashboardData),
    runOrLog(getChainStats, {
      bondedTokens: undefined,
      totalSupply: undefined,
      communityPool: undefined,
      inflation: undefined,
      stakingAPR: undefined
    }),
    runOrLog(getNetworkCapacity),
    runOrLog(() => getProviderGraphData("count")),
    runOrLog(() => getBlocks(5)),
    runOrLog(() => getTransactions(5))
  ]);

  const chainStats = {
    ...chainStatsQuery,
    height: latestBlocks && latestBlocks.length > 0 ? latestBlocks[0].height : undefined,
    transactionCount: latestBlocks && latestBlocks.length > 0 ? latestBlocks[0].totalTransactionCount : undefined
  };
```

apps/api/src/services/external/apiNodeService.ts
```diff
@@ -28,45 +29,79 @@ import { CosmosMintInflationResponse } from "@src/types/rest/cosmosMintInflation
...
-      const distributionQuery = axios.get<CosmosDistributionParamsResponse>(`${apiNodeUrl}/cosmos/distribution/v1beta1/params`);
-
-      const [bondedTokensResponse, supplyResponse, communityPoolResponse, inflationResponse, distributionResponse] = await Promise.all([
-        bondedTokensQuery,
-        supplyQuery,
-        communityPoolQuery,
-        inflationQuery,
-        distributionQuery
-      ]);
+      const bondedTokens = await runOrLog(async () => {
```
ygrishajev: Same here, promise.all should be of good help

```typescript
const bondedTokensAsPromised = runOrLog(async () => {
        const bondedTokensQuery = await axios.get<CosmosStakingPoolResponse>(`${apiNodeUrl}/cosmos/staking/v1beta1/pool`);
        return parseInt(bondedTokensQuery.data.pool.bonded_tokens);
      });

      const totalSupplyAsPromised = runOrLog(async () => {
        const supplyQuery = await axios.get<CosmosBankSupplyResponse>(`${apiNodeUrl}/cosmos/bank/v1beta1/supply?pagination.limit=1000`);
        return parseInt(supplyQuery.data.supply.find(x => x.denom === "uakt")?.amount || "0");
      });

      const communityPoolAsPromised = runOrLog(async () => {
        const communityPoolQuery = await axios.get<CosmosDistributionCommunityPoolResponse>(`${apiNodeUrl}/cosmos/distribution/v1beta1/community_pool`);
        return parseFloat(communityPoolQuery.data.pool.find(x => x.denom === "uakt")?.amount || "0");
      });

      const inflationAsPromised = runOrLog(async () => {
        const inflationQuery = await axios.get<CosmosMintInflationResponse>(`${apiNodeUrl}/cosmos/mint/v1beta1/inflation`);
        return parseFloat(inflationQuery.data.inflation || "0");
      });

      const communityTaxAsPromised = runOrLog(async () => {
        const distributionQuery = await axios.get<CosmosDistributionParamsResponse>(`${apiNodeUrl}/cosmos/distribution/v1beta1/params`);
        return parseFloat(distributionQuery.data.params.community_tax || "0");
      });

      const [bondedTokens, totalSupply, communityPool, inflation, communityTax] = await Promise.all([
        bondedTokensAsPromised,
        totalSupplyAsPromised,
        communityPoolAsPromised,
        inflationAsPromised,
        communityTaxAsPromised
      ]);

      return {
        communityPool,
        inflation,
        communityTax,
        bondedTokens,
        totalSupply
      };
```

ygrishajev (approved):
Thanks for all your changes @forbesus! This looks so much better!

I'm adding a couple of comments, that you may improve further if you like. Anyways lgtm.
Let's please make sure it's also good with @Redm4x

ygrishajev:
oh, 1 thing before we can merge. Please squash the commits and use a scope for the resulting commits. e.g.

```
feat(stats): ...message
```

This should fail on pre-commit hook coz this scope is not added to our [rules](https://github.com/akash-network/console/blob/73ff83843947c16a75085b7d7e2ddcead49bc67a/.commitlintrc.json). Please feel free to add it.

@Redm4x does this scope make sense to you?

forbesus:
All things are finished @Redm4x  cc: @ygrishajev

ygrishajev:
We follow rebase strategy, can you squash the merge commit too pls?

forbesus:
How about now? @ygrishajev  cc: @Redm4x

ygrishajev (approved):
Lgtm! Thanks for all your changes!

Redm4x (approved):

PR #392, opened 2024-09-30
test(deployment): e2e testing create deployment

baktun14:
- Added Leap wallet extension for integration tests
- Added a test that deploys hello world app and closes it
---
You can ignore all the files under the Leap folder, which is basically all the files downloaded from the extension.


apps/deploy-web/tests/deploy-hello-world.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+
+// test.describe.configure({ mode: "serial" });
+
+test.only("deploy hello world", async ({ extPage: page, context }) => {
+  await setupLeap(context, page);
+
+  const customTemplatePage = new DeployHelloWorldPage(context, page, "new-deployment", "hello-world-card");
+
+  // Create deployment
+  await customTemplatePage.gotoInteractive();
```
baktun14: This could probably go in the `DeployHelloWorldPage` class under a single function `deploy()` wdyt? @ygrishajev

apps/deploy-web/playwright.config.ts
```diff
@@ -36,17 +33,17 @@ export default defineConfig({
...
+    // },
 
-    {
-      name: "webkit",
-      use: { ...devices["Desktop Safari"] }
-    }
+    // {
+    //   name: "webkit",
+    //   use: { ...devices["Desktop Safari"] }
+    // }
```
ygrishajev: ✂️

apps/deploy-web/src/components/new-deployment/BidRow.tsx
```diff
@@ -21,6 +21,7 @@ import { PriceEstimateTooltip } from "../shared/PriceEstimateTooltip";
 import { PricePerMonth } from "../shared/PricePerMonth";
 
 type Props = {
+  index: number;
```
ygrishajev: what would you say about this?
```suggestion
  testIndex: number;
```

apps/deploy-web/tests/deploy-hello-world.spec.ts
```diff
@@ -0,0 +1,27 @@
+import { DeployHelloWorldPage } from "./pages/DeployHelloWorldPage";
+import { test } from "./utils/fixture";
+import { setupLeap } from "./utils/wallet";
+
+// test.describe.configure({ mode: "serial" });
```
ygrishajev: ?

apps/deploy-web/tests/deploy-hello-world.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+
+test.only("deploy hello world", async ({ extPage: page, context }) => {
```
ygrishajev: is `only` needed?

```suggestion
test("deploy hello world", async ({ extPage: page, context }) => {
```

apps/deploy-web/tests/deploy-hello-world.spec.ts
ygrishajev: It can be done for sure. The comments and grouping of the method calls below also asks for more methods)) e.g.

```typescript
async validateLeaseAndClose() {
   await customTemplatePage.validateLease();
   await customTemplatePage.closeDeploymentDetail();
   await customTemplatePage.signTransaction();
}
```

apps/deploy-web/tests/pages/DeployBasePage.tsx
```diff
@@ -13,8 +14,8 @@ export class DeployBasePage {
 
   async gotoInteractive() {
     if (this.cardTestId) {
-      await this.page.goto("http://localhost:3000");
-      await this.page.getByTestId("welcome-modal-accept-button").click();
+      // await this.page.goto("http://localhost:3000");
+      // await this.page.getByTestId("welcome-modal-accept-button").click();
```
ygrishajev: ?

apps/deploy-web/tests/pages/DeployBasePage.tsx
```diff
@@ -30,7 +31,33 @@ export class DeployBasePage {
...
+  }
+
+  async createLease() {
+    await this.page.getByTestId("create-lease-filter-audited").click();
+    await this.page.getByTestId("bid-list-row-radio-0").click();
+    await this.page.getByTestId("create-lease-button").click();
+  }
+
+  async validateLease() {
+    await this.page.waitForURL(/http:\/\/localhost:3000\/deployments\/\d+/);
```
ygrishajev: should port be an env var?

ygrishajev (commented):
Why so much leap files are needed there?
Can we rename utils into fixtures? I'd avoid names like `utils`/`helpers` as it's generally a white noise that would cause anyone to still inspect what's inside.

baktun14:
> Why so much leap files are needed there? Can we rename utils into fixtures? I'd avoid names like `utils`/`helpers` as it's generally a white noise that would cause anyone to still inspect what's inside.

It's basically the extension code that we need to run it.

apps/deploy-web/tests/pages/DeployBasePage.tsx
baktun14: I wasn't sure about leaving it there or not, because this is done in the leap setup. However, we might need it for other tests that don't need a wallet.

apps/deploy-web/tests/pages/DeployBasePage.tsx
baktun14: actually yea the url should be an env var if we run the tests on staging or other live env.

apps/deploy-web/tests/deploy-hello-world.spec.ts
baktun14: Left this when I was doing some testing, I'll remove it.

apps/deploy-web/tests/deploy-hello-world.spec.ts
baktun14: Forgot about removing it. 👍

apps/deploy-web/playwright.config.ts
baktun14: Yea, I'll remove it because chrome extensions don't work with firefox or safari.

apps/deploy-web/src/components/new-deployment/BidRow.tsx
baktun14: Sure

apps/deploy-web/tests/fixture/test-env.config.ts
```diff
@@ -0,0 +1,11 @@
+import { z } from "zod";
+
+export const testEnvSchema = z.object({
+  BASE_URL: z.string().default("http://localhost:3000"),
+  TEST_WALLET_MNEMONIC: z.string()
+});
+
+export const testEnvConfig = testEnvSchema.parse({
```
ygrishajev, line 8: I think this is running in node env, so you can safely do
```suggestion
export const testEnvConfig = testEnvSchema.parse(process.env)
```

apps/deploy-web/tests/fixture/test-env.config.ts
ygrishajev, line 1: this is not really a fixture, right?) I'd suggest we treat these tests as an app, it will quite often replicate one with page objects. If so, we'd better do same way as we do in the app. E.g. in this case, we'd create a dir `config` and a file `env.config.ts` in it.

apps/deploy-web/tests/fixture/wallet-setup.ts
```diff
@@ -8,7 +10,7 @@ export const setupLeap = async (context: BrowserContext, page: Page) => {
 
   await page.getByTestId("import-seed-phrase").click();
 
-  const mnemonic = process.env.TEST_WALLET_MNEMONIC;
+  const mnemonic = testEnvConfig.TEST_WALLET_MNEMONIC;
   if (!mnemonic) {
```
ygrishajev, line 14: I guess this check is redundant now as the schema would throw

apps/deploy-web/tests/pages/DeployBasePage.tsx
ygrishajev: so perhaps we can run it conditionally based on the args?

```typescript
async gotoInteractive(options?: { withLeap: boolean }) {
    if (!this.cardTestId) {
        return
    }

    if (options?.withLeap)
         await this.page.goto("http://localhost:3000");
         await this.page.getByTestId("welcome-modal-accept-button").click();
    }
      await this.page.getByTestId("sidebar-deploy-button").first().click();
      await this.page.getByTestId(this.cardTestId).click();
    }
}
```

apps/deploy-web/src/components/deployments/DeploymentDetail.tsx
```diff
@@ -164,11 +164,27 @@ export function DeploymentDetail({ dseq }: React.PropsWithChildren<{ dseq: strin
...
+              <TabsTrigger value="LEASES" data-testid="deployment-tab-leases">
+                Leases
+              </TabsTrigger>
+              {isActive && (
+                <TabsTrigger value="LOGS" data-testid="deployment-tab-logs">
+                  Logs
+                </TabsTrigger>
+              )}
+              {isActive && (
+                <TabsTrigger value="SHELL" data-testid="deployment-tab-events">
```
Redm4x:
```suggestion
                <TabsTrigger value="SHELL" data-testid="deployment-tab-shell">
```

Redm4x (approved):

PR #396, opened 2024-10-03
Feature/config

ygrishajev:
PR commits are describing what's done. It's also a step-by-step guide to apply to provider-web


apps/stats-web/src/app/addresses/[address]/deployments/[dseq]/page.tsx
```diff
@@ -39,8 +48,12 @@ async function fetchDeploymentData(address: string, dseq: string, network: strin
   return response.json();
 }
 
-export default async function DeploymentDetailPage({ params: { address, dseq }, searchParams: { network } }: IProps) {
-  const deployment = await fetchDeploymentData(address, dseq, network as string);
+export default async function DeploymentDetailPage(props: DeploymentDetailPageProps) {
+  const {
+    params: { address, dseq },
+    searchParams: { network }
+  } = DeploymentDetailPageSchema.parse(props);
```
Comment (baktun14) [inline apps/stats-web/src/app/addresses/[address]/deployments/[dseq]/page.tsx:55]: This will throw an error if the props are not correct?

baktun14 (approved):

apps/stats-web/src/app/addresses/[address]/deployments/[dseq]/page.tsx
Comment (ygrishajev) [reply apps/stats-web/src/app/addresses/[address]/deployments/[dseq]/page.tsx:55]: yea

Comment (baktun14) [reply apps/stats-web/src/app/addresses/[address]/deployments/[dseq]/page.tsx:55]: I think network id should have a default value of mainnet.

Comment (ygrishajev) [reply apps/stats-web/src/app/addresses/[address]/deployments/[dseq]/page.tsx:55]: so let's imagine someone implements a link to this page on the frontend. The one may notice that it works without the network params and not bother why it's needed. Thus we may get a bug when a mainnet link is opened from, say, staging.
I'd say it should fail in dev.

Comment (baktun14) [reply apps/stats-web/src/app/addresses/[address]/deployments/[dseq]/page.tsx:55]: I suppose yea, we do append the network param to the path everywhere is use it so it should be fine.

PR #423, opened 2024-10-29
fix(wallet): ensure proper network and manage wallet switch

ygrishajev:


apps/deploy-web/src/context/WalletProvider/WalletProvider.tsx
```diff
@@ -102,12 +103,6 @@ export const WalletProvider = ({ children }) => {
     });
   }, [addEndpoints, settings.apiEndpoint, settings.rpcEndpoint]);
 
-  useEffect(() => {
```
ygrishajev, line 105: this needs to happen based on localstorage value

apps/deploy-web/src/context/WalletProvider/WalletProvider.tsx
```diff
@@ -158,17 +153,25 @@ export const WalletProvider = ({ children }) => {
   }
 
   async function loadWallet(): Promise<void> {
-    let currentWallets = getStorageWallets();
+    const networkId =
```
ygrishajev, line 156: if we switch from a net that managed wallets are not supposed to operate on the below updates are run on the prev net and then reload happens. So now it would be running on the net that is to be changed to.

packages/network-store/src/network.store.ts
```diff
@@ -78,6 +78,21 @@ export class NetworkStore {
     this.initiateNetworks();
   }
 
+  private getInitialNetworkId() {
```
ygrishajev, line 81: so initial value was default one and then it was reading a different value from localstorage

baktun14 (approved):

PR #439, opened 2024-10-31
feat(deployment): implements custodial deployments top up data collection

ygrishajev:


baktun14 (approved):

baktun14 (approved):

apps/api/src/deployment/services/top-up-custodial-deployments/top-up-custodial-deployments.service.ts
```diff
@@ -0,0 +1,136 @@
...
+
+  private async retrieveFeesLimit(granter: string, grantee: string, denom: string) {
+    const feesAllowance = await this.allowanceHttpService.getFeeAllowanceForGranterAndGrantee(granter, grantee);
+    const feesSpendLimit = feesAllowance.allowance.spend_limit.find(limit => limit.denom === denom);
+
+    return feesSpendLimit ? parseFloat(feesSpendLimit.amount) : 0;
+  }
+
+  private async retrieveDrainingDeployments(owner: string, denom: string): Promise<DrainingDeploymentOutput[]> {
+    const currentHeight = await this.blockHttpService.getCurrentHeight();
```
Redm4x: This does an axios call to a node and it is being called concurrently for multiple deployments. Should we add caching since it will most likely return de same value for every calls?

apps/api/src/deployment/services/top-up-custodial-deployments/top-up-custodial-deployments.service.ts
```diff
@@ -0,0 +1,136 @@
...
+
+  async topUpDeployments() {
+    const wallets = [
+      { wallet: this.uaktMasterWalletService, client: this.uaktMasterSigningClientService },
+      { wallet: this.usdtMasterWalletService, client: this.usdtMasterSigningClientService }
+    ];
+
+    const topUpAllCustodialDeployments = wallets.map(async ({ wallet, client }) => {
+      const address = await wallet.getFirstAddress();
+      await this.allowanceHttpService.paginateDeploymentGrants({ grantee: address, limit: this.CONCURRENCY }, async grants => {
```
Redm4x, line 52: Isn't the master wallet the **granter** in this case?

apps/api/src/deployment/services/top-up-custodial-deployments/top-up-custodial-deployments.service.ts
ygrishajev: yea, i initially added this for simplicity but forgot to change

apps/api/src/deployment/services/top-up-custodial-deployments/top-up-custodial-deployments.service.ts
ygrishajev: no, custodial wallets are granting master wallet to top up on their behalf

apps/api/src/deployment/services/top-up-custodial-deployments/top-up-custodial-deployments.service.ts
ygrishajev: https://github.com/akash-network/console/pull/439/files#diff-4f235eb0e292279ed8db718ddd636e379b7bf3aa6c8858de1a9234099fdb89abR116

Redm4x (approved):

PR #441, opened 2024-11-02
Update trial-providers.json

deathlessdd:
There needs to be a fix for the trial there is bots that are spamming the trial This needs to be fixed before trial can happen because they will lease anything at any price. It is endless they will keep leasing all of the gpus.


deathlessdd (commented):
Let me explain You can start a Trial > rinse repeat unlimited times. The Bots figured a way to do this and they will lease every gpu at any price.

baktun14:
Please create an issue if you feel the system needs improvement.

Issue #407, opened 2024-10-14
New "new deployment" page


anilmurty:
# Background
We've been working on Frictionless Free Trials, Credit Card Payments and a workflow (similar to vercel) that takes care of building a container and deployment SDL for the user. Previously we built a way for users to get a "VM-like" experience using the "container-vm" template (that let's them provide an SSH pub key and then get access to a plain linux container that they can then install what they want into).

All these features combined together make Akash network usable by non-crypto users who may not be familiar with akash or SDLs or docker containers. There is a good opportunity to revisit the new deployment page, with these features in place - so that we're meeting these non-crypto, non-akash-familar developers where they are coming from.

# Goal
The goal of this task is to redesign the new deployment page (https://console.akash.network/new-deployment) so that there are clear workflows for the 4 developer user personas we want to target:

1. For **Pure web devs** who just have a react app and don't know how or know but don't want to deal with containers or infra (and use vercel or equivalent PaaS today) => "Buid & Deploy"  (which will lead to the new "Build & Deploy" template that is under development and soon to be released https://github.com/akash-network/console/issues/226)

![Screenshot 2024-10-14 at 12 00 01 PM](https://github.com/user-attachments/assets/2c91229d-1ead-4cfb-b4c9-1f22102fcbd0)

2. For **DevOps Engineers who use EC2 or equivalent VM instances**, to deploy apps for their teams => "Container-VM" (which will lead to this existing workflow under the "Plain-Linux" template https://console.akash.network/deploy-linux?step=edit-deployment)

![Screenshot 2024-10-14 at 12 00 36 PM](https://github.com/user-attachments/assets/32dacfa7-9549-4c97-904a-851dad8f3024)

3. For **DevOps engineers who already have microservices and are already container native** with their applications => "Custom Container" which will lead to this existing workflow: https://console.akash.network/new-deployment?step=edit-deployment

![Screenshot 2024-10-14 at 12 04 22 PM](https://github.com/user-attachments/assets/e16d3318-f494-409f-a5ce-a7e2b6dcc3fb)

4. For **Akash champions (aka power users) who already have their own SDLs** and don't need help building containers or SDLs => "Upload SDL" (this will be a minimalist element on the top right of this page that Denis will add in a subsequent iteration)

# UI Design

The UI design for the new "new-deployment" page is as shown below

- Each call to action (CTA) below will lead to one of the 4 workflows above. 
- The templates are "placeholders" and will be specific AI templates (and will frequently be changed based on what's trending)
[...]

PR #444, opened 2024-11-04
feat(deployment): improve new deployment page

baktun14:
https://github.com/akash-network/console/issues/407

image


apps/deploy-web/src/components/new-deployment/DeployOptionBox.tsx
```diff
@@ -1,41 +1,41 @@
...
-              </AvatarFallback>
-            </Avatar>
-          )}
-
-          <div className="ml-4 font-bold tracking-tight">{title}</div>
+    <Card className="min-h-[100px] cursor-pointer text-center hover:bg-secondary/60 dark:hover:bg-secondary/30" onClick={onClick} data-testid={testId}>
+      <CardHeader className="pb-2">
+        <div className="mb-2 flex items-center justify-center">
+          <div className="flex items-center space-x-2 rounded-sm bg-secondary p-1">
+            {topIcons?.map((icon, i) => <Image src={icon} alt="icon" width={28} height={28} key={i} className="max-h-[28px] object-contain" />)}
```
ygrishajev: array index shouldn't be used as key, name of the icon is a better choice here

apps/deploy-web/src/components/new-deployment/NewDeploymentContainer.tsx
```diff
@@ -159,10 +159,16 @@ export const NewDeploymentContainer: FC = () => {
   }
 
   return (
-    <Layout isLoading={isLoadingTemplates} isUsingSettings isUsingWallet containerClassName="pb-0">
-      <div className="flex w-full items-center">{activeStep !== null && <CustomizedSteppers activeStep={activeStep} />}</div>
+    <Layout isLoading={isLoadingTemplates} isUsingSettings isUsingWallet containerClassName="pb-0 h-full">
+      {activeStep !== null && activeStep > 0 && (
```
ygrishajev: not a blocker
```suggestion
      {!!activeStep && (
```

apps/deploy-web/src/components/new-deployment/TemplateList.tsx
```diff
@@ -34,15 +33,20 @@ const previewTemplateIds = [
   "akash-network-awesome-akash-grok",
   "akash-network-awesome-akash-FastChat"
 ];
+
 type Props = {
   onChangeGitProvider: (gh: boolean) => void;
+  onTemplateSelected: Dispatch<TemplateCreation | null>;
+  setEditedManifest: Dispatch<SetStateAction<string>>;
```
ygrishajev: Perhaps this would be a better thing to do?
```suggestion
  onManifestEdited: Dispatch<SetStateAction<string>>;
```
exposing as `Dispatch` kinda breaks the boundary of concerns too
```suggestion
  setEditedManifest: string => void;
```

apps/deploy-web/src/components/new-deployment/TemplateList.tsx
```diff
@@ -56,102 +60,119 @@ export const TemplateList: React.FunctionComponent<Props> = ({ onChangeGitProvid
...
+        description: "Custom uploaded file",
+        content: event.target?.result as string
+      });
+      setEditedManifest(event.target?.result as string);
+      setSdlEditMode("yaml");
+
+      router.push(UrlService.newDeployment({ step: RouteStep.editDeployment }));
+    };
+
+    reader.readAsText(fileUploaded);
```
ygrishajev: there is such logic already. I understand if you want to copypaste for now. However ideally there should be a hook we could re-use

apps/deploy-web/src/components/new-deployment/DeployOptionBox.tsx
baktun14: yea you're right, forgot to change it when I initially only had the same icon everywhere.

apps/deploy-web/src/components/new-deployment/NewDeploymentContainer.tsx
baktun14: true

apps/deploy-web/src/components/new-deployment/TemplateList.tsx
baktun14: Yea might as well do it now

PR #446, opened 2024-11-05
fix(auth): fix logout button not logging out completely

Redm4x:
An issue with [nextjs-auth0](https://github.com/auth0/nextjs-auth0) was causing the logout process to not be completed. That caused the user to be immediately logged back in when clicking "login" again without the need to select an account and enter password.

There is still an [open issue](https://github.com/auth0/nextjs-auth0/issues/1744) for this. This PR fixes the behavior by doing the redirect with `window.location` and using `<a>` tag instead of using the nextjs router and `<Link>` tag.

### DIFF
--- a/apps/deploy-web/src/components/layout/SidebarRouteButton.tsx
+++ b/apps/deploy-web/src/components/layout/SidebarRouteButton.tsx
@@ -1,5 +1,5 @@
 "use client";
-import { ReactNode } from "react";
+import React, { ReactNode } from "react";
 import { buttonVariants } from "@akashnetwork/ui/components";
 import { cn } from "@akashnetwork/ui/utils";
 import Link from "next/link";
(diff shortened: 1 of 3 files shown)


apps/deploy-web/src/components/layout/SidebarRouteButton.tsx
```diff
@@ -13,36 +13,62 @@ type Props = {
...
-      >
-        {!!route.icon && (
-          <span className={cn("z-[100] min-w-0", { ["m-[initial]"]: isNavOpen, ["mx-auto"]: !isNavOpen })}>
-            {route.icon({ className: cn({ ["text-primary font-bold"]: isSelected, ["mx-auto"]: !isNavOpen }, "text-xs") })}
-          </span>
-        )}
-        {isNavOpen && <span className="mb-1 ml-4 mt-1 min-w-0 flex-auto whitespace-nowrap">{route.title}</span>}
-      </Link>
+      {useNextLinkTag ? (
+        <Link
```
baktun14: Probably better to use something like `React.createElement(useNextLinkTag ? "Link" : "a", {attribute: stuff}, <div>Inner content</div>)`

baktun14 (approved):

Issue #429, opened 2024-10-31
Extract logger into packages


ygrishajev:
We have a [logger](https://github.com/akash-network/console/blob/main/apps/api/src/core/services/logger/logger.service.ts) in API which provides a proper interface and functionality for the purpose. We need to be able to use it in other packages. For this it needs to be moved to a dedicated package.

Acceptance criteria:

1. a new `logging` package is created
2. the logger is moved there
3. adjustments to the constructor need to be made to ensure we can still pass the parameters which are currently imported via config
4. api is updated to depend on that package
5. should be covered with unit tests

forbesus:
Hello @ygrishajev I will take this issue
Please assign me this issue

ygrishajev:
hey @forbesus! thanks for your initiative. Sure, go for it!

ygrishajev:
So, thanks to @forbesus most of this task is done.

To close it we need to add unit tests to the logger.

forbesus:
okay, i will add unit test  @ygrishajev

PR #452, opened 2024-11-07
test(observability): cover logger service with unit tests

forbesus:
ref #429

### DIFF
--- a/packages/logging/jest.config.js
+++ b/packages/logging/jest.config.js
@@ -1,0 +1,7 @@
+module.exports = {
+    preset: "ts-jest",
+    testEnvironment: "node",
+    testMatch: ["**/src/**/*.spec.ts"],
+    verbose: true,
+  };
+  
--- a/packages/logging/package.json
+++ b/packages/logging/package.json
@@ -6,18 +6,22 @@
   "scripts": {
     "format": "prettier --write ./*.{ts,json} **/*.{ts,json}",
     "lint": "eslint .",
-    "update-apps-local-deps": "mvm-update"
+    "update-apps-local-deps": "mvm-update",
+    "test": "jest"
   },
   "author": "",
   "license": "ISC",
   "dependencies": {
+    "@types/jest": "^29.5.14",
     "dotenv": "^16.4.5",
     "http-errors": "^2.0.0",
+    "jest": "^29.7.0",
     "pino": "^9.5.0",
     "pino-cloud-logging": "^1.0.6",
     "pino-fluentd": "^0.2.4",
     "pino-pretty": "^11.3.0",
     "stream": "^0.0.3",
+    "ts-jest": "^29.1.4",
     "zod": "^3.23.8"
   },
   "devDependencies": {
--- a/packages/logging/src/servicies/logger/logger.service.spec.ts
+++ b/packages/logging/src/servicies/logger/logger.service.spec.ts
@@ -1,0 +1,122 @@
+import createHttpError from "http-errors";
+import pino from "pino";
+import pinoFluentd from "pino-fluentd";
+import pretty from "pino-pretty";
+
+import { config } from "../../config";
+import { Logger, LoggerService } from "./logger.service";
+
+jest.mock("pino");
+jest.mock("pino-fluentd");
+jest.mock("pino-pretty");
+
+describe("LoggerService", () => {
+  let loggerService: LoggerService;
+  let mockLogger: jest.Mocked<Logger>;
+  const defaultLogFormat = config.STD_OUT_LOG_FORMAT;
+  const defaultFluentdTag = config.FLUENTD_TAG;
+  const defaultFluentdHost = config.FLUENTD_HOST;
+  const defaultFluentdPort = config.FLUENTD_PORT;
+
+  beforeEach(() => {
+    mockLogger = {
+      info: jest.fn(),
+      error: jest.fn(),
+      warn: jest.fn(),
+      debug: jest.fn(),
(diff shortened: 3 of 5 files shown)


forbesus:
hello @ygrishajev 
i created unit test
would you check it?

packages/logging/jest.config.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+    preset: "ts-jest",
+    testEnvironment: "node",
+    testMatch: ["**/test/**/*.test.ts"],
```
ygrishajev: let's make it like in other apps/packages pls
```suggestion
    testMatch: ["**/test/**/*.spec.ts"],
```

packages/logging/package.json
```diff
@@ -6,18 +6,22 @@
...
+    "@types/jest": "^29.5.14",
     "dotenv": "^16.4.5",
     "http-errors": "^2.0.0",
+    "jest": "^29.7.0",
     "pino": "^9.5.0",
     "pino-cloud-logging": "^1.0.6",
     "pino-fluentd": "^0.2.4",
     "pino-pretty": "^11.3.0",
     "stream": "^0.0.3",
+    "ts-jest": "^29.2.5",
```
ygrishajev: could you please make sure the versions of new packages here are the same as in other projects? you may check against `apps/api`

packages/logging/test/LoggerService.test.ts
ygrishajev, line 1: please move this file to `packages/logging/src/servicies/logger/logger.service.spec.ts`

packages/logging/test/LoggerService.test.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  afterEach(() => {
+    config.STD_OUT_LOG_FORMAT = defaultLogFormat;
+    config.FLUENTD_TAG = defaultFluentdTag;
+    config.FLUENTD_HOST = defaultFluentdHost;
+    config.FLUENTD_PORT = defaultFluentdPort;
+    jest.clearAllMocks();
+  });
+
+  describe("initPino", () => {
```
ygrishajev:
```suggestion
  describe("constructor", () => {
```

packages/logging/test/LoggerService.test.ts
```diff
@@ -0,0 +1,156 @@
...
+      expect(loggable).toEqual({
+        status: 404,
+        message: "Not found",
+        stack: "stack trace",
+        data: { key: "value" },
+        originalError: "stack trace"
+      });
+    });
+
+    it("should return status, message, stack, and data for HttpError", () => {
```
ygrishajev: seems to be a duplicate of the above

packages/logging/test/LoggerService.test.ts
```diff
@@ -0,0 +1,156 @@
...
+      config.FLUENTD_HOST = "";
+      config.FLUENTD_PORT = 0;
+      config.FLUENTD_TAG = "";
+
+      new LoggerService();
+      expect(pinoFluentd).not.toHaveBeenCalled();
+    });
+  });
+
+  describe("Logging methods", () => {
```
ygrishajev: I suggest to adjust like so:

```typescript
const methods: (keyof Logger)[] = ["info", "error", "warn", "debug"];
  describe.each(methods)("prototype.%s", method => {
    const logMessage = "Test message";

    it(`should call pino.${method} on info method`, () => {
      loggerService[method](logMessage);
      expect(mockLogger[method]).toHaveBeenCalledWith(logMessage);
    });
  });
```

for this Logger would need to implement common interface, which can be done like so:

```typescript
import pino, { Logger as PinoLogger } from "pino";
export type Logger = Pick<PinoLogger, "info" | "error" | "warn" | "debug">;

export class LoggerService implements Logger {
```

ygrishajev (commented):
hey, generally looks good. Adding some comments.

also commit should have type `test` and scope `observability`:

`test(observability): cover logger service with unit tests`

packages/logging/test/LoggerService.test.ts
```diff
@@ -0,0 +1,156 @@
...
+      expect(mockLogger.warn).toHaveBeenCalledWith(logMessage);
+    });
+
+    it("should call pino.debug on debug method", () => {
+      loggerService.debug(logMessage);
+      expect(mockLogger.debug).toHaveBeenCalledWith(logMessage);
+    });
+  });
+
+  describe("toLoggableInput", () => {
```
ygrishajev:
```suggestion
  describe("prototype.toLoggableInput", () => {
```

packages/logging/src/servicies/logger/logger.service.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  afterEach(() => {
+    config.STD_OUT_LOG_FORMAT = defaultLogFormat;
+    config.FLUENTD_TAG = defaultFluentdTag;
+    config.FLUENTD_HOST = defaultFluentdHost;
+    config.FLUENTD_PORT = defaultFluentdPort;
+    jest.clearAllMocks();
+  });
+
+  describe("constructor", () => {
```
ygrishajev: ouch, I've just noticed there's the actual method `initPino`. So the describe you had before actually makes more sense! apologies!
```suggestion
  describe("prototype.initPino", () => {
```

If you don't want to change it, I'll just do it myself later :)

ygrishajev (approved):

forbesus:
@ygrishajev all things are changed

ygrishajev:
thanks!

PR #453, opened 2024-11-07
fix(deployment): trial provider attributes

baktun14:


apps/api/src/billing/services/trial-validation/trial-validation.service.ts
```diff
@@ -1,20 +1,35 @@
-import * as v1beta4 from "@akashnetwork/akash-api/v1beta4";
+import * as v1beta3 from "@akashnetwork/akash-api/v1beta3";
 import { EncodeObject } from "@cosmjs/proto-signing";
 import { singleton } from "tsyringe";
 
 import { UserWalletOutput } from "@src/billing/repositories";
-import { getTrialProviders } from "@src/services/external/githubService";
+
```
ygrishajev, line 6: Let's follow screaming case convention for hardcoded constants?

apps/deploy-web/src/queries/useAnonymousUserQuery.ts
```diff
@@ -21,9 +21,14 @@ export function useAnonymousUserQuery(id?: string, options?: { enabled?: boolean
...
   useWhen(options?.enabled && !userState.user, async () => {
-    const { data: fetched, ...rest } = await userHttpService.getOrCreateAnonymousUser(id);
-    const token = "token" in rest ? rest.token : undefined;
-    setUserState({ user: fetched, token, isLoading: false });
+    try {
+      const { data: fetched, ...rest } = await userHttpService.getOrCreateAnonymousUser(id);
+      const token = "token" in rest ? rest.token : undefined;
+      setUserState({ user: fetched, token, isLoading: false });
+    } catch (error) {
+      console.error(error);
```
ygrishajev: I kinda missed how this log ended up here. However just want to point out that it's more/less useless. If there's an error it should be added to the state above and handled somewhere in ui.

apps/deploy-web/src/utils/deploymentData/v1beta3.ts
```diff
@@ -22,6 +25,51 @@ const getDenomFromSdl = (groups: any[]): string => {
   return denoms[0];
 };
 
+export function appendTrialAttribute(yamlStr: string) {
+  const sdl = getSdl(yamlStr, "beta3", networkStore.selectedNetworkId);
+
+  Object.keys(sdl.data.profiles.placement).forEach(placement => {
+    if (!sdl.data.profiles.placement[placement].attributes) {
```
ygrishajev: `sdl.data.profiles.placement[placement]` is used like 10+ times. would be easier to read if it's aliased by a var

apps/deploy-web/src/utils/deploymentData/v1beta3.ts
```diff
@@ -22,6 +25,51 @@ const getDenomFromSdl = (groups: any[]): string => {
...
+      sdl.data.profiles.placement[placement].attributes = [];
+    }
+
+    if (sdl.data.profiles.placement[placement].attributes.find(attr => attr.key === trialAttribute)) {
```
ygrishajev: real hard to understand why this is a break condition. Storing it into into a verbose var would help.
```suggestion
    const isBreakCondition = sdl.data.profiles.placement[placement].attributes.find(attr => attr.key === trialAttribute);
    
    if (isBreakCondition) {
```

ygrishajev: after second look I got it's missing trial attr. However this is still the reason to approach this like so :)

apps/deploy-web/src/utils/deploymentData/v1beta3.ts
```diff
@@ -22,6 +25,51 @@ const getDenomFromSdl = (groups: any[]): string => {
...
+
+  const result = yaml.dump(sdl.data, {
+    indent: 2,
+    quotingType: '"',
+    styles: {
+      "!!null": "empty" // dump null as emtpy value
+    },
+    replacer: (key, value) => {
+      // Attributes is a key value pair object, but we store it as an array of objects
+      if (key === "attributes" && Array.isArray(value) && value.some(attr => attr.key === trialAttribute)) {
```
ygrishajev: the condition above is hard to understand, I'd suggest to store it in a var

apps/deploy-web/src/utils/deploymentData/v1beta3.ts
```diff
@@ -22,6 +25,51 @@ const getDenomFromSdl = (groups: any[]): string => {
...
+        return value?.reduce((acc, curr) => ((acc[curr.key] = curr.value), acc), {});
```
ygrishajev: imo this can be a function called smth descriptive, perhaps comments above would not be needed then

apps/deploy-web/src/queries/useAnonymousUserQuery.ts
baktun14: This is to handle the case where the id of the anonymous user is in the browser but not in db anymore, like staging where we clear the users from time to time and it just makes the ui stuck in loading screen. I'm not sure what good it would do to show the user that this didn't work as it's supposed to be a hidden logic. It is supposed to create one if it doesn't find it so I'm not sure why it throws.

apps/deploy-web/src/utils/deploymentData/v1beta3.ts
baktun14: Sure

apps/deploy-web/src/queries/useAnonymousUserQuery.ts
ygrishajev: i'm saying about console.logs. But actually logger would solve this. As we wouldn't have debug level in prod but would locally. scratch the comment)

baktun14: I can change it to `Sentry.captureException(error);`

Redm4x (approved):

PR #454, opened 2024-11-08
feat(deployment): implement top up message sending

ygrishajev:
Also rework managed top up logic to rely on user wallets instead of grants

refs #395


apps/api/src/deployment/services/top-up-managed-deployments/top-up-managed-deployments.service.ts
```diff
@@ -1,59 +1,67 @@
...
+    private readonly balancesService: BalancesService,
+    private readonly rpcClientService: RpcMessageService,
     private readonly errorService: ErrorService
   ) {}
 
-  async topUpDeployments() {
-    const address = await this.managedMasterWalletService.getFirstAddress();
-    await this.allowanceHttpService.paginateDeploymentGrants({ granter: address, limit: this.CONCURRENCY }, async grants => {
+  async topUpDeployments(options: TopUpDeploymentsOptions) {
+    await this.userWalletRepository.paginate({ limit: this.CONCURRENCY }, async wallets => {
```
baktun14, line 38: Ideally, we should iterate over only the wallets that need top up, because otherwise we have to iterate over all and query deployments for each which will be heavy for the database.

ygrishajev: Just dropping a response discussed in private for visibility.

there's no concurrency for this job, so it's a single query per several seconds. It uses the index on the query. So it's not expensive.
This is an approach we discussed concluding this is efficient enough for the time being.

baktun14 (approved):

PR #455, opened 2024-11-08
feat(deployment): implement clean up of managed deployments

ygrishajev:
refs #395

### DIFF
--- a/apps/api/src/deployment/repositories/deployment/deployment.repository.ts
+++ b/apps/api/src/deployment/repositories/deployment/deployment.repository.ts
@@ -1,0 +1,46 @@
+import { Deployment, Lease } from "@akashnetwork/database/dbSchemas/akash";
+import { literal, Op } from "sequelize";
+import { singleton } from "tsyringe";
+
+export interface StaleDeploymentsOptions {
+  createdHeight: number;
+  owner: string;
+}
+
+export interface StaleDeploymentsOutput {
+  dseq: number;
+}
+
+@singleton()
+export class DeploymentRepository {
+  async findStaleDeployments(options: StaleDeploymentsOptions): Promise<StaleDeploymentsOutput[]> {
+    const deployments = await Deployment.findAll({
+      attributes: ["dseq"],
+      include: [
+        {
+          model: Lease,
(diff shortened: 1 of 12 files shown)


apps/api/src/deployment/repositories/deployment/deployment.repository.ts
```diff
@@ -0,0 +1,46 @@
...
+          },
+          {
+            closedHeight: null
+          }
+        ]
+      },
+      group: ["deployment.dseq"],
+      having: literal(`COUNT("leases"."deploymentId") = 0`),
+      raw: true
+    });
```
ygrishajev, line 42: this would produce smth lilke

```sql
SELECT "deployment"."dseq"
FROM "deployment" AS "deployment"
         LEFT OUTER JOIN "lease" AS "leases" ON "deployment"."id" = "leases"."deploymentId"
WHERE (
    "deployment"."createdHeight" < 18839498 - 97
        AND "deployment"."closedHeight" IS NULL
    )
  AND "deployment"."owner" = 'akash1kgplcsz3jzgu8e69pwh3dj6dhfn5lfkl6gwhdn'
GROUP BY "deployment"."dseq"
HAVING COUNT("leases"."deploymentId") = 0;
```

baktun14 (approved):

apps/api/src/deployment/repositories/deployment/deployment.repository.ts
```diff
@@ -0,0 +1,46 @@
...
+        [Op.and]: [
+          {
+            createdHeight: {
+              [Op.lt]: options.createdHeight
+            }
+          },
+          {
+            closedHeight: null
+          }
+        ]
```
Redm4x, line 37: Could the conditions be all on the same level like this?
```
{
        owner: options.owner,
        createdHeight: {
          [Op.lt]: options.createdHeight
        },
        closedHeight: null
}
```

ygrishajev: Yea, you're right, will fix with another pr. merged this before I saw this comment

PR #458, opened 2024-11-08
Update trial-providers.json

88plug:
Add Crypto and Coffee managed providers


baktun14:
The process is changing and this will be obselete.

PR #461, opened 2024-11-11
feat(billing): add community providers to whitelist

chainzero:


ygrishajev:
pls update commit to have `billing` scope under which the whole feature was delivered. There is no trial scope https://github.com/akash-network/console/blob/main/.commitlintrc.json

it seems like hooks are not enabled locally.

baktun14:
I updated the commit name in the squash commit

Issue #436, opened 2024-10-31
Utilise new logger in stats-web


ygrishajev:
Depends on #430 

See milestone for more description.

Acceptance criteria:

- logger module from packages is added to statw-web
- all console logs are replaced with properly bound logger instance calls
- redundant or not useful logs should be removed

forbesus:
are you available to chat via discord? @ygrishajev

ygrishajev:
@forbesus yea, same username as here - ygrishajev

However I'm not sure about my availability till mon

PR #462, opened 2024-11-12
feat(observability): utilise new logger in stats-web

forbesus:
ref #436


apps/stats-web/src/app/transactions/[hash]/page.tsx
```diff
@@ -33,7 +36,7 @@ export async function generateMetadata({ params: { hash } }: TransactionDetailPa
 
 async function fetchTransactionData(hash: string, network: Network["id"]): Promise<TransactionDetail | null> {
   const apiUrl = serverApiUrlService.getBaseApiUrlFor(network);
-  console.log("DEBUG apiUrl", apiUrl);
+  transactionLogger.info(`DEBUG apiUrl: ${apiUrl}`);
```
Comment (ygrishajev) [inline apps/stats-web/src/app/transactions/[hash]/page.tsx]:
```suggestion
```

apps/stats-web/src/app/transactions/[hash]/error.tsx
```diff
@@ -1,15 +1,18 @@
...
 import PageContainer from "@/components/PageContainer";
 import { Title } from "@/components/Title";
 
+const transactionLogger = new LoggerService({ context: "transaction" });
+
 export default function Error({ error, reset }: { error: Error & { digest?: string }; reset: () => void }) {
   useEffect(() => {
     // Log the error to an error reporting service
-    console.error(error);
+    transactionLogger.error(error);
```
Comment (ygrishajev) [inline apps/stats-web/src/app/transactions/[hash]/error.tsx]:
```suggestion
    transactionLogger.debug(error);
```

apps/stats-web/src/app/error.tsx
```diff
@@ -1,15 +1,18 @@
...
 import PageContainer from "@/components/PageContainer";
 import { Title } from "@/components/Title";
 
+const errorLogger = new LoggerService({ context: "error" });
+
 export default function Error({ error, reset }: { error: Error & { digest?: string }; reset: () => void }) {
   useEffect(() => {
     // Log the error to an error reporting service
-    console.error(error);
+    errorLogger.error(error);
```
ygrishajev:
```suggestion
    errorLogger.debug(error);
```

apps/stats-web/src/app/blocks/[height]/errors.tsx
```diff
@@ -1,11 +1,14 @@
...
 import { useEffect } from "react";
+import { LoggerService } from "@akashnetwork/logging";
+
+const blockErrorLogger = new LoggerService({ context: "block" });
 
 export default function Error({ error, reset }: { error: Error & { digest?: string }; reset: () => void }) {
   useEffect(() => {
     // Log the error to an error reporting service
-    console.error(error);
+    blockErrorLogger.error(error);
```
Comment (ygrishajev) [inline apps/stats-web/src/app/blocks/[height]/errors.tsx]:
```suggestion
    blockErrorLogger.debug(error);
```

apps/stats-web/src/lib/copyClipboard.ts
```diff
@@ -14,9 +18,9 @@ function fallbackCopyTextToClipboard(text: string) {
   try {
     const successful = document.execCommand("copy");
     const msg = successful ? "successful" : "unsuccessful";
-    console.log("Fallback: Copying text command was " + msg);
+    libLogger.info("Fallback: Copying text command was " + msg);
```
ygrishajev:
```suggestion
    libLogger.debug("Fallback: Copying text command was " + msg);
```

apps/stats-web/src/lib/copyClipboard.ts
```diff
@@ -14,9 +18,9 @@ function fallbackCopyTextToClipboard(text: string) {
...
   } catch (err) {
-    console.error("Fallback: Oops, unable to copy", err);
+    libLogger.error(`Fallback: Oops, unable to copy: ${err}`);
```
ygrishajev:
```suggestion
    libLogger.debug(`Fallback: Oops, unable to copy: ${err}`);
```

apps/stats-web/src/lib/copyClipboard.ts
```diff
@@ -28,10 +32,10 @@ export const copyTextToClipboard = (text: string) => {
   }
   navigator.clipboard.writeText(text).then(
     () => {
-      console.log("Async: Copying to clipboard was successful!");
+      libLogger.info("Async: Copying to clipboard was successful!");
```
ygrishajev:
```suggestion
      libLogger.debug("Async: Copying to clipboard was successful!");
```

apps/stats-web/src/lib/copyClipboard.ts
```diff
@@ -28,10 +32,10 @@ export const copyTextToClipboard = (text: string) => {
...
     },
     err => {
-      console.error("Async: Could not copy text: ", err);
+      libLogger.error(`Async: Could not copy text: ${err}`);
```
ygrishajev:
```suggestion
      libLogger.debug(`Async: Could not copy text: ${err}`);
```

packages/logging/src/servicies/logger/logger.service.ts
```diff
@@ -2,8 +2,6 @@ import { context, trace } from "@opentelemetry/api";
 import { isHttpError } from "http-errors";
 import pino, { Bindings, Logger as PinoLogger, LoggerOptions } from "pino";
 import { gcpLogOptions } from "pino-cloud-logging";
-import pinoFluentd from "pino-fluentd";
-import pretty from "pino-pretty";
```
ygrishajev: this is still needed, this is for readability when developing locally

forbesus:
would you check this pr? @ygrishajev

apps/stats-web/src/app/blocks/[height]/errors.tsx
```diff
@@ -1,11 +1,14 @@
 "use client"; // Error components must be Client Components
 
 import { useEffect } from "react";
+import { LoggerService } from "@akashnetwork/logging";
+
+const blockErrorLogger = new LoggerService({ context: "block" });
```
Comment (ygrishajev) [inline apps/stats-web/src/app/blocks/[height]/errors.tsx]: this is generally used for dev. So we need to be able to look it up faster. prob smth like this would make sense more

```suggestion
const blockErrorLogger = new LoggerService({ context: "apps/stats-web/src/app/blocks/[height]/errors.tsx" });
```

wdut?

packages/logging/src/servicies/logger/logger.service.ts
forbesus: actually, `pinoFluentd ` and `pretty` is not working on client side @ygrishajev

ygrishajev: well, this was the point of the other task to make it work nicely in both envs.
Basically we need to load pretty conditionally, or in the worst case scenario have 2 loggers.

ygrishajev: i see from the doc it's possible to use it without import https://github.com/pinojs/pino-pretty?tab=readme-ov-file#programmatic-integration

meaning we can add this option conditionally based on the runtime env

forbesus:
all things are completed @ygrishajev

packages/logging/src/servicies/logger/logger.service.ts
```diff
@@ -28,20 +26,17 @@ export class LoggerService implements Logger {
       }
     };
 
-    if (config.STD_OUT_LOG_FORMAT === "pretty") {
-      destinations.push(pretty({ sync: true }));
-    } else {
-      options = gcpLogOptions(options as any) as LoggerOptions;
+    if (typeof window === 'undefined') {
```
ygrishajev: it should still respect `STD_OUT_LOG_FORMAT`

ygrishajev (commented):
so how does it work with process.env and config within logger?

forbesus:
how about now? @ygrishajev

packages/logging/src/servicies/logger/logger.service.ts
```diff
@@ -29,19 +27,18 @@ export class LoggerService implements Logger {
     };
 
     if (config.STD_OUT_LOG_FORMAT === "pretty") {
-      destinations.push(pretty({ sync: true }));
+      if (typeof window !== 'undefined') {
+        options.transport = {
+          target: 'pino-pretty',
+          options: { colorize: true }
+        };
+        destinations.push(process.stdout);
```
ygrishajev: `destinations` is redundant and not needed anymore, right?

packages/logging/src/servicies/logger/logger.service.ts
```diff
@@ -29,19 +27,18 @@ export class LoggerService implements Logger {
     };
 
     if (config.STD_OUT_LOG_FORMAT === "pretty") {
-      destinations.push(pretty({ sync: true }));
+      if (typeof window !== 'undefined') {
```
ygrishajev: lets do it like so

```typescript
if (typeof window !== "undefined" && config.STD_OUT_LOG_FORMAT === "pretty") {
      options.transport = {
        target: "pino-pretty",
        options: { colorize: true }
      };
      destinations.push(process.stdout);
    } else {
      options = gcpLogOptions(options as any) as LoggerOptions;
    }
```

in browser it's fine to log raw objects

ygrishajev (commented):
so I have `LOG_LEVEL=debug` but it's not picked up by config in the browser. I think config needs to be provided to the LoggerService constructor.

forbesus:
> so I have `LOG_LEVEL=debug` but it's not picked up by config in the browser. I think config needs to be provided to the LoggerService constructor.

i am not sure what this meaning exactly
could you tell me more detail?

ygrishajev:
@forbesus LOG_LEVEL is an env var

```ts
LOG_LEVEL: z.enum(["fatal", "error", "warn", "info", "debug", "trace"]).optional().default("info"),
```

which controls which logs are done. It is a great way to configure logger on the backend, but it wouldn't work in most frontend implementations.

So LoggerService should probably implement both env vars and config object as param in constructor.

forbesus:
how about now? @ygrishajev

packages/logging/src/servicies/logger/logger.service.ts
```diff
@@ -16,7 +16,7 @@ export class LoggerService implements Logger {
 
   private initPino(bindings?: Bindings): Logger {
     let options: LoggerOptions = {
-      level: config.LOG_LEVEL,
+      level: bindings?.logLevel || config.LOG_LEVEL,
```
ygrishajev: bindings is what goes to log object with every log, config is used to parametrize pino and logger. so the interface should be smth like

```ts
constructor(options?: LoggerOptions & { context: string })
```

this way existing client code wouldn't break. In the init though we'd need to put context into bindings

forbesus:
please check again @ygrishajev

packages/logging/src/servicies/logger/logger.service.ts
```diff
@@ -13,35 +10,30 @@ export type Logger = Pick<PinoLogger, "info" | "error" | "warn" | "debug">;
...
-
-    let options: LoggerOptions = {
-      level: config.LOG_LEVEL,
+  private initPino(options?: LoggerOptions, bindings?: Bindings): Logger {
+    let loggerOptions: LoggerOptions = {
+      level: bindings?.level || config.LOG_LEVEL,
       mixin: () => {
         const currentSpan = trace.getSpan(context.active());
-        return { ...currentSpan?.spanContext() };
+        return { ...currentSpan?.spanContext(), context: bindings?.context };
```
ygrishajev: no need to add it here

packages/logging/src/servicies/logger/logger.service.ts
```diff
@@ -13,35 +10,30 @@ export type Logger = Pick<PinoLogger, "info" | "error" | "warn" | "debug">;
 export class LoggerService implements Logger {
   protected pino: Logger;
 
-  constructor(bindings?: Bindings) {
-    this.pino = this.initPino(bindings);
+  constructor(options?: LoggerOptions & { context: string }) {
```
ygrishajev: we need to make sure this is taken into account https://github.com/pinojs/pino/blob/HEAD/docs/api.md#bindings

forbesus:
how about now? @ygrishajev

packages/logging/src/servicies/logger/logger.service.ts
```diff
@@ -34,7 +34,7 @@ export class LoggerService implements Logger {
       ...inputOptions
     };
 
-    if (typeof window === "undefined" && config.STD_OUT_LOG_FORMAT === "pretty") {
+    if (typeof window !== "undefined" && config.STD_OUT_LOG_FORMAT === "pretty") {
```
ygrishajev: I'm pretty sure this should stay as is. Why'd you change this?
```suggestion
    if (typeof window === "undefined" && config.STD_OUT_LOG_FORMAT === "pretty") {
```

apps/stats-web/src/config/browser-env.config.ts
```diff
@@ -6,5 +6,6 @@ export const browserEnvConfig = validateStaticEnvVars({
   NEXT_PUBLIC_NODE_ENV: process.env.NEXT_PUBLIC_NODE_ENV,
   NEXT_PUBLIC_BASE_API_TESTNET_URL: process.env.NEXT_PUBLIC_BASE_API_TESTNET_URL,
   NEXT_PUBLIC_BASE_API_SANDBOX_URL: process.env.NEXT_PUBLIC_BASE_API_SANDBOX_URL,
-  NEXT_PUBLIC_BASE_API_MAINNET_URL: process.env.NEXT_PUBLIC_BASE_API_MAINNET_URL
+  NEXT_PUBLIC_BASE_API_MAINNET_URL: process.env.NEXT_PUBLIC_BASE_API_MAINNET_URL,
+  NEXT_PUBLIC_BASE_API_LOG_LEVEL: process.env.NEXT_PUBLIC_LOG_LEVEL
```
ygrishajev: how is this var utilised?

packages/logging/src/servicies/logger/logger.service.ts
forbesus: sorry, i just confused

apps/stats-web/src/config/browser-env.config.ts
forbesus: it's not using now

packages/logging/src/servicies/logger/logger.service.ts
forbesus: image
i am getting this error when i use LoggerService 
would you check this? @ygrishajev

forbesus: how about using `this.pino.child()` instead of `this.pino.setBindings()` ?
```js
setContext(context: string) {
    this.pino.child({ context });

    return this;
  }
```
like this @ygrishajev

forbesus: and i think making two logger may good choice but it's worst case  because next.js is running on both of ssr and csr

forbesus:
hello @ygrishajev i just fixed this issue
would you check again?
thanks

packages/logging/src/servicies/logger/logger.service.ts
```diff
@@ -35,7 +35,7 @@ export class LoggerService implements Logger {
       ...inputOptions
     };
 
-    if (typeof window === "undefined" && config.STD_OUT_LOG_FORMAT === "pretty") {
+    if (typeof window === "undefined" && config.STD_OUT_LOG_FORMAT === "pretty" && config.NODE_ENV !== "production") {
```
ygrishajev: this class shouldn't depend on NODE_ENV, only on env vars specific to it and direct config passed to the constructor.
```suggestion
    if (typeof window === "undefined" && config.STD_OUT_LOG_FORMAT === "pretty") {
```

packages/logging/src/config/env.config.ts
```diff
@@ -6,6 +6,7 @@ dotenv.config();
 
 const envSchema = z.object({
   LOG_LEVEL: z.enum(["fatal", "error", "warn", "info", "debug", "trace"]).optional().default("info"),
+  NODE_ENV: z.enum(["development", "production", "test"]).optional().default("development"),
```
ygrishajev: this package shouldn't depend on NODE_ENV, only on env vars specific to it and direct config passed to the constructor.

```suggestion
```

apps/stats-web/next.config.js
```diff
@@ -13,6 +13,14 @@ try {
 
 /** @type {import('next').NextConfig} */
 const nextConfig = {
+  webpack: (config, { isServer }) => {
```
ygrishajev: could you please explain why is this needed?

forbesus: @ygrishajev 
now the Next.js bundler is interfering with how thread-stream resolves its worker files. because worker threads used in server-side libraries are not bundled correctly, i excluded it

ygrishajev: what issue was it causing exactly? throwing some kind of exception? at build time? at run time?

forbesus: image
i am getting this error at runtime

forbesus: i did some research on it and found a solution by changing `next.config.js` 
@ygrishajev

ygrishajev: i don't mind this in next.config. I just was curious what is it

packages/logging/src/servicies/logger/logger.service.ts
```diff
@@ -48,13 +48,13 @@ export class LoggerService implements Logger {
   }
 
   setContext(context: string) {
-    this.pino.setBindings({ context });
+    this.pino.child({ context });
```
ygrishajev: why so?

forbesus: using `setBindings` had been reported as bug
https://github.com/pinojs/pino/issues/2021
@ygrishajev

ygrishajev:
@forbesus I'm getting this error on the backend

```
stats-web-1  |  ⨯ Error: Cannot find module 'pino'
stats-web-1  | Require stack:
stats-web-1  | - /app/apps/stats-web/.next/server/app/_not-found/page.js
stats-web-1  |     at Object.pino (/app/apps/stats-web/.next/server/app/_not-found/page.js:176:18)
stats-web-1  |     at __webpack_require__ (/app/apps/stats-web/.next/server/webpack-runtime.js:33:43)
stats-web-1  |     at eval (../../packages/logging/src/servicies/logger/logger.service.ts:7:62)
stats-web-1  |     at (ssr)/../../packages/logging/src/servicies/logger/logger.service.ts (/app/apps/stats-web/.next/server/app/_not-found/page.js:775:1)
stats-web-1  |     at __webpack_require__ (/app/apps/stats-web/.next/server/webpack-runtime.js:33:43)
stats-web-1  |     at eval (../../packages/logging/src/index.ts:5:90)
stats-web-1  |     at (ssr)/../../packages/logging/src/index.ts (/app/apps/stats-web/.next/server/app/_not-found/page.js:764:1)
stats-web-1  |     at __webpack_require__ (/app/apps/stats-web/.next/server/webpack-runtime.js:33:43)
stats-web-1  |     at eval (./src/app/error.tsx:9:79)
stats-web-1  |     at (ssr)/./src/app/error.tsx (/app/apps/stats-web/.next/server/app/_not-found/page.js:324:1)
stats-web-1  |     at Object.__webpack_require__ [as require] (/app/apps/stats-web/.next/server/webpack-runtime.js:33:43)
stats-web-1  |     at JSON.parse (<anonymous>)
stats-web-1  |     at JSON.parse (<anonymous>)
```

packages/logging/src/servicies/logger/logger.service.ts
```diff
@@ -60,8 +60,10 @@ export class LoggerService implements Logger {
     }
 
     try {
-      // eslint-disable-next-line @typescript-eslint/no-var-requires
-      return require("pino-pretty")({ colorize: true, sync: true });
```
forbesus: in order to prevent loading server-specific libraries on the client side (browser), we should set `if (typeof window === "undefined")` in front of `require`  @ygrishajev

ygrishajev: yea, take a look few lines higher ;)

forbesus: return is working on runtime
at compile time, require is working csr and ssr , and return is not working
pino-pretty should be worked with ssr
but `pino-pretty` is always importing

so we should change this code like this
```
private getPrettyIfPresent(): PinoPretty.PrettyStream | undefined {
    if (typeof window !== "undefined" || LoggerService.config.STD_OUT_LOG_FORMAT !== "pretty") {
      return;
    } else {
      try {
        // eslint-disable-next-line @typescript-eslint/no-var-requires
        return require("pino-pretty")({ colorize: true, sync: true });
      } catch (e) {
        this.debug({ context: LoggerService.name, message: "Failed to load pino-pretty", error: e });
        /* empty */
      }
    }
  }```

forbesus: @ygrishajev

forbesus:
moved to #516

Issue #430, opened 2024-10-31
Ensure logger works in all environments


ygrishajev:
Depends on #429

The logger as described in #429 wold only work in nodejs environment. We need to make sure it's hybrid and can work in any environment. To do so we need to adjust it to import dependencies dynamically based on environment and packages present in node_modules.

Acceptance criteria:

- logger works in both nodejs and browser env
- it dynamically loads relevant modules or uses relevant browser apis based on environment, configurations and modules present in node_modules

forbesus:
hello @ygrishajev i checked this issue and package is working well on browser and nodejs
so i think we can integrate this in another apps

ygrishajev:
@forbesus ok, cool. Let's start with #436 then. I think some issues might still pop up. E.g. config might not work with process env properly. We'll also need it on external packages once it is published. So this needs to be tested against built dist.

forbesus:
hello @ygrishajev previous weekend, i checked all pino documentation again 
we are using pino-fluentd and pino-pretty modules for logging package now and fluentd and pretty all were designed for server side and working with only nodejs

i am sorry for previous confusing and didn't check about fluentd and pretty

ygrishajev:
@forbesus No prob. Is it only fluentd? I don't think it's needed anymore. Let's just remove it

PR #466, opened 2024-11-13
refactor(observability): ensure logger can be configured with options and env

ygrishajev:
This implementation ensures that all pino options can be passed to the LoggerService and would have higher prio comared to env configuration. Env configuration would still take effect.

Also it adds a couple of utility and factory methods.

refs #430


ygrishajev:
@forbesus so the changes needed for the logger appeared a bit more tedious to explain. Decided to quickly do it myself. Your PR is still valid though. Just make sure to rebase when this one is merged and discard your LoggerService changes.

baktun14 (approved):

Issue #467, opened 2024-11-13
Implement private container support in SDL Builder


baktun14:
- Add fields to support private container support in SDL Builder form
  - https://akash.network/docs/getting-started/stack-definition-language/#private-container-registry-support
- Implement SDL generation to include private containers

PR #479, opened 2024-11-18
feat(sdl): can set credentials

jzsfkzm:
One can select an image repository and enter credentials after this. Available in the SDL Builder, but not in the deployment UI.

ref #467

### DIFF
--- a/apps/deploy-web/src/utils/sdl/data.ts
+++ b/apps/deploy-web/src/utils/sdl/data.ts
@@ -17,17 +17,10 @@ export const defaultHttpOptions = {
   nextTimeout: 60000
 };
 
-const defaultCredentials = {
-  host: 'docker.io',
-  username: '',
-  password: '',
-};
-
 export const defaultService: ServiceType = {
   id: nanoid(),
   title: "service-1",
   image: "",
-  credentials: defaultCredentials,
   sshPubKey: "",
   profile: {
     cpu: 0.1,
@@ -109,7 +102,6 @@ export const defaultRentGpuService: ServiceType = {
   id: nanoid(),
   title: "service-1",
   image: "",
-  credentials: defaultCredentials,
   profile: {
     cpu: 0.1,
     gpu: 1,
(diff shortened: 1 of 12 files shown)


baktun14 (commented):
The credentials for private repos are optional, since not all repos are private. The way you implemented it is great, we just need to add a flag [like "hasGpu"](https://github.com/akash-network/console/blob/main/apps/deploy-web/src/types/sdlBuilder.ts#L80) and conditionally add the validation. So a simple checkbox like the `hasGpu` to toggle the form controls should do it.

Is there a reason behind the `canChangeCredentials`? We would like to have it in all the SDL builder instances ([SDL Builder page](https://console.akash.network/sdl-builder) and [new deployment page](https://console.akash.network/new-deployment?step=edit-deployment)).

apps/deploy-web/src/utils/sdl/data.ts
```diff
@@ -17,10 +17,17 @@ export const defaultHttpOptions = {
...
+  host: 'docker.io',
+  username: '',
+  password: '',
+};
+
 export const defaultService: ServiceType = {
   id: nanoid(),
   title: "service-1",
   image: "",
+  credentials: defaultCredentials,
```
baktun14: This is throwing and error in the build as the `username` and `password` are required.

jzsfkzm:
> The credentials for private repos are optional, since not all repos are private. The way you implemented it is great, we just need to add a flag [like "hasGpu"](https://github.com/akash-network/console/blob/main/apps/deploy-web/src/types/sdlBuilder.ts#L80) and conditionally add the validation. So a simple checkbox like the `hasGpu` to toggle the form controls should do it.
> 
> Is there a reason behind the `canChangeCredentials`? We would like to have it in all the SDL builder instances ([SDL Builder page](https://console.akash.network/sdl-builder) and [new deployment page](https://console.akash.network/new-deployment?step=edit-deployment)).

I added the flag `canChangeCredentials` because I figured it is not needed for templates on the Deploy page. Now I see it is very much needed there too, so removed. Now it is all the same for Deploy and SDL Builder.
Added a checkbox with label "Private" on top the box, revealing credential fields only when you click it.

baktun14 (approved):
I just tested it locally and it works great, great work!

PR #481, opened 2024-11-18
refactor: splits templates API into list and fetch by id

### LINKED ISSUE #102: Improve templates API
- The templates API are getting heavy as more templates are added to the awesome akash repo. Also, the endpoint is called many times throughout the UX.
image
---
TODO
- [x] Modifiy the [templates list api endpoint](https://github.com/akash-network/console/blob/main/apps/api/src/routes/v1/templates.ts#L44) to only return the info needed to display in a list (without the readme, etc) with goal to reduce the amount of uncessary data sent to the browser
- [x] Add a new api endpoint for a template detail with all the info

stalniy:
1. Splits templates API into 2: list and getById
2. Reduces amount of properties for every template in response. It contains only 4 properties that is used on UI:
    * id
    * summary
    * logoUrl
    * name

~Not tested.~ Fixes #102


stalniy:
~DO NOT MERGE it. Because very likely it will break template detail page on UI as UI currently does not fetch template by id from API.~

baktun14:
Thanks for your work!

Sorry if it wasn't in the requirements, but following your comments about the impact on the frontend; I suggest we keep the old `/templates` api endpoint as-is and create a new one `/template-list` with your changes. That way it won't affect the frontend and we'll be able to merge your pr without effect and implement it on the frontend in another pr.

stalniy:
Probably I could create this extra api endpoints under /v2. To make this  evolution more clear from api perspective and at the same time keeping /v1/templates to keep old clients work without issues. Some people also may have cached version of SPA in their browsers (long lived browser tabs), and their version of client app may try to use older API version

baktun14:
> Probably I could create this extra api endpoints under /v2. To make this evolution more clear from api perspective and at the same time keeping /v1/templates to keep old clients work without issues. Some people also may have cached version of SPA in their browsers (long lived browser tabs), and their version of client app may try to use older API version

Great suggestion and I discussed this internally, but for the sake of keeping the scope of this pr small, I suggest we just keep the old version and add the 2 new endpoints you added.

stalniy:
Changed, now it's safe to merge it

apps/api/src/routes/v1/templates/list.ts
```diff
@@ -0,0 +1,42 @@
...
+          schema: responseSchema
+        }
+      }
+    }
+  }
+});
+
+export default new OpenAPIHono().openapi(route, async c => {
+  const templatesPerCategory = await getCachedTemplatesGallery();
+  // TODO: remove manual response filtering when https://github.com/honojs/middleware/issues/181 is done
```
ygrishajev, line 36: 🔥

baktun14 (approved):
Tested it locally, works great. Good job!

Issue #486, opened 2024-11-19
Api endpoint trial providers


baktun14:
- Add an api endpoint that will fetch the providers with trial attribute `console/trials` and is signed with auditor `akash1365yvmc4s7awdyj3n2sav7xfx76adc6dnmlx63`
- Remove [whitelist](https://github.com/akash-network/console/blob/main/config/trial-providers.json) from codebase and api endpoint

PR #488, opened 2024-11-21
feat(provider): new provider trial endpoint

baktun14:
https://github.com/akash-network/console/issues/486


apps/api/src/deployment/repositories/provider/provider.repository.ts
```diff
@@ -0,0 +1,21 @@
...
+      attributes: ["provider"],
+      where: {
+        auditor: AUDITOR,
+        key: TRIAL_ATTRIBUTE,
+        value: "true"
+      },
+      raw: true
+    });
+
+    return trialProviders ? (trialProviders.map(provider => provider.provider) as unknown as string[]) : [];
```
ygrishajev: It seems like it's properly typed and if no records found it should be an empty array

```suggestion
    return trialProviders.map(provider => provider.provider);
```

apps/api/src/deployment/services/trial-providers/trial-providers.service.ts
```diff
@@ -0,0 +1,12 @@
+import { singleton } from "tsyringe";
+
+import { ProviderRepository } from "@src/deployment/repositories/provider/provider.repository";
+
+@singleton()
+export class TrialProvidersService {
+  constructor(private readonly providerRepository: ProviderRepository) {}
+
+  async findTrialProviders(): Promise<string[]> {
```
ygrishajev: I actually like to not use repose in controllers directly. A service can contain more logic later. However I also think it's fine to still do so in cases like this to avoid overthinking. I do not request any changes by this comment :)

apps/api/src/utils/constants.ts
```diff
@@ -23,3 +23,6 @@ export const defaultNodeUrlMapping: { [key: string]: string } = {
 export const apiNodeUrl = env.REST_API_NODE_URL ?? defaultNodeUrlMapping[env.NETWORK] ?? defaultNodeUrlMapping.mainnet;
 export const betaTypeVersion = "v1beta3";
 export const betaTypeVersionMarket = "v1beta4";
+
+export const TRIAL_ATTRIBUTE = "console/trials";
```
ygrishajev: `https://github.com/akash-network/console/tree/main/apps/api/src/deployment/config` pls

apps/deploy-web/src/queries/useBidQuery.ts
```diff
@@ -11,7 +11,7 @@ async function getBidList(apiEndpoint: string, address: string, dseq: string): P
   if (!address || !dseq) return null;
 
   const response = await axios.get(ApiUrlService.bidList(apiEndpoint, address, dseq));
```
ygrishajev:
```suggestion
  const response = await axios.get<{ bids: RpcBid[] }>(ApiUrlService.bidList(apiEndpoint, address, dseq));
```

apps/deploy-web/src/queries/useBidQuery.ts
```diff
@@ -11,7 +11,7 @@ async function getBidList(apiEndpoint: string, address: string, dseq: string): P
...
-  let bids = response.data.bids as RpcBid[];
+  const bids = response.data.bids as RpcBid[];
```
ygrishajev:
```suggestion
  const { bids } = response.data;
```

apps/api/src/deployment/services/trial-providers/trial-providers.service.ts
baktun14: Yea I was thinking the same. I'll leave it like this even if it's just more boilerplate code.

PR #491, opened 2024-11-21
fix(deployment): latest processed height deployment stale

baktun14:


Redm4x (approved):

apps/api/src/chain/repositories/block.repository.ts
```diff
@@ -0,0 +1,11 @@
+import { Block } from "@akashnetwork/database/dbSchemas/base";
+import { singleton } from "tsyringe";
+
+@singleton()
+export class BlockRepository {
+  async findLatestProcessedHeight(): Promise<number> {
+    const height = await Block.max("height", { where: { isProcessed: true } });
```
ygrishajev, line 7: Theoretically it can be null, so makes sense to handle this case. It would be an issue if we, for instance, manage to adjust our apps to boot with empty dbs.

apps/api/src/chain/services/block-http/block-http.service.ts
```diff
@@ -2,14 +2,23 @@ import { BlockHttpService as BlockHttpServiceCommon } from "@akashnetwork/http-s
...
+    private readonly blockRepository: BlockRepository
+  ) {}
 
   @Memoize({ ttlInSeconds: averageBlockTime })
   async getCurrentHeight() {
     return await this.blockHttpService.getCurrentHeight();
   }
+
+  @Memoize({ ttlInSeconds: averageBlockTime })
+  async getLatestProcessedHeight() {
```
ygrishajev: This service is intended to serve as an abstraction over chain rest interface for blocks as suggested by its name. Other http services do the same for other domains. Db dep is breaking the boundaries of concerns here

apps/api/src/deployment/services/stale-managed-deployments-cleaner/stale-managed-deployments-cleaner.service.ts
```diff
@@ -50,7 +50,7 @@ export class StaleManagedDeploymentsCleanerService {
   }
 
   private async cleanUpForWallet(wallet: UserWalletOutput) {
-    const currentHeight = await this.blockHttpService.getCurrentHeight();
+    const currentHeight = await this.blockHttpService.getLatestProcessedHeight();
```
ygrishajev: I would suggest to use repo directly here.

ygrishajev: If the method replaced isn't used, I'd suggest to drop the service.

ygrishajev (changes requested):

apps/api/src/chain/services/block-http/block-http.service.ts
baktun14: Ok I'll use the repo directly in the cleaner service as you suggested.

ygrishajev (approved):
Thx )

PR #499, opened 2024-11-26
Save last ip, user-agent and fingerprint on users

Redm4x:
## Implement basic fingerprinting
- Added the `clientInfoMiddleware` middleware to all routes to add the ip, user-agent and fingerprint to the request context.
- Save the last ip, user-agent and fingerprint in the database (`userSetting`).
- Updated hono to the latest version to get `getConnInfo` from `hono/node-server`.
  - Changes to the `predictedBlockDate`, `predictedDateHeight`, `providerDeployments` and `transactionList` endpoints were to fix typing issues that surfaced after upgrading.

The fingerprinting is done server-side and is inspired by [express-fingerprint](https://github.com/yusukeshib/express-fingerprint). It uses information from the following sources to generate a hash:
- Client IP
- User-Agent
- Accept, Accept-Encoding and Accept-Language headers


apps/api/drizzle/0007_big_black_widow.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE "userSetting" ADD COLUMN "lastIp" varchar(255);--> statement-breakpoint
```
ygrishajev:
```suggestion
ALTER TABLE "userSetting" ADD COLUMN "last_ip" varchar(255);--> statement-breakpoint
```

ygrishajev: cols are named in snake case

apps/api/src/middlewares/clientInfoMiddleware.ts
ygrishajev, line 1: don't you want to implement it this way https://github.com/akash-network/console/blob/main/apps/api/src/core/services/request-context-interceptor/request-context.interceptor.ts ?

apps/api/src/services/db/userDataService.ts
```diff
@@ -86,7 +99,10 @@ export async function getSettingsOrInit({ anonymousUserId, userId, wantedUsernam
           email: email,
           emailVerified: emailVerified,
           stripeCustomerId: null,
-          subscribedToNewsletter: subscribedToNewsletter
+          subscribedToNewsletter: subscribedToNewsletter,
```
ygrishajev:
```suggestion
          subscribedToNewsletter,
```

apps/api/src/services/db/userDataService.ts
```diff
@@ -125,14 +141,26 @@ export async function getSettingsOrInit({ anonymousUserId, userId, wantedUsernam
       email: email,
       emailVerified: emailVerified,
       stripeCustomerId: null,
-      subscribedToNewsletter: subscribedToNewsletter
+      subscribedToNewsletter: subscribedToNewsletter,
```
ygrishajev:
```suggestion
      subscribedToNewsletter,
```

apps/api/src/user/controllers/user/user.controller.ts
```diff
@@ -20,8 +20,12 @@ export class UserController {
     private readonly staleAnonymousUsersCleanerService: StaleAnonymousUsersCleanerService
   ) {}
 
-  async create(): Promise<AnonymousUserResponseOutput> {
-    const user = await this.userRepository.create();
+  async create(ip: string | undefined, userAgent: string | undefined, lastFingerprint: string | undefined): Promise<AnonymousUserResponseOutput> {
+    const user = await this.userRepository.create({
+      lastIp: ip,
+      lastUserAgent: userAgent,
+      lastFingerprint: lastFingerprint
```
ygrishajev:
```suggestion
      lastFingerprint
```

apps/api/src/user/controllers/user/user.controller.ts
```diff
@@ -20,8 +20,12 @@ export class UserController {
     private readonly staleAnonymousUsersCleanerService: StaleAnonymousUsersCleanerService
   ) {}
 
-  async create(): Promise<AnonymousUserResponseOutput> {
-    const user = await this.userRepository.create();
+  async create(ip: string | undefined, userAgent: string | undefined, lastFingerprint: string | undefined): Promise<AnonymousUserResponseOutput> {
```
ygrishajev: let's pass it as an object with params? controller method would generally accept a payload like so

You can access request via execution context here https://github.com/akash-network/console/blob/main/apps/api/src/core/services/request-context-interceptor/request-context.interceptor.ts

baktun14 (commented):
I think this is fine for now, we log this and we can do better analysis on the usage and decide if we want to apply it or not.

baktun14 (approved):

Issue #477, opened 2024-11-14
Improve template data/state management


baktun14:
- [x] Based on the API changes from #102 , modifiy the app to query and display with the new endpoints
  - [x] Template list
  - [x] Template detail

stalniy:
As far as I remember, the only leftover here is to kill the old API endpoint. Or maybe modify it to be an alias to `/templates-list`. @baktun14 wdyt?

baktun14:
> As far as I remember, the only leftover here is to kill the old API endpoint. Or maybe modify it to be an alias to `/templates-list`. [@baktun14](https://github.com/baktun14) wdyt?

Which endpoint are you talking about removing?

stalniy:
I think I created a separate task for this -> https://github.com/akash-network/console/issues/665

baktun14:
Ok cool I'll close this then 👍

PR #506, opened 2024-11-28
refactor(template): utilize new GET /v1/templates/{id} endpoint for template and deployment detail

ygrishajev:
refs #477


apps/deploy-web/src/pages/deployments/[dseq]/index.tsx
```diff
@@ -1,19 +1,25 @@
...
-};
+const contextSchema = z.object({
+  params: z.object({
+    dseq: z.string().regex(/^\d+$/)
+  })
+});
 
-export default DeploymentDetailPage;
+export const getServerSideProps = getValidatedServerSideProps(contextSchema, async ({ params }) => {
+  const remoteDeployTemplate = await services.template.findById(CI_CD_TEMPLATE_ID);
```
Comment (baktun14) [inline apps/deploy-web/src/pages/deployments/[dseq]/index.tsx]: I don't think we necessarily need to have loaded on the server. Especially since it's for a specific feature (Git repo deploy) that not that many deployments have currently.

baktun14 (commented):
Cool! Were you planning on changing the template list endpoint in another pr?

apps/deploy-web/src/pages/deployments/[dseq]/index.tsx
Comment (ygrishajev) [reply apps/deploy-web/src/pages/deployments/[dseq]/index.tsx]: looking into the logic in the components it seems like this template is still needed there. Perhaps this could be changed ofc.

But I was thinking of how often that template is changed? perhaps let's move it to static props and call it a day?

ygrishajev:
> Cool! Were you planning on changing the template list endpoint in another pr?

This PR doesn't complete the issue, yes. So more is to come. May be I'll switch to some other outstanding things though. I was just pushing what's done. It's ready to merge though

apps/deploy-web/src/pages/deployments/[dseq]/index.tsx
Comment (baktun14) [reply apps/deploy-web/src/pages/deployments/[dseq]/index.tsx]: Previously it was loaded through a hook,, but now it'll wait for the template to load to render the page. It should be fast enough that it doesn't affect the ux, but generally I'd put it as a fetch hook in the component.

baktun14:
> > Cool! Were you planning on changing the template list endpoint in another pr?
> 
> This PR doesn't complete the issue, yes. So more is to come. May be I'll switch to some other outstanding things though. I was just pushing what's done. It's ready to merge though

Ok sounds good!

baktun14 (approved):

PR #507, opened 2024-11-29
feat(observability): implements client side logging configuration

ygrishajev:
Stats-web uses experimental app dir. It hydrates client side in a certain way. To make sure logger is instantiated with proper config it should be configured under `use client` directive. Further more, logger instantiated in module scope (outside react context) would actually be instantiated on the backend and would ignore client side config. So to make sure it works, it should be in the context. For this a `useLogger` hook is added.


ygrishajev:
@forbesus FYI

forbesus:
> @forbesus FYI

https://github.com/akash-network/console/pull/462#discussion_r1863468758
how about this?

forbesus:
we can use logger without using custom hook and changing next configuration, i think @ygrishajev

ygrishajev:
> we can use logger without using custom hook and changing next configuration, i think @ygrishajev 

Demo?

package-lock.json
baktun14, line 1: Something wrong with the lock file

baktun14 (approved):

PR #514, opened 2024-12-04
features/credit card banner

baktun14:
- Added credit card payment banner
- Fixed not found template bug
- Removed beta welcome


apps/deploy-web/src/components/layout/Layout.tsx
```diff
@@ -102,42 +87,58 @@ const LayoutApp: React.FunctionComponent<Props> = ({ children, isLoading, isUsin
...
-                <ErrorBoundary FallbackComponent={ErrorFallback}>
-                  {!isUsingSettings || isSettingsInit ? (
-                    !isUsingWallet || isWalletLoaded ? (
-                      <div className={cn({ ["container pb-8 pt-4"]: !disableContainer }, containerClassName)}>{children}</div>
-                    ) : (
-                      <Loading text="Loading wallet..." />
-                    )
+    <div className="h-full">
+      {/* Banner for Free Trial and Credit Card payments */}
+      {hasCreditCardBanner && (
```
ygrishajev: I don't mind using it this way as it's temporal. But it is itching :) like this is a general all pages component and there is smth custom. Needs to be a separate one ideally.

apps/deploy-web/src/components/layout/Nav.tsx
```diff
@@ -24,9 +25,14 @@ export const Nav = ({
   const theme = useCookieTheme();
   const [isSignedInWithTrial] = useAtom(walletStore.isSignedInWithTrial);
   const { user } = useCustomUser();
+  const hasCreditCardBanner = useHasCreditCardBanner();
 
   return (
-    <header className="fixed top-0 z-50 w-full border-b border-border bg-popover dark:bg-background">
+    <header
+      className={cn("fixed top-0 z-50 w-full border-b border-border bg-popover dark:bg-background", {
+        "top-[40px]": hasCreditCardBanner
```
ygrishajev: so now both layout and nav have a dep on this banner. This here should be accepting a `className` prop and be controlled on the parent.

apps/deploy-web/src/components/layout/Sidebar.tsx
```diff
@@ -303,7 +305,8 @@ export const Sidebar: React.FunctionComponent<Props> = ({ isMobileOpen, handleDr
         PaperProps={{
           className: cn("border-none ease z-[1000] bg-header/95 transition-[width] duration-300 box-border overflow-hidden mt-[57px]", {
             ["md:w-[240px]"]: _isNavOpen,
-            ["md:w-[57px]"]: !_isNavOpen
+            ["md:w-[57px]"]: !_isNavOpen,
+            ["h-[calc(100%-40px)] mt-[97px]"]: hasCreditCardBanner
```
ygrishajev: Same, should be accepting a `className`.

apps/deploy-web/src/pages/new-deployment/index.tsx
```diff
@@ -18,11 +18,19 @@ export const getServerSideProps = getValidatedServerSideProps<NewDeploymentConta
...
-  if (template && query.templateId) {
-    return { props: { template, templateId: query.templateId } };
+    if (template && query.templateId) {
+      return { props: { template, templateId: query.templateId } };
+    }
+  } catch (error) {
+    if (error?.response?.status === 404) {
+      console.log("Template not found");
+    } else {
+      console.error(error);
```
ygrishajev: logger?

apps/deploy-web/src/components/layout/Layout.tsx
```diff
@@ -102,42 +87,58 @@ const LayoutApp: React.FunctionComponent<Props> = ({ children, isLoading, isUsin
...
-
-                <ErrorBoundary FallbackComponent={ErrorFallback}>
-                  {!isUsingSettings || isSettingsInit ? (
-                    !isUsingWallet || isWalletLoaded ? (
-                      <div className={cn({ ["container pb-8 pt-4"]: !disableContainer }, containerClassName)}>{children}</div>
-                    ) : (
-                      <Loading text="Loading wallet..." />
-                    )
+    <div className="h-full">
+      {/* Banner for Free Trial and Credit Card payments */}
```
ygrishajev: it says the same with the text below? is this comment needed? )

apps/deploy-web/src/components/layout/Layout.tsx
```diff
@@ -102,42 +87,58 @@ const LayoutApp: React.FunctionComponent<Props> = ({ children, isLoading, isUsin
...
+      {hasCreditCardBanner && (
+        <>
+          <div className="fixed top-0 flex h-[40px] w-full items-center justify-center space-x-4 bg-primary px-3 py-2">
+            <span className="text-sm font-semibold text-white">Credit Card payments are now available!</span>
+
+            {!hasManagedWallet && <ConnectManagedWalletButton className="mb-2 mr-2 w-full md:mb-0 md:w-auto" size="sm" />}
+          </div>
+          <style jsx global>{`
```
ygrishajev: is this styling needed? Flex would do it's job on a permanent basis in a more elegant way. It would also remove the scrollbar where it's not needed.

baktun14: True, didn't think of that!

apps/deploy-web/src/components/layout/Layout.tsx
baktun14: Guess not

apps/deploy-web/src/components/layout/Layout.tsx
baktun14: I can extract it in a component

PR #515, opened 2024-12-04
refactor(billing): keep banner deps in a single component

ygrishajev:


apps/deploy-web/src/components/layout/Layout.tsx
```diff
@@ -87,33 +87,30 @@ const LayoutApp: React.FunctionComponent<Props> = ({ children, isLoading, isUsin
...
       {/* Banner for Free Trial and Credit Card payments */}
       {hasCreditCardBanner && (
         <>
-          <div className="fixed top-0 flex h-[40px] w-full items-center justify-center space-x-4 bg-primary px-3 py-2">
+          <div className="fixed top-0 z-10 flex h-[40px] w-full items-center justify-center space-x-4 bg-primary px-3 py-2">
             <span className="text-sm font-semibold text-white">Credit Card payments are now available!</span>
 
             {!hasManagedWallet && <ConnectManagedWalletButton className="mb-2 mr-2 w-full md:mb-0 md:w-auto" size="sm" />}
           </div>
-          <style jsx global>{`
```
baktun14, line 99: Why did you remove that?

ygrishajev: I might have misunderstood the purpose of it. But if I did, then flex would do this better which I've added. It would also work for any kind of layouts. There's an instant improvement: scroll bar is gone where it isn't needed.

baktun14 (approved):

PR #516, opened 2024-12-04
feat(observability): utilise new logger in stats-web

forbesus:
ref #436

### DIFF
--- a/apps/stats-web/src/app/transactions/[hash]/page.tsx
+++ b/apps/stats-web/src/app/transactions/[hash]/page.tsx
@@ -10,6 +10,7 @@ import PageContainer from "@/components/PageContainer";
 import { Title } from "@/components/Title";
 import { TxMessageRow } from "@/components/transactions/TxMessageRow";
 import { networkId } from "@/config/env-config.schema";
+import { useLogger } from "@/hooks/useLogger";
 import { getSplitText } from "@/hooks/useShortText";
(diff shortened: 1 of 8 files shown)


apps/stats-web/src/app/transactions/[hash]/page.tsx
```diff
@@ -31,18 +32,18 @@ export async function generateMetadata({ params: { hash } }: TransactionDetailPa
   };
 }
 
-async function fetchTransactionData(hash: string, network: Network["id"]): Promise<TransactionDetail | null> {
+async function fetchTransactionData(hash: string, network: Network["id"], logger: ReturnType<typeof useLogger>): Promise<TransactionDetail | null> {
   const apiUrl = serverApiUrlService.getBaseApiUrlFor(network);
-  console.log("DEBUG apiUrl", apiUrl);
+  logger.debug(`DEBUG apiUrl: ${apiUrl}`);
```
Comment (ygrishajev) [inline apps/stats-web/src/app/transactions/[hash]/page.tsx]:
```suggestion
```

ygrishajev (commented):
found another log 
image

also logger needs to be configured on the front end. There should be a file created in `src/config/logger.config.ts` which should call `Logger.configure()` with a log level from env config, should "use client" . That file should be imported somewhere on the in the entry point.

forbesus:
how about now? @ygrishajev

ygrishajev (approved):

Issue #434, opened 2024-10-31
Utilise new logger in indexer


ygrishajev:
Depends on #430

What
- Integrate the shared logging package into the Indexer app.
- Use createOtelLogger for the Indexer (it’s a backend service) to automatically include OpenTelemetry trace context in logs.
- Replace all console.* calls with the bound logger instance.
- Remove redundant/noisy logs; prefer structured logs with an event field and contextual attributes.

Why
- Consistent, structured, and parseable logs across apps.
- OTEL correlation for backend services (traceId, spanId) greatly improves debugging and observability.
- Reduces noise and aligns with practices already used in apps/api and apps/deploy-web.

Examples

apps/api
1) Per-module/service logger with context
```ts
import { LoggerService } from "@akashnetwork/logging";

export class StaleAnonymousUsersCleanerService {
  private readonly logger = LoggerService.forContext(StaleAnonymousUsersCleanerService.name);

  async run() {
    this.logger.debug({ event: "STALE_ANONYMOUS_USERS_CLEANUP_START" });
    // ...
    this.logger.debug({ event: "STALE_ANONYMOUS_USERS_CLEANUP_END" });
  }
}
```

2) Process-level error handling in server bootstrap
```ts
import { LoggerService } from "@akashnetwork/logging";

const logger = LoggerService.forContext("SERVER");

process.on("unhandledRejection", (reason, promise) => {
  logger.error({ event: "UNHANDLED_REJECTION", reason, promise });
});

process.on("uncaughtException", error => {
  logger.error({ event: "UNCAUGHT_EXCEPTION", error });
});
```

3) Hono HTTP logging interceptor
```ts
import { LoggerService } from "@akashnetwork/logging";
import { HttpLoggerIntercepter } from "@akashnetwork/logging/hono";
// ...
const httpLogger = LoggerService.forContext("HTTP");
const httpLoggerInterceptor = new HttpLoggerIntercepter(httpLogger);
// attach interceptor to Hono app as done in apps/api/src/rest-app.ts
```

apps/deploy-web
1) Next.js middleware logger
```ts
import { LoggerService } from "@akashnetwork/logging";

const logger = new LoggerService({ name: "middleware" });

export function middleware(request: NextRequest) {
  logger.info({ message: `Redirecting from ${request.nextUrl.pathname}` });
  // ...
}
```

2) Injecting and using a logger within a service
```ts
import type { LoggerService } from "@akashnetwork/logging";

export class ErrorHandlerService {
[...]

PR #522, opened 2024-12-05
feat(observability): utilise new logger in indexer

forbesus:
ref #434

### DIFF
--- a/apps/indexer/src/chain/chainSync.ts
+++ b/apps/indexer/src/chain/chainSync.ts
@@ -1,6 +1,7 @@
 import { activeChain } from "@akashnetwork/database/chainDefinitions";
 import { Block, Message } from "@akashnetwork/database/dbSchemas";
 import { Day, Transaction } from "@akashnetwork/database/dbSchemas/base";
+import { LoggerService } from "@akashnetwork/logging";
 import { fromBase64 } from "@cosmjs/encoding";
 import { decodeTxRaw } from "@cosmjs/proto-signing";
 import { asyncify, eachLimit } from "async";
@@ -24,6 +25,8 @@ import {
 } from "./dataStore";
 import { nodeAccessor } from "./nodeAccessor";
 import { statsProcessor } from "./statsProcessor";
+
+const logger = LoggerService.forContext("ChainSync");
 
 export const setMissingBlock = (height: number) => (missingBlock = height);
 let missingBlock: number;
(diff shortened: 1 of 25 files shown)


apps/indexer/src/chain/chainSync.ts
```diff
@@ -25,6 +26,8 @@ import {
 import { nodeAccessor } from "./nodeAccessor";
 import { statsProcessor } from "./statsProcessor";
 
+const logger = LoggerService.forContext("apps/indexer/chain/ChainSync.ts");
```
ygrishajev: in the backend it's a bit different. We will consume these in some kind of logging tool like grafana and we have a standard approach to naming. let it be something like a uniqe name in CamelCase. In this case 
```suggestion
const logger = LoggerService.forContext("ChainSync");
```

In case it's a class, it would be a class name.

ygrishajev (commented):
In the browser all of the logs were `debug` coz we need no logs when it's served to a user, only during development. Here all the logs should have relevant log level.

forbesus:
Would you check again? @ygrishajev

forbesus:
Would you check this? @ygrishajev

apps/indexer/src/chain/dataStore.ts
```diff
@@ -6,6 +7,7 @@ import { dataFolderPath } from "@src/shared/constants";
 import { bytesToHumanReadableSize } from "@src/shared/utils/files";
 
 const LevelNotFoundCode = "LEVEL_NOT_FOUND";
+const logger = LoggerService.forContext("dataStore");
```
ygrishajev: let's call all of them in the same notation pls
```suggestion
const logger = LoggerService.forContext("DataStore");
```

apps/indexer/src/chain/chainSync.ts
```diff
@@ -101,13 +104,13 @@ export async function syncBlocks() {
...
+      logger.info("Limiting download to " + maxDownloadGroupSize + " blocks");
       latestBlockToDownload = startHeight + maxDownloadGroupSize;
     }
 
-    console.log("Starting download at block #" + startHeight);
-    console.log("Will end download at block #" + latestBlockToDownload);
-    console.log(latestBlockToDownload - startHeight + 1 + " blocks to download");
+    logger.info("Starting download at block #" + startHeight);
+    logger.info("Will end download at block #" + latestBlockToDownload);
+    logger.info(latestBlockToDownload - startHeight + 1 + " blocks to download");
```
ygrishajev: let's compress the above 3 lines into and logged object:
```ts
logger.info({ event: 'DOWNLOAD', startHeight, endHeight, blocksCount });
```

apps/indexer/src/index.ts
```diff
@@ -27,6 +28,8 @@ const app = express();
 
 const { PORT = 3079 } = process.env;
 
+const logger = LoggerService.forContext("indexer-app");
```
ygrishajev:
```suggestion
const logger = LoggerService.forContext("App");
```

ygrishajev (commented):
@Redm4x @max could one of you verify all the log levels used make sense pls?

forbesus:
please check again @Redm4x @max @ygrishajev

forbesus:
> @Redm4x @max could one of you verify all the log levels used make sense pls?

i think @baktun14 should check this
how about @ygrishajev ?

PR #533, opened 2024-12-09
feat(deployment): auto top up custodial deployment with available amount

### LINKED ISSUE #524: Auto Top Up Custodial Deployment with Available amount
When auto top up job checks for available funds it throws an insufficient error in case there's not enough for the desired period. We need to top up with all available funds in this case. This is done for managed wallets https://github.com/akash-network/console/pull/523 - logic should be similar.

This task also requires to fix a bug:
this check `balances.balance > amount + this.MIN_FEES_AVAILABLE` doesn't make sense when the deployment is in usdc https://github.com/akash-network/console/blob/main/apps/api/src/deployment/services/top-up-custodial-deployments/top-up-custodial-deployments.service.ts#L147
Instead it should fetch akt balance and check it against relevant balance. For akt deployments should prob stay the same.

ygrishajev:
closes #524

### DIFF
--- a/apps/api/src/deployment/services/top-up-custodial-deployments/top-up-custodial-deployments.service.ts
+++ b/apps/api/src/deployment/services/top-up-custodial-deployments/top-up-custodial-deployments.service.ts
@@ -7,6 +7,7 @@ import { BlockHttpService } from "@src/chain/services/block-http/block-http.serv
 import { ErrorService } from "@src/core/services/error/error.service";
 import { TopUpSummarizer } from "@src/deployment/lib/top-up-summarizer/top-up-summarizer";
 import { DrainingDeploymentService } from "@src/deployment/services/draining-deployment/draining-deployment.service";
+import { TopUpCustodialBalanceService } from "@src/deployment/services/top-up-custodial-balance/top-up-custodial-balance.service";
 import { TopUpToolsService } from "@src/deployment/services/top-up-tools/top-up-tools.service";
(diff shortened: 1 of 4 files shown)


apps/api/src/deployment/services/top-up-custodial-deployments/top-up-custodial-deployments.service.ts
```diff
@@ -124,27 +124,50 @@ export class TopUpCustodialDeploymentsService implements DeploymentsRefiller {
...
+  }
+
+  private canTopUp(amount: number, balances: Balances) {
+    if (!amount) {
+      return false;
+    }
+
+    const hasSufficientDeploymentLimit = amount <= balances.deploymentLimit;
+    const hasSufficientFeesLimit = balances.feesLimit >= this.MIN_FEES_AVAILABLE;
+    const hasSufficientFeesBalance = typeof balances.feesBalance === "undefined" || balances.feesBalance >= this.MIN_FEES_AVAILABLE;
```
baktun14, line 167: Not sure I understand the diff between balance and limit.

baktun14 (approved):

apps/api/src/deployment/services/top-up-custodial-deployments/top-up-custodial-deployments.service.ts
ygrishajev: Limit is a grant

PR #557, opened 2024-12-13
refactor(billing): make all transactions via same signer implementation

### LINKED ISSUE #426: Calculate fees in master signing client
While implementing billing features it was easier to have static fees value. It was forgotten and is still static https://github.com/akash-network/console/blob/main/apps/api/src/billing/services/master-signing-client/master-signing-client.service.ts#L138

We need look into this and properly calculate fees if it makes sense.

Acceptance criteria:
- works for both uakt and usdc
- performance is not heavily affected. This can be checked via present functional test

ygrishajev:
closes #426


baktun14 (approved):

baktun14 (approved):
🚀

PR #572, opened 2024-12-18
feat(billing): resolve with valid only grants and allowances from htt…

ygrishajev:
…p service


baktun14 (approved):

packages/http-sdk/src/authz/authz-http.service.ts
```diff
@@ -0,0 +1,136 @@
...
+      await cb(response.grants.filter(grant => this.isValidDepositDeploymentGrant(grant)));
+    } while (nextPageKey);
+  }
+
+  private isValidFeeAllowance({ allowance }: FeeAllowance) {
+    return allowance["@type"] === this.FEE_ALLOWANCE_TYPE && (!allowance.expiration || isFuture(new Date(allowance.expiration)));
+  }
+
+  private isValidDepositDeploymentGrant({ authorization, expiration }: ExactDepositDeploymentGrant) {
+    return authorization["@type"] === this.DEPOSIT_DEPLOYMENT_GRANT_TYPE && (!expiration || isFuture(new Date(expiration)));
```
baktun14, line 134: Technically undefined expiration is not supported with our custom type

ygrishajev: Do u think I should remove the check?

baktun14: I think we can leave it there, it should be fixed in the coming releases. We don't use empty expiration in our codebase for that matter.

PR #634, opened 2025-01-14
refactor(template): replaces /v1/templates with /v1/templates-list

stalniy:
refs #477

What was done:

1. replaces /v1/templates with /v1/templates-list
2. Fixes types in `ApiHttpService`
3. Needed to add `deploy` field in /v1/templates-list  because it's used by [./apps/deploy-web/src/components/sdl/ImageSelect.tsx](https://github.com/akash-network/console/blob/main/apps/deploy-web/src/components/sdl/ImageSelect.tsx#L27) and [apps/deploy-web/src/hooks/useGpuTemplates.ts](https://github.com/akash-network/console/blob/main/apps/deploy-web/src/hooks/useGpuTemplates.ts#L11)
4. Search in templates list was previously done by template `name` or `readme`. And since `readme` is not available in response anymore, I used template's `summary`.

### DIFF
--- a/apps/deploy-web/src/components/templates/TemplateGallery.tsx
+++ b/apps/deploy-web/src/components/templates/TemplateGallery.tsx
@@ -1,15 +1,15 @@
 "use client";
-import { useEffect, useState } from "react";
-import { MdSearchOff } from "react-icons/md";
 import { Button, buttonVariants, Spinner } from "@akashnetwork/ui/components";
 import { cn } from "@akashnetwork/ui/utils";
 import IconButton from "@mui/material/IconButton";
 import TextField from "@mui/material/TextField";
 import { FilterList, Xmark } from "iconoir-react";
 import { useRouter, useSearchParams } from "next/navigation";
+import { useEffect, useState } from "react";
+import { MdSearchOff } from "react-icons/md";
 
 import { LinkTo } from "@src/components/shared/LinkTo";
-import { ApiTemplate } from "@src/types";
+import { TemplateOutputSummaryWithCategory } from "@src/queries/useTemplateQuery";
 import { domainName, UrlService } from "@src/utils/urlUtils";
 import { useTemplates } from "../../context/TemplatesProvider";
 import Layout from "../layout/Layout";
@@ -23,7 +23,7 @@ let timeoutId: NodeJS.Timeout | null = null;
 export const TemplateGallery: React.FunctionComponent = () => {
   const [selectedCategoryTitle, setSelectedCategoryTitle] = useState<string | null>(null);
   const [searchTerms, setSearchTerms] = useState("");
-  const [shownTemplates, setShownTemplates] = useState<ApiTemplate[]>([]);
+  const [shownTemplates, setShownTemplates] = useState<TemplateOutputSummaryWithCategory[]>([]);
   const { isLoading: isLoadingTemplates, categories, templates } = useTemplates();
   const router = useRouter();
   const [isMobileSearchOpen, setIsMobileSearchOpen] = useState(false);
@@ -43,9 +43,9 @@ export const TemplateGallery: React.FunctionComponent = () => {
   }, []);
 
   useEffect(() => {
-    const queryCategory = searchParams?.get("category") as string;
-    const querySearch = searchParams?.get("search") as string;
-    let _templates: ApiTemplate[] = [];
+    const queryCategory = searchParams?.get("category");
+    const querySearch = searchParams?.get("search");
+    let _templates: TemplateOutputSummaryWithCategory[] = [];
 
     if (queryCategory) {
       const selectedCategory = categories.find(x => x.title === queryCategory);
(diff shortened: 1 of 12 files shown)


apps/deploy-web/src/queries/useTemplateQuery.tsx
```diff
@@ -1,14 +1,15 @@
...
 import { useSnackbar } from "notistack";
+import { QueryKey, useMutation, useQuery, useQueryClient, UseQueryOptions, UseQueryResult } from "react-query";
 
 import { useCustomUser } from "@src/hooks/useCustomUser";
+import { services } from "@src/services/http/http-browser.service";
 import { ITemplate } from "@src/types";
-import { ApiUrlService } from "@src/utils/apiUtils";
 import { UrlService } from "@src/utils/urlUtils";
 import { QueryKeys } from "./queryKeys";
+import { TemplateCategory, TemplateOutputSummary } from "@akashnetwork/http-sdk/src/template/template-http.service";
```
ygrishajev: would this be fine?
```suggestion
import type { TemplateCategory, TemplateOutputSummary } from "@akashnetwork/http-sdk";
```

apps/deploy-web/src/components/templates/TemplateGallery.tsx
```diff
@@ -55,8 +55,9 @@ export const TemplateGallery: React.FunctionComponent = () => {
     }
 
     if (querySearch) {
+      // TODO: use minisearch instead https://lucaong.github.io/minisearch/
```
ygrishajev, line 58: nice, would you add an issue?

ygrishajev (approved):

ygrishajev (commented):
One thing i missed from the prev pr prob, why is the endpoint named `template-list`? Shouldn't we continue following the naming approach and name it just `templates`? If it's taken - bump the version?

apps/deploy-web/src/components/templates/TemplateGallery.tsx
baktun14: and once the issue is create, you can implement it :) but I'm curious, what would be the direct benifit of using this library?

apps/deploy-web/src/context/TemplatesProvider/TemplatesProviderContext.tsx
```diff
@@ -1,28 +1,22 @@
...
-  templates: Array<ApiTemplate>;
-  getTemplateById: (id: string) => ApiTemplate;
+  categories: EnhancedTemplateCategory[];
+  templates: TemplateOutputSummaryWithCategory[];
 };
 
 const TemplatesProviderContext = React.createContext<ContextType>({} as ContextType);
 
 export const TemplatesProvider = ({ children }) => {
   const { data, isFetching: isLoading } = useTemplatesQuery();
```
baktun14, line 15: Perhaps we could remove this context provider entirely and only call the query where we need to load the templates. That way it won't add more http requests globally when we might not need it. Wdyt?

stalniy: it's hard to judge from my standpoint since I don't have a full picture of the app. I'd merge this first, so we get performance benefits and only then do more refactoring if needed.

I saw that there are 3 http requests which are calling the same endpoint on page load. This obviously not good and needs to be optimized with caching and probably [async lock](https://developer.mozilla.org/en-US/docs/Web/API/Web_Locks_API) or globally on http service layer with traffic warden 

Screenshot 2025-01-15 at 11 17 12

apps/deploy-web/src/components/templates/TemplateGallery.tsx
stalniy: yes, this is something to discuss, currently the code splits query into search terms but doesn't do term normalization and doesn't ignore stop words. That minysearch library is actually full text search engine that implements inverted index under the hood and has search capabilities like elasticsearch. Usually such libraries are used to search over content of statically built websites

This definitely will improve searching experience, the question is whether we need/want this

apps/deploy-web/src/queries/useTemplateQuery.tsx
stalniy: one more thing for eslint to check. I use autoimporting a lot and do not review what actually was inserted by IDE.

will fix this one, it additionally requires exporting of template service from  `@akashnetwork/http-sdk`

stalniy:
> One thing i missed from the prev pr prob, why is the endpoint named `template-list`? Shouldn't we continue following the naming approach and name it just `templates`? If it's taken - bump the version?

@ygrishajev  We discussed this with @baktun14, you can take a look at his comment [here](https://github.com/akash-network/console/pull/481#issuecomment-2484173504)

apps/deploy-web/src/components/templates/TemplateGallery.tsx
stalniy: https://github.com/akash-network/console/issues/642

apps/deploy-web/src/context/TemplatesProvider/TemplatesProviderContext.tsx
baktun14: I don't mind if you do it in another branch, however I think we should address this now rather than later as minor refactors like this tend to never get done.

baktun14 (approved):

apps/deploy-web/src/context/TemplatesProvider/TemplatesProviderContext.tsx
stalniy: https://github.com/akash-network/console/issues/644

Issue #645, opened 2025-01-16
Add donut charts for persistent and ephemeral storage


anilmurty:
## Problem Statement

Console only shows one storage chart that aggregates both persistent and ephemeral storage across all providers on the network. This was fine until a few months ago when the persistent storage was a small fraction of ephemeral. But over the past year several GPU providers have been onboarded with a lot of persistent storage. It'll be great to separate the two types so it is clear to users how much of each type is available 

## To-Do
 
- [x] Figure out how to utilize data from the new persistent storage tables in ProviderSnapshot (ref: https://github.com/akash-network/console/pull/267)
- [x] Either figure out how to change existing "Storage" donut chart to display both types of storgae in the same (use different color codes and/ or labels to distinguish) or (if too complicated) Add a new donut chart to https://console.akash.network/providers next to existing "Storage" one and identical component to it. Left one will be "Emphemeral Storage" and right one will be "Persistent Storage"
<img width="1151" alt="Screenshot 2025-01-15 at 6 33 10 PM" src="https://github.com/user-attachments/assets/61326c97-2df3-431c-9be8-4b8f95e0cf91" />

anilmurty:
@aktdenis - any thoughts on design for distinguishing persistent and ephemeral storage in the same donut chart?

stalniy:
@anilmurty this is what I was able to achieve locally.

https://github.com/user-attachments/assets/efb10160-1918-49e0-8d08-316ae9372020

stalniy:
@anilmurty we have the same network capacity chart on 2 pages: providers list page and particular provider page. 

I assume that I need to change on both.

aktdenis:
@anilmurty i agree, colors would be one way to achieve differentiation.

@stalniy this could work. i would recommend using different colors. full gray, faded gray and full red and faded red. along that, positioning similar colors next to each other. 

preview:

<img width="445" alt="Image" src="https://github.com/user-attachments/assets/7cb404a4-670f-44c7-a5c0-667bb77563fc" />

also, what do you think about changing the label "active" to "used"?

anilmurty:
@stalniy - that (single char with multiple stats) looks great. And yes please on same for per provider (provider details) page - thanks for noticing.

@aktdenis - "active" and "available" in the convention set previously (for other charts) so if we make the change we will need to make it across the app. I'd suggest we stick to the convention for now and do the change (if we want to) separately (while reviewing other such changes together). 
[...]

PR #647, opened 2025-01-16
feat(stats): shows ephemeral and persistent storage separately on pie chart

stalniy:
refs #645, #665

Minor stuff needs to be finialized from requirements point of view, like colors and arc label positioning

### DIFF
--- a/apps/api/src/services/db/providerStatusService.ts
+++ b/apps/api/src/services/db/providerStatusService.ts
@@ -1,6 +1,7 @@
 import { Provider, ProviderAttribute, ProviderAttributeSignature, ProviderSnapshotNode, ProviderSnapshotNodeGPU } from "@akashnetwork/database/dbSchemas/akash";
 import { ProviderSnapshot } from "@akashnetwork/database/dbSchemas/akash/providerSnapshot";
 import { add, sub } from "date-fns";
+import uniqBy from "lodash/uniqBy";
 import { Op } from "sequelize";
 
 import { ProviderDetail } from "@src/types/provider";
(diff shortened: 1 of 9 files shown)


apps/api/src/services/db/providerStatusService.ts
```diff
@@ -24,31 +25,59 @@ export async function getNetworkCapacity() {
     ]
   });
 
-  const filteredProviders = providers.filter((value, index, self) => self.map(x => x.hostUri).indexOf(value.hostUri) === index);
+  const filteredProviders = uniqBy(providers, provider => provider.hostUri);
+  // TODO: use aggregated stats format per stats entry as in ProviderList['stats']
```
ygrishajev: **suggestion(non-blocking)**: create an issue and include it in here. otherwise a todo might stay here for indefinite time.

```suggestion
  // TODO: use aggregated stats format per stats entry as in ProviderList['stats']
  //    Issue: https://github.com/akash-network/console/issues/${id}
```

ygrishajev (dismissed):

apps/api/src/services/db/providerStatusService.ts
stalniy: makes sense, I created issue. Will update PR https://github.com/akash-network/console/issues/665

apps/deploy-web/src/types/provider.ts
```diff
@@ -208,24 +208,36 @@ export interface ApiProviderList {
   lastOnlineDate: string;
   isAudited: boolean;
   gpuModels: { vendor: string; model: string; ram: string; interface: string }[];
+  /** @deprecated use `stats` instead */
```
baktun14, line 211: **thought:** I was thinking about this for a while, we could create a `types` package and share the same types in both api and frontend without duplication. It would be more efficient and less prone to forget about updating the frontend if we change something in the backend.

baktun14: https://github.com/akash-network/console/issues/667

baktun14 (dismissed):

apps/deploy-web/src/types/provider.ts
stalniy: Maybe instead, we could auto generate a client for our API. We can use routes schema from hono (or resulting OpenAPI spec) in order to generate TS types for client and corresponding calls. What do you think?

stalniy: https://hono.dev/docs/concepts/stacks

ygrishajev (approved):

PR #649, opened 2025-01-16
Feature/turnstile

ygrishajev:


apps/deploy-web/src/config/env-config.schema.ts
```diff
@@ -27,7 +27,7 @@ export const browserEnvSchema = z.object({
   NEXT_PUBLIC_GITLAB_CLIENT_ID: z.string().optional(),
   NEXT_PUBLIC_GITHUB_CLIENT_ID: z.string().optional(),
   NEXT_PUBLIC_CI_CD_IMAGE_NAME: z.string(),
-  NEXT_PUBLIC_TURNSTILE_SITE_KEY: z.string()
+  NEXT_PUBLIC_TURNSTILE_SITE_KEY: z.string().optional()
```
stalniy: **suggestion(non-blocking):**  It makes sense to have predefined value like 'disable' or 'ignore' to disable captcha and ensure that `NEXT_PUBLIC_TURNSTILE_SITE_KEY` is specified and not forgotten.

[Python zen](https://peps.python.org/pep-0020/) :) 
> Explicit is better than implicit

stalniy (approved):

PR #652, opened 2025-01-16
chore(dx): add lint staged and pre-commit formatting and linting

ygrishajev:


packages/dev-config/.eslintrc.base.js
```diff
@@ -7,7 +7,7 @@ module.exports = {
   plugins: ["simple-import-sort"],
   ignorePatterns: ["node_modules", "dist", "build", "public", "Leap"],
   rules: {
-    "@typescript-eslint/no-unused-vars": ["error", { ignoreRestSiblings: true }],
+    "@typescript-eslint/no-unused-vars": ["error", { ignoreRestSiblings: true, argsIgnorePattern: "^_" }],
```
stalniy, line 10: **question(non-blocking):** what is this used for?

stalniy (approved):

packages/dev-config/.eslintrc.base.js
ygrishajev: args like `_` or `_foo` which aren't used

PR #675, opened 2025-01-20
fix(deployment): ensure a single deployment top up error doesn't prevent others

ygrishajev:


apps/api/src/deployment/services/top-up-custodial-deployments/top-up-custodial-deployments.service.ts
```diff
@@ -97,27 +99,37 @@ export class TopUpCustodialDeploymentsService implements DeploymentsRefiller {
...
+      const topUpParams = {
+        dseq,
+        amount,
+        denom,
+        owner,
+        grantee
+      };
+      const logParams = { params: { topUpParams, masterWallet: grantee }, dryRun: options.dryRun };
+
+      try {
```
ygrishajev, line 111: the issue was that without this error handler the loop would stop for the current wallet resulting into the rest not being topped up. same for the other service below

apps/api/src/deployment/services/top-up-custodial-deployments/top-up-custodial-deployments.service.ts
```diff
@@ -41,31 +42,31 @@ export class TopUpCustodialDeploymentsService implements DeploymentsRefiller {
...
-    summary.set("endBlockHeight", await this.blockHttpService.getCurrentHeight());
-    this.logger.info({ event: "TOP_UP_SUMMARY", summary: summary.summarize() });
+    summarizer.set("endBlockHeight", await this.blockHttpService.getCurrentHeight());
+
+    const summary = summarizer.summarize();
+    this.logger.info({ event: "TOP_UP_SUMMARY", summary, dryRun: options.dryRun });
+
+    if (summary.walletsTopUpErrorCount || summary.deploymentTopUpErrorCount) {
+      const eventId = this.sentry.captureEvent(this.sentryEventService.toEvent({ message: "Top up deployments error", summary }));
+      this.logger.debug({ event: "SENTRY_EVENT_REPORTED", eventId });
```
ygrishajev, line 68: make sure, just one sentry event is reported, the errors could be found via logs

baktun14 (approved):

Issue #663, opened 2025-01-16
Rework email verification prompt for payments


ygrishajev:
Current verification prompt is bulky and could be improved in terms of UI. In some places it breaks the layout.

![Image](https://github.com/user-attachments/assets/427612d3-b717-4d0b-ab5b-b148f2503413)

![Image](https://github.com/user-attachments/assets/97685f06-56c1-4bee-a24d-1c7b4e3f2bfd)

This task suggests to rework it in a way similar to [Sign-in prompt](https://github.com/akash-network/console/blob/main/apps/deploy-web/src/hooks/useLoginRequiredEventHandler.tsx). A way to do this may be several composable hooks which could be used on a handler. Another way may be to just update a referenced hook.

![Image](https://github.com/user-attachments/assets/bc31266b-9d62-46f0-b131-4d3cab15b804)

**Acceptance criteria**:

- rework the warning into a modal prompt
- the prompt should have a resend email button (auth0 api call to be implemented on the backend)

PR #676, opened 2025-01-20
feat(auth): implement verification email re-send and rework relevant UI

ygrishajev:
The purpose is described in #663

Now the flow when user clicks `Add Funds` button would be as follows
1) if user is not logged in the one would be prompted to do so as before
2) if user is logged in and not verified the one would be prompted to do so similarly

Screenshot 2025-01-20 at 13 53 18


apps/deploy-web/src/components/user/VerifiedLoginRequiredLink.tsx
```diff
@@ -2,23 +2,23 @@ import React from "react";
...
+> = props => {
+  const whenLoggedInAndVerified = useAddFundsVerifiedLoginRequiredEventHandler();
+
   return props.disabled ? (
     <Button className={props.className} disabled>
       {props.children}
     </Button>
   ) : (
-    <Link {...props} onClick={whenLoggedIn(props.onClick || (() => {}), message)} />
+    <Link {...props} onClick={whenLoggedInAndVerified(props.onClick || (() => {}))} />
```
baktun14, line 22: So everywhere we required login it'll require for email verification as well? I think this makes sense yea

baktun14 (approved):

apps/deploy-web/src/components/user/VerifiedLoginRequiredLink.tsx
ygrishajev: well, everywhere this component is used. there's another one where it's used too.

PR #682, opened 2025-01-21
chore: removes cloud sql proxy from api deployment

stalniy:


ygrishajev (approved):

baktun14 (commented):
I think we should keep this for the user db

stalniy:
OK, then I convert to draft it for now

PR #693, opened 2025-01-21
fix: change logo for dark mode

ShinySyntax:


baktun14 (changes requested):
These are svg's with .png extension.

baktun14:
So these new logos will be shown in both light/dark mode? Also, you need to sign your commits for it to be mergeable.

ShinySyntax:
Sure, these new logos will be shown in both light/ dark mode

baktun14:
How does it look in light mode? Have you tested?

Issue #170, opened 2024-04-24
Validate provider certificate


baktun14:
- In the provider proxy where [we create the https agent](https://github.com/akash-network/console/blob/c5efbff4bb45ea7eac58c174ca1456a70f2ae556/apps/provider-proxy/server.ts#L95), validate the response certificate with the on-chain certificate.
- Same with the [uptime checks](https://github.com/akash-network/console/blob/c5efbff4bb45ea7eac58c174ca1456a70f2ae556/apps/indexer/src/providers/statusEndpointHandlers/grpc.ts#L108) using grpc
- Throw an error if it doesn't match and gracefully handle the error

## Acceptance
- [x] implement validation in provider-proxy for HTTP requests
- [x] implement validation in websocket connection in provider-proxy
- [x] adjust deploy-web to send additional information to provider-proxy in order to do validation
- [ ] implement cert validation in indexer

stalniy:
this is how I see implementation after looking into the code:

1. Fetch provider certificates that is available on chain
2. To do this, I need 2 things: chain api endpoint and provider address
3. I can send chain api endpoint and provider address as part of payload for every provider proxy request (it can either be URL or network name, network name is a safer option)
4. Using chain api endpoint and provider address, I can request valid certificates of this provider
5. validate if request response certificate is one of those fetched and check stuff of whether it's expired/etc (I think `https.Agent` can do it by himself if I pass those certs in `ca` option)

stalniy:
`TLSSocket` has a method to retrieve certificate after connection has been established with remote host https://nodejs.org/api/tls.html#tlssocketgetpeercertificatedetailed

stalniy:
Changes related to provider-proxy has been released today to production

baktun14:
@stalniy Nice! I'm not entirely sure if we need to implement certificate validation in the indexer. We only fetch the version and status. Wdyt?

stalniy:
A note related to implementation of SSL certificates validation in indexer:

The current version of @grpc/grpc-js doesn't support async validation of certificate. I mean that with the entities/configs it provides, there is no way to intercept client request and do validation. 

However, [the latest code state in master](https://github.com/grpc/grpc-node/blob/master/packages/grpc-js/src/transport.ts#L743) has support for this. We can create custom credentials implementation that has own secure connecter which can validate socket in async way.

[...]

PR #696, opened 2025-01-22
refactor: splits provider proxy into services and routes to improve future testing

stalniy:
## Why

Currently the whole codebase is in single file what makes it harder to extend it.  refs #170 

## What

1. Splits codebase into multiple files, keeping functionality as close as possible to original version
2. Adds functional tests to ensure proxy properly works.


stalniy:
Potentially, we will need to use some trusted provider for testing. 

Though it's possible to create self-signed certificate for testing but it's impossible to create CA trusted certificate on letsencrypt if server is not publicly available

ygrishajev (approved):
looks a lot better now!

apps/provider-proxy/src/app.ts
```diff
@@ -0,0 +1,33 @@
+import cors from "cors";
+import express from "express";
+
+import { getAppStatus } from "./routes/getAppStatus";
+import { proxyProviderRequest } from "./routes/proxyProviderRequest";
+
+export const app = express();
+
+const whitelist = [
```
baktun14, line 9: It would be nice to receive those as env vars, it's something we've postponed for a while, but doesn't take too much effort to implement.

apps/provider-proxy/test/functional/provider-proxy.spec.ts
```diff
@@ -0,0 +1,29 @@
+import { request } from "../setup/apiClient";
+
+describe("Provider proxy", () => {
+  it("should proxy request if provider uses self-signed certificate", async () => {
+    const response = await request("/", {
+      method: "POST",
+      body: JSON.stringify({
+        method: "GET",
+        url: "https://provider.tevuicdasialmareny.es:8443/status"
```
baktun14, line 9: who's provider is this?

stalniy: this provider was taken from sandbox

PR #700, opened 2025-01-22
feat: adds certificate validation on provider proxy

stalniy:
## Why

We need to validate provider self-signed certificate

## What 

1. Fetch provider certificates from chain and check whether SSL certificate returned by provider's host is one of those on chain 
2. adds `nodemon` for easier local development
3. removes unused packages
4. creates internal `@akasnetwork/net` package for sharing blockchain configuration
5. added unit and functional tests
6. modifies GHA for provider proxy to run both types of tests 
7. added esbuild to bundle internal packages together with provider-proxy

refs: #170

### DIFF
--- a/apps/provider-proxy/src/app.ts
+++ b/apps/provider-proxy/src/app.ts
@@ -16,6 +16,7 @@ const whitelist = [
   "https://console-beta.akash.network"
 ];
 
+app.disable("x-powered-by");
 app.use(
   cors({
     origin: function (origin, callback) {
--- a/apps/provider-proxy/src/services/ProviderProxy.ts
+++ b/apps/provider-proxy/src/services/ProviderProxy.ts
@@ -1,24 +1,118 @@
+import { SupportedChainNetworks } from "@akashnetwork/net";
+import { X509Certificate } from "crypto";
+import { IncomingMessage } from "http";
 import https, { RequestOptions } from "https";
-import fetch, { BodyInit, Headers, Response } from "node-fetch";
+import { LRUCache } from "lru-cache";
+import { TLSSocket } from "tls";
+
+import { CertValidationResultError, validateCertificate } from "../utils/validateCertificate";
+import { ProviderService } from "./ProviderService";
 
 export class ProviderProxy {
-  fetch(url: string, options: FetchOptions): Promise<Response> {
-    const httpsAgent = new https.Agent({
-      cert: options.cert,
-      key: options.key,
-      rejectUnauthorized: false
-    });
+  private readonly knownCertificatesCache = new LRUCache<string, boolean>({
+    max: 100_000,
+    ttl: 30 * 60 * 1000
+  });
 
-    return fetch(url, {
-      method: options.method,
-      body: options.body,
-      headers: new Headers(options.headers),
-      agent: httpsAgent
+  constructor(
+    private readonly now: () => number,
+    private readonly providerService: ProviderService
+  ) {}
+
+  connect(url: string, options: ProxyConnectOptions): Promise<ProxyConnectionResult> {
+    return new Promise<ProxyConnectionResult>((resolve, reject) => {
+      const req = https.request(
+        url,
+        {
+          method: options.method,
+          headers: {
+            "Content-Type": "application/json",
+            ...options.headers
+          },
+          timeout: options.timeout,
+          cert: options.cert,
+          key: options.key,
+          rejectUnauthorized: false
+        },
+        async res => {
+          try {
+            res.setEncoding("utf8");
+            const socket = res.socket as TLSSocket;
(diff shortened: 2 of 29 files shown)


stalniy:
This PR lacks logic for blockchain API url detection. It's just hardcoded for now. Additionally I need to improve test coverage for `ProviderProxy` class

apps/provider-proxy/src/services/ProviderProxy.ts
```diff
@@ -1,24 +1,151 @@
...
+            // @see https://nodejs.org/api/tls.html#session-resumption
+            // for servers which support TLS session resumption, handshake phase is skipped for subsequent requests
+            // to improve performance and in this case certicate is not available because it is not requested.
+            // There is a way to disable session resumption but it will hurt performance.
+            // To disable either create a new `https.Agent` for every request or reduce sesion related options in it
+            // sessionTimeout & maxCachedSessions. In this case, we will do TLS handshake on every request and
+            // will have certificate for every request
+            const didHandshake = Object.keys(serverCert).length > 0;
+
+            if (didHandshake && options.network && options.providerAddress) {
```
stalniy, line 51: temporary ignore validation if `options.network` and `options.providerAddress` are empty because `deploy-web` also require adjustments, just didn't want to create massive PR

apps/provider-proxy/src/app.ts
```diff
@@ -16,6 +16,7 @@ const whitelist = [
   "https://console-beta.akash.network"
 ];
 
+app.disable("x-powered-by");
```
baktun14, line 19: What does that do?

apps/provider-proxy/src/services/ProviderProxy.ts
```diff
@@ -1,24 +1,161 @@
...
+    const cacheKey = knownCertsCacheKey(network, providerAddress);
+    const cachedCerts = this.knownCertificatesCache.get(cacheKey);
+    if (cachedCerts && !forceRefresh) return cachedCerts;
 
-    return fetch(url, {
-      method: options.method,
-      body: options.body,
-      headers: new Headers(options.headers),
-      agent: httpsAgent
+    const queryParams = `?filter.state=valid&filter.owner=${providerAddress}&pagination.limit=${this.certificatesAmountToFetch}`;
```
baktun14: One thing to note is that it's possible for addresses to have more than 100 certificates. Although very unlikely, it has happened in the past and resulted in hard to find bugs. [This is how](https://github.com/akash-network/console/blob/main/apps/deploy-web/src/utils/apiUtils.ts#L130) we load them in  the frontend and we have the same function in the indexer as well.

apps/provider-proxy/test/seeders/createX509CertPair.ts
```diff
@@ -0,0 +1,46 @@
...
+  cert.publicKey = keys.publicKey;
+  cert.serialNumber = options.serialNumber ?? "01";
+  cert.validity.notBefore = options.validFrom || new Date();
+  cert.validity.notAfter = options.validTo || nextDay(cert.validity.notBefore);
+
+  const attrs = [
+    { name: "commonName", value: options?.commonName ?? "example.org" },
+    { name: "countryName", value: "US" },
+    { shortName: "ST", value: "Virginia" },
+    { name: "localityName", value: "Blacksburg" },
```
baktun14: Where are these values taken from?

baktun14 (commented):
Very well done and solid implementation! 🚀

apps/provider-proxy/src/app.ts
stalniy: this disabled `X-Powered-by` header which express server exposes in headers and actually revealing the technology used by our server 
Screenshot 2025-01-27 at 16 56 06

Though in case of Akash it may not be necessary as this repository is open. But let's not make life of attackers easy anyway :)

apps/provider-proxy/src/services/ProviderProxy.ts
stalniy: I filter by valid certificates only and expected to actually have like 10 or so of them :) 

Is there any reason to have so much valid certificates?

apps/provider-proxy/test/seeders/createX509CertPair.ts
stalniy: This is copy/pasted example values. For us they do not have any value. I think I can remove them and leave only commonName and countryName. 

Thanks for pointing on it

apps/provider-proxy/src/app.ts
baktun14: Ahhh ok yes in our case nothing is hidden indeed but good to know 👍

baktun14 (approved):

PR #706, opened 2025-01-23
feat(provider): added cicd

### LINKED ISSUE #543: Implement CICD for Provider Console
Once we have #539 closed we'd have most of our basic CICD needs covered. At that point we would have only a couple of our services ready and need to spread it to other apps.

Acceptance criteria:
- automated versioning and changelogs should be added to the referred app
- `.github/workflows/create-pre-release-pr.yml` should be updated to also include a referred app
- create `release-${app}.yml` workflow after existing examples
- create workflows to deploy after existing examples

Depends on #539

jigar-arc10:


ygrishajev (approved):

apps/provider-console/package.json
```diff
@@ -8,10 +8,12 @@
     "dev": "next -p 3000",
     "format": "prettier --write ./*.{ts,js,json} **/*.{ts,tsx,js,json}",
     "lint": "eslint .",
+    "release": "release-it",
```
stalniy, line 11: **issue:** you need to change package name to `@akashnetwork/provider-console` otherwise changelog generator will not be able to find corresponding tags and will always generate changelog starting from the 1st commit in this package

PR #733, opened 2025-01-28
feat: adds more reporting data

ygrishajev:


apps/api/src/core/services/http-logger/http-logger.service.ts
```diff
@@ -12,18 +27,35 @@ export class HttpLoggerService implements HonoInterceptor {
...
+            requestUrl: c.req.url,
+            status: c.res.status,
+            referrer: c.req.raw.referrer,
+            protocol: c.req.header("x-forwarded-proto"),
+            duration: `${(performance.now() - timer).toFixed(3)}ms`
+          }
+        };
+
+        if (clientInfo) {
+          log.httpRequest.userAgent = clientInfo.userAgent;
```
baktun14, line 48: Nice that we're using it :D

baktun14 (approved):

Issue #714, opened 2025-01-24
Auto Top-Up Feature Redesign


ygrishajev:
The Auto Top-Up feature redesign aims to provide users with granular control over their deployment funding mechanisms. Instead of the current one-size-fits-all approach where enabling auto top-up affects all deployments, users will have the flexibility to manage top-up settings on a per-deployment basis. This change addresses key user concerns about fund distribution control and prevents scenarios where high-cost deployments could unexpectedly drain wallet resources. The new system will allow users to strategically allocate their funds by choosing which specific deployments should receive automatic funding and for how long, while maintaining the simplicity of the master wallet grant mechanism. For managed wallet users, this introduces needed flexibility in controlling their deployment costs. The redesign also improves visibility into funding status through new notification systems, helping users better manage their deployment resources and prevent unexpected service interruptions.

## Current Behavior
- Auto top-up works as a global setting for all deployments
- Users grant permission to master wallet for deposits
- Managed wallet users have this enabled by default with no opt-out
- All deployments are automatically topped up to maintain a 7-day runtime buffer
- No flexibility in adjusting the runtime buffer period per deployment

## Issues Identified
1. Users want selective control over which deployments get auto top-up and for which expected runtime duration
2. High-cost deployments can drain wallet funds

### 1. Database Changes
Create new table `user_deployment`:
- Fields:
  - `user_id` (FK)
  - `address` (VARCHAR)
  - `deployment_id` (FK)
  - `auto_top_up_enabled` (BOOLEAN)
  - `target_runtime_duration` (BIGINT) (milliseconds)

### 2. API Implementation
New REST endpoint: `/v1/deployments`
- Methods: GET, POST, UPDATE, DELETE
- Query parameter: `userId`, `address`
- Settings should be validated against on-chain data (balance, grant)
- Response should include all user-specific deployment settings

#### Settings Page
- Retain existing top-up grant form
- Add disclaimer about per-deployment settings

#### Deployment Details Page
- Add auto top-up widget
  - Should have a toggle
  - Should have a `target_runtime_duration` modifier
  - Show current auto top-up status

#### Deployment List Page
- Show auto top-up status indicator

#### Client-side Validation
- Verify grant status before enabling
- Check wallet balance against deployment costs
- Validate duration settings

#### Warning System
Implement notifications for:
- Low balance warnings
- Grant expiration alerts
- Multiple deployment funding risks
- Display warning indicators on:
  - Deployment details page
  - Deployment list view

#### Current Behavior Modification
[...]

PR #738, opened 2025-01-28
feat(deployment): reworks top up deployments to rely on db setting

ygrishajev:
refs #714


baktun14 (dismissed):

baktun14 (approved):
Nice! So when users will enable the feature, we'll either create or update the `DeploymentSetting`?

ygrishajev:
> Nice! So when users will enable the feature, we'll either create or update the `DeploymentSetting`?

Yea, will work on that next. I want to update the setting manually and test the job first

PR #739, opened 2025-01-28
chore: db schema docs

baktun14:
The goal is to add the new queries generated to example query when they're not in the codebase and document the schema to improve the query generation with cursor

### DIFF
--- a/packages/database/dbSchemas/akash/akashBlock.ts
+++ b/packages/database/dbSchemas/akash/akashBlock.ts
@@ -4,20 +4,59 @@ import { Column, Table } from "sequelize-typescript";
 import { Block } from "../base";
 import { tableConfig } from "../base/block";
 
+/**
+ * Custom Block model for Akash
+ *
+ * This model extends the base Block model and adds additional columns for Akash specific metrics updated by the indexer for every block.
+ * It makes it easier to query for Akash specific metrics with the most precision.
+ */
 @Table({
   ...tableConfig,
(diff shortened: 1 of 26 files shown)


packages/database/dbSchemas/akash/akashBlock.ts
```diff
@@ -4,20 +4,61 @@ import { Column, Table } from "sequelize-typescript";
...
+ */
 @Table({
   ...tableConfig,
   indexes: [...tableConfig.indexes, { name: "block_totaluusdspent_is_null", unique: false, fields: ["height"], where: { totalUUsdSpent: null } }]
 })
 export class AkashBlock extends Block {
+  /**
+   * Total amount of AKT spent at current block height in uakt
+   */
   @Column(DataTypes.DOUBLE) totalUAktSpent?: number;
```
stalniy, line 21: question(non-blocking): storing money in double will cause 1 cent issues sooner or later. Is it important in this situation? 

**PS**: usually money are stored in currency minor units as integers to mitigate this issues

stalniy (approved):

packages/database/dbSchemas/akash/akashBlock.ts
baktun14: The reason for this if I remember correctly is because it's possible to have fractional uAkt, like `100.5uakt` which makes it possible to have smaller values than the smallest denom. It's used in bids to mitigate against the volatility of AKT, for example if AKT goes to 100$.

PR #740, opened 2025-01-28
feat(provider): changed user field non editable and default to root

### LINKED ISSUE #723: Only Root user allowed to create provider using provider console
Currently, the provider console does not block non-root users. Helm and other provider services require users to be root to perform specific tasks.

Make the user field root and make it non-editable.

jigar-arc10:
close #723


devalpatel67:
LGTM.

devalpatel67 (approved):
LGTM

PR #742, opened 2025-01-29
feat: adjusts logic to send provider and chain details to provider proxy

stalniy:
## Why

refs #170 

## What

1. adds provide proxy service
2. adds `useScopedProviderProxy` which automatically uses selected chain network for subsequent provider proxy requests
3. adds `ServicesProvider` to make it possible to substitute them in unit tests later

### DIFF
--- a/apps/deploy-web/src/context/ServicesProvider/index.tsx
+++ b/apps/deploy-web/src/context/ServicesProvider/index.tsx
@@ -1,0 +1,13 @@
+import React, { useContext } from "react";
+
+import { services as defaultServices } from "@src/services/http/http-browser.service";
+
+const ServicesContext = React.createContext(defaultServices);
+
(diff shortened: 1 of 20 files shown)


apps/deploy-web/src/context/ServicesProvider/index.tsx
```diff
@@ -0,0 +1,13 @@
+import React, { useContext } from "react";
+
+import { services as defaultServices } from "@src/services/http/http-browser.service";
+
+const ServicesContext = React.createContext(defaultServices);
+
+export const ServicesProvider: React.FC<{ children: React.ReactNode; services?: Partial<typeof defaultServices> }> = ({ children, services }) => {
```
ygrishajev, line 7: non-blocking: there are services like this https://github.com/akash-network/console/blob/main/apps/deploy-web/src/hooks/useAuthZService.tsx

perhaps it makes sense to merge them here too at some point

ygrishajev (approved):

Issue #719, opened 2025-01-25
Add test cases to cover full Console API surface (functional testing)


anilmurty:
Current Functional (API) tests are https://github.com/akash-network/console/tree/main/apps/api/test/functional

This task is to add test cases for the rest of the endpoints listed here https://console-api.akash.network/v1/swagger (@ygrishajev is this up to date?)

baktun14:
The endpoints in the swagger docs are public ones without auth. 

I think the most important ones are the managed wallets api services like this one: https://github.com/akash-network/console/blob/main/apps/api/src/deployment/services/top-up-managed-deployments/top-up-managed-deployments.service.spec.ts

ygrishajev:
@baktun14 billing is mostly covered. do we want to specify some endpoints to start with?

ygrishajev:
I think stripe webhook might be a good start. wdut @baktun14 ?

baktun14:
Yea stripe webhook is a critical endpoint and good start @ygrishajev

ygrishajev:
great! @jzsfkzm this can be used for payload signature https://github.com/stripe/stripe-node?tab=readme-ov-file#testing-webhook-signing

PR #744, opened 2025-01-29
test(billing): functional tests for Stripe webhook

jzsfkzm:
refs #719

### DIFF
--- a/apps/api/src/billing/services/stripe-webhook/stripe-webhook.service.ts
+++ b/apps/api/src/billing/services/stripe-webhook/stripe-webhook.service.ts
@@ -36,13 +36,13 @@ export class StripeWebhookService {
       return;
     }
 
-    const checkoutSession = await this.stripe.checkout.sessions.retrieve(event.data.object.id, {
+    const checkoutSession = await this.stripe.checkout.sessions.retrieve(sessionId, {
       expand: ["line_items"]
     });
 
     if (checkoutSession.payment_status !== "unpaid") {
       await this.refillService.topUpWallet(checkoutSession.amount_subtotal, checkoutSessionCache.userId);
-      await this.checkoutSessionRepository.deleteBy({ sessionId: event.data.object.id });
+      await this.checkoutSessionRepository.deleteBy({ sessionId });
     } else {
       this.logger.error({ event: "PAYMENT_NOT_COMPLETED", sessionId });
     }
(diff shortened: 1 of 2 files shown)


apps/api/src/billing/services/stripe-webhook/stripe-webhook.service.ts
```diff
@@ -22,27 +21,26 @@ export class StripeWebhookService {
     this.logger.info({ event: "STRIPE_EVENT_RECEIVED", type: event.type });
 
     if (event.type === "checkout.session.completed" || event.type === "checkout.session.async_payment_succeeded") {
-      await this.tryToTopUpWallet(event);
+      await this.tryToTopUpWallet(event.data.object.id);
```
ygrishajev: non-blocking. this was meant to serve kinda like router. so every method would have a similar interface accepting `Event` and resolving with `void`. Although I see the cause for a change, perhaps it would make sense to extract `sessionId` within the method?

ygrishajev:
blocking: looks like you don't have git hooks installed. I think there was an issue before. You might need to run `npm run prepare` to set husky up. The conventional commit lint wouldn't allow `webhook` scope. There's another one we use for this: `billing`

apps/api/src/billing/services/stripe-webhook/stripe-webhook.service.ts
jzsfkzm: Ah, ok, wasn't aware of this reasoning, just saw it like a spot that can be simplified. Dropping this part of the change then.

apps/api/test/functional/stripe-webhook.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+  const db = container.resolve<ApiPgDatabase>(POSTGRES_DB);
+  const userWalletsQuery = db.query.UserWallets;
+  const checkoutSessionRepository = container.resolve(CheckoutSessionRepository);
+  const dbService = container.resolve(DbTestingService);
+
+  afterEach(async () => {
+    await dbService.cleanAll();
+  });
+
+  const signPayload = (payload: string, secret: string) => {
```
ygrishajev: non-blocking: there is a helper in stripe lib https://github.com/stripe/stripe-node?tab=readme-ov-file#testing-webhook-signing

do you think it could be used?

jzsfkzm: Sure, thanks, using it now. I wasn't aware.

ygrishajev (approved):
lgtm! let's just change commit scope pls and discuss other stu

jzsfkzm:
> lgtm! let's just change commit scope pls and discuss other stu

Ah, I did already, just the PR title needed a manual update too. Now it's also done.

PR #751, opened 2025-01-29
fix(provider): added missing types and added error handling

### LINKED ISSUE #705: Client Error when the user tries to import a different wallet from the one connected to the browser plugin
A couple of providers experienced client error issues when they tried to create a provider multiple times with the same wallet.
This error happens during the last step of provider creation.

Initial assessment - Local storage corruption happened due to multiple provider creation processes. We need more testing to find the root cause.

jigar-arc10:
- fixed missing types in queries
- fixed error-handling issues
- added toast shadcn component

### DIFF
--- a/packages/ui/components/toast.tsx
+++ b/packages/ui/components/toast.tsx
@@ -1,0 +1,95 @@
+"use client";
+
+import * as React from "react";
+import * as ToastPrimitives from "@radix-ui/react-toast";
+import { cva, type VariantProps } from "class-variance-authority";
+import { X } from "lucide-react";
+
+import { cn } from "../utils";
+
+const ToastProvider = ToastPrimitives.Provider;
+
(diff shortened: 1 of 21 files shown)


packages/ui/components/toast.tsx
```diff
@@ -0,0 +1,95 @@
+"use client";
```
baktun14, line 1: We use `notistack` for toasts in deploy-web. I removed the this shadcn component in the beginning because notistack had a better api I think, I'm not 100% sure. We can leave it like for now for efficiency, but let's keep it in mind.

baktun14 (approved):

packages/ui/components/toast.tsx
jigar-arc10: Sure, I will put an issue for change to notistack in future for the provider console.

PR #759, opened 2025-01-30
refactor: simplifies api dbs config and removes redundant connections

ygrishajev:


apps/api/src/db/dbConnection.ts
```diff
@@ -43,28 +29,7 @@ export const chainDb = new Sequelize(csMap[env.NETWORK], {
   models: chainModels
 });
 
-export const chainDbs: { [key: string]: Sequelize } = Object.keys(chainDefinitions)
```
baktun14, line 46: We can remove the non-akash chains from `chainDefinitions`

apps/api/env/.env.unit.test
```diff
@@ -1,5 +1,4 @@
-AKASH_SANDBOX_DATABASE_CS=postgres://postgres:password@localhost:5432/console-akash-sandbox
-USER_DATABASE_CS=postgres://postgres:password@localhost:5432/console-users
```
baktun14, line 2: We could change `POSTGRES_DB_URI` to `USER_POSTGRES_DB_URI`

apps/api/src/db/dbConnection.ts
```diff
@@ -43,28 +29,7 @@ export const chainDb = new Sequelize(csMap[env.NETWORK], {
...
-        logging,
-        logQueryParameters: true,
-        repositoryMode: true,
-        transactionType: DbTransaction.TYPES.IMMEDIATE,
-        define: {
-          timestamps: false,
-          freezeTableName: true
-        },
-        models: getChainModels(chain)
-      })
```
baktun14, line 62: We don't need these configs?

apps/api/env/.env.unit.test
ygrishajev: It's not only users db. It's a more general db for an API as a whole atm. Isn't it?

baktun14: Ok sounds good

apps/api/src/db/dbConnection.ts
ygrishajev: I wasn't able to find any refs

baktun14: Ok let's just make sure it doesn't break anything :)

ygrishajev: Well, tests are passing)

baktun14 (approved):

stalniy:
🥳 🚀

stalniy (approved):

PR #764, opened 2025-01-30
refactor: extracts certificate validation into a separate class

stalniy:
## Why

To reuse for websocket certificate validation, refs #170 

## What

1. extracts certificate validation to a separate class
2. adds more unit tests 
3. removes `cert` and `key` from Proxy agent cache because we do not want to reuse the same client's cert for all provider proxy clients (it's a bug)
4. adds instrumentation to certificate validation for easier debugging on the prod/staging
5. adds async-lock


baktun14 (dismissed):

apps/provider-proxy/src/container.ts
```diff
@@ -13,9 +14,15 @@ const providerService = new ProviderService((network: SupportedChainNetworks) =>
   // @see https://github.com/mswjs/msw/discussions/2416
   return process.env.TEST_CHAIN_NETWORK_URL || netConfig.getBaseAPIUrl(network);
 }, fetch);
-const providerProxy = new ProviderProxy(Date.now, providerService);
+const certificateValidator = new CertificateValidator(
```
baktun14, line 17: Great, we will be able to use it in the indexer as well!

baktun14: Just for context, in the indexer we have scheduled tasks where we query the providers periodically using the status and version endpoint. https://github.com/akash-network/console/blob/main/apps/indexer/src/index.ts#L98

baktun14: Actually, I'm not sure if this is needed since we only check the status and we don't query with TLS requests for a specific deployment/lease.

We use this https agent config though
image

apps/provider-proxy/package.json
```diff
@@ -21,6 +21,7 @@
   },
   "dependencies": {
     "@akashnetwork/net": "*",
+    "async-lock": "^1.4.1",
```
baktun14: **suggestion:** We have https://www.npmjs.com/package/async-sema semaphore in api, we should use the same package in proxy.

stalniy: There is a [standard around async-lock ](https://developer.mozilla.org/en-US/docs/Web/API/Web_Locks_API) which is available in browsers. Very likely it will become a part of nodejs sooner or later in the same way as fetch API did. 

That's why I used async-lock.

baktun14: Ah ok interesting I didn't know. Ok sounds good

baktun14 (dismissed):

apps/provider-proxy/package.json
stalniy: but I agree to keep it consistent and  use async-sema

baktun14 (approved):

PR #771, opened 2025-02-03
feat(deployment): implement deployment settings api

ygrishajev:
refs #714

### DIFF
--- a/apps/api/src/core/services/hono-error-handler/hono-error-handler.service.ts
+++ b/apps/api/src/core/services/hono-error-handler/hono-error-handler.service.ts
@@ -1,4 +1,5 @@
 import { LoggerService } from "@akashnetwork/logging";
+import { ForbiddenError } from "@casl/ability";
 import { context, trace } from "@opentelemetry/api";
 import type { Event } from "@sentry/types";
 import type { Context, Env } from "hono";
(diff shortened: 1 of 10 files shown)


apps/api/src/core/services/hono-error-handler/hono-error-handler.service.ts
```diff
@@ -36,6 +37,10 @@ export class HonoErrorHandlerService {
       return c.json({ error: "BadRequestError", data: error.errors }, { status: 400 });
     }
 
+    if (error instanceof ForbiddenError) {
+      return c.json({ error: "ForbiddenError", message: "Forbidden" }, { status: 403 });
```
stalniy, line 41: thought(non blocking): GitHub for example, returns 404 on private repositories which user don’t have access to. This adds additional level of information hiding

stalniy (approved):

apps/api/src/core/services/hono-error-handler/hono-error-handler.service.ts
ygrishajev: yea, that's true, however this is not like this on endpoints that create records. There you'd actually want 403. If you check out endpoints which include search [here](https://github.com/akash-network/console/pull/771/files#diff-ee10d8e39b9042dcfd16ef6aa2dfbd351912873e7ae0cfc2bdd05bdfe02cd7e2R20) - I do reject with 404

ygrishajev: when using accessibleBy-like logic - it resolves with empty array/null on searches and throws on creation. so manual assertion on queries and this fallback in the global handler kinda make sense to me. got better idea?

baktun14 (approved):

apps/api/src/core/services/hono-error-handler/hono-error-handler.service.ts
stalniy: Nice! I looked into it but missed ) thanks for pointing out

PR #772, opened 2025-02-03
Add transaction events indexing

Redm4x:
- Added indexing of transaction events and attributes
- Added logic to flag deployments & leases as closed when close events are present
- Removed unnecessary dseq cache which was causing memory issues
- Reinitialize indexer cache when a rollback occurs

### DIFF
--- a/apps/indexer/src/chain/chainSync.ts
+++ b/apps/indexer/src/chain/chainSync.ts
@@ -1,6 +1,6 @@
 import { activeChain } from "@akashnetwork/database/chainDefinitions";
 import { Block, Message } from "@akashnetwork/database/dbSchemas";
-import { Day, Transaction } from "@akashnetwork/database/dbSchemas/base";
+import { Day, Transaction, TransactionEvent, TransactionEventAttribute } from "@akashnetwork/database/dbSchemas/base";
 import { fromBase64 } from "@cosmjs/encoding";
 import { decodeTxRaw } from "@cosmjs/proto-signing";
 import { asyncify, eachLimit } from "async";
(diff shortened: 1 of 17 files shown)


apps/indexer/src/chain/chainSync.ts
```diff
@@ -272,7 +295,7 @@ async function insertBlocks(startHeight: number, endHeight: number) {
 
     blocksToAdd.push(blockEntry);
 
-    if (blocksToAdd.length >= 500 || i === endHeight) {
+    if (blocksToAdd.length >= 50 || i === endHeight) {
```
baktun14, line 298: We still want to keep this to 50?

apps/indexer/src/chain/statsProcessor.ts
```diff
@@ -128,6 +128,29 @@ class StatsProcessor {
           ["transactions", "messages", "index", "ASC"]
         ]
       });
+
+      let allTransactionEvents: TransactionEvent[] = [];
+
+      await sequelize.transaction(async dbTransaction => {
+        await sequelize.query("SET LOCAL enable_seqscan = OFF;", { transaction: dbTransaction });
```
baktun14, line 136: It would be nice to add a comment here to explain why we have this

baktun14 (dismissed):

apps/indexer/src/chain/chainSync.ts
Redm4x: Yeah, it keeps the sql transaction size more reasonable now that we also inserts the events & attributes.

baktun14 (approved):

PR #778, opened 2025-02-03
fix(provider): change dynamic message for signing

jigar-arc10:


apps/provider-console/src/utils/walletUtils.ts
```diff
@@ -75,5 +75,7 @@ export function updateLocalStorageWalletName(address: string, name: string) {
 }
 
 export function getNonceMessage(nonce: string, walletAddress: string) {
-  return `provider-console-beta.akash.network wants you to sign in with your Keplr account - ${walletAddress} using Nonce - ${nonce}`;
-}
\ No newline at end of file
+  const domain = window.location.origin;
+  const baseDomain = domain === "http://localhost:3000" ? "https://provider-console-beta.akash.network" : domain;
```
baktun14: This should be an env var so we can easily change between beta and prod

jigar-arc10: added dynamic url for localhost and beta.

baktun14 (approved):

PR #800, opened 2025-02-06
chore(ci): adds unit tests + coverage for deploy-web

stalniy:
## Why

to ensure that changes do not break existing functionality

## What

1. adds unit tests for deploy-web + coverage
2. installed testing-library for testing
3. wrote opinionated unit test with possibility to substitute used components in render function
4. disable eslint warnings printing in ci and lint-staged

### DIFF
--- a/apps/deploy-web/src/components/authorizations/GranteeRow.tsx
+++ b/apps/deploy-web/src/components/authorizations/GranteeRow.tsx
@@ -15,7 +15,7 @@ type Props = {
 
 export const GranteeRow: React.FunctionComponent<Props> = ({ grant }) => {
   const limit = grant?.authorization?.spend_limit;
-  const denomData = limit ? useDenomData(limit.denom) : null;
+  const denomData = useDenomData(limit?.denom);
 
   return (
     <TableRow className="[&>td]:px-2 [&>td]:py-1">
--- a/apps/deploy-web/src/components/providers/ProviderRawData.tsx
+++ b/apps/deploy-web/src/components/providers/ProviderRawData.tsx
@@ -11,11 +11,19 @@ import Layout from "../layout/Layout";
 import { CustomNextSeo } from "../shared/CustomNextSeo";
 import ProviderDetailLayout, { ProviderDetailTabs } from "./ProviderDetailLayout";
 
-type Props = {
-  owner: string;
+export const COMPONENTS = {
+  Layout,
+  CustomNextSeo,
+  ProviderDetailLayout,
+  DynamicReactJson
 };
 
-export const ProviderRawData: React.FunctionComponent<Props> = ({ owner }) => {
+type Props = {
+  owner: string;
+  components?: typeof COMPONENTS;
+};
+
+export const ProviderRawData: React.FunctionComponent<Props> = ({ owner, components: c = COMPONENTS }) => {
   const [provider, setProvider] = useState<Partial<ClientProviderDetailWithStatus> | null>(null);
   const { isLoading: isLoadingProvider, refetch: getProviderDetail } = useProviderDetail(owner, {
     enabled: false,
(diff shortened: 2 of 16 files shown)


apps/deploy-web/src/components/authorizations/GranteeRow.tsx
```diff
@@ -15,6 +15,7 @@ type Props = {
 
 export const GranteeRow: React.FunctionComponent<Props> = ({ grant }) => {
   const limit = grant?.authorization?.spend_limit;
+  // eslint-disable-next-line react-hooks/rules-of-hooks
   const denomData = limit ? useDenomData(limit.denom) : null;
```
stalniy: this is a buggy place, need to fix, because hooks cannot be conditionally called

apps/deploy-web/src/components/providers/ProviderRawData.tsx
```diff
@@ -11,11 +11,19 @@ import Layout from "../layout/Layout";
 import { CustomNextSeo } from "../shared/CustomNextSeo";
 import ProviderDetailLayout, { ProviderDetailTabs } from "./ProviderDetailLayout";
 
+export const COMPONENTS = {
+  Layout,
+  CustomNextSeo,
+  ProviderDetailLayout,
+  DynamicReactJson
+};
```
stalniy, line 19: this is added for easier mocking of component dependencies in unit tests. Follows similar pattern as Vue/Angular component dependencies.

Benefits can be found in apps/deploy-web/tests/unit/ProviderRawData.spec.tsx

**PS:** I'm not a fan of `jest.mock` stuff where we need to provide paths everywhere.

apps/deploy-web/src/components/providers/ProviderRawData.tsx
```diff
@@ -11,11 +11,19 @@ import Layout from "../layout/Layout";
...
+
 type Props = {
   owner: string;
+  components?: typeof COMPONENTS;
```
stalniy, line 23: optional property to pass components used in rendering, fallbacks to default value, so in reality this thing will be only used inside unit tests. Potentially can be used in some advanced cases

apps/deploy-web/src/utils/apiUtils.ts
```diff
@@ -127,7 +127,7 @@ export class ApiUrlService {
 
 // TODO: implement proper pagination on clients
 //   Issue: https://github.com/akash-network/console/milestone/7
-export async function loadWithPagination<T>(baseUrl: string, dataKey: string, limit: number) {
+export async function loadWithPagination<T>(baseUrl: string, dataKey: string, limit: number, httpClient = axios) {
```
stalniy, line 130: dirty hack to pass httpClient from context, so then it can be replaced in unit tests. "dirty" hack because it's a temporary solution until we refactor to services which are provided via single context

package.json
```diff
@@ -28,7 +28,7 @@
   },
   "lint-staged": {
     "*.{js,jsx,ts,tsx}": [
-      "eslint --fix",
+      "eslint --fix --quiet",
```
stalniy, line 31: warnings are not actionable, so no need to show them during commit and in CI logs because they clatter output and with big amount of warnings, it's very hard to find the actual error, especially in GHA logs

apps/deploy-web/src/components/authorizations/GranteeRow.tsx
baktun14: suggestion(blocking): Yea true, can probably be just `const denomData = useDenomData(limit?.denom);`

apps/deploy-web/src/components/providers/ProviderRawData.tsx
baktun14: thought(non-blocking): Interesting, I never used this pattern before. But if it helps for unit testing I'm open to it.

baktun14 (dismissed):

ygrishajev (approved):

Issue #715, opened 2025-01-24
Investigate/implement feature flagging/rollout solution


ygrishajev:
Recently we faced a need to:
- be flexible in toggling some features in the console
- control rollout of some features (e.g. only for testers, some certain users)

There are some ready to use solutions. This task is to investigate potential options, pick one and create and set it up.

Should be followed by another task to integrate it into our apps (console api/web) once done.

ygrishajev:
One potential tool we could use for this is https://www.getunleash.io/. I have been using it in the past and can give a positive feedback about it. The functionality of it is pretty rich. Some of it which I used:
1) possibility to create user segments to roll out features to
2) rolling feature releases
3) feature toggles
4) A/B testing
4) can be deployed on premises. there're helm charts which should make kick off easy https://github.com/Unleash/helm-charts

ygrishajev:
ok, unleash it is

PR #801, opened 2025-02-06
feat(deployment): adds auto top feature flag

ygrishajev:
refs #714


apps/deploy-web/env/.env.staging
```diff
@@ -6,6 +6,7 @@ NEXT_PUBLIC_TURNSTILE_SITE_KEY=0x4AAAAAAA5Stdp3_0Q5FClX
 
 NEXT_PUBLIC_DEFAULT_NETWORK_ID=mainnet
 NEXT_PUBLIC_BILLING_ENABLED=true
+NEXT_PUBLIC_AUTO_TOP_UP_ENABLED=true
```
stalniy, line 9: **thought(non-blocking):** I think it makes sense to create a service on UI which requests feature flags from API and then we can adapt it over the app and later choose some platform for feature flags.

Anyway, it's better to make integration via our API because their API may be well known and blocked in ad blockers

stalniy (approved):

apps/deploy-web/env/.env.staging
ygrishajev: it totally does https://github.com/akash-network/console/issues/715

Issue #628, opened 2025-01-11
Allow credit card users to add funds at deploy time


anilmurty:
## Problem Statement

Deployments created by credit card users default to $5 in the escrow, which (depending on resources leased) can sometimes allow the deployment to run for only a few minutes. Users have the option to "Add Funds" but sometimes forget to do so and have their lease closed.

## Proposed Solution

Allow credit card users to optionally add/ increase the funds that go into escrow at deploy time. 

The changes would be
1. Show a text box (similar to the one we show for wallet users) where a default value of $5 is populated and the user has the option of increasing it. This would be added under the resources box shown below
![Screenshot 2025-01-10 at 4 46 34 PM](https://github.com/user-attachments/assets/5ed4a9da-f6ac-4eaf-aab7-7fa4c5be99a9)

2. Check credit balance to ensure there are funds to fund the escrow account.
3. Give user the option to purchase more funds

One thing I realized @aktdenis  @baktun14 @ygrishajev is that we call both "Adding funds to the account" and "adding funds to the deployment" as "Add Funds" - we may need some distinction between the two

aktdenis:
@anilmurty 
how about "Top Up Deployment"?... or the other way around and say "Top Up Account"?

anilmurty:
hey @aktdenis - those work but I think they may be a tad long. After thinking over this a bit, I'm thinking we leave this one as Add "Funds" and change the credit purchase one to "Buy Credits" - wdyt? - this way both are just 2 words

![Image](https://github.com/user-attachments/assets/cdccf3c0-05db-44f3-8f99-355bde67dbb1)

aktdenis:
hey @anilmurty, good point. it's distinguishable and works for me

jzsfkzm:
Hi there, I created a PR for this. I thought it would make sense to unify ManagedDeploymentConfirm and DeploymentDepositModal. In this change we'd be rendering the list of services on DeploymentDepositModal and use that modal for managed wallets. After this the experience would be really similar for users with managed or custodial wallets.

@aktdenis I added the list of services inside a scrollable container, is that ok for you? My feeling was we should not let deployments with multiple (n>5) services push the modal over the limits. Here's a screenshot about:
<img width="534" alt="Image" src="https://github.com/user-attachments/assets/1e401bc7-f247-4b6b-a211-5803163cec53" />

[...]

PR #804, opened 2025-02-06
feat(billing): unify deployment modals

jzsfkzm:
refs #628

### DIFF
--- a/apps/deploy-web/src/components/deployments/DeploymentDepositModal.tsx
+++ b/apps/deploy-web/src/components/deployments/DeploymentDepositModal.tsx
@@ -29,14 +29,13 @@ import { UAKT_DENOM } from "@src/config/denom.config";
 import { usePricing } from "@src/context/PricingProvider";
 import { useSettings } from "@src/context/SettingsProvider";
 import { useWallet } from "@src/context/WalletProvider";
-import { useUser } from "@src/hooks/useUser";
+import { useAddFundsVerifiedLoginRequiredEventHandler } from "@src/hooks/useAddFundsVerifiedLoginRequiredEventHandler";
 import { useDenomData, useWalletBalance } from "@src/hooks/useWalletBalance";
 import { useGranteeGrants } from "@src/queries/useGrantsQuery";
 import { analyticsService } from "@src/services/analytics/analytics.service";
 import { ServiceType } from "@src/types";
 import { denomToUdenom, udenomToDenom } from "@src/utils/mathHelpers";
 import { coinToUDenom } from "@src/utils/priceUtils";
-import { UrlService } from "@src/utils/urlUtils";
 import { LeaseSpecDetail } from "../shared/LeaseSpecDetail";
 import { LinkTo } from "../shared/LinkTo";
 import { GranteeDepositMenuItem } from "./GranteeDepositMenuItem";
@@ -103,10 +102,12 @@ export const DeploymentDepositModal: React.FunctionComponent<DeploymentDepositMo
   const { handleSubmit, control, watch, setValue, clearErrors, unregister } = form;
   const { amount, useDepositor, depositorAddress } = watch();
   const validGrants = granteeGrants?.filter(x => compareAsc(new Date(), new Date(x.expiration)) !== 1 && x.authorization.spend_limit.denom === denom) || [];
-  const user = useUser();
+  const whenLoggedInAndVerified = useAddFundsVerifiedLoginRequiredEventHandler();
 
-  const goToSignIn = () => {
-    window.location.href = UrlService.login();
+  const closePopupAndGoToCheckoutIfPossible = (event: React.MouseEvent) => {
+    handleCancel();
+
+    whenLoggedInAndVerified(goToCheckout)(event);
   };
 
   const goToCheckout = () => {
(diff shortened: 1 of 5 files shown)


baktun14:
https://github.com/akash-network/console/issues/628#issuecomment-2641016269

baktun14:
Do we need to remove https://github.com/akash-network/console/blob/8fc4523034e06036bddc75378d1503d7ba8ead64/apps/deploy-web/src/hooks/useManagedDeploymentConfirm.tsx#L39?

apps/deploy-web/src/components/deployments/DeploymentDepositModal.tsx
```diff
@@ -225,10 +242,29 @@ export const DeploymentDepositModal: React.FunctionComponent<DeploymentDepositMo
...
+      {services.length > 0 && (
+        <div className="mb-3 max-h-[300px] overflow-scroll">
+          {services.map(service => {
+            return (
+              <Alert key={service.title} className="mb-1">
+                <div className="mb-2 text-sm">
+                  <span className="font-bold">{service.title}</span>:{service.image}
+                </div>
+                <div className="flex items-center space-x-4 whitespace-nowrap">
+                  <LeaseSpecDetail type="cpu" className="flex-shrink-0" value={service.profile?.cpu as number} />
```
ygrishajev: non-blocking: perhaps it's just copy/paste. seems like type casting isn't necessary here
```suggestion
                  <LeaseSpecDetail type="cpu" className="flex-shrink-0" value={service.profile?.cpu} />
```

apps/deploy-web/src/components/deployments/DeploymentDepositModal.tsx
```diff
@@ -225,10 +242,29 @@ export const DeploymentDepositModal: React.FunctionComponent<DeploymentDepositMo
...
+                  {service.profile?.hasGpu && <LeaseSpecDetail type="gpu" className="flex-shrink-0" value={service.profile?.gpu as number} />}
```
ygrishajev: non-blocking: If done like so type casting doesn't seem to be necessary
```suggestion
                  {!!service.profile?.gpu && <LeaseSpecDetail type="gpu" className="flex-shrink-0" value={service.profile?.gpu} />}
```

apps/deploy-web/src/components/deployments/DeploymentDepositModal.tsx
```diff
@@ -259,6 +295,17 @@ export const DeploymentDepositModal: React.FunctionComponent<DeploymentDepositMo
                 );
               }}
             />
+            <div className="mt-1 flex justify-end text-xs">
+              {user?.userId ? (
+                <LinkTo onClick={() => goToCheckout()} className="text-primary">
```
ygrishajev: blocking: only users which are logged in and verified can use the checkout and top up. we already have utilities for this. e.g. here in [YourAccount](https://github.com/akash-network/console/blob/main/apps/deploy-web/src/components/home/YourAccount.tsx#L235)

ygrishajev: if done like so the link to sign might not be needed

ygrishajev: another concern: is this visible to all the users? should only be for managed wallet users

ygrishajev (changes requested):

apps/deploy-web/src/components/deployments/DeploymentDepositModal.tsx
jzsfkzm: Nice, thanks, fixed. It was indeed copy-paste from `createDeploymentConfirm`.

apps/deploy-web/src/components/deployments/DeploymentDepositModal.tsx
jzsfkzm: I found this the other day but then took the wrong decision not using, solely because I felt like switching from one modal to another would look weird. I implemented it now and it doesn't look weird, feels ok instead.
Other solution would be to render the link only when the user is logged in and is verified.

ygrishajev: I was not thinking about multiple modals honestly. There might be a better decision. However what's important is the logic that user has to login and verify email to be able to top up

ygrishajev:
It feels a bit weird to have "buy more credits" not along "balance". wdut?

Image

apps/deploy-web/src/components/sdl/RentGpusForm.tsx
```diff
@@ -298,13 +292,23 @@ export const RentGpusForm: React.FunctionComponent = () => {
...
-                To create a deployment, you need to have at least <b>{minDeposit.akt} AKT</b> or <b>{minDeposit.usdc} USDC</b> in an escrow account.{" "}
+                {isManaged ? (
+                  <>
+                    To create a deployment, you need to have at least <b>${depositData?.min}</b> in an escrow account.{" "}
+                  </>
+                ) : (
+                  <>
+                    To create a deployment, you need to have at least <b>{minDeposit.akt} AKT</b> or <b>{minDeposit.usdc} USDC</b> in an escrow account.{" "}
+                  </>
+                )}
```
baktun14: This is repeated twice, perhaps we could have a hook that returns this message in both places

jzsfkzm: Right, thanks, done.

apps/deploy-web/src/hooks/useDeploymentMinimumEscrowAlertText.tsx
```diff
@@ -0,0 +1,24 @@
...
+export const useDeploymentMinimumEscrowAlertText = () => {
+  const { isManaged } = useWallet();
+  const [sdlDenom] = useState("uakt");
+  const depositData = useDenomData(sdlDenom);
+  const { minDeposit } = useChainParam();
+
+  return useMemo(() => {
+    return isManaged ? (
+      <>
+        To create a deployment, you need to have at least <b>${depositData?.min}</b> in an escrow account.{" "}
```
baktun14: I didn't think there would be rendering involved, but in this case I think it makes sense to make it a component instead of hook. `<DeploymentMinimumEscrowAlertText />`

baktun14 (dismissed):

baktun14:
> It feels a bit weird to have "buy more credits" not along "balance". wdut?
> 
> Image

I suggest we put this "Buy credits" CTA to the bottom of the deposit modal along the other actions.

baktun14 (approved):

ygrishajev (approved):

baktun14:
Thanks for you work @jzsfkzm !

PR #809, opened 2025-02-06
feat(deployment): marks closed deployments settings

ygrishajev:


apps/api/drizzle/0012_quiet_exiles.sql
```diff
@@ -0,0 +1,2 @@
+DROP INDEX IF EXISTS "auto_top_up_enabled_id_idx";--> statement-breakpoint
```
baktun14: question(non-blocking))): It could be done in the same migration above?

baktun14 (dismissed):

baktun14:
Ah so if a deployment was closed we didn't handle disabling the deployment setting?

apps/api/drizzle/0012_quiet_exiles.sql
ygrishajev: It could, why is it blocking? It's totally fine this way too imo

ygrishajev:
> Ah so if a deployment was closed we didn't handle disabling the deployment setting?

Yes, but i was planning to finish this in the end from the start

apps/api/drizzle/0012_quiet_exiles.sql
baktun14: It's not really blocking, it does work for sure, I just think for a single pr schema change doing just one migration makes more sense.

stalniy: well, for databases which does not support Transactions for DDL changes, it's actually critical to have 1 migration per transactional statement, otherwise you will need to do migration manually if it fails somewhere in the middle.

BUT Postgres actually supports transactions for DDL changes and drizzle-kitt automatically wraps it in transaction. So, having multiple DDL changes in single file is OK.

To solve the issue of conflicts in file names, we need to use timestamp instead of sequential prefix -> Screenshot 2025-02-07 at 04 40 58

Having timestamps in prefix, we will not have collisions and will not actually care about how many files are used for migration

stalniy (approved):

PR #813, opened 2025-02-07
fix(provider): better error handling on non provider calls (404 error)

### LINKED ISSUE #811: Provider Console Error msg when user connects a wallet without a provider
## Problem Statement

This error message is displayed in the bottom right for a few seconds on many pages when they load if the wallet connected is not associated with a provider

Image

The UI is trying to fetch data about the provider (like dash, attributes etc) and doesn't get it. 

## Solution

I'd suggest replacing it with a more graceful message that says "no provider data found" and "likely because this wallet does not have a provider set up yet"

Alternatively, we can suppress the error message entirely in cases where there is no provider found and instead add an empty state to all pages that says "No Provider Associated with this Wallet. Click on [Create Provider](https://provider-console.akash.network/become-provider) to set up a provider or connect a different wallet that is associated with a provider"

Thoughts @aktdenis ?

jigar-arc10:
closes #811


devalpatel67 (approved):
LGTM

PR #819, opened 2025-02-08
feat(deployment): tops up deployments for the same owner in a single tx

ygrishajev:
refs #714

### DIFF
--- a/apps/api/src/deployment/services/top-up-managed-deployments/top-up-managed-deployments.service.ts
+++ b/apps/api/src/deployment/services/top-up-managed-deployments/top-up-managed-deployments.service.ts
@@ -1,19 +1,29 @@
 import { LoggerService } from "@akashnetwork/logging";
+import groupBy from "lodash/groupBy";
 import { singleton } from "tsyringe";
 
 import { InjectWallet } from "@src/billing/providers/wallet.provider";
-import { DepositDeploymentMsgOptions, RpcMessageService, Wallet } from "@src/billing/services";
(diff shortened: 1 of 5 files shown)


apps/api/src/deployment/services/top-up-managed-deployments/top-up-managed-deployments.service.ts
```diff
@@ -26,61 +36,115 @@ export class TopUpManagedDeploymentsService implements DeploymentsRefiller {
...
+      deployments.map(async deployment => {
+        this.summarizer.inc("deploymentCount");
+        this.summarizer.trackWallet(deployment.address);
+
+        try {
+          const { address, predictedClosedHeight } = deployment;
+          this.summarizer.ensurePredictedClosedHeight(predictedClosedHeight);
+
+          const balance = await this.cachedBalanceService.get(address);
+          const desiredAmount = await this.drainingDeploymentService.calculateTopUpAmount(deployment);
```
stalniy, line 69: question(non-blocking): can be parallelized?

stalniy (approved):

baktun14 (approved):

apps/api/src/deployment/services/top-up-managed-deployments/top-up-managed-deployments.service.ts
ygrishajev: I'm not sure I follow. This whole thing is parallelized. This is a callback of a `map` inside a `Promise.all`

PR #823, opened 2025-02-10
chore: improves constructor types of BaseRepository in api

stalniy:
## Why

To make coding less error prone. Last parameter is now corresponds to the table type which is passed as the 2nd argument:

Screenshot 2025-02-10 at 17 41 38

## What

improves constructor types of BaseRepository in api


ygrishajev (approved):
🔥

baktun14 (approved):

Issue #760, opened 2025-01-30
Improve api functional tests performance


ygrishajev:
Currently we have to run tests sequentially utilising jest's `--runInBand` option. This is to avoid DB data checks and master wallet sequence mismatch. It would be great to address these issues and run tests in parallel. This way we'd save our own time running tests locally, consume less of CI resources and reduce PR checks time.

DB acceptance criteria:
- create new dbs for every test run
- override relevant env vars
- drop dbs after finish

Master wallet acceptance criteria:
- generate a new wallet for a test run
- send tokens to the every wallet created
- override relevant env vars

ygrishajev:
As DB part is pretty trivial, wallet part might need some thinking. Some points I'm concluding so far:

- It might be a good idea to store wallets into some json file checked into git, to avoid wallet creation on every run
- The above also would reduce frequency of funding if implemented properly
- Every run should lock the cached wallet

PR #825, opened 2025-02-10
test(wallet): add functional tests for /blocks and /transactions

jzsfkzm:
refs #719


apps/api/test/functional/transactions.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+        method: "GET",
+        headers: new Headers({ "Content-Type": "application/json" })
+      });
+
+      expect(response.status).toBe(400);
+    });
+  });
+
+  describe("GET /v1/transactions/{hash}", () => {
+    it("resolves trancation by hash", async () => {
```
ygrishajev:
```suggestion
    it("resolves transaction by hash", async () => {
```

ygrishajev:
I think tests are failing in the CI as CI db doesn't have data. Let's try removing [this var](https://github.com/akash-network/console/blob/main/.github/workflows/validate-n-build-api.yml#L55)

baktun14 (dismissed):

baktun14:
Yea that's a good point @ygrishajev we should focus on tests that don't require indexer data and that ultimately will need to be refactored in the module architecture. We do have [this task](https://github.com/akash-network/console/issues/760) that we should resolve before adding more tests.

ygrishajev:
@jzsfkzm I think with the latest migration additions on the indexer this ~~not~~now can be adjusted and used

baktun14 (approved):

baktun14:
> @jzsfkzm I think with the latest migration additions on the indexer this not can be adjusted and used

Is this resolved? I'm not entirely sure I understand the meaning

jzsfkzm:
> > @jzsfkzm I think with the latest migration additions on the indexer this not can be adjusted and used
> 
> Is this resolved? I'm not entirely sure I understand the meaning

I guess there was a typo, @ygrishajev meant "now", not "not". Made the changes he suggested, I think it's fine now, what do you think @ygrishajev?

PR #831, opened 2025-02-11
refactor: changes structure and reduce side-effects in provider proxy

stalniy:
## Why

To improve the structure of the project

## What

1. changes structure and reduce side-effects in provider proxy
2. adds opentelemetry but it doesn't work for websocket. Will do it in a separate task
3. fixes bug of proxying request via hono


apps/provider-proxy/src/container.ts
```diff
@@ -2,33 +2,42 @@ import { LoggerService } from "@akashnetwork/logging";
...
-const certificateValidator = new CertificateValidator(
-  Date.now,
-  providerService,
-  process.env.NODE_ENV === "test" ? undefined : createCertificateValidatorInstrumentation(new LoggerService({ name: "cert-validator" }))
-);
-const providerProxy = new ProviderProxy(certificateValidator);
-const httpLogger = new LoggerService({ name: "http-proxy" });
-const createWsLogger = process.env.NODE_ENV === "test" ? undefined : (context: string) => new LoggerService({ name: "ws-proxy" }).setContext(context);
+export function createContainer() {
+  const isLoggingDisabled = process.env.NODE_ENV === "test";
```
ygrishajev, line 11: I think it's better not to rely on this kind of logic. Logging should have it's own flag. You might need logs in tests sometimes.
All it takes is another `.env.test` file which imo is more declarative and clean approach. If there's no common way to load env files yet, ok, this is non-blocking.

apps/provider-proxy/src/services/HttpLoggerInterceptor.ts
```diff
@@ -0,0 +1,62 @@
...
+    referrer?: string;
+    protocol?: string;
+    remoteIp?: string;
+    duration: string;
+  };
+  fingerprint?: string;
+  userId?: string;
+};
+
+export class HttpLoggerInterceptor implements HonoInterceptor {
```
ygrishajev: perhaps it's a good idea to extract this into logging package to avoid duplicating?

apps/provider-proxy/src/container.ts
```diff
@@ -2,33 +2,42 @@ import { LoggerService } from "@akashnetwork/logging";
...
-}, fetch);
-const certificateValidator = new CertificateValidator(
-  Date.now,
-  providerService,
-  process.env.NODE_ENV === "test" ? undefined : createCertificateValidatorInstrumentation(new LoggerService({ name: "cert-validator" }))
-);
-const providerProxy = new ProviderProxy(certificateValidator);
-const httpLogger = new LoggerService({ name: "http-proxy" });
-const createWsLogger = process.env.NODE_ENV === "test" ? undefined : (context: string) => new LoggerService({ name: "ws-proxy" }).setContext(context);
+export function createContainer() {
```
ygrishajev, line 10: this getting more and more into it. Perhaps using a container like in api would be a cleaner approach. non-blocking

apps/provider-proxy/src/services/HttpLoggerInterceptor.ts
ygrishajev: non-blocking, but let's discuss and align on this

apps/provider-proxy/src/container.ts
stalniy: just to ensure that open-telementry can do its job before I instantiate some service. just function wrapper.

apps/provider-proxy/src/services/HttpLoggerInterceptor.ts
stalniy: yes, this was one of the reasons why I wanted to migrate this service to Hono. But in api project there is additional decorator on this class and here we do not have this dependency injection library.

apps/provider-proxy/src/container.ts
ygrishajev: decorator can be applied in the api or provided with a custom provider. there's similar approach with https://github.com/akash-network/console/blob/main/apps/api/src/core/providers/http-sdk.provider.ts

apps/provider-proxy/src/services/HttpLoggerInterceptor.ts
ygrishajev: decorator can be applied in the api or provided with a custom provider. there's similar approach with https://github.com/akash-network/console/blob/main/apps/api/src/core/providers/http-sdk.provider.ts

apps/provider-proxy/src/container.ts
stalniy: Agree, it's a dirty hack

ygrishajev (approved):

PR #850, opened 2025-02-14
feat(analytics): integrates amplitude

ygrishajev:
also hides analytic reporters behind a single interface


baktun14 (approved):

anilmurty:
awesome! 
For a moment I thought you are disabling GA but seems like it's only disabled for dev and staging but not for prod?

ygrishajev:
> awesome! 
> 
> For a moment I thought you are disabling GA but seems like it's only disabled for dev and staging but not for prod?

Yes. GA stays the same as before. I've just added explicit toggles

Issue #767, opened 2025-01-31
(V1) Implement API endpoints to manage deployments protected by the API key


baktun14:
For each of the following endpoint api-key protected endpoints, we want to abtract away all the blockchain logic of creating the proper transaction payload to sign a transaction. We have example code in both console and [akashjs](https://github.com/akash-network/akashjs/blob/main/examples/create_deployment.ts)

These should be done individually 

  - [x] Create certificate
    - POST /v1/certificates
    - returns base64 encoded certificate (token)
  - [x] Create deployment
    - POST /v1/deployments
      - Parse and validate SDL
      - Create transaction payload [example](https://github.com/akash-network/console/blob/main/apps/api/test/functional/create-deployment.spec.ts#L48)
      - Sign transaction
    - Payload: technically we should only need the SDL
```js
{
  sdl: string
}
```
  - [x] Get bids
    - GET /v1/bids/{dseq}
    - proxy the request to a rpc using the address of the user account associated with the api key
  - [x] Create lease
    - POST /v1/leases
      - Create lease payload [example](https://github.com/akash-network/console/blob/c9d3b4528a7a1e864f98541d1a93fbf7622f7cde/apps/deploy-web/src/components/new-deployment/CreateLease.tsx#L202)
      - Sign transaction
      - Send manifest
    - Payload:
 ```js
{
  sdl: string,
  dseq: string,
  gseq: string,
  oseq: string,
  provider: string,
  token: string // (base64 cert)
}
```
  - [x] Deployment details
    - GET /v1/deployments/{dseq}
    - Formulate a response with data coming from db indexer or rpc node with the following example return payload
```js
{
  deployment: DeploymentDetails,
  leases: LeaseDetails[],
  escrow: EscrowBalance
}
```
  - [x] Close Deployment
    - DELETE /v1/deployments/{dseq}

ygrishajev:
I'd suggest a more unified interface like:

POST /v1/deployments
GET /v1/bids/{dseq}
POST /v1/leases

baktun14:
@ygrishajev makes sense

baktun14:
@jzsfkzm I will work on the `/v1/certificates` endpoint

PR #859, opened 2025-02-18
feat(network): API endpoint listing bids

jzsfkzm:
refs #767


apps/api/src/bid/services/bid.service.ts
```diff
@@ -0,0 +1,15 @@
...
+import { singleton } from "tsyringe";
+
+import { RestAkashBidListResponseType } from "@src/types/rest";
+import { apiNodeUrl } from "@src/utils/constants";
+import { ListBidsResponse } from "../http-schemas/bid.schema";
+
+@singleton()
+export class BidService {
+  public async list(owner: string, dseq: string): Promise<ListBidsResponse['data']> {
+    const response = await axios.get<RestAkashBidListResponseType>(`${apiNodeUrl}/akash/market/v1beta4/bids/list?filters.owner=${owner}&filters.dseq=${dseq}`);
```
baktun14: Let's extract this in the [http package](https://github.com/akash-network/console/tree/f1a0c32422b3716010f936aa7d0a4abe38ee4762/packages/http-sdk)

apps/api/src/bid/routes/bids/bids.router.ts
```diff
@@ -0,0 +1,34 @@
+import { createRoute } from "@hono/zod-openapi";
+import { container } from "tsyringe";
+
+import { BidController } from "@src/bid/controllers/bid/bid.controller";
+import { ListBidsParamsSchema, ListBidsResponseSchema } from "@src/bid/http-schemas/bid.schema";
+import { OpenApiHonoHandler } from "@src/core/services/open-api-hono-handler/open-api-hono-handler";
+
+const listRoute = createRoute({
+  method: "get",
+  path: "/v1/bids/{dseq}",
```
ygrishajev: `dseq` is not a unique identifier for a deployment. `userId` + `dseq` or `address/owner` + `dseq` is. In our case user seems to be more appropriate. So the interface should prob be something like:

```suggestion
  path: "/v1/bids/${userId}/{dseq}",
```

It's not perfect so it's totally open to discussion, however user should be taken into account

apps/deploy-web/tests/import-sdl.spec.ts
```diff
@@ -0,0 +1,100 @@
+import { expect, test } from "@playwright/test";
+import { readFileSync } from "fs";
+import path from "path";
+
+import { BuildTemplatePage } from "./pages/BuildTemplatePage";
+
+test("import sdl", async ({ page, context }) => {
```
ygrishajev: just to understand. is this related to the PR purpose? I mean according to it's title?

jzsfkzm: Silly me, added files not belongig to the PR. Sorry, cleaned up.

apps/api/src/bid/routes/bids/bids.router.ts
jzsfkzm: Right, but if you [check the controller](https://github.com/jzsfkzm/console/blob/d4aa97f2d4b367a5c9340dd6e257f3292f3cd6b0/apps/api/src/bid/controllers/bid/bid.controller.ts#L18), we're grabbing the user's info there, leaning on x-api-key.
Isn't that what we need here?

ygrishajev: yes, I understand this. However it's generally better to specify a user id anyway and then use permissions to limit the query or check if authorized. this endpoint then could be used according to permissions, for instance for admin access if needed too

jzsfkzm: Wouldn't it be better to have dseq and userId as query string params?
`GET /v1/bids?dseq={dseq}&userId={userId}`

and in the menwhile
- dseq would be mandatory
- userId would be optional
  - when not set, we would load bids of the user in auth
  - when set, would be checked against auth

This way regular users would not need to send their userId, but we could still use it for admin purposes.

ygrishajev: yea, sounds good

apps/api/src/bid/services/bid.service.ts
jzsfkzm: Thanks for the heads up, I did this and some other part of the code could've been changed too, now using BidHttpService.

apps/api/src/bid/routes/bids/bids.router.ts
jzsfkzm: Made changes according to this, could you re-review pls?

apps/deploy-web/tests/import-sdl.spec.ts
ygrishajev: I mean, it's good to have these. Just got confused looking for reasoning here

apps/api/src/bid/controllers/bid/bid.controller.ts
```diff
@@ -0,0 +1,36 @@
...
+
+@singleton()
+export class BidController {
+  constructor(
+    private readonly bidHttpService: BidHttpService,
+    private readonly authService: AuthService,
+    private readonly userWalletRepository: UserWalletRepository,
+  ) {}
+
+  @Protected([{ action: "create", subject: "DeploymentSetting" }])
```
ygrishajev: why `DeploymentSetting`?

```suggestion
  @Protected([{ action: "sign", subject: "UserWallet" }])
```

apps/api/src/bid/controllers/bid/bid.controller.ts
```diff
@@ -0,0 +1,36 @@
...
+  async list(dseq: string, userId?: string): Promise<ListBidsResponse> {
+    const { currentUser } = this.authService;
+    assert(this.isValidUserId(userId), 401);
+
+    const wallets = await this.userWalletRepository.findByUserId(userId ?? currentUser.userId);
```
ygrishajev: we better use permissions here and remove the check above.

```suggestion
    const wallets = await this.userWalletRepository.accessibleBy(ability, "sign").findByUserId(userId ?? currentUser.userId);
```

apps/api/src/bid/controllers/bid/bid.controller.ts
```diff
@@ -0,0 +1,36 @@
...
+    const allBids = await Promise.all(wallets.map(wallet => this.bidHttpService.list(wallet.address, dseq)));
```
ygrishajev: according to current setup it would be just a single wallet.

apps/api/src/deployment/services/gpu-bids-creator/gpu-bids-creator.service.ts
```diff
@@ -107,9 +110,9 @@ export class GpuBidsCreatorService {
 
       await sleep(30_000);
 
-      const bids = await this.getBids(walletAddress, dseq);
+      const bids = await this.bidHttpService.list(walletAddress, dseq);
 
-      this.logger.info(`Got ${bids.bids.length} bids. Closing deployment...`);
+      this.logger.info(`Got ${bids.length} bids. Closing deployment...`);
```
ygrishajev: non-blocking
```suggestion
      this.logger.info({ event: "DEPLOYMENT_CLOSING", bidsCount: bids.length });
```

apps/api/src/bid/controllers/bid/bid.controller.ts
ygrishajev: based on this https://github.com/akash-network/console/blob/main/apps/api/src/billing/controllers/wallet/wallet.controller.ts#L40

jzsfkzm: No particular reason, It was just me picking an arbitrary permission. Fixed it.

apps/api/src/bid/controllers/bid/bid.controller.ts
jzsfkzm: We can add the accessibleBy part, added it now.
But: if we remove the decortor call, we'll have to manually check `ability.can("sign", "UserWallet")`, for an early 401. I would rather keep the decorator version instead of an assert call inside, because I think a decorator on top of the method is an easier read next time. What do you think?

apps/api/src/bid/controllers/bid/bid.controller.ts
jzsfkzm: Ah, ok, wasn't sure about it, reverted it to a single wallet.

apps/api/src/deployment/services/gpu-bids-creator/gpu-bids-creator.service.ts
jzsfkzm: Quite a few string-like logs in the same file, should I also fix those while at it?

apps/api/src/bid/controllers/bid/bid.controller.ts
ygrishajev: Yea, I think decorator is fine, I'm referring to the assert

apps/api/src/deployment/services/gpu-bids-creator/gpu-bids-creator.service.ts
ygrishajev: as you will honestly, this is not your changes :)

apps/api/src/bid/controllers/bid/bid.controller.ts
jzsfkzm: Ah, that one! You're right, it's not really needed after we have `.accessibleBy`, removed it.

apps/api/src/deployment/services/gpu-bids-creator/gpu-bids-creator.service.ts
jzsfkzm: Ok, did it, hope event keys are descriptive enough.

ygrishajev (approved):

baktun14 (approved):

PR #860, opened 2025-02-18
feat(network): add API endpoints for deployment

jzsfkzm:
refs #767

### DIFF
--- a/apps/api/src/deployment/controllers/deployment/deployment.controller.ts
+++ b/apps/api/src/deployment/controllers/deployment/deployment.controller.ts
@@ -1,22 +1,39 @@
 import { singleton } from "tsyringe";
 
-import { StaleManagedDeploymentsCleanerService } from "@src/deployment/services/stale-managed-deployments-cleaner/stale-managed-deployments-cleaner.service";
-import { TopUpManagedDeploymentsService } from "@src/deployment/services/top-up-managed-deployments/top-up-managed-deployments.service";
-import { TopUpDeploymentsOptions } from "@src/deployment/types/deployments-refiller";
-import { CleanUpStaleDeploymentsParams } from "@src/deployment/types/state-deployments";
+import { AuthService, Protected } from "@src/auth/services/auth.service";
+import { UserWalletRepository } from "@src/billing/repositories";
+import { CreateDeploymentRequest, CreateDeploymentResponse, GetDeploymentResponse } from "@src/deployment/http-schemas/deployment.schema";
+import { DeploymentService } from "@src/deployment/services/deployment/deployment.service";
 
 @singleton()
-export class TopUpDeploymentsController {
+export class DeploymentController {
   constructor(
-    private readonly topUpManagedDeploymentsService: TopUpManagedDeploymentsService,
-    private readonly staleDeploymentsCleanerService: StaleManagedDeploymentsCleanerService
+    private readonly deploymentService: DeploymentService,
+    private readonly authService: AuthService,
+    private readonly userWalletRepository: UserWalletRepository
   ) {}
 
-  async topUpDeployments(options: TopUpDeploymentsOptions) {
-    await this.topUpManagedDeploymentsService.topUpDeployments(options);
+  @Protected([{ action: "sign", subject: "UserWallet" }])
+  async findByDseqAndUserId(dseq: string, userId?: string): Promise<GetDeploymentResponse> {
+    const { currentUser, ability } = this.authService;
+
+    const wallets = await this.userWalletRepository.accessibleBy(ability, "sign").findByUserId(userId ?? currentUser.userId);
+    const deployment = await this.deploymentService.findByOwnerAndDseq(wallets[0].address, dseq);
+
+    return {
+      data: deployment
+    }
   }
 
-  async cleanUpStaleDeployment(options: CleanUpStaleDeploymentsParams) {
-    await this.staleDeploymentsCleanerService.cleanup(options);
+  @Protected([{ action: "sign", subject: "UserWallet" }])
+  async create(input: CreateDeploymentRequest['data']): Promise<CreateDeploymentResponse> {
+    const { currentUser, ability } = this.authService;
+
+    const wallets = await this.userWalletRepository.accessibleBy(ability, "sign").findByUserId(currentUser.userId);
+    const result = await this.deploymentService.create(wallets[0], input);
+
+    return {
+      data: result
+    };
   }
 }
--- a/apps/api/src/deployment/routes/deployments/deployments.router.ts
+++ b/apps/api/src/deployment/routes/deployments/deployments.router.ts
@@ -1,0 +1,66 @@
+import { createRoute } from "@hono/zod-openapi";
+import { container } from "tsyringe";
+
+import { OpenApiHonoHandler } from "@src/core/services/open-api-hono-handler/open-api-hono-handler";
+import { DeploymentController } from "@src/deployment/controllers/deployment/deployment.controller";
+import { CreateDeploymentRequestSchema, CreateDeploymentResponseSchema, GetDeploymentQuerySchema, GetDeploymentResponseSchema } from "@src/deployment/http-schemas/deployment.schema";
+
+const getRoute = createRoute({
+  method: "get",
+  path: "/v1/deployments",
+  summary: "Get a deployment",
+  tags: ["Deployments"],
+  request: {
+    query: GetDeploymentQuerySchema
+  },
+  responses: {
+    200: {
+      description: "Returns deployment info",
+      content: {
+        "application/json": {
+          schema: GetDeploymentResponseSchema
+        }
+      }
+    }
+  }
+});
+
+const postRoute = createRoute({
+  method: "post",
+  path: "/v1/deployments",
+  summary: "Create new deployment",
+  tags: ["Deployments"],
+  request: {
+    body: {
+      content: {
+        "application/json": {
+          schema: CreateDeploymentRequestSchema
+        }
+      }
+    }
+  },
+  responses: {
+    201: {
+      description: "API key created successfully",
+      content: {
+        "application/json": {
+          schema: CreateDeploymentResponseSchema
+        }
+      }
+    }
+  }
+});
+
+export const deploymentsRouter = new OpenApiHonoHandler();
+
+deploymentsRouter.openapi(getRoute, async function routeGetDeployment(c) {
+  const { dseq, userId } = c.req.valid("query");
+  const result = await container.resolve(DeploymentController).findByDseqAndUserId(dseq, userId);
+  return c.json(result, 200);
+});
+
+deploymentsRouter.openapi(postRoute, async function routeCreateDeployment(c) {
+  const { data } = c.req.valid("json");
+  const result = await container.resolve(DeploymentController).create(data);
+  return c.json(result, 201);
+});
--- a/apps/api/src/deployment/services/deployment/deployment.service.ts
+++ b/apps/api/src/deployment/services/deployment/deployment.service.ts
@@ -1,0 +1,67 @@
+import { SDL } from "@akashnetwork/akashjs/build/sdl";
+import { BlockHttpService, DeploymentHttpService, LeaseHttpService } from "@akashnetwork/http-sdk";
+import { BadRequest, InternalServerError, NotFound } from "http-errors";
+import { singleton } from "tsyringe";
+
+import { InjectWallet } from "@src/billing/providers/wallet.provider";
+import { UserWalletOutput } from "@src/billing/repositories";
+import { ManagedSignerService, RpcMessageService, Wallet } from "@src/billing/services";
+import { BillingConfigService } from "@src/billing/services/billing-config/billing-config.service";
+import { CreateDeploymentRequest, CreateDeploymentResponse, GetDeploymentResponse } from "@src/deployment/http-schemas/deployment.schema";
+
+@singleton()
+export class DeploymentService {
+  constructor(
+    private readonly blockHttpService: BlockHttpService,
+    private readonly deploymentHttpService: DeploymentHttpService,
+    private readonly leaseHttpService: LeaseHttpService,
+    private readonly signerService: ManagedSignerService,
+    @InjectWallet("MANAGED") private readonly masterWallet: Wallet,
+    private readonly billingConfigService: BillingConfigService,
+    private readonly rpcMessageService: RpcMessageService,
+  ) { }
+
+  public async findByOwnerAndDseq(owner: string, dseq: string): Promise<GetDeploymentResponse['data']> {
+    const deploymentResponse = await this.deploymentHttpService.findByOwnerAndDseq(owner, dseq);
+    if ("code" in deploymentResponse) {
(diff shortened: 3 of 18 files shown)


apps/api/src/deployment/services/deployment/deployment.service.ts
```diff
@@ -0,0 +1,83 @@
...
+        id: {
+          owner: account.address,
+          dseq: dseq
+        },
+        groups: sdl.groups(),
+        version: manifestVersion,
+        deposit: {
+          denom: "uakt",
+          amount: "500000"
+        },
```
baktun14: This will be the USDC ibc denom for managed wallets, you can use it [from here](https://github.com/akash-network/console/blob/f1a0c32422b3716010f936aa7d0a4abe38ee4762/apps/api/src/billing/config/network.config.ts#L2) 

Also, the amount should be `"5000000"`, but to test and use sandbox uakt it's probably to have parametrized.

apps/api/src/deployment/services/deployment/deployment.service.ts
```diff
@@ -0,0 +1,83 @@
...
+        depositor: account.address
+      })
+    };
+
+    return await this.messageTransmitterService.signAndBroadcast(account.address, client, [message]);
```
baktun14: We already have a [sign and broadcast transaction](https://github.com/akash-network/console/blob/f1a0c32422b3716010f936aa7d0a4abe38ee4762/apps/api/src/billing/controllers/wallet/wallet.controller.ts#L41-L45) that handles wallet for the master wallet. The gpu bot wallet mnemonic is used to another purpose.

apps/api/test/functional/deployments.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+            sdl: "invalid-sdl",
+          }
+        }),
+        headers: new Headers({ "Content-Type": "application/json", 'x-api-key': apiKey })
+      });
+
+      expect(response.status).toBe(400);
+    });
+
+    xit("creates a deployment", async () => {
```
baktun14: `xit`?

apps/api/src/deployment/services/deployment/deployment.service.ts
```diff
@@ -0,0 +1,83 @@
...
+
+@singleton()
+export class DeploymentService {
+  constructor(
+    private readonly config: BillingConfigService,
+    private readonly messageTransmitterService: MessageTransmitterService,
+  ) { }
+
+  public async findByOwnerAndDseq(owner: string, dseq: string): Promise<SingleDeploymentResponse['data']> {
+    const response = await axios.get<RestAkasheploymentInfoResponse>(`${apiNodeUrl}/akash/deployment/${betaTypeVersion}/deployments/info`, {
```
baktun14: This will be the perfect candidates for @stalniy's new chain-sdk implemention. Technically we [already have an endpoint implemented](https://github.com/akash-network/console/blob/f1a0c32422b3716010f936aa7d0a4abe38ee4762/apps/api/src/services/external/apiNodeService.ts#L361) for this one, but not refactored in the new services architecture. We should abstract those http calls in the [http package](https://github.com/akash-network/console/tree/f1a0c32422b3716010f936aa7d0a4abe38ee4762/packages/http-sdk).

baktun14 (changes requested):

apps/api/src/deployment/services/deployment/deployment.service.ts
ygrishajev: managed wallet denom is configured by env var https://github.com/akash-network/console/blob/main/apps/api/src/billing/config/env.config.ts#L12

apps/api/src/deployment/routes/deployments/deployments.router.ts
```diff
@@ -0,0 +1,71 @@
...
+import { z } from "zod";
+
+import { OpenApiHonoHandler } from "@src/core/services/open-api-hono-handler/open-api-hono-handler";
+import { DeploymentController } from "@src/deployment/controllers/deployment/deployment.controller";
+import { CreateDeploymentRequestSchema, CreateDeploymentResponseSchema, GetDeploymentParamsSchema, GetDeploymentResponseSchema } from "@src/deployment/http-schemas/deployment.schema";
+
+const getRoute = createRoute({
+  method: "get",
+  path: "/v1/deployments/{dseq}",
+  summary: "List deployments",
```
ygrishajev:
```suggestion
  summary: "Get a single deployment",
```

apps/api/src/deployment/routes/deployments/deployments.router.ts
```diff
@@ -0,0 +1,71 @@
...
+import { container } from "tsyringe";
+import { z } from "zod";
+
+import { OpenApiHonoHandler } from "@src/core/services/open-api-hono-handler/open-api-hono-handler";
+import { DeploymentController } from "@src/deployment/controllers/deployment/deployment.controller";
+import { CreateDeploymentRequestSchema, CreateDeploymentResponseSchema, GetDeploymentParamsSchema, GetDeploymentResponseSchema } from "@src/deployment/http-schemas/deployment.schema";
+
+const getRoute = createRoute({
+  method: "get",
+  path: "/v1/deployments/{dseq}",
```
ygrishajev: `dseq` is not a unique identifier for a deployment. `userId` + `dseq` or `address/owner` + `dseq` is. In our case user seems to be more appropriate. So the interface should prob be something like:

```suggestion
  path: "/v1/deployments/${userId}/{dseq}",
```

It's not perfect so it's totally open to discussion, however user should be taken into account

apps/api/src/deployment/controllers/deployment/deployment.controller.ts
```diff
@@ -1,15 +1,24 @@
...
+import { CreateDeploymentRequest, CreateDeploymentResponse, SingleDeploymentResponse } from "@src/deployment/http-schemas/deployment.schema";
+import { DeploymentService } from "@src/deployment/services/deployment/deployment.service";
 import { StaleManagedDeploymentsCleanerService } from "@src/deployment/services/stale-managed-deployments-cleaner/stale-managed-deployments-cleaner.service";
 import { TopUpManagedDeploymentsService } from "@src/deployment/services/top-up-managed-deployments/top-up-managed-deployments.service";
 import { TopUpDeploymentsOptions } from "@src/deployment/types/deployments-refiller";
 import { CleanUpStaleDeploymentsParams } from "@src/deployment/types/state-deployments";
 
 @singleton()
-export class TopUpDeploymentsController {
+export class DeploymentController {
```
ygrishajev, line 9: non-blocking: why do you think we should merge these? Why not having 2 specific controllers?

apps/api/src/deployment/controllers/deployment/deployment.controller.ts
```diff
@@ -19,4 +28,33 @@ export class TopUpDeploymentsController {
   async cleanUpStaleDeployment(options: CleanUpStaleDeploymentsParams) {
     await this.staleDeploymentsCleanerService.cleanup(options);
   }
+
+  @Protected([{ action: "create", subject: "DeploymentSetting" }])
```
ygrishajev: blocking: this doesn't seem right. Why is it `DeploymentSetting`?

apps/api/src/deployment/services/deployment/deployment.service.ts
```diff
@@ -0,0 +1,83 @@
...
+      params: {
+        "id.owner": owner,
+        "id.dseq": dseq,
+      }
+    });
+
+    if ('code' in response.data) {
+      return {
```
ygrishajev: why do we need to resolve with this response? I assume code may be different? controller should throw appropriate http error. E.g. if deployment not found I assume this return value should be `undefined` and controller should throw 404. If it's smth else it should be 500 perhaps.

apps/api/src/deployment/controllers/deployment/deployment.controller.ts
```diff
@@ -19,4 +28,33 @@ export class TopUpDeploymentsController {
...
+    }
+  }
+
+  @Protected([{ action: "create", subject: "DeploymentSetting" }])
+  async create(input: CreateDeploymentRequest["data"]): Promise<CreateDeploymentResponse> {
+    let sdl: SDL;
+    try {
+      sdl = SDL.fromString(input.sdl, 'beta3');
+    } catch (_error) {
+      assert(false, 400, "Invalid SDL");
```
ygrishajev: should the error be checked? I mean we should ensure this is a validation issue and not some code bug in which case it should be 500.

apps/api/src/deployment/services/deployment/deployment.service.ts
```diff
@@ -0,0 +1,83 @@
...
+    const wallet = await DirectSecp256k1HdWallet.fromMnemonic(env.GPU_BOT_WALLET_MNEMONIC, { prefix: "akash" });
+    const [account] = await wallet.getAccounts();
+    const myRegistry = new Registry([...getAkashTypeRegistry()]);
+    const client = await SigningStargateClient.connectWithSigner(this.config.get("RPC_NODE_ENDPOINT"), wallet, {
+      registry: myRegistry,
+      broadcastTimeoutMs: 30_000
+    });
+
+    const message = {
+      typeUrl: `/akash.deployment.v1beta3.MsgCreateDeployment`,
```
ygrishajev: messages are kept in https://github.com/akash-network/console/blob/main/apps/api/src/billing/services/rpc-message-service/rpc-message.service.ts

apps/api/src/deployment/routes/deployments/deployments.router.ts
jzsfkzm: How about if it had the same interface as `GET /v1/bids`, with `?dseq&userId`, userId being optional as it can be derived from auth, but also can be forced when you're an admin?
Maybe it should be `GET /v1/x/{dseq}?userId` now that I think about it. What do you think?

apps/api/src/deployment/services/deployment/deployment.service.ts
jzsfkzm: That function is returning a somewhat different type than the one described for this endpoint, escrow_account for example is missing, or I'd rather say balance is the only thing picked from escrow_account. I'm all for using the same interface where it is possible, so happy to use that function, but should we right away?

apps/api/test/functional/deployments.spec.ts
jzsfkzm: Was pretty unsure about that part of the code, how it should work and what should it actually do. Can we chat about it @ygrishajev?

apps/api/src/deployment/routes/deployments/deployments.router.ts
jzsfkzm: Fixed, thanks.

ygrishajev (dismissed):

apps/api/src/deployment/controllers/deployment/deployment.controller.ts
jzsfkzm: What would be the proper permission here, `sign::UserWallet`?

apps/api/src/deployment/controllers/deployment/deployment.controller.ts
jzsfkzm: Seemed logical to me as both do something with deployments, but I was unsure too, we can keep those separated, so separating.

apps/api/src/deployment/controllers/deployment/deployment.controller.ts
ygrishajev: At least this is what used currently to sign a create deployment message on user wallet. I'm not sure it is needed to add a new entity/action.

apps/api/src/deployment/services/deployment/deployment.service.ts
jzsfkzm: Right, changed it, though I'm pretty unsure what is coming in the error when it is an error response (like `{code, message}`).

apps/api/src/deployment/controllers/deployment/deployment.controller.ts
jzsfkzm: I have a hard time checking this error, maybe we should check pairing.

apps/api/src/deployment/routes/deployments/deployments.router.ts
jzsfkzm: Ok, changed it so now it is working the same as `GET /v1/bids` (both params in query string).

apps/api/src/deployment/services/deployment/deployment.service.ts
jzsfkzm: Amount now should be in the payload and the denom is coming from config.

apps/api/src/deployment/services/deployment/deployment.service.ts
jzsfkzm: Changed it that way, thanks.

apps/api/src/deployment/controllers/deployment/deployment.controller.ts
jzsfkzm: Changed it to `sign::userWallet`.

apps/api/src/deployment/controllers/deployment/deployment.controller.ts
jzsfkzm: The error is checked now, whenever it is an `SdlValidationError`, we throw a 400 to the client with an actionable message from the error.

apps/api/src/deployment/services/deployment/deployment.service.ts
baktun14: @jzsfkzm Just this and lgtm!

apps/api/src/deployment/services/deployment/deployment.service.ts
```diff
@@ -0,0 +1,84 @@
...
+        },
+        depositor: await this.masterWallet.getFirstAddress()
+      })
+    };
+    const encodedMessage = {
+      typeUrl: message.typeUrl,
+      value: Buffer.from(this.registry.encode(message)).toString("base64")
+    };
+
+    return await this.signerService.executeEncodedTxByUserId(wallet.userId, [encodedMessage]);
```
baktun14: I think we should have a similar method like `executeDecodedTxByUserId(...)` so we don't have to encode/decode/encode in the api

jzsfkzm: That's indeed a good idea, will make this change.

apps/api/src/deployment/services/deployment/deployment.service.ts
jzsfkzm: Moved the creation of the message to the service.

apps/api/src/deployment/services/deployment/deployment.service.ts
jzsfkzm: Made this change too, now we have two methods, running (almost) the same path.

baktun14 (approved):

apps/api/src/deployment/services/deployment/deployment.service.ts
baktun14: Perfect

Issue #64, opened 2023-12-19
Support AMD GPUs


baktun14:
- Adapt GPU code to support any GPU vendor, there is some hardcoded vendor code to nvidia
- Adapt the provider attributes schema?

PR #861, opened 2025-02-18
fix(sdl): can use AMD GPUs

jzsfkzm:
Also added some logic preselecting memory and interface when there's only one option, and clearing selections when you change vendors.

refs #64

### DIFF
--- a/apps/deploy-web/src/components/sdl/GpuFormControl.tsx
+++ b/apps/deploy-web/src/components/sdl/GpuFormControl.tsx
@@ -1,5 +1,5 @@
 "use client";
-import { ReactNode } from "react";
+import { ReactNode, useCallback, useMemo } from "react";
 import { Control, Controller, useFieldArray, UseFormSetValue } from "react-hook-form";
 import { MdSpeed } from "react-icons/md";
 import { Button, Checkbox, CustomTooltip, FormField, FormItem, FormMessage, Input, Slider, Spinner } from "@akashnetwork/ui/components";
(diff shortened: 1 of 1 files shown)


baktun14 (dismissed):

apps/deploy-web/src/components/sdl/GpuFormControl.tsx
```diff
@@ -43,6 +43,33 @@ export const GpuFormControl: React.FunctionComponent<Props> = ({ gpuModels, cont
...
 
+  const gpuVendors = useMemo(() => {
+    return gpuModels
+      ? gpuModels.map((vendor, index) => {
+          return { id: index, value: vendor.name };
+        })
+      : hardCodedGpuVendors;
+  }, [gpuModels]);
+
+  const gpuModelsByName: Record<string, GpuModel> = useMemo(() => {
```
ygrishajev, line 57: non-blocking: some code could be dropped.

```suggestion
import keyBy from "lodash/keyBy";
// ....
const gpuModelsByName: Record<string, GpuModel> = keyBy(gpuModels, "name");
```

ygrishajev (dismissed):

apps/deploy-web/src/components/sdl/GpuFormControl.tsx
jzsfkzm: It is a bit more complicated than that, because gpuModels is an array of GpuVendors:

```
Array<{
  models: Array<{
    name: string;
  }>
]>
```

But you're right about we can simplify it with lodash functions, I just did.

ygrishajev: oh, I did miss that. new code looks cleaner indeed though, thx

ygrishajev (approved):

PR #897, opened 2025-02-21
feat(provider): added version to sidebar

### LINKED ISSUE #877: Add version to provider Console
Display a version so that when users report issues we can see what version it happens with

Copy what we do for Deploy Console

Image

jigar-arc10:
added version to sidebar
closes #877


devalpatel67 (approved):
LGTM

PR #898, opened 2025-02-23
fix(provider): render longer provider names, make them copiable

### LINKED ISSUE #858: Increase displayed provider hostname to 40 characters
## Problem Statement

Customers/ Users of Console frequently report issues with deployments and must provider hostnames of the provider that the deployment is running on. Unfortunately because Console truncates the provider hostname, most of the GPU providers (that start with "provider" and end with "akash.pub" end up looking the same (as shown below) in the deployments list as well in the deployment details page

Image

Image

## Solution
I'd suggesting making some or all of these changes to the UI ( @aktdenis - your thoughts on solving this UI issue?)
1. Increase displayed hostname field to allow for up to 40 characters if possible. This would cover the longest of the GPU hostnames we have today - which is `provider.rtx4090.evergreen.akash.pub`
2. Include a "copy" UI function to allow the user to just copy it to the clipboard.
3. Do the above things for both the deployments list page as well as the deployments detail page

jzsfkzm:
closes #858

### DIFF
--- a/apps/deploy-web/src/components/deployments/LeaseChip.tsx
+++ b/apps/deploy-web/src/components/deployments/LeaseChip.tsx
@@ -1,13 +1,14 @@
 "use client";
 import { useEffect, useState } from "react";
-import { Badge, CustomTooltip, Spinner } from "@akashnetwork/ui/components";
+import { Badge, Spinner } from "@akashnetwork/ui/components";
 import Link from "next/link";
 
-import { getSplitText } from "@src/hooks/useShortText";
 import { LeaseDto } from "@src/types/deployment";
 import { ApiProviderList } from "@src/types/provider";
 import { UrlService } from "@src/utils/urlUtils";
+import { CopyTextToClipboardButton } from "../copy-text-to-clipboard-button/CopyTextToClipboardButton";
 import { StatusPill } from "../shared/StatusPill";
+import { ShortenedValue } from "../shortened-value/ShortenedValue";
 
 type Props = {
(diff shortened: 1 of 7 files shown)


baktun14 (dismissed):
Nice!🔥 Can you just post a quick screenshot of how it looks like please?

apps/deploy-web/src/components/deployments/LeaseChip.tsx
```diff
@@ -27,25 +28,24 @@ export const LeaseChip: React.FunctionComponent<Props> = ({ lease, providers })
...
+        {providerName ? (
+          <ShortenedValue value={providerName} maxLength={40} headLength={14} />
+        ) : (
+          <Spinner size="xSmall" />
+        )}
         <StatusPill state={lease.state} size="small" />
       </Badge>
     </Link>
+
+    {providerName && <CopyTextToClipboardButton value={providerName} />}
```
baktun14, line 48: Is that shown as just an empty button?

jzsfkzm:
> Nice!🔥 Can you just post a quick screenshot of how it looks like please?

Thanks, sure, here you are @baktun14. Somewhat played around with col widths on this table, all those had the same width, now some columns are wider:
Screenshot 2025-02-24 at 0 18 03

Added it here too:
Screenshot 2025-02-24 at 0 19 01

And here as well, also changed the icon next to URIs:
Screenshot 2025-02-24 at 0 20 56

baktun14 (approved):

PR #903, opened 2025-02-25
feat(certificate): managed wallet api create certificates

baktun14:
https://github.com/akash-network/console/issues/767

### DIFF
--- a/apps/api/src/certificate/controllers/certificate.controller.ts
+++ b/apps/api/src/certificate/controllers/certificate.controller.ts
@@ -1,0 +1,18 @@
+import assert from "http-assert";
+import { singleton } from "tsyringe";
+
+import { Protected } from "@src/auth/services/auth.service";
+import { CreateCertificateResponse } from "../http-schemas/create-certificate.schema";
+import { CertificateService } from "../services/certificate.service";
+
+@singleton()
+export class CertificateController {
+  constructor(private readonly certificateService: CertificateService) {}
+
+  @Protected([{ action: "read", subject: "DeploymentSetting" }])
+  async create(): Promise<CreateCertificateResponse> {
+    const cert = await this.certificateService.create();
+    assert(cert, 404, "Certificate not found");
+    return cert;
+  }
+}
--- a/apps/api/src/certificate/controllers/certificate.controller.ts
+++ b/apps/api/src/certificate/controllers/certificate.controller.ts
@@ -1,4 +1,3 @@
-import assert from "http-assert";
 import { singleton } from "tsyringe";
 
 import { Protected } from "@src/auth/services/auth.service";
@@ -12,7 +11,6 @@ export class CertificateController {
   @Protected([{ action: "sign", subject: "UserWallet" }])
   async create(): Promise<CreateCertificateResponse> {
     const cert = await this.certificateService.create();
-    assert(cert, 404, "Certificate not found");
     return { data: cert };
   }
 }
--- a/apps/api/src/certificate/controllers/certificate.controller.ts
+++ b/apps/api/src/certificate/controllers/certificate.controller.ts
@@ -9,7 +9,10 @@ import { CertificateService } from "../services/certificate.service";
 export class CertificateController {
   constructor(private readonly certificateService: CertificateService) {}
 
-  @Protected([{ action: "read", subject: "DeploymentSetting" }])
+  @Protected([
+    { action: "sign", subject: "UserWallet" },
+    { action: "read", subject: "ApiKey" }
+  ])
   async create(): Promise<CreateCertificateResponse> {
     const cert = await this.certificateService.create();
     assert(cert, 404, "Certificate not found");
--- a/apps/api/src/certificate/controllers/certificate.controller.ts
+++ b/apps/api/src/certificate/controllers/certificate.controller.ts
@@ -9,13 +9,10 @@ import { CertificateService } from "../services/certificate.service";
 export class CertificateController {
   constructor(private readonly certificateService: CertificateService) {}
 
-  @Protected([
-    { action: "sign", subject: "UserWallet" },
-    { action: "read", subject: "ApiKey" }
-  ])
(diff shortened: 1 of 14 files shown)


apps/api/src/certificate/services/certificate.service.ts
```diff
@@ -0,0 +1,49 @@
...
+  async create(): Promise<CertificateOutput> {
+    const userWallet = await this.userWalletRepository.findOneByUserId(this.authService.currentUser.id);
+
+    if (userWallet) {
+      const { cert: crtpem, publicKey: pubpem, privateKey: encryptedKey } = certificateManager.generatePEM(userWallet.address);
+      const createCertificateMsg = this.rpcMessageService.getCreateCertificateMsg(userWallet.address, crtpem, pubpem);
+      const messages = [createCertificateMsg];
+
+      try {
+        await this.managedSignerService.executeManagedTx(userWallet.id, messages);
```
baktun14: I will change this to `executeDecodedTxByUserId(...)` so it refreshes the user balances as well. (https://github.com/akash-network/console/pull/860/files#r1970537622)

apps/api/src/certificate/http-schemas/create-certificate.schema.ts
```diff
@@ -0,0 +1,9 @@
+import { z } from "zod";
+
+export const CreateCertificateResponseSchema = z.object({
```
ygrishajev, line 3: can we wrap it in `data` as a common approach?

apps/api/src/certificate/services/certificate.service.ts
```diff
@@ -0,0 +1,43 @@
...
+export class CertificateService {
+  constructor(
+    private readonly userWalletRepository: UserWalletRepository,
+    private readonly authService: AuthService,
+    private readonly rpcMessageService: RpcMessageService,
+    private readonly managedSignerService: ManagedSignerService
+  ) {}
+
+  async create(): Promise<CertificateOutput> {
+    const userWallet = await this.userWalletRepository.findOneByUserId(this.authService.currentUser.id);
```
ygrishajev:
```suggestion
    const userWallet = await this.userWalletRepository.accessibleBy(ability, "sign").findOneByUserId(this.authService.currentUser.id);
    
    assert(userWallet, 404, "UserWallet not found")
```

apps/api/src/certificate/controllers/certificate.controller.ts
```diff
@@ -0,0 +1,21 @@
...
+import { CreateCertificateResponse } from "../http-schemas/create-certificate.schema";
+import { CertificateService } from "../services/certificate.service";
+
+@singleton()
+export class CertificateController {
+  constructor(private readonly certificateService: CertificateService) {}
+
+  @Protected([
+    { action: "sign", subject: "UserWallet" },
+    { action: "read", subject: "ApiKey" }
```
ygrishajev: I'm not sure I follow what this permission is for here

baktun14: Yea you're right it doesn't make sense and is redundant hehe

apps/api/src/certificate/http-schemas/create-certificate.schema.ts
baktun14: Oh yea true

apps/api/src/certificate/services/certificate.service.ts
```diff
@@ -0,0 +1,46 @@
...
+    private readonly rpcMessageService: RpcMessageService,
+    private readonly managedSignerService: ManagedSignerService
+  ) {}
+
+  async create(): Promise<CertificateOutput> {
+    const userWallet = await this.userWalletRepository.accessibleBy(this.authService.ability, "sign").findOneByUserId(this.authService.currentUser.id);
+
+    assert(userWallet, 404, "UserWallet not found");
+
+    if (userWallet) {
```
ygrishajev: this check is redundant now thanks to the `assert` above
```suggestion
```

apps/api/src/certificate/controllers/certificate.controller.ts
```diff
@@ -0,0 +1,18 @@
...
+import { CertificateService } from "../services/certificate.service";
+
+@singleton()
+export class CertificateController {
+  constructor(private readonly certificateService: CertificateService) {}
+
+  @Protected([{ action: "sign", subject: "UserWallet" }])
+  async create(): Promise<CreateCertificateResponse> {
+    const cert = await this.certificateService.create();
+    assert(cert, 404, "Certificate not found");
```
ygrishajev: I don't think this is to ever work. 404 on a requested resource for a create op is controversial a bit :)
anyways the called function will throw the proper error if the wallet not found
```suggestion
```

ygrishajev (changes requested):

apps/api/src/certificate/controllers/certificate.controller.ts
baktun14: True

apps/api/test/functional/certificate.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+          "x-api-key": apiKey
+        })
+      });
+
+      expect(response.status).toBe(200);
+      const result = (await response.json()) as CreateCertificateResponse;
+
+      expect(result.data?.certPem).toBeDefined();
+      expect(result.data?.pubkeyPem).toBeDefined();
+      expect(result.data?.encryptedKey).toBeDefined();
```
ygrishajev, line 100: non-blocking: this could prob be tested a bit further. like that's a defined string:
```suggestion
      expect(result.data).toMatchObject({
        certPem: expect.any(String),
        pubkeyPem: expect.any(String),
        encryptedKey: expect.any(String)
      });
```

ygrishajev (approved):

apps/api/test/functional/certificate.spec.ts
baktun14: Will do in another branch

Issue #836, opened 2025-02-11
Handle CloudFlare challenge error on all the requests


ygrishajev:
Currently if CL "Under Attack" mode is enabled it causes requests to start failing silently after 30m when the cookie is expired. We only handle that on ["start-trial" endpoint](https://github.com/akash-network/console/blob/main/apps/deploy-web/src/components/turnstile/Turnstile.tsx#L25). This should be fine when "Under Attack" is not enabled. However we need to make sure it's also fine when it is.

Acceptance criteria:
1) ensure all instances of axios/http-services are provided with relevant response interceptors

ygrishajev:
The issue still persists for some endpoints. E.g. Amplitude is affected. After 30min of page not being refreshed requests are rejected by cf.

We should patch xhr and fetch globally to address the issue better.

PR #909, opened 2025-02-26
fix(auth): patches fetch to retry on cloudflare challenge

ygrishajev:
refs #836


apps/deploy-web/src/components/turnstile/Turnstile.tsx
```diff
@@ -40,7 +42,26 @@ export const Turnstile: FC = () => {
   const dismissedSubject = useRef(new Subject<void>());
 
   useEffect(() => {
-    return addResponseInterceptor(async error => {
+    if (originalFetch) {
+      window.fetch = async (...args) => {
```
stalniy, line 46: question(non-blocking): do we need a tech debt issue or this is kind of hack that must live with the code forever?

stalniy (approved):

Issue #887, opened 2025-02-21
Add signup prompt for no bids state


anilmurty:
## Problem Statement

Once all trial providers are used up new trials receive no bids and probably think that the product/ service is broken. To overcome this - one of the actions/ experiments (there are others for sure) we can try is to display a banner that prompts the user to sign up and indicates that doing so will give them access to a larger list of providers

## Solution

Two parts to the solution:

Frontend/ UI:

If a user has not received any bids for 13 seconds (past 2 blocks) AND if they are a trial user, then display a banner that says:
"Looks like you are not getting any bids. This is likely because all trial providers are currently being used by others. Console has 10 providers for trial users but more available for non-trial users. To view all providers, we recommend signing up" (with [signing up] hyperlinked to the sign up flow)

Backend:
Implement a second "trials-user" signed-attribute check that filters to the list of trial providers to a second (larger) list similar https://console-api.akash.network/v1/trial-providers

One attribute for users who have signed up but have not made a credit card transaction - `console/trials-user` (ref: https://github.com/akash-network/console/blob/0cf3eb0dd15534d1f62ca33d987a0e9c34880c79/apps/api/src/billing/services/trial-validation/trial-validation.service.ts#L7 for the console/trials attribute)

Code Ref:

On the frontend we append the attribute: https://github.com/akash-network/console/blob/5a58c298e6ad15944a93b7cd1d82837b78235c9a/apps/deploy-web/src/utils/deploymentData/v1beta3.ts#L30

On the backend we validate that the attribute is there: [https://github.com/akash-network/console/blob/5a58c298e6ad15944a93b7cd1d82837b7823[…]c/billing/services/trial-validation/trial-validation.service.ts](https://github.com/akash-network/console/blob/5a58c298e6ad15944a93b7cd1d82837b78235c9a/apps/api/src/billing/services/trial-validation/trial-validation.service.ts#L11)

stalniy:
Hey @anilmurty 

I'd like to clarify this requirement
> Implement a second "trials-user" signed-attribute check that filters to the list of trial providers to a second (larger) list similar

I assume I need to add one more attribute which is called `console/trials-user`. In what case should this attribute be associated with a user? I can't understand how is it different from existing one `console/trials`.

Later I need to add this attribute in [/trial-providers](https://console-api.akash.network/v1/trial-providers) filter to return providers from both attributes (existing one and new one) and ensure that it's validated in trial validation service, right? 

[...]

PR #922, opened 2025-02-28
feat(deployment): show banner for trial users with no bids

jzsfkzm:
refs #887

### DIFF
--- a/apps/api/src/billing/services/trial-validation/trial-validation.service.ts
+++ b/apps/api/src/billing/services/trial-validation/trial-validation.service.ts
@@ -1,10 +1,11 @@
+import { GroupSpec } from "@akashnetwork/akash-api/akash/deployment/v1beta3";
 import * as v1beta3 from "@akashnetwork/akash-api/v1beta3";
 import { EncodeObject } from "@cosmjs/proto-signing";
 import assert from "http-assert";
 import { singleton } from "tsyringe";
 
 import { UserWalletOutput } from "@src/billing/repositories";
-import { AUDITOR, TRIAL_ATTRIBUTE } from "@src/deployment/config/provider.config";
+import { AUDITOR, TRIAL_ATTRIBUTE, TRIAL_REGISTERED_ATTRIBUTE } from "@src/deployment/config/provider.config";
 import { getAddressDeployments } from "@src/services/external/apiNodeService";
 
 const TRIAL_DEPLOYMENT_LIMIT = 5;
@@ -12,19 +13,13 @@ const TRIAL_DEPLOYMENT_LIMIT = 5;
 @singleton()
 export class TrialValidationService {
   async validateLeaseProviders(decoded: EncodeObject, userWallet: UserWalletOutput) {
-    if (userWallet.isTrialing && decoded.typeUrl === "/akash.deployment.v1beta3.MsgCreateDeployment") {
+    if (decoded.typeUrl === "/akash.deployment.v1beta3.MsgCreateDeployment") {
       const value = decoded.value as v1beta3.MsgCreateDeployment;
-
-      value.groups.forEach(group => {
-        const hasTrial = group.requirements.attributes.some(attribute => {
-          return attribute.key === TRIAL_ATTRIBUTE && attribute.value === "true";
-        });
-
-        const hasSignedByAllOf = group.requirements.signedBy.allOf.every(signedBy => {
-          return signedBy === AUDITOR;
-        });
-        assert(hasTrial && hasSignedByAllOf, 400, `provider not authorized: ${group.requirements.attributes}`);
-      });
+      if (userWallet.isTrialing) {
+        this.validateAttribute(value.groups, TRIAL_ATTRIBUTE);
+      } else if (!userWallet.stripeCustomerId) {
+        this.validateAttribute(value.groups, TRIAL_REGISTERED_ATTRIBUTE);
+      }
     }
 
     return true;
@@ -35,5 +30,18 @@ export class TrialValidationService {
       const deployments = await getAddressDeployments(userWallet.address, 0, 1, false, {});
       assert(deployments.count < TRIAL_DEPLOYMENT_LIMIT, 402, "Trial limit reached. Add funds to your account to deploy more.");
     }
+  }
+
+  private validateAttribute(groups: GroupSpec[], key: string) {
+    groups.forEach(group => {
+      const hasAttribute = group.requirements.attributes.some(attribute => {
+        return attribute.key === key && attribute.value === "true";
+      });
+
+      const hasSignedByAllOf = group.requirements.signedBy.allOf.every(signedBy => {
+        return signedBy === AUDITOR;
+      });
+      assert(hasAttribute && hasSignedByAllOf, 400, `provider not authorized: ${group.requirements.attributes}`);
+    });
   }
 }
--- a/apps/deploy-web/src/components/new-deployment/CreateLease.tsx
+++ b/apps/deploy-web/src/components/new-deployment/CreateLease.tsx
@@ -5,6 +5,8 @@ import {
   AlertDescription,
   AlertTitle,
   Button,
+  Card,
+  CardContent,
   Checkbox,
   CustomTooltip,
   DropdownMenu,
@@ -26,6 +28,7 @@ import { useWallet } from "@src/context/WalletProvider";
 import { useManagedDeploymentConfirm } from "@src/hooks/useManagedDeploymentConfirm";
 import { useWhen } from "@src/hooks/useWhen";
 import { useBidList } from "@src/queries/useBidQuery";
+import { useBlock } from "@src/queries/useBlocksQuery";
 import { useDeploymentDetail } from "@src/queries/useDeploymentQuery";
 import { useProviderList } from "@src/queries/useProvidersQuery";
 import { analyticsService } from "@src/services/analytics/analytics.service";
@@ -33,6 +36,7 @@ import networkStore from "@src/store/networkStore";
 import { BidDto } from "@src/types/deployment";
 import { RouteStep } from "@src/types/route-steps.type";
 import { deploymentData } from "@src/utils/deploymentData";
+import { TRIAL_ATTRIBUTE } from "@src/utils/deploymentData/v1beta3";
 import { getDeploymentLocalData } from "@src/utils/deploymentLocalDataUtils";
 import { sendManifestToProvider } from "@src/utils/deploymentUtils";
 import { addScriptToHead } from "@src/utils/domUtils";
@@ -58,6 +62,7 @@ const REFRESH_BIDS_INTERVAL = 7000;
 const MAX_NUM_OF_BID_REQUESTS = Math.floor((5.5 * 60 * 1000) / REFRESH_BIDS_INTERVAL);
 // Show a warning after 1 minute
 const WARNING_NUM_OF_BID_REQUESTS = Math.round((60 * 1000) / REFRESH_BIDS_INTERVAL);
+const TRIAL_SIGNUP_WARNING_TIMEOUT = 33000;
 
 export const CreateLease: React.FunctionComponent<Props> = ({ dseq }) => {
   const [isSendingManifest, setIsSendingManifest] = useState(false);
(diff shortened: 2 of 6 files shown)


apps/api/src/billing/services/trial-validation/trial-validation.service.ts
```diff
@@ -1,29 +1,38 @@
...
 @singleton()
 export class TrialValidationService {
   async validateLeaseProviders(decoded: EncodeObject, userWallet: UserWalletOutput) {
-    if (userWallet.isTrialing && decoded.typeUrl === "/akash.deployment.v1beta3.MsgCreateDeployment") {
+    if (decoded.typeUrl === "/akash.deployment.v1beta3.MsgCreateDeployment") {
       const value = decoded.value as v1beta3.MsgCreateDeployment;
+      if (userWallet.isTrialing) {
+        this.validateAttribute(value.groups, TRIAL_ATTRIBUTE);
+      } else if (!userWallet.stripeCustomerId) {
+        this.validateAttribute(value.groups, TRIAL_ATTRIBUTE);
```
baktun14: issue (blocking): We only need to have one or the other trial attribute, so we can remove this one.

apps/deploy-web/src/components/new-deployment/ManifestEdit.tsx
```diff
@@ -226,8 +226,11 @@ export const ManifestEdit: React.FunctionComponent<Props> = ({
         return;
       }
 
-      if (isTrialing) {
-        sdl = appendTrialAttribute(sdl);
+      if (isTrialing || isOnboarding) {
+        sdl = appendTrialAttribute(sdl, TRIAL_ATTRIBUTE);
+        if (isOnboarding) {
+          sdl = appendTrialAttribute(sdl, TRIAL_REGISTERED_ATTRIBUTE);
```
baktun14: issue (blocking): same here, one or the other

apps/api/src/billing/services/trial-validation/trial-validation.service.ts
baktun14: I just tested and I'm pretty sure provider attributes are applied `AND` not `OR`

apps/deploy-web/src/components/new-deployment/CreateLease.tsx
```diff
@@ -236,6 +239,44 @@ export const CreateLease: React.FunctionComponent<Props> = ({ dseq }) => {
...
+    const intervalId = setInterval(() => {
+      const blockTime = new Date(block.block.header.time).getTime();
+      const now = new Date().getTime();
+
+      setZeroBidsForTrialWarningDisplayed(now - blockTime > TRIAL_SIGNUP_WARNING_TIMEOUT);
+      clearInterval(intervalId);
+    }, 1000);
+
+    return () => clearInterval(intervalId);
+  }, [block, bids, isTrialing, numberOfRequests]);
```
baktun14: suggestion (non-blocking): I tend to put all the state variables and useEffect's at the top of the file

apps/deploy-web/src/context/WalletProvider/WalletProvider.tsx
```diff
@@ -314,6 +315,7 @@ export const WalletProvider = ({ children }) => {
         isCustodial: !isManaged,
         isWalletLoading: isLoading,
         isTrialing: isManaged && !!managedWallet?.isTrialing,
+        isOnboarding: isManaged && !managedWallet?.stripeCustomerId,
```
baktun14: question (blocking): shouldn't this be to check if the user has a user account?

baktun14 (changes requested):

apps/deploy-web/src/context/WalletProvider/WalletProvider.tsx
jzsfkzm: My understanding was someone is considered "onboarding" when already has a managed wallet, but doesn't have a Stripe customer id (yet). But is that the case? And is this value properly covering that case?

apps/deploy-web/src/components/new-deployment/ManifestEdit.tsx
jzsfkzm: Changed it, now it's either one or the other.
Btw what is the custom here, should I resolve issues you raise, or is it usually closed by the one who raised the issue?

apps/deploy-web/src/components/new-deployment/CreateLease.tsx
jzsfkzm: I see your point, moved it up.

apps/deploy-web/src/components/new-deployment/ManifestEdit.tsx
baktun14: It depends on the complexy of the change, if it's a simple fix you can just resolve it.

apps/deploy-web/src/context/WalletProvider/WalletProvider.tsx
baktun14: There are 3 user wallet states:
- anonymous trial account with just a managed wallet
- user account with email with a managed wallet, still with trial
- user account with email with a managed wallet, with added funds with credit card

The one we want to cover here is the 2nd one.

jzsfkzm: But isn't it covering that case? isManaged and has no Stripe ID.

baktun14: Having a wallet doesn't mean that a user has a user account. We allow users to start trial anonymously and it creates a userSetting row in db without `userId` so we need to check if the user has `.userId` defined.

baktun14: For reference: https://github.com/akash-network/console/blob/ad173e2f74248f2acb73d81a61065f0360c0cd80/apps/deploy-web/src/hooks/useUser.ts#L18

apps/api/src/billing/services/trial-validation/trial-validation.service.ts
```diff
@@ -1,30 +1,25 @@
...
-        });
-
-        const hasSignedByAllOf = group.requirements.signedBy.allOf.every(signedBy => {
-          return signedBy === AUDITOR;
-        });
-        assert(hasTrial && hasSignedByAllOf, 400, `provider not authorized: ${group.requirements.attributes}`);
-      });
+      if (userWallet.isTrialing) {
+        this.validateAttribute(value.groups, TRIAL_ATTRIBUTE);
+      } else if (!userWallet.stripeCustomerId) {
```
ygrishajev, line 20: why is this check needed? `isTrialing` should be enough. it's put to `false` once user does a first top up, meaning there is a stripe customer at this point.

ygrishajev:
```suggestion
      } else {
```

apps/deploy-web/src/components/new-deployment/CreateLease.tsx
```diff
@@ -189,6 +194,32 @@ export const CreateLease: React.FunctionComponent<Props> = ({ dseq }) => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [search, bids, providers, isFilteringFavorites, isFilteringAudited, favoriteProviders]);
 
+  const [zeroBidsForTrialWarningDisplayed, setZeroBidsForTrialWarningDisplayed] = useState(false);
+  const { data: block, refetch: getBlock } = useBlock(dseq, {
+    disabled: true
+  });
+
+  useEffect(() => {
+    getBlock();
```
ygrishajev: what is this call for?

apps/deploy-web/src/components/new-deployment/CreateLease.tsx
```diff
@@ -189,6 +194,32 @@ export const CreateLease: React.FunctionComponent<Props> = ({ dseq }) => {
...
+  }, [getBlock]);
+
+  useEffect(() => {
+    if (!isTrialing || numberOfRequests === 0 || (bids && bids?.length > 0)) {
```
ygrishajev:
```suggestion
    if (!isTrialing || numberOfRequests === 0 || (bids && bids.length > 0)) {
```

```suggestion
    if (!isTrialing || numberOfRequests === 0 || bids?.length > 0) {
```

apps/deploy-web/src/components/new-deployment/CreateLease.tsx
```diff
@@ -459,6 +502,30 @@ export const CreateLease: React.FunctionComponent<Props> = ({ dseq }) => {
...
       )}
+
+      {zeroBidsForTrialWarningDisplayed && (
+        <div className="pt-4">
+          <Card>
+            <CardContent>
+              <div className="px-16 pb-4 pt-6 text-center">
+                <h3 className="mb-4 text-xl font-bold">Waiting for bids</h3>
+                <p className="mb-8">
+                  Looks like you are not getting any bids. This is likely because all trial providers are currently being used by others. Console has{" "}
```
ygrishajev: Just passed this to ChatGPT :)

"It looks like you’re not receiving any bids. This is likely because all trial providers are currently in use. Console offers {trialProviderCount} providers for trial users, but many more are available for non-trial users. To access the full list of providers, we recommend signing up."

apps/api/src/billing/services/trial-validation/trial-validation.service.ts
jzsfkzm: Are you sure? My intention was to validate for this attribute only when the user is onboarding (already has a userId but hasn't paid yet).

apps/deploy-web/src/components/new-deployment/CreateLease.tsx
jzsfkzm: Chose the prior one, as just `bids?.length > 0` fails on type checks with "'bids.length' is possibly 'undefined'".

apps/deploy-web/src/components/new-deployment/CreateLease.tsx
jzsfkzm: Sounds better indeed, changed it.

apps/deploy-web/src/components/new-deployment/CreateLease.tsx
jzsfkzm: Uhm, was a residue from local testing, not even there any more.

apps/api/src/billing/services/trial-validation/trial-validation.service.ts
ygrishajev: https://github.com/akash-network/console/blob/main/apps/api/src/billing/services/refill/refill.service.ts#L70

it doesn't hurt to leave it like that rn prob, but it might be unexpected that we rely on 2 different fields for the same purpose if we start changing smth

apps/deploy-web/src/context/WalletProvider/WalletProvider.tsx
```diff
@@ -320,6 +321,7 @@ export const WalletProvider: React.FC<{ children: React.ReactNode }> = ({ childr
         isCustodial: !isManaged,
         isWalletLoading: isLoading,
         isTrialing: isManaged && !!managedWallet?.isTrialing,
+        isOnboarding: !!user?.id && (!user?.planCode || user?.planCode === "COMMUNITY"),
```
baktun14:
```suggestion
        isOnboarding: !user?.userId,
```

apps/deploy-web/src/components/new-deployment/CreateLease.tsx
```diff
@@ -467,6 +506,31 @@ export const CreateLease: React.FunctionComponent<Props> = ({ dseq }) => {
...
+      {zeroBidsForTrialWarningDisplayed && (
+        <div className="pt-4">
+          <Card>
+            <CardContent>
+              <div className="px-16 pb-4 pt-6 text-center">
+                <h3 className="mb-4 text-xl font-bold">Waiting for bids</h3>
+                <p className="mb-8">
+                  It looks like you’re not receiving any bids. This is likely because all trial providers are currently in use. Console offers{" "}
+                  {trialProviderCount} providers for trial users, but many more are available for non-trial users. To access the full list of providers, we
+                  recommend signing up.
```
baktun14:
```suggestion
                  It looks like you’re not receiving any bids. This is likely because all trial providers are currently in use. Console offers{" "}
                  {trialProviderCount} providers for trial users, but many more are available for non-trial users. To access the full list of providers, we recommend signing up and adding funds to your account.
```

jzsfkzm: Thanks, changed it.

apps/deploy-web/src/context/WalletProvider/WalletProvider.tsx
jzsfkzm: Thank you, changed this one too.

apps/deploy-web/src/context/WalletProvider/WalletProvider.tsx
```diff
@@ -320,6 +321,7 @@ export const WalletProvider: React.FC<{ children: React.ReactNode }> = ({ childr
         isCustodial: !isManaged,
         isWalletLoading: isLoading,
         isTrialing: isManaged && !!managedWallet?.isTrialing,
+        isOnboarding: !user?.userId,
```
baktun14: Sorry I just realized this doesn't make sense, we want to check if the user has an account, but not paid so trialing.
```suggestion
        isOnboarding: !!user?.userId && isManaged && !!managedWallet?.isTrialing,
```

baktun14 (approved):

PR #939, opened 2025-03-04
fix(provider): fixed loading issue on some pages, fixed some guides

### LINKED ISSUE #878: Guide users about Public IP requirements
From the initial onboarding feedback, users were unclear if their nodes need to have public IPs are not. To clarify this, we should add another bullet point in this checklist

Image

- "At least one control node needs to have a public IP and add the nodes need to be on the same layer-2 network. Ideally all the control plane nodes should have public IPs

### LINKED ISSUE #880: Provider Guidance on Control Node connection
Users (from the onboarding feedback) are sometimes unclear on what the control machine is and why they need to provide auth to connect to it. To clarify let's make two changes:

1. Update the description here to say “Provider Console need to connect to log into your control node, in order to make any configuration changes to your provider”

Image

2. Investigate if we can pull and pre-populate the IP of the control node (for users who set up their provider using provider console)

jigar-arc10:
closes #878 
closes #880


devalpatel67 (approved):
LGTM

PR #941, opened 2025-03-04
refactor: enables strict ts types for http-sdk

stalniy:
## Why 

To iteratively migrate to strict types

## What 

Migrates @akasnetwork/http-sdk to strict types.

### DIFF
--- a/packages/http-sdk/package.json
+++ b/packages/http-sdk/package.json
@@ -8,7 +8,8 @@
   "main": "src/index.ts",
   "scripts": {
     "format": "prettier --write ./*.{ts,json} **/*.{ts,json}",
-    "lint": "eslint ."
+    "lint": "eslint .",
+    "validate:types": "tsc --noEmit && echo"
   },
   "dependencies": {
     "axios": "^1.7.2",
--- a/packages/http-sdk/src/authz/authz-http.service.ts
+++ b/packages/http-sdk/src/authz/authz-http.service.ts
@@ -3,6 +3,7 @@ import { isFuture } from "date-fns";
 
 import { HttpService } from "../http/http.service";
 import type { Denom } from "../types/denom.type";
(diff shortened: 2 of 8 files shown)


packages/http-sdk/package.json
```diff
@@ -9,7 +9,8 @@
   "scripts": {
     "format": "prettier --write ./*.{ts,json} **/*.{ts,json}",
     "lint": "eslint .",
-    "update-apps-local-deps": "mvm-update"
+    "update-apps-local-deps": "mvm-update",
+    "validate:types": "tsc --noEmit && echo"
```
stalniy, line 12: note: `&& echo` at the end to ignore staged files which `lint-staged` passes to command. This is a temporary thing and after migration this npm command will be removed

packages/http-sdk/src/authz/authz-http.service.ts
```diff
@@ -114,50 +116,45 @@ export class AuthzHttpService extends HttpService {
...
-    return !!(await this.getValidDepositDeploymentGrantsForGranterAndGrantee(granter, grantee))
+    return !!(await this.getValidDepositDeploymentGrantsForGranterAndGrantee(granter, grantee));
   }
 
   async paginateDepositDeploymentGrants(
     options: ({ granter: string } | { grantee: string }) & { limit: number },
     cb: (page: DepositDeploymentGrantResponse["grants"]) => Promise<void>
   ): Promise<void> {
-    let nextPageKey: string | null = null;
+    let nextPageKey: unknown;
```
stalniy, line 130: note: required to get proper type inference work in a loop. Probably due to type narrowing/widening inside loop ts cannot properly infer type of `response` variable if `nextPageKey` is of union type

baktun14 (dismissed):

PR #942, opened 2025-03-04
feat(deployment): close deployment api endpoint

baktun14:


apps/api/test/functional/deployments.spec.ts
```diff
@@ -145,19 +127,67 @@ describe("Deployments API", () => {
     return { adminApiKeySecret };
   }
 
-  afterEach(async () => {
-    await dbService.cleanAll();
-  });
+  function createMockDeploymentInfo(wallets: UserWalletOutput[], dseq: string, deploymentInfo?: RestAkashDeploymentInfoResponse) {
```
ygrishajev, line 130: non-blocking: it would be nice if extracted to a seeder

ygrishajev: non-blocking: also faker is a good thing where it makes sense

ygrishajev (approved):

apps/api/test/functional/deployments.spec.ts
baktun14: I will keep working on those tests in my current create lease branch 👍

PR #954, opened 2025-03-06
refactor: migrates database package to strict types

stalniy:
## Why 

To iteratively migrate to strict types

## What 

Migrates @akasnetwork/database to strict types.

### DIFF
--- a/packages/database/dbSchemas/akash/deploymentGroupResource.ts
+++ b/packages/database/dbSchemas/akash/deploymentGroupResource.ts
@@ -15,47 +15,47 @@ export class DeploymentGroupResource extends Model {
   /**
    * The unique identifier for the database deployment group resource
    */
-  @Required @Column(DataTypes.UUID) deploymentGroupId: string;
+  @Required @Column(DataTypes.UUID) deploymentGroupId!: string;
   /**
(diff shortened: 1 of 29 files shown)


packages/database/dbSchemas/akash/deploymentGroupResource.ts
```diff
@@ -15,47 +15,47 @@ export class DeploymentGroupResource extends Model {
...
    * The gpu units of the deployment group resource
    * 1 GPU unit = 1 GPU
    */
-  @Required @Column gpuUnits: number;
+  @Required @Column gpuUnits!: number;
   /**
    * The gpu vendor of the deployment group resource
    */
-  @Column gpuVendor: string;
+  @Column gpuVendor!: string;
```
baktun14, line 32: question (blocking): Does this make sense for non required fields?

stalniy: this is just how it is right now. The bed thing about decorators is that they do not have influence on the final type. So, this doesn't break and doesn't improve types by itself but just makes this compilable in strict mode

baktun14 (dismissed):

PR #955, opened 2025-03-06
chore(ci): adds auto detection of local dependencies in builds

stalniy:
## Why

Currently when app depends on a package and there are package changes, our CI doesn't whether app is buildable. This means somebody can make a change in packages but will discover possible consequences only app files are changed which may be done by another person or in another PR causing deployment delays and increasing time to delivery

## What

Automatically detects dependencies on local packages inside apps to check whether app related build job needs to be triggered


packages/logging/src/index.ts
```diff
@@ -1 +1,2 @@
-export * from './servicies/logger/logger.service';
+export * from "./servicies/logger/logger.service";
+console.log("logging");
```
ygrishajev:
```suggestion

```

ygrishajev (approved):

PR #959, opened 2025-03-07
test(wallet): creates a unique wallet for each spec for each run

ygrishajev:
This PR introduces a tests wallet setup for functional tests. This should solve 2 issues:
1) account sequence mismatch issues when running same tests at the same time
2) once we have a pg db generated for each run we'll be able to run tests in parallel and improve run time

Before each run:
1) it generates a local faucet wallet
2) fills it via sandbox faucet
3) generates a unique wallet for every functional spec
4) sends tokens from local faucet wallet to spec wallets
5) sets spec wallet mnemonics in env


apps/api/src/chain/chain-events/chain-events.module.ts
stalniy, line 1: **question(blocking):** is this empty file committed accidentally?

apps/api/test/services/test-wallet.service.ts
```diff
@@ -0,0 +1,148 @@
...
+    const initialBalance = await this.balanceHttpService.getBalance(faucetAddress, "uakt");
+    const initialAmount = initialBalance?.amount;
+    let updatedAmount = initialAmount;
+
+    await this.topUpFaucetWallet(faucetAddress);
+
+    while (initialAmount === updatedAmount) {
+      const updatedBalance = await this.balanceHttpService.getBalance(faucetAddress, "uakt");
+      updatedAmount = updatedBalance?.amount;
+      await new Promise(resolve => setTimeout(resolve, 1000));
```
stalniy: **suggestion(non-blocking):** you can use promisified `setTimeout` from `import { setTimeout as delay } from 'timers/promises'`
```suggestion
      await delay(resolve, 1000);
```

apps/api/jest.config.js
```diff
@@ -29,7 +29,9 @@ module.exports = {
       ...common,
       testMatch: ["<rootDir>/test/functional/**/*.spec.ts"],
       setupFilesAfterEnv: ["./test/setup-functional-tests.ts"],
-      setupFiles: ["./test/setup-functional-env.ts"]
+      setupFiles: ["./test/setup-functional-env.ts"],
+      globalSetup: "./test/setup-global-functional.ts",
+      testEnvironment: "./test/custom-jest-environment.ts"
```
stalniy, line 35: **praise**: 💪

stalniy (dismissed):

apps/api/src/chain/chain-events/chain-events.module.ts
ygrishajev: Yup

baktun14 (dismissed):
👌

stalniy (dismissed):

stalniy (approved):

Issue #917, opened 2025-02-27
Track and prebuild templates instead of fetching them from github API


stalniy:
## Why

There are few drawbacks with the current solution:
1. Cold start is slow. Currently API fetches all templates from few github repos, this is a slow process which may take up to few seconds. 
2. Github API is a dependency which may not work or may work slow

This templates are rarely changed and can be prebuilt and exposed via some static hosting or just added as an asset to console-api

## What

There are few ways to do it. 
1. We can prebuilt them in GHA, this requires adding the same GHA in every repository. May not work if we do not own some of them
2. Prebuild on every commit to API, this will allow us to release templates on every commit. If somebody will add template to repository, eventually it will land on production. If we need this to happen quicker, we can release a new version manually
3. Prebuild periodically, for example every day or few times per day.

## Acceptance criteria

- [x] analyze and choose the best strategy, potentially it can be a combination of 2 & 3
- [x] implement this strategy 
- [x] make sure that generated templates structure is efficient for further processing in runtime

stalniy:
I was able to receive [524 status code](https://http.dev/524) in response after deployment in console-beta

<img width="587" alt="Image" src="https://github.com/user-attachments/assets/59abdd13-7133-49af-ae78-68a1b4eb2064" />

stalniy:
I decided not to change the actual code but just move part of it at docker build time which happens in GHA what allows us to reuse github actions temporary access token. Every time we build and deploy an image, templates will be prebuilt. Additionally the old logic was optimized to be much faster and in case templates change happens while we haven't deployed anything, it will detect it and re-collect all templates.

PR #962, opened 2025-03-07
refactor: adds warmUpTemplatesCache script

stalniy:
## Why

ref: #917

## What

1. parallelized fetching templates logic to speed up it. The current implementation fetches all templates in 180+secs on my laptop
2. created `build:akash-templates` npm script which uses existing templates fetching logic to fetch templates at docker build time. The nice thing here is that we can re-use GHA provided GITHUB_TOKEN. It's short-lived and cannot be used after job is done
3. adds default local dependency to all apps -> packages/docker, whenever there are changes in that folder, we need to rebuild all apps to ensure they are buildable
4. splits `templatesReposService.ts` into 2 files: one depends on `env` service and another one doesn't. This is done to ensure that env var validation will not cause issues in future when some env var becomes required. Anyway, this needs to be refactored to service/repository pattern + DI

to show real numbers with old and optimized code:
```
new =>   4.85s user 0.37s system 76% cpu 6.833 total
old =>  5.29s user 0.43s system 3% cpu 3:01.40 total
```


apps/api/src/utils/mapInBatches.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * Calls callback `options.batchSize` times in parallel for every item in array sequentially.
+ * This is useful for rate limiting requests to external services.
+ *
+ * @returns an array of mapped results
+ */
+export async function mapInBatches<T, U>(items: T[], callback: (item: T, index: number) => Promise<U>, options?: MapInBatchesOptions): Promise<U[]> {
```
baktun14: question (non-blocking): I fail to see where this is used

baktun14 (dismissed):

apps/api/src/utils/mapInBatches.ts
stalniy: Yes, nice catch! too much stashes in the late evening

stalniy:
I compared the output in data folder for new and old approach with `diff -u` and the json files are identical

baktun14 (approved):

Issue #175, opened 2024-04-30
SDL builder improvements


Redm4x:
- [x] Add support for RAM storage in the SDL Builder.
- [x] Support multiple ephemeral storages
- [x] Support multiple persistent storage
- [x] Fix command import/export, [use this sdl to test](https://gist.githubusercontent.com/andy108369/d20f554cc251a8622651f1b0ecfb74c5/raw/2071f6ff512de1062ebad5a44f6665e2256f8ef8/dolphin-sdl.yaml)

jzsfkzm:
Hi @baktun14, I added a change where command is a multi-line string on the UI. However I'm not convinced auto-adding "sh -c" is always what the user wants. Maybe sometimes they want to use "bash -c" instead. That can be achieved though if they enter bash -c in the command input and the rest as args, like they did before.
Also I made a bit a bit of a refactoring here, adding label to the TextArea component. Let me know if this fits the needs, or should work otherwise.

PR #968, opened 2025-03-08
feat: introduce multi-line commands

jzsfkzm:
refs #175

### DIFF
--- a/apps/deploy-web/src/utils/sdl/sdlGenerator.ts
+++ b/apps/deploy-web/src/utils/sdl/sdlGenerator.ts
@@ -3,6 +3,25 @@ import yaml from "js-yaml";
 import { ExposeType, ProfileGpuModelType, ServiceType } from "@src/types";
 import { defaultHttpOptions } from "./data";
 
+export const buildCommand = (command: string) => {
+  if (command.trim() === "sh -c") {
+    return ["sh", "-c"];
+  }
+
+  const lines = command.split("\n");
+
+  if (lines.length > 1 || command.startsWith("sh -c")) {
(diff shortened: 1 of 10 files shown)


apps/deploy-web/src/utils/sdl/sdlGenerator.ts
```diff
@@ -3,6 +3,25 @@ import yaml from "js-yaml";
...
 import { defaultHttpOptions } from "./data";
 
+export const buildCommand = (command: string) => {
+  if (command.trim() === "sh -c") {
+    return ["sh", "-c"];
+  }
+
+  const lines = command.split("\n");
+
+  if (lines.length > 1 || command.startsWith("sh -c")) {
```
ygrishajev, line 13: I might be missing the point here but there might be other executables out there then `sh`, should this be addressed?

baktun14: I haven't tested this in a long time

baktun14 (approved):
question (non-blocking): I assume you tested [the sdl](https://gist.githubusercontent.com/andy108369/d20f554cc251a8622651f1b0ecfb74c5/raw/2071f6ff512de1062ebad5a44f6665e2256f8ef8/dolphin-sdl.yaml) in question for the command?

apps/deploy-web/src/utils/sdl/sdlGenerator.ts
ygrishajev: ok, do you think it's safe to merge without considering this?

PR #969, opened 2025-03-10
feat(deployment): managed api create leases

baktun14:
#767


apps/api/scripts/lease-flow.ts
```diff
@@ -0,0 +1,155 @@
+import axios from "axios";
+import * as dotenv from "dotenv";
```
stalniy: **issue(blocking):** we do not have direct dependency on dotenv. You need to use @dotenvx/dotenvx. Also you can simplify call by doing [this](https://github.com/dotenvx/dotenvx?tab=readme-ov-file#multiple-environments): 

```ts
dotenvx run -f .env.local -- ts-node scripts/lease-flow.ts
```

apps/api/scripts/lease-flow.ts
```diff
@@ -0,0 +1,155 @@
+import axios from "axios";
```
stalniy: **suggestion(blocking):** would be nice to describe the purpose of this script at least in comments at the top of the file

apps/api/src/auth/services/api-key/api-key-auth.service.ts
```diff
@@ -25,8 +26,9 @@ export class ApiKeyAuthService {
     const [prefix, type, env] = apiKey.split(".");
     assert(prefix === this.API_KEY_PREFIX && type === this.API_KEY_TYPE && env === this.DEPLOYMENT_ENV, 401, "Invalid API key format");
 
-    const obfuscatedKey = this.apiKeyGenerator.obfuscateApiKey(apiKey);
-    const key = await this.apiKeyRepository.findOneBy({ keyFormat: obfuscatedKey });
+    const apiKeys = await this.apiKeyRepository.find();
+    const key = apiKeys.find(key => compare(apiKey, key.hashedKey));
```
stalniy: **issue(blocking):** bcryptjs `compare` function returns `Promise<boolean>`, this means find will return the first truthy item

apps/api/src/deployment/services/provider/provider-proxy.service.ts
```diff
@@ -0,0 +1,49 @@
...
+  keyPem?: string;
+  body?: string;
+  timeout?: number;
+  chainNetwork: string;
+  providerIdentity: ProviderIdentity;
+  headers?: Record<string, string>;
+}
+
+@singleton()
+export class ProviderProxyService {
```
stalniy, line 23: **question(non-blocking):** should we move it http-sdk? I wrote similar one in deploy-web

apps/api/src/deployment/services/provider/provider.service.ts
```diff
@@ -0,0 +1,64 @@
...
+          keyPem: options.keyPem,
+          chainNetwork: this.chainNetwork,
+          providerIdentity,
+          timeout: 60000
+        });
+
+        if (result) return result;
+      } catch (err) {
+        if (err.message?.includes("no lease for deployment") && i < this.MANIFEST_SEND_MAX_RETRIES) {
+          await new Promise(resolve => setTimeout(resolve, this.MANIFEST_SEND_RETRY_DELAY));
```
stalniy: **suggestion(non-blocking):** you can use `setTimeout` from `timers/promises` which is already promisified.

```ts
import { setTimeout as delay } from 'timers/promises'

await delay(1000) 
```

apps/api/src/deployment/services/sdl/sdl.service.ts
```diff
@@ -0,0 +1,55 @@
...
+  public getManifestYaml(sdlConfig: v2Sdl, networkType: NetworkType) {
+    const sdl = this.getSdl(sdlConfig, networkType);
+    return sdl.manifestSortedJSON();
+  }
+
+  public validateSdl(yamlJson: string) {
+    try {
+      SDL.fromString(yamlJson, "beta3");
+      return true;
+    } catch (error) {
```
stalniy: **suggestion(non-blocking):** you can write empty `catch` if you do not care about error. This shows that you do not care about the error and not just forgot to handle it somehow:

```ts
try {
 ///....
} catch { // <------ no error variable
  //...
}
```

apps/api/test/services/test-wallet.service.ts
```diff
@@ -141,7 +142,7 @@ export class TestWalletService {
   }
 
   private getFileName(path: string) {
-    return path.split("/").pop();
+    return path.split(/[/\\]/).pop();
```
stalniy: **suggestion(non-blocking):** I guess it's an attempt to handle OS dependent slashes. If so, you can use 

```ts
import { sep as FOLDER_SEP } from 'path';

return path.split(FOLDER_SEP);
```

stalniy (changes requested):

apps/api/src/deployment/services/provider/provider-proxy.service.ts
baktun14: Yea we should, I didn't want to start to refactor too many unrelated things. Ideally I think we should move out the certification validator so we could call the providers directly in the api without having to go through the proxy.

apps/api/test/services/test-wallet.service.ts
baktun14: Yea it is, oh cool didn't know about this.

apps/api/src/auth/services/api-key/api-key-auth.service.ts
baktun14: oh damn good catch

apps/api/scripts/lease-flow.ts
baktun14: Sounds good, it's just a script to test the flow and also acts as some kind of documentation

apps/api/scripts/lease-flow.ts
stalniy: by the way, the blocking part relates only to use dotenvx in the code. You can keep it the old. The code sample is just a non-blocking suggestion :)

baktun14: Sure, I'll change it to dotenvx

apps/api/src/deployment/services/provider/provider-proxy.service.ts
baktun14: https://github.com/akash-network/console/issues/1011

apps/api/src/deployment/services/deployment/deployment.service.ts
```diff
@@ -17,8 +18,9 @@ export class DeploymentService {
     private readonly leaseHttpService: LeaseHttpService,
     private readonly signerService: ManagedSignerService,
     @InjectWallet("MANAGED") private readonly masterWallet: Wallet,
-    private readonly billingConfigService: BillingConfigService,
-    private readonly rpcMessageService: RpcMessageService
+    private readonly rpcMessageService: RpcMessageService,
+    private readonly sdlService: SdlService,
+    @InjectBillingConfig() private readonly billingConfig: BillingConfig
```
ygrishajev: non-blocking
```suggestion
    private readonly billingConfig: BillingConfigService
```

apps/api/src/deployment/services/provider/provider.service.ts
```diff
@@ -0,0 +1,65 @@
...
+    if (!providerResponse) {
+      throw new Error(`Provider ${provider} not found`);
+    }
+
+    const providerIdentity: ProviderIdentity = {
+      owner: provider,
+      hostUri: providerResponse.provider.host_uri
+    };
+
+    // Send manifest
```
ygrishajev: non-blocking
```suggestion
```

ygrishajev (dismissed):

PR #971, opened 2025-03-10
chore(ci): refactors reusable-build-image into action

stalniy:
## Why

Currently release-api workflow fails because of #966 

## What

It requires passing GITHUB_TOKEN to release-api workflow for it to prefetch akash templates at build time. Because I need to pass this only for console-api app, I re-wrote reusable-build-image.yml into reusable action


.github/workflows/console-api-release.yml
```diff
@@ -25,6 +25,8 @@ jobs:
       packages: write
     with:
       tag: ${{ needs.release.outputs.git_tag }}
+    secrets:
```
ygrishajev: The workflow should inherit secrets as configured above

stalniy: ah, missed that sorry

ygrishajev (dismissed):

stalniy:
Close since benefits of this PR gives are not such big in comparison to risks

PR #1000, opened 2025-03-12
feat: dynamic amount of storages

jzsfkzm:
refs #175

Changed SDL form so one can add more storages. Will render a block of storage input for each storages when importing an SDL.

Default view:
Screenshot 2025-03-12 at 1 27 38

After clicking the "Add storage" button:
Screenshot 2025-03-12 at 1 28 34


apps/deploy-web/src/utils/akash/units.ts
```diff
@@ -39,7 +39,9 @@ export const storageUnits = [
...
 
 export const persistentStorageTypes = [
-  { id: 1, className: "beta1", name: "hdd" },
-  { id: 2, className: "beta2", name: "ssd" },
+  { id: 1, className: "beta1", name: "HDD" },
+  { id: 2, className: "beta2", name: "SSD" },
   { id: 3, className: "beta3", name: "NVMe" }
 ];
+
+export const ephemeralStorageTypes = [...persistentStorageTypes, { id: 4, className: "ram", name: "RAM" }];
```
baktun14, line 47: I'm not entirely sure about this, @troian can ephemeral storage be of type persistent as well?

baktun14:
Pretty cool, thanks!

apps/deploy-web/src/utils/akash/units.ts
jzsfkzm: Just to be clear, it's about the media it is stored on (eg. RAM or HDD or SSD).

ygrishajev (dismissed):
lgtm. @baktun14 would you mind to re-iterate on this one?

apps/deploy-web/src/utils/akash/units.ts
baktun14: Yea I'm just not sure if we can apply a specific media for ephemeral storage

baktun14 (changes requested):
I talked with the team and we can't have ephemeral storage treated the same way as persistent storage is configured (with mount and other configs)

Also, can you make the default type for persistent storage to nvme instead of ssd please?

Edit: I will write the spec to what we can and cannot do.

baktun14:
Ok so here's the specs:
- We can only have 1 ephemeral storage
- We can have multiple persistent storage, but they must have different names with different mount configs
- We can technically have multiple shm storage (ram, even though is doesn't make sense in practice), but they also must have different names
  - When the user specifies SHM (ram) it's always going to be `persistent: false`. We should make it as a dropdown between persistent or shm instead of adding the type "ram" in the list of media types. We can hide the read only option of SHM.
- We need to make sure there are no storage who use the same `name`
- Default value for persistent storage will be nvme

jzsfkzm:
> Ok so here's the specs:
> 
> * We can only have 1 ephemeral storage
> * We can have multiple persistent storage, but they must have different names with different mount configs
> * We can technically have multiple shm storage (ram, even though is doesn't make sense in practice), but they also must have different names
>   
>   * When the user specifies SHM (ram) it's always going to be `persistent: false`. We should make it as a dropdown between persistent or shm instead of adding the type "ram" in the list of media types. We can hide the read only option of SHM.
> * We need to make sure there are no storage who use the same `name`
> * Default value for persistent storage will be nvme

Thanks for the specs, I made some changes to conform these. Looks like this now (if you add a few storages):
Screenshot 2025-03-27 at 15 12 55

baktun14 (approved):

PR #1003, opened 2025-03-12
refactor: refactors part of deploy-web to strict typescript types

stalniy:
## Why 

to improve quality of the project and reliance on existing tools. To catch bugs at early stage

## What

1. introduced `tsconfig.strict.json` which includes only those files which were refactored and should adhere to strict type policy which is checked by lint-staged in case there are changes in those files
2. lint-staged validation check which ensures that refactored files do not degrade into prev state
3. types changes/additions

## Additional notes

If file inside tsconfig.strict.json references file which is not in tsconfig.strict.json that file will be type checked as well. There is no way to workaround this, that's why it's possible to refactor only the whole import subtree. The easiest way to do it is to refactor from bottom to top, starting from files which has the least imports. In case of modular structure, it's possible to refactor the whole module at once as well because it follows bottom-up approach

### DIFF
--- a/apps/deploy-web/src/context/SettingsProvider/SettingsProviderContext.tsx
+++ b/apps/deploy-web/src/context/SettingsProvider/SettingsProviderContext.tsx
@@ -197,7 +197,7 @@ export const SettingsProvider: FCWithChildren = ({ children }) => {
     return fastestNode;
   };
 
-  const updateSettings = newSettings => {
+  const updateSettings = (newSettings: any) => {
     setSettings(prevSettings => {
       clearQueries(prevSettings, newSettings);
       setLocalStorageItem("settings", JSON.stringify(newSettings));
@@ -206,7 +206,7 @@ export const SettingsProvider: FCWithChildren = ({ children }) => {
     });
   };
 
-  const clearQueries = (prevSettings, newSettings) => {
+  const clearQueries = (prevSettings: Settings, newSettings: Settings) => {
     if (prevSettings.apiEndpoint !== newSettings.apiEndpoint || (prevSettings.isCustomNode && !newSettings.isCustomNode)) {
       // Cancel and remove queries from cache if the api endpoint is changed
       queryClient.resetQueries();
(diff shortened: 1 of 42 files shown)


apps/deploy-web/src/services/analytics/analytics.service.ts
```diff
@@ -146,7 +146,7 @@ export class AnalyticsService {
 
     for (const key in user) {
       if (key !== "id") {
-        event.set(AMPLITUDE_USER_PROPERTIES_MAP[key] || AMPLITUDE_USER_PROPERTIES_MAP, user[key]);
+        event.set(AMPLITUDE_USER_PROPERTIES_MAP[key as keyof AnalyticsUser] || key, String(user[key as keyof typeof user]));
```
stalniy, line 149: @ygrishajev I need your attention here. Since I couldn't understand purpose of the code, and it doesn't match `event.set` types, I assumed there is a bug

apps/deploy-web/src/context/SettingsProvider/SettingsProviderContext.tsx
```diff
@@ -197,7 +197,7 @@ export const SettingsProvider: FCWithChildren = ({ children }) => {
     return fastestNode;
   };
 
-  const updateSettings = newSettings => {
+  const updateSettings = (newSettings: any) => {
```
baktun14, line 200:
```suggestion
  const updateSettings = (newSettings: Settings) => {
```

baktun14 (approved):

apps/deploy-web/src/context/SettingsProvider/SettingsProviderContext.tsx
stalniy: I tried but then there is another type related issue appears, that's why I specified it as `any`

apps/deploy-web/src/services/analytics/analytics.service.ts
ygrishajev: ur change looks correct to me

ygrishajev (approved):

PR #1012, opened 2025-03-12
feat(notifications): adds basic alerts service

ygrishajev:


.github/workflows/notifications-validate-n-build.yml
stalniy, line 1: **question(non-blocking):** we currently have 2 conventions for service github CI workflow:
* - ${service}-build.yml
* - ${service}-validate-n-build.yml

Does it makes to unify them? It's easier to rename to `${service}-build` because all others uses the same convention but I'm ok with any consistent naming.

apps/notifications/README.md
```diff
@@ -0,0 +1,98 @@
...
+  <a href="https://paypal.me/kamilmysliwiec" target="_blank"><img src="https://img.shields.io/badge/Donate-PayPal-ff3f59.svg" alt="Donate us"/></a>
+    <a href="https://opencollective.com/nest#sponsor"  target="_blank"><img src="https://img.shields.io/badge/Support%20us-Open%20Collective-41B883.svg" alt="Support us"></a>
+  <a href="https://twitter.com/nestframework" target="_blank"><img src="https://img.shields.io/twitter/follow/nestframework.svg?style=social&label=Follow" alt="Follow us on Twitter"></a>
+</p>
+  <!--[![Backers on Open Collective](https://opencollective.com/nest/backers/badge.svg)](https://opencollective.com/nest#backer)
+  [![Sponsors on Open Collective](https://opencollective.com/nest/sponsors/badge.svg)](https://opencollective.com/nest#sponsor)-->
+
+## Description
+
+[Nest](https://github.com/nestjs/nest) framework TypeScript starter repository.
```
stalniy: **question(non-blocking):** does it makes sense to update README to actually reflect what this service is for?

apps/notifications/jest.config.js
```diff
@@ -0,0 +1,29 @@
+const MAP_ALIASES = {
```
stalniy: **suggestion(non-blocking):** let's change it to be ts file because then we get options autocompletions in IDE

apps/notifications/src/app.module.js
stalniy, line 1: **thought(blocking):** I guess this is committed by mistake

apps/notifications/src/app.module.js.map
stalniy, line 1: **thought(blocking):** I guess this is committed by mistake

apps/notifications/src/broker/providers/pg-boss/pg-boss.provider.ts
```diff
@@ -0,0 +1,32 @@
+import { Provider } from '@nestjs/common';
+import { ConfigService } from '@nestjs/config';
+import { Client } from 'pg';
```
stalniy, line 3: **question(non-blocking):** what do you think about https://github.com/porsager/postgres ?

apps/notifications/src/broker/services/broker/broker.service.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+      providers: [
+        BrokerService,
+        { provide: PgBoss, useValue: pgBossMock },
+        { provide: Client, useValue: pgClientMock },
+        { provide: ConfigService, useValue: configServiceMock },
+        { provide: LoggerService, useValue: loggerMock },
+      ],
+    }).compile();
+
+    service = module.get<BrokerService>(BrokerService);
```
stalniy: **idea(non-blocking):** Just want to share experience here of using `ng-mocks` in Angular project. This is how this would look like in Angular service test with ng-mocks. Just :

```ts
function setup(input: TestInput) {
  const module: TestingModule = await Test.createTestingModule({
      providers: [
        BrokerService,
        MockProvider(PgBoss),
        MockProvider(Client),
        MockProvider(ConfigService),
        MockProvider(LoggerService)
      ],
  }).compile();
  return module;
}
```

this `MockProvider` helper is very nice because it automatically creates mock based on the provider methods. And it allows to partially change some of them, for example:

```ts
MockProvider(LoggerService, {
   info: jest.fn()
})
```

This opens possibility to use general mocks but pass custom implementation details to `setup` function:

```ts
function setup(input) {
   const module: TestingModule = await Test.createTestingModule({
      providers: [
        BrokerService,
        MockProvider(PgBoss),
        MockProvider(Client),
        MockProvider(ConfigService),
        MockProvider(LoggerService, { info: input.loggerInfo }) // <----- passing custom `info` method from test
      ],
  }).compile();
  return module;
}
```

And then in test it's possible to create and pass custom spies without sharing them between all tests:

```ts
it('....', () => {
  const loggerInfo = jest.fn();
  const module = setup({ loggerInfo });
})
```

Also just reminder to take a look at https://github.com/akash-network/console/discussions/910

apps/notifications/README.md
ygrishajev: ofc it does, as you can see it's WIP

apps/notifications/src/broker/providers/pg-boss/pg-boss.provider.ts
ygrishajev: I agree it is better. but pg-boss is using that by default, but now when I think about it, I guess we could use anything here.

.github/workflows/notifications-validate-n-build.yml
ygrishajev: I'm thinking it's better when the name reflect the sense, so if we are to unify, the longer name is better imo

apps/notifications/src/broker/providers/pg-boss/pg-boss.provider.ts
ygrishajev: I think I'll go with this anyway for the above reason. also looks like the tx syntax better matches the usecase

apps/notifications/package.json
```diff
@@ -0,0 +1,66 @@
+{
+  "name": "@akashnetwork/notifications",
+  "version": "0.0.1",
+  "private": true,
+  "description": "",
+  "license": "UNLICENSED",
```
baktun14:
```suggestion
  "license": "Apache-2.0",
```

PR #1013, opened 2025-03-13
chore: update mainnet-nodes.json

baktun14:


config/mainnet-nodes.json
```diff
@@ -1,8 +1,8 @@
 [
   {
     "id": "akashnet.net",
-    "api": "https://api.akashnet.net:443",
-    "rpc": "https://rpc.akashnet.net:443"
+    "api": "https://consoleapi.akashnet.net:443",
+    "rpc": "https://consolerpc.akashnet.net:443"
```
stalniy, line 5: **question:** why domain were changed? was this done by core team as part of network upgrade?

baktun14: It's another rpc dedicated for console, the other one had issues during the upgrade so I removed it. We could re-add it.

PR #1014, opened 2025-03-13
fix(provider): replace rpc for proxy

baktun14:


packages/net/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@akashnetwork/net",
-  "version": "0.0.1",
+  "version": "0.0.2",
```
stalniy, line 3: **nitpick:** there is no need to update versions of internal packages after https://github.com/akash-network/console/pull/955. GHA automatically checks for changes in internal packages and re-tests corresponding application

PR #1017, opened 2025-03-13
fix(wallet): fix fetching of authz

baktun14:


apps/deploy-web/src/queries/useGrantsQuery.ts
```diff
@@ -29,7 +29,7 @@ export function useGranteeGrants(address?: string, options: { enabled?: boolean;
 
   // TODO: ensure app is not loaded till settings are fetched
   //   Issue: https://github.com/akash-network/console/issues/600
-  options.enabled = !!options.enabled && !!address && !!settings.apiEndpoint;
```
stalniy, line 32: **thought(blocking):** I see that `enabled` prop is used on [apps/deploy-web/src/components/deployments/DeploymentDepositModal.tsx](https://github.com/akash-network/console/blob/main/apps/deploy-web/src/components/deployments/DeploymentDepositModal.tsx#L92)

Either you need to remove it there or keep it here. Probably if `enabled` is not passed it should be treated as this hook is enabled by default

packages/http-sdk/src/authz/authz-http.service.ts
```diff
@@ -137,7 +137,7 @@ export class AuthzHttpService extends HttpService {
           params: { "pagination.key": nextPageKey, "pagination.limit": options.limit }
         })
       );
-      nextPageKey = response.pagination.next_key;
+      nextPageKey = response?.pagination?.next_key;
```
stalniy: **question(blocking):** do you need to update `DepositDeploymentGrantResponse` type in this case, to ensure that nullish pagination is handled properly everywhere?

stalniy (changes requested):

packages/http-sdk/src/authz/authz-http.service.ts
baktun14: This is such a weird case, I want to check this with you when I'm online tomorrow lol, but when I log the values in console it's defined and for some reason it can't read next_key of null

apps/deploy-web/src/queries/useGrantsQuery.ts
baktun14: I will check, but the default to true was being overridden in the list of authz page

packages/http-sdk/src/authz/authz-http.service.ts
baktun14: Actually yea I updated the type

apps/deploy-web/src/components/authorizations/Authorizations.tsx
```diff
@@ -54,7 +54,8 @@ export const Authorizations: React.FunctionComponent = () => {
     select: selectNonMaster
   });
   const { data: granteeGrants, isLoading: isLoadingGranteeGrants } = useGranteeGrants(address, {
-    refetchInterval: isRefreshing === "granteeGrants" ? refreshingInterval : defaultRefetchInterval
+    refetchInterval: isRefreshing === "granteeGrants" ? refreshingInterval : defaultRefetchInterval,
+    enabled: true
```
stalniy: **thought(non-blocking):** I think we need to make `enabled` to be true in `useGranteeGrants` hook to reflect the behavior of `useQuery` hook, its `enabled` option is true if not passed

stalniy (dismissed):

apps/deploy-web/src/components/authorizations/Authorizations.tsx
baktun14: Yea that was the purpose of having it set to true by default in the params, but it's an object and gets overridden by whatever options passed. I can make it true if undefined

PR #1031, opened 2025-03-14
chore: improves cache utilization in docker build

stalniy:
## Why

Very often we need to install npm dependencies, this is time consuming and currently we refetch dependencies on every release of the every app. Basically every change in package.json invalidates docker caching mostly because of package version change. However the actual dependencies are not frequently changed. 

This both speeds up local dev rebuild and GHA builds. For local I use:

```
npm run dc:up:dev -- api deploy-web --build --force-recreate
```

after this changes it's super fast, so I can be sure that my node_modules are up to date with the latest package.json changes

## What

1. Removes unused external dependency in API's webpack file
2. Updates Dockerfiles to ignore version change and reuse cached layers 
3. Cleans npm cache to reduce docker image size
4. Removes ./packages from `Dockerfile.node` because we compile bundle in dist
5. Reduces docker image sizes

### DIFF
--- a/packages/docker/Dockerfile.nextjs
+++ b/packages/docker/Dockerfile.nextjs
@@ -12,17 +12,24 @@ ENV NEXT_TELEMETRY_DISABLED 1
 
 WORKDIR /app
 
+FROM base as prepare
+
+COPY package*.json /app
+COPY /$WORKSPACE/package.json /app/$WORKSPACE/package.json
+COPY --parents /packages/*/package.json /app
+# reset app version to 1.0.0 to avoid cache busting after every release
+RUN npm version --allow-same-version 1.0.0 -w $WORKSPACE --no-workspaces-update && \
+    npm install --package-lock-only
+
 FROM base AS development
 
 ENV NODE_ENV development
 
 RUN apk add --no-cache libc6-compat
 
-COPY package*.json /app
-COPY /$WORKSPACE/package.json /app/$WORKSPACE/package.json
-COPY --parents /packages/*/package.json /app
+COPY --from=prepare /app .
 
-RUN npm ci
+RUN npm ci && npm cache clean --force
 
 COPY $WORKSPACE ./$WORKSPACE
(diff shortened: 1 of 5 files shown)


packages/docker/Dockerfile.nextjs
```diff
@@ -12,17 +12,23 @@ ENV NEXT_TELEMETRY_DISABLED 1
 
 WORKDIR /app
 
+FROM base as prepare
+
+COPY package*.json /app
+COPY /$WORKSPACE/package.json /app/$WORKSPACE/package.json
+COPY --parents /packages/*/package.json /app
+# reset app version to 1.0.0 to avoid cache busting after every release
+RUN npm version --allow-same-version 1.0.0 -w $WORKSPACE --no-workspaces-update
```
baktun14: question (non-blocking): will that affect the version of the app once deployed?

stalniy: No, it's just to have a stable version to run npm ci and make sure docker caches this. Later versions are overwritten back to actual values at these lines -> https://github.com/akash-network/console/pull/1031/files#diff-a75f5b03c61fa0c6a799dea7b23933d2cee82c2baf971036b727cac4ccd162dcR33-R34

baktun14 (dismissed):

packages/docker/Dockerfile.nextjs
stalniy: **BLOCKING**: though this requires a bit more investigation. The issue is that `npm version` installs dependencies if they are not installed, could you imagine that?)) 

Addition of `--no-workspaces-update` disables this behavior but it also does not update package-lock in this case. But I need to update it as well. Without this option `npm version` updates both json files but also installs dependencies 😄

stalniy: the fix is to use `npm install --package-lock-only` which updates only package lock file and do not install dependencies

ygrishajev (approved):

PR #1033, opened 2025-03-14
chore(observability): adds healthz to console-web

ygrishajev:


apps/deploy-web/src/context/CustomChainProvider/CustomChainProvider.tsx
```diff
@@ -1,6 +1,5 @@
 "use client";
 import "@interchain-ui/react/styles";
-import "@interchain-ui/react/globalStyles";
```
ygrishajev: there's no such file neither in doc or in node_modules

apps/deploy-web/src/pages/api/healthz.ts
```diff
@@ -0,0 +1,5 @@
+import type { NextApiRequest, NextApiResponse } from "next";
+
+export default function handler(req: NextApiRequest, res: NextApiResponse) {
```
ygrishajev, line 3: this is a formal endpoint to be used in k8s. we might add something more meaningful here if there is anything (like db check)

apps/deploy-web/src/context/CustomChainProvider/CustomChainProvider.tsx
stalniy: it's not a file, it's from `exports` in package.json

stalniy: Screenshot 2025-03-14 at 17 45 43

apps/deploy-web/src/pages/api/healthz.ts
ygrishajev: ok, I see now that the path is broken there. anyway, i couldn't spot how it affects ui so just dropped it

ygrishajev: https://github.com/hyperweb-io/interchain-ui/issues/214

```
"./globalStyles": {
      "import": "./dist/styles/global.css.css",
      "require": "./dist/styles/global.css.css"
}
```

image

baktun14 (dismissed):

stalniy (approved):

PR #1050, opened 2025-03-17
refactor: migrates console-web components to strict types

stalniy:
## Why

Improve app stability

## What 

migrates console-web components to strict types


apps/deploy-web/src/components/sdl/AdvancedConfig.tsx
```diff
@@ -74,13 +74,13 @@ export const AdvancedConfig: React.FunctionComponent<Props> = ({ control, curren
...
-                <ExposeList currentService={currentService} setIsEditingExpose={setIsEditingExpose} />
+                <ExposeList currentService={currentService} setIsEditingExpose={() => setIsEditingExpose(true)} />
               </div>
               <div className="mb-4">
-                <EnvVarList currentService={currentService} setIsEditingEnv={setIsEditingEnv} />
+                <EnvVarList currentService={currentService} setIsEditingEnv={() => setIsEditingEnv(true)} />
               </div>
               <div className="mb-4">
-                <CommandList currentService={currentService} setIsEditingCommands={setIsEditingCommands} />
+                <CommandList currentService={currentService} setIsEditingCommands={() => setIsEditingCommands(true)} />
```
baktun14, line 83: question (blocking): shouldn't we pass the value from the function? It can be the service index or boolean

stalniy: not in case `serviceIndex` is undefined. 

This behavior with `serviceIndex` property which is then passed back to another property causes type incompatibilities which I don't know how to properly fix without type casting. I tried different ways. Eventually I came to conclusion that this components actually must not receive `serviceIndex` at all and it should be passed on calling side. But this requires more refactoring which I don't want to do in scope of this PR. These are the minimal changes I can do in order to have proper typesafety and to keep old behavior

baktun14: Ahh ok I see these don't receive the `serviceIndex`, got it.

baktun14 (approved):

PR #1051, opened 2025-03-17
feat(analytics): move lease created script

baktun14:


apps/deploy-web/src/utils/domUtils.ts
```diff
@@ -3,3 +3,17 @@ export function hasSomeParentTheClass(element: HTMLElement, classname: string):
...
   return !!element.parentNode && hasSomeParentTheClass(element.parentNode as HTMLElement, classname);
 }
+
+/**
+ * Adds a script tag to the document head
+ * @param src The source URL of the script
+ * @param async Whether the script should be loaded asynchronously
+ * @returns The created script element
+ */
+export function addScriptToHead(src: string, async: boolean = true): HTMLScriptElement {
```
stalniy, line 13: suggestion(non-blocking): better to pass options object because we could reuse this in different contexts. For example there may be cors restrictions or might need some data attribute on script, etc

stalniy (approved):

PR #1055, opened 2025-03-18
feat(deployment): managed api deposit

baktun14:
#989

### DIFF
--- a/apps/api/src/deployment/controllers/deployment/deployment.controller.ts
+++ b/apps/api/src/deployment/controllers/deployment/deployment.controller.ts
@@ -6,6 +6,8 @@ import {
   CloseDeploymentResponse,
   CreateDeploymentRequest,
   CreateDeploymentResponse,
+  DepositDeploymentRequest,
+  DepositDeploymentResponse,
   GetDeploymentResponse
 } from "@src/deployment/http-schemas/deployment.schema";
 import { DeploymentService } from "@src/deployment/services/deployment/deployment.service";
@@ -48,6 +50,16 @@ export class DeploymentController {
 
     const wallets = await this.userWalletRepository.accessibleBy(ability, "sign").findByUserId(currentUser.id);
     const result = await this.deploymentService.close(wallets[0], dseq);
+
+    return { data: result };
+  }
+
+  @Protected([{ action: "sign", subject: "UserWallet" }])
+  async deposit(dseq: string, input: DepositDeploymentRequest["data"]): Promise<DepositDeploymentResponse> {
+    const { currentUser, ability } = this.authService;
+
+    const wallets = await this.userWalletRepository.accessibleBy(ability, "sign").findByUserId(currentUser.id);
+    const result = await this.deploymentService.deposit(wallets[0], dseq, input.deposit);
 
     return { data: result };
   }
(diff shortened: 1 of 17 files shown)


apps/api/src/deployment/controllers/deployment/deployment.controller.ts
```diff
@@ -51,4 +53,14 @@ export class DeploymentController {
 
     return { data: result };
   }
+
+  @Protected([{ action: "sign", subject: "UserWallet" }])
+  async deposit(dseq: string, input: DepositDeploymentRequest["data"]): Promise<DepositDeploymentResponse> {
+    const { currentUser, ability } = this.authService;
+
+    const wallets = await this.userWalletRepository.accessibleBy(ability, "sign").findByUserId(currentUser.id);
+    const result = await this.deploymentService.deposit(wallets[0], dseq, input.deposit);
```
stalniy: **question(non-blocking):** is it possible for `wallets` to be an empty array?

apps/api/test/functional/deployments.spec.ts
```diff
@@ -460,4 +463,114 @@ describe("Deployments API", () => {
...
+
+      expect(response.status).toBe(404);
+      const result = await response.json();
+      expect(result).toEqual({
+        error: "NotFoundError",
+        message: "Deployment not found"
+      });
+    });
+
+    it("should return 500 if transaction fails", async () => {
```
stalniy: thought(non-blocking): usually on 500 error service must be restarted because 500 error is an unexpected crash in program behavior and we cannot trust program state anymore. 

So, either failure in signing transaction is expected error which should be handled with 503 for example

stalniy (dismissed):

apps/api/src/deployment/routes/deployments/deployments.router.ts
```diff
@@ -78,6 +81,33 @@ const deleteRoute = createRoute({
   }
 });
 
+const depositRoute = createRoute({
+  method: "post",
+  path: "/v1/deployments/{dseq}/deposit",
```
ygrishajev: this endpoint is kinda relates to an entity in terms of REST. However this is rather a business flow endpoint. There is no need to complicate the path and better name the endpoint as a command. we discussed it [here](https://github.com/akash-network/console/discussions/888). E.g.

```suggestion
  path: "/v1/deposit-deployment",
```

and pass a dseq as a part of a body

ygrishajev (dismissed):

apps/api/test/functional/deployments.spec.ts
stalniy: What I want to say is that definition of 500 error is "unexpected error", if you test that you get 500 this means it's not a 500 error. Because the test implies that this thing is expected, it cannot be 500 error

apps/api/src/deployment/routes/deployments/deployments.router.ts
baktun14: Makes sense

apps/api/src/deployment/controllers/deployment/deployment.controller.ts
baktun14: Good point, fixed.

apps/api/test/functional/deployments.spec.ts
baktun14: Well transactions can fail for many unknown reasons, so we should return 503 in the case of unknown tx fail?

ygrishajev: imo doesn't make sense to test :) it would make testing if it was some certain handled error

baktun14: Ok yea makes sense

Issue #973, opened 2025-03-10
Fix and setup console web e2e tests


ygrishajev:
E2E tests are failing again as we got them outdated. the reason we don't have them running in CI.

Acceptance criteria:
1) fix the tests to meed latest codebase
2) add them to npm scripts
3) run them in CI after container is deployed to beta.

stalniy:
Tests have been added https://github.com/akash-network/console/actions/runs/13974102418/job/39123833481

PR #1057, opened 2025-03-18
chore: enable e2e tests for console-web on PR level

stalniy:
## Why

to improve stability of the app ref #973 

## What

1. enable e2e tests for console-web on PR level
2. Extracts nodejs and npm dependencies setup into a separate action, for ease of reuse in other jobs
3. run tests against production console web image which uses production APIs (console-api, provider-proxy, etc)

## Additional info

I think we need to run it on every PR instead of on deployment because we will get feedback faster and will not need to make additional PR with a fix, after deployment. Ideally, we need to test console-web during CI on mock servers (or mock all requests on playwright level). BUT this requires more work.

After updating description, I think that running against production services may not always work because of API <-> UI protocol change. So, I need to rework it and run only after deployment


.github/actions/setup-app-deps/action.yml
```diff
@@ -0,0 +1,30 @@
+name: 'Setup app dependencies'
```
ygrishajev, line 1: 👍

.github/workflows/console-web-docker-build.yml
```diff
@@ -73,3 +62,69 @@ jobs:
...
+        with:
+          path: console-deploy-web.tar.gz
+          key: ${{ needs.build-deploy-web.outputs.deploy-web-image }}
+      - name: Load Docker image from cache
+        run: docker load -i console-deploy-web.tar.gz
+      - name: Run e2e tests
+        env:
+          TEST_WALLET_MNEMONIC: ${{ secrets.CONSOLE_WEB_E2E_TEST_WALLET_MNEMONIC }}
+        run: |
+          docker run -d \
```
ygrishajev: non-blockin. having an env file would make it more reliable. we'd use same values for running tests locally and in ci.

.github/workflows/console-web-docker-build.yml
```diff
@@ -73,3 +62,69 @@ jobs:
...
+      - name: Restore Docker image from cache
+        uses: actions/cache@v4
+        with:
+          path: console-deploy-web.tar.gz
+          key: ${{ needs.build-deploy-web.outputs.deploy-web-image }}
+      - name: Load Docker image from cache
+        run: docker load -i console-deploy-web.tar.gz
+      - name: Run e2e tests
+        env:
+          TEST_WALLET_MNEMONIC: ${{ secrets.CONSOLE_WEB_E2E_TEST_WALLET_MNEMONIC }}
```
ygrishajev: need to make sure this wallet is unique for gh

ygrishajev (dismissed):

baktun14 (dismissed):

PR #1071, opened 2025-03-19
fix(deployment): ignore ssl verification for api to proxy

baktun14:


apps/api/src/deployment/services/provider/provider-proxy.service.ts
```diff
@@ -42,7 +43,12 @@ export class ProviderProxyService {
         providerAddress: providerIdentity.owner,
         network: chainNetwork
       },
-      { timeout }
+      {
+        timeout,
+        httpsAgent: new https.Agent({
+          rejectUnauthorized: false
+        })
```
stalniy, line 50: **suggestion(blocking)**: let's add tech-debt task for this. If this is k8 provided certificate, maybe we could pass `ca` option with some root certificate, so then it will do validation even for self-signed certs

stalniy (approved):

apps/api/src/deployment/services/provider/provider-proxy.service.ts
baktun14: https://github.com/akash-network/console/issues/1075

PR #1094, opened 2025-03-22
fix: ignore zero provider stats data for last day

### LINKED ISSUE #721: Stats dashboard network capacity new day bug
When a new day starts (0:00 UTC), there's a short period of time (~15m) where the stats show -100% for all the network capacity. It auto-fixes itself after that period when the provider snapshots are taken. 

The reason for this is we do a query on day and join on the provider snapshots and when the new day starts, there's no provider snapshots that is flagged as `isLastSuccessOfDay` yet.  [Link to the query in question](https://github.com/akash-network/console/blob/917a7097b8abc880a325594f254e01ecb01429de/apps/api/src/services/db/statsService.ts#L210-L229)

This causes confusion for anyone looking at the dashboard in that period of time.

- [ ] Return the previous day `isLastSuccessOfDay` provider snapshot when it's a new day and there's no provider snapshot yet.

jzsfkzm:
fixes #721

### DIFF
--- a/apps/api/src/services/db/statsService.ts
+++ b/apps/api/src/services/db/statsService.ts
@@ -1,6 +1,6 @@
 import { AkashBlock as Block } from "@akashnetwork/database/dbSchemas/akash";
 import { Day } from "@akashnetwork/database/dbSchemas/base";
-import { subHours } from "date-fns";
+import { format, subHours } from "date-fns";
 import { Op, QueryTypes } from "sequelize";
 
 import { cacheKeys, cacheResponse } from "@src/caching/helpers";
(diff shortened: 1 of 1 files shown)


apps/api/src/services/db/statsService.ts
```diff
@@ -202,40 +202,42 @@ export const getProviderGraphData = async (dataName: ProviderStatsKey) => {
...
-         INNER JOIN (
-            SELECT DISTINCT ON("hostUri",DATE("checkDate")) 
-              DATE("checkDate") AS date, 
-              ps."activeCPU", ps."pendingCPU", ps."availableCPU", 
-              ps."activeGPU", ps."pendingGPU", ps."availableGPU", 
-              ps."activeMemory", ps."pendingMemory", ps."availableMemory", 
-              ps."activeEphemeralStorage" + COALESCE(ps."activePersistentStorage", 0) AS "activeStorage", ps."pendingEphemeralStorage" + COALESCE(ps."pendingPersistentStorage", 0) AS "pendingStorage", ps."availableEphemeralStorage" + COALESCE(ps."availablePersistentStorage", 0) AS "availableStorage", 
+  const result = removeLastIfZeroCount(
+    await cacheResponse(
+      60 * 5, // 5 minutes
```
ygrishajev: non-blocking: a constant from `date-fns` could be used - no commend would be needed

ygrishajev (dismissed):
I'm not sure how to test this, but the change seems to be doing what it declares :)

baktun14 (changes requested):
https://github.com/akash-network/console/issues/721#issuecomment-2755087688

baktun14 (approved):

Issue #1104, opened 2025-03-25
Throttle last_active_at on console-api


stalniy:
## Why

Currently this field is updated on every request to API and this causes db write overhead. We want to optimize it and put load off the db

## What

Throttle last_active_at on console-api, so it's updated not more than once every 30 mins

PR #1108, opened 2025-03-25
refactor: updates last_used_at once in 30 minutes

stalniy:
## Why

ref #1104 

## What

updates last_used_at once in 30 minutes

### DIFF
--- a/apps/api/src/auth/services/auth.interceptor.ts
+++ b/apps/api/src/auth/services/auth.interceptor.ts
@@ -81,6 +81,10 @@ export class AuthInterceptor implements HonoInterceptor {
     };
   }
 
+  clearLastUserActivityCache(): void {
+    this.lastUserActivityCache.clear();
+  }
+
   private async auth(user?: UserOutput) {
     this.authService.currentUser = user;
     if (user) {
--- a/apps/api/src/auth/services/auth.interceptor.ts
+++ b/apps/api/src/auth/services/auth.interceptor.ts
@@ -1,29 +1,36 @@
 import { LoggerService } from "@akashnetwork/logging";
+import { secondsInMinute } from "date-fns";
 import { Context, Next } from "hono";
 import { HTTPException } from "hono/http-exception";
+import { LRUCache } from "lru-cache";
 import { singleton } from "tsyringe";
 
 import { AbilityService } from "@src/auth/services/ability/ability.service";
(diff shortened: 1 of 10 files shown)


apps/api/src/auth/services/auth.interceptor.ts
```diff
@@ -1,29 +1,35 @@
...
-import { kvStore } from "@src/middlewares/userMiddleware";
 import { UserOutput, UserRepository } from "@src/user/repositories";
-import { toUTC } from "@src/utils";
-import { env } from "@src/utils/env";
-import { getJwks, useKVStore, verify } from "@src/verify-rsa-jwt-cloudflare-worker-main";
 import { ApiKeyRepository } from "../repositories/api-key/api-key.repository";
 import { ApiKeyAuthService } from "./api-key/api-key-auth.service";
+import { UserAuthTokenService } from "./user-auth-token/user-auth-token.service";
+
+const LAST_USER_ACTIVITY_THROTTLE_TIME_SECONDS = 30 * 60;
```
ygrishajev: non-blocking: `date-fns` has handy constants for durations

apps/api/src/auth/services/auth.interceptor.ts
```diff
@@ -56,14 +62,14 @@ export class AuthInterceptor implements HonoInterceptor {
...
-          await this.auth(currentUser);
-          c.set("user", currentUser);
 
-          if (apiKeyOutput && currentUser) {
-            await this.apiKeyRepository.updateById(apiKeyOutput.id, {
-              lastUsedAt: toUTC(new Date())
-            });
-          }
+          await Promise.all([
+            apiKeyOutput && currentUser && this.shouldMarkUserAsActive(apiKeyOutput?.userId, new Date())
```
ygrishajev: non-blocking: took me some minutes to understand why the check is needed here and below in `this.auth`, perhaps having a dedicated method to `markApiKeyAsActive` would help. Totally not a blocker - just a thought

ygrishajev (dismissed):

apps/api/src/auth/services/auth.interceptor.ts
stalniy: yes, I tried to refactor this but all attempts were not successful. So, I decided to keep it as is and suggest to add token prefixes or Authorization header prefixes to simplify all that.

ygrishajev (dismissed):

PR #1110, opened 2025-03-25
feat(provider): added upgrade button for provider

### LINKED ISSUE #1106: Users should able to upgrade provider from provider console
Current users don't have a way to upgrade their provider if a provider upgrade is available.
This should include the provider version and helm chart version upgrade as well.

jigar-arc10:
closes #1106


devalpatel67 (approved):
LGTM

PR #1113, opened 2025-03-25
feat: adds basic feature flags support to api and deploy-web

stalniy:
## Why

We need to toggle some functionality off and on when it doesn't work as expected or abused

## What

Implements basic feature flagging on top of env vars. API exposes features over endpoint to make it possible for deploy-web to adjust its UI based on enabled features without redeployment/rebuilding.

Also added `ALLOW_ANONYMOUS_USER_TRIAL` feature flag which is disabled in production but enabled in functional tests


baktun14 (dismissed):

ygrishajev (dismissed):
very nice!

ygrishajev (approved):

Issue #1065, opened 2025-03-19
SDL Editor Provokes Frequent Parsing Errors


chainzero:
Receiving frequent errors when attempting to use SDL editor screen in Console.

Example:

- Create new deployment
- Select Run Custom Container
- Enter YAML tab
- Select all text in editor > delete >  copy in SDL of interest
- Attempt to copy/paste SDL results in the following error

```
Error
Cannot read properties of undefined (reading 'resources')
```

Errors attempting to edit SDL are not unique to above, example flow.  Receive similar errors in attempting to use a template from Awesome Akash and when editing the SDL in that flow.

chainzero:
SDL used in testing cited example if useful:

```
---
version: "2.0"

services:
  obtaingpuone:
    image: ubuntu:22.04
    command:
      - "sh"
      - "-c"
    args:
      - 'uptime;
        nvidia-smi;
        sleep infinity'
    expose:
      - port: 8080
        as: 80
        to:
          - global: true

profiles:
  compute:
    obtaingpu:
      resources:
        cpu:
          units: 0.1
        memory:
          size: 256Mi
        gpu:
          units: 1
          attributes:
            vendor:
              nvidia:
        storage:
          size: 256Mi
  placement:
    akash:
      pricing:
        obtaingpu: 
          denom: uakt
          amount: 10000000

deployment:
  obtaingpuone:
    akash:
      profile: obtaingpu
      count: 1
```

stalniy:
This one also a duplicate https://github.com/akash-network/console/issues/985

PR #1127, opened 2025-03-27
fix(deployment): fix import sdl in render

baktun14:
#1065

### DIFF
--- a/apps/deploy-web/src/hooks/useImportSimpleSdl.ts
+++ b/apps/deploy-web/src/hooks/useImportSimpleSdl.ts
@@ -1,0 +1,16 @@
+import { useMemo } from "react";
+
+import { importSimpleSdl } from "@src/utils/sdl/sdlImport";
+
+export function useImportSimpleSdl(sdl: string | null | undefined) {
+  return useMemo(() => {
+    if (!sdl) return [];
+
+    try {
+      return importSimpleSdl(sdl);
+    } catch (error) {
+      console.error("Failed to parse SDL:", error);
+      return [];
+    }
+  }, [sdl]);
+}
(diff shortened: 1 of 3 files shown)


apps/deploy-web/src/hooks/useImportSimpleSdl.ts
```diff
@@ -0,0 +1,16 @@
...
+import { importSimpleSdl } from "@src/utils/sdl/sdlImport";
+
+export function useImportSimpleSdl(sdl: string | null | undefined) {
+  return useMemo(() => {
+    if (!sdl) return [];
+
+    try {
+      return importSimpleSdl(sdl);
+    } catch (error) {
+      console.error("Failed to parse SDL:", error);
```
stalniy, line 12: **suggestion(non-blocking):** I think that if we should also show at least general error message to the user because currently we show JS error which doesn't seem professional/reliable

Screenshot 2025-03-28 at 06 07 18

The current behavior is obviously much better than runtime exception, that's why it's non-blocking comment

stalniy (approved):

apps/deploy-web/src/hooks/useImportSimpleSdl.ts
baktun14: Yea I agree it makes sense, however we already do [validation of the sdl](https://github.com/akash-network/console/blob/main/apps/deploy-web/src/components/new-deployment/ManifestEdit.tsx#L119) in a useEffect hook every time the sdl changes and show the error to the user.

This `importSimpleSdl` function was moved from the create function to the render level when we added that the services are displayed in the deposit modal and doesn't really serve as a validation function.

baktun14: We could just ignore the error here, I'll update this in another branch since this is breaking console right now

PR #1129, opened 2025-03-28
refactor: moves trial authorization spending out of db transaction

stalniy:
## Why

Because calling external service in transaction prolongs resource locking time and causes db to become slow

## What

1. moves trial authorization spending out of db transaction
2. moves docker compose options in front of service names to make it properly work with `npx dc up:dev -- deploy-web`

### DIFF
--- a/apps/api/src/billing/services/wallet-initializer/wallet-initializer.service.ts
+++ b/apps/api/src/billing/services/wallet-initializer/wallet-initializer.service.ts
@@ -14,10 +14,11 @@ export class WalletInitializerService {
 
   async initializeAndGrantTrialLimits(userId: UserWalletInput["userId"]): Promise<UserWalletPublicOutput> {
     let userWallet = await this.userWalletRepository.findOneByUserId(userId);
+    if (userWallet) return this.userWalletRepository.toPublic(userWallet);
 
-    if (!userWallet) {
-      userWallet = await this.userWalletRepository.accessibleBy(this.authService.ability, "create").create({ userId });
+    userWallet = await this.userWalletRepository.accessibleBy(this.authService.ability, "create").create({ userId });
 
+    try {
       const wallet = await this.walletManager.createAndAuthorizeTrialSpending({ addressIndex: userWallet.id });
       userWallet = await this.userWalletRepository.updateById(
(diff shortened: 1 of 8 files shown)


ygrishajev (dismissed):

apps/api/src/billing/services/wallet-initializer/wallet-initializer.service.ts
```diff
@@ -28,6 +29,9 @@ export class WalletInitializerService {
         },
         { returning: true }
       );
+    } catch (error) {
+      await this.userWalletRepository.deleteById(userWallet.id);
```
baktun14, line 33: thought (blocking): I think we should revoke the authz as well here. And should we move the userWallet creation on line 19 in the try catch? Or I guess if it fails it just throws and returns 500 and user can try again

stalniy: yes, if wallet creation fails, it's just 500 error and everything else won't be executed. So, wallet is not created and we don't need to delete it

stalniy: regarding revoking, we don't know the reason why it failed, very likely because trial spending was not granted. I haven't dig deep into implementation details of `createAndAuthorizeTrialSpending`. It can potentially fail on `updateById` which is less likely then yes, we need to revoke trial spending

baktun14: Yea I agree the `updateById` has a very low chance of failing

baktun14 (dismissed):

apps/api/src/billing/services/wallet-initializer/wallet-initializer.service.ts
ygrishajev: On the other hand, if we drop the wallet the pk is already incremented. It will never be used again. The grants will expire in a month on their own.

baktun14: Good point

ygrishajev (approved):

PR #1136, opened 2025-03-30
fix: update gas pricing

ygrishajev:


ghost (changes requested):
Not sure that's the way to go. I think it should match what we have in the Chain Registry, e.g.
https://github.com/cosmos/chain-registry/blob/master/akash/chain.json#L14-L23

Probably `fixed_min_gas_price` should be just set to `0.025` (aka the "high gas")

PR #1152, opened 2025-04-01
chore: adds eslint rule that separates type imports for value imports

stalniy:
## Why

This helps to reduce import/no-cycle errors because it's allowed to have cyclic deps between types

## What

adds eslint rule that separates type imports for value imports


ygrishajev (dismissed):

stalniy:
It appears that `@typescript-eslint/consistent-type-imports` doesn't work with TS experimental decorators ([more info](https://typescript-eslint.io/blog/changes-to-consistent-type-imports-with-decorators/)). Moreover, if improperly configured it can cause runtime errors like this:

```
Cannot inject the dependency "executionContextService" at position #0 of "RequestContextInterceptor" constructor. Reason:
        TypeInfo not known for "Function"
```

There is a workaround but it's basically makes esilnt ignore all files that contain decorators. I still plan to use it because it can be applied in other apps successfully

ygrishajev (dismissed):

Issue #1147, opened 2025-03-31
Trial providers endpoint


baktun14:
- Add the provider capacity per provider and total to the existing trials provider endpoint (cpu, ram, disk, gpu)
  - https://console-api.akash.network/v1/trial-providers
  - Add provider host names
  - Add a new query param to this endpoint to filter for `console/trials-registered` only
    - https://console-api.akash.network/v1/trial-providers?registered=true

PR #1158, opened 2025-04-02
feat: load trial providers with resource info, can filter for trial-registered

jzsfkzm:
refs #1147


baktun14:
@jzsfkzm I have the fix for the tests in this pr https://github.com/akash-network/console/pull/1183

apps/api/test/functional/trial-providers.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+    availableEphemeralStorage
+  });
+};
+
+describe("Trial Providers", () => {
+  const provider1 = "akash1rk090a6mq9gvm0h6ljf8kz8mrxglwwxsk4srxh";
+  const provider2 = "akash1d4fletej4cwn9x8jzpzmnk6zkqeh90ejjskpmu";
+
+  beforeAll(async () => {
+    await expect.getState().dbService.copyIndexerTables(["providerAttributeSignature", "provider"]);
```
baktun14: question(non-blocking): can you explain how this works? We copy these tables for the test in the database of the test run?

baktun14 (dismissed):
Nice!

apps/api/test/functional/trial-providers.spec.ts
jzsfkzm: Yes, that method will copy all sequences from the original database, and the tables listed, so for each test file we'll have separate indexer database too. Copying sequences is cheap, but didn't want to copy all the tables, because some tables are huge, it's easier to feed some synthetic test data instead. Maybe if we did `pg_dump` and loaded that, that could be faster, but I didn't try it.
Let me know what you think.

baktun14: I think it's fine to copy on the tables that we need for tests for now, thanks for the explanation.

baktun14:
The tests are failing in the actions

baktun14 (dismissed):

jzsfkzm:
@baktun14 let's not merge this yet, it will fail on every build otherwise. Will talk to @ygrishajev so we can debug why it is failing here.

apps/api/test/functional/trial-providers.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+
+    await createProviderSnapshot(provider1, 1, 1, 2, 3, 4, 5);
+    await createProviderSnapshot(provider1, 2, 11, 12, 13, 14, 15);
+    await createProviderSnapshot(provider2, 1, 21, 22, 23, 24, 25);
+    await createProviderSnapshot(provider2, 2, 31, 32, 33, 34, 35);
```
stalniy: **suggestion(blocking):** please pass numbers as an object, otherwise there is no understanding what they mean without looking into source code of `createProviderSnapshot`. Also I think you can run them with `Promise.all` to speed up testing

apps/api/test/services/test-database.service.ts
```diff
@@ -77,4 +77,113 @@ export class TestDatabaseService {
       await sql.end();
     }
   }
+
```
ygrishajev: all this would not work in CI env where there is no indexer db imported. should be solved differently. I got something locally gonna push a PR shortly.

apps/api/test/functional/trial-providers.spec.ts
ygrishajev: as I mentioned in a different comment, this wouldn't work as it relies on a db existence which is not imported in CI.
```suggestion
```

You can use this once merged to pre-create providers so that seeders below work https://github.com/akash-network/console/pull/1257/files#diff-0107cc7c80b1a3372b2d16da819414c1856899b7866b8ddc5a663489cacc9932R5

ygrishajev (changes requested):

apps/api/test/functional/trial-providers.spec.ts
jzsfkzm: Oh my god this is so great, thanks. Added a similar seeder, for ProviderSnapshots in this case, and it just works.

apps/api/test/functional/trial-providers.spec.ts
jzsfkzm: You're right! It was bad coding. Dropped the whole thing and added a ProviderSnapshot seeder instead.

jzsfkzm:
@baktun14 @stalniy @ygrishajev thanks for all the help, now tests are green on this PR, care for another review?

apps/api/test/services/test-database.service.ts
```diff
@@ -88,4 +88,113 @@ export class TestDatabaseService {
...
+  public async copyIndexerTables(tableNames: string[]): Promise<void> {
```
ygrishajev: I think all of the changes in this files aren't needed anymore, right?

apps/api/test/setup-functional-tests.ts
```diff
@@ -7,6 +7,7 @@ const dbService = new TestDatabaseService(testPath);
 
 beforeAll(async () => {
   await dbService.setup();
+  expect.setState({ dbService });
```
ygrishajev: and this one not needed too, is it?

apps/api/test/services/test-database.service.ts
jzsfkzm: That's true, dropped these.

apps/api/test/setup-functional-tests.ts
jzsfkzm: Dropped this as well.

apps/api/src/deployment/repositories/provider/provider.repository.ts
```diff
@@ -1,22 +1,65 @@
...
-  async getTrialProviders(): Promise<string[]> {
-    const trialProviders = await ProviderAttributeSignature.findAll({
-      attributes: ["provider"],
-      where: {
-        auditor: AUDITOR,
-        key: TRIAL_ATTRIBUTE,
-        value: "true"
-      },
-      raw: true
+  async getTrialProviders(registered: boolean): Promise<TrialProviders> {
```
ygrishajev, line 9:
```suggestion
  async getTrialProviders({ registered }: { registered: boolean }): Promise<TrialProviders> {
```

apps/api/src/routes/v1/trialProviders.ts
```diff
@@ -7,20 +7,46 @@ const route = createRoute({
   method: "get",
   path: "/trial-providers",
   tags: ["Trial", "Providers"],
+  request: {
+    query: z.object({
+      registered: z.string().optional()
```
ygrishajev, line 12:
```suggestion
      registered: z
    .enum(["true", "false"])
    .default("true")
    .transform(v => v === "true")
    .optional()
```

apps/api/src/deployment/repositories/provider/provider.repository.ts
ygrishajev: to make sure a bool means smth on the client code

baktun14 (changes requested):
@jzsfkzm Change of plans for this endpoint.

Instead of creating a new endpoint, we're going to update the [current one](https://github.com/akash-network/console/blob/main/apps/api/src/routes/v1/providers/list.ts) and support filterting by `scope`.

For example: `/v1/providers?scope=all|trial`

I also created a bunch of tasks to refactor the old routes into the module architecture, so at the same time we could move/refactor the old provider list route.

The issue in question: https://github.com/akash-network/console/issues/1272, but for the sake of keeping the scope small, just do the [provider list route here](https://github.com/akash-network/console/tree/main/apps/api/src/routes/v1/providers).

You can probably re-use most of the code you wrote for tests so no loss here. Sorry for the confusion and change of plans, let me know if you have any questions.

ygrishajev:
> For example: `/v1/providers/scope=all|trial`

Should be a single endpoint with query params `/v1/providers?scope=all|trial`. Not for this task, in general.

jzsfkzm:
Closing this in favor [of the other PR](https://github.com/akash-network/console/pull/1288).

Issue #337, opened 2024-08-26
Upgrade react-query to v5


baktun14:
We're using an older version (3.39.x) and now the library is at v5 https://tanstack.com/query/latest
- Upgrade the package
- Refactor everywhere we use react-query
- Test for breaking changes

ShinySyntax:
May i get this issue?

forbesus:
TanStack/react-query is upgraded 5.52.x now

do you want this version? @baktun14

baktun14:
Hey @ShinySyntax and @forbesus thanks for your interesting in contributing. Since @ShinySyntax you asked first, I will assign this task to you. Are you in the Akash discord?

ShinySyntax:
Sure, Thanks

ShinySyntax:
Hello, @baktun14 
I have done this issue.
Plz check https://github.com/akash-network/console/pull/350

ShinySyntax:
Hello, @baktun14 
I sent PR
Plz check https://github.com/akash-network/console/pull/356

anilmurty:
@baktun14 - I see that the two PRs are closed. Can this issue be closed?

baktun14:
@anilmurty They were closed but not completed.

jzsfkzm:
@baktun14 should I give a try at this? sizeable change, but should be pretty straightforward

ygrishajev:
@jzsfkzm go for it

PR #1159, opened 2025-04-02
refactor: update react-query in a few places

jzsfkzm:
refs #337


apps/deploy-web/src/pages/_app.tsx
```diff
@@ -3,12 +3,12 @@ import "nprogress/nprogress.css";
 import "../styles/index.css";
 
 import React from "react";
-import { dehydrate, Hydrate, QueryClient as LegacyQueryClient, QueryClientProvider as LegacyQueryClientProvider } from "react-query";
+import { Hydrate, QueryClientProvider as LegacyQueryClientProvider } from "react-query";
```
baktun14: question (non-blocking): We'er still using the old one?

baktun14 (dismissed):

apps/deploy-web/src/pages/_app.tsx
jzsfkzm: Oh, good question actually. Yes, we still use the old one at a few places, didn't want to change everything at once because when I did that before, it yielded a huge changeset hard to review.
But it is good question because it led me to the fact we should use the boundary from the new version here instead of the one from the old version. Fixed.

stalniy:
@jzsfkzm did you try to use [codemod](https://tanstack.com/query/latest/docs/framework/react/guides/migrating-to-v5#codemod)?

stalniy (approved):

jzsfkzm:
> @jzsfkzm did you try to use [codemod](https://tanstack.com/query/latest/docs/framework/react/guides/migrating-to-v5#codemod)?

I did not back then, tried it now just for the sake of trying, but it wouldn't really do the job. Eg. places where we pass `onSuccess` to useQuery, callbacks we need to turn into `useEffect` calls under the new version.

baktun14 (approved):

PR #1161, opened 2025-04-04
feat(deployment): managed wallet api lease status

baktun14:
https://github.com/akash-network/console/issues/1025


apps/api/src/deployment/services/deployment/deployment.service.ts
```diff
@@ -120,7 +157,7 @@ export class DeploymentService {
     await this.ensureDeploymentIsUpToDate(wallet, dseq, manifestVersion, deployment);
     await this.sendManifestToProviders(dseq, manifest, certificate as { certPem: string; keyPem: string }, deployment.leases);
 
-    return await this.findByOwnerAndDseq(wallet.address, dseq);
+    return await this.findByOwnerAndDseq(wallet.address, dseq, { certificate: { certPem: certificate.certPem, keyPem: certificate.keyPem } });
```
stalniy, line 160: **nitpick:** seems you can just pass `certificate` object to the options

apps/api/src/core/decorators/benchmark.decorator.ts
```diff
@@ -0,0 +1,172 @@
...
+        heapTotal: process.memoryUsage().heapTotal
+      }
+    };
+
+    // Log enhanced metrics
+    context.logger.info({
+      message: `Method ${context.methodName} completed`,
+      duration: `${totalDuration.toFixed(2)}ms`,
+      metrics
+    });
```
stalniy: thought(non-blocking): would be nice to use prometheus to log this metrics, then we could make different kind of graphs out of it in grafana

apps/api/src/core/decorators/benchmark.decorator.ts
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  static validateContext(context: BenchmarkContext | undefined, methodName: string): asserts context is BenchmarkContext {
+    if (!context) {
+      throw new BenchmarkError(`No benchmark context found for method ${methodName}`);
+    }
+  }
+}
+
+export function Benchmark() {
```
stalniy: **thought(non-blocking):** there is a big issue in JS related to decorators:
1. Typescript supports 2 types of decorators: legacy decorators and latest EcmaScript decorators
2. Legacy decorators sooner or later will become a legacy burden, so I'd like to discourage us from adding more decorators
3. I asked a [question in typescript-go](https://github.com/microsoft/typescript-go/discussions/741) whether legacy decorators will be supported. TypeScript GoLang version is a planned TypeScript 7 version, we are currently at TypeScript 5.8. 

There are DI frameworks that support new decorators version in TypeScript but there are not much of them. Major players resist to migrate to new spec (e..g, Nest.js), making arguments that they should be compliant with all the features in the legacy decorators implemented by TypeScript.

There are few hints which let's me make a conclusion that decorators should be avoided:
1. Mess in community related to this
2. EcmaScript spec for new decorators is a stage-3 which is very close to the final. Thought it was in progress for 7 years
3. Angular, one of the major frameworks bloated with decorators **moves away!** from using decorators in their components. For example, they get rid of decorators for DI injection
4. some of eslint rules do not work for files which uses decorators. So, those rules are not applied to some files -> https://typescript-eslint.io/blog/changes-to-consistent-type-imports-with-decorators. The main argument is that this decorators are legacy and will be removed at some point, so they don't want to support them, making their life easier

stalniy (dismissed):

apps/api/src/core/decorators/benchmark.decorator.ts
ygrishajev: but we can use it as a wrapper function, right?

ygrishajev: @baktun14 so, now when we have tracing implemented this approach is generally loosing sense. We already have tools installed in the app and can easily utilise them. It all drills down to:

```ts
import { trace } from '@opentelemetry/api';
  const span = trace.getTracer('default').startSpan('yourSpanName');
  // ..... logic to trace down
  span.end();
```

image

You'll then get data like so. If you add your custom span, it will be accessible in the hierarchy

baktun14: Ok I'll remove it since we have otel, thanks for the comments

apps/api/src/deployment/services/deployment/deployment.service.ts
baktun14: one is optional so they don't match

PR #1181, opened 2025-04-09
feat(observability): adds tracing utilities and spans tx signing method

ygrishajev:
This implementation results in
Screenshot 2025-04-09 at 14 39 20

### DIFF
--- a/apps/api/src/core/services/tracing/tracing.service.ts
+++ b/apps/api/src/core/services/tracing/tracing.service.ts
@@ -1,0 +1,182 @@
+import { SpanStatusCode, trace } from "@opentelemetry/api";
+
+/**
+ * Decorator for tracing method execution
+ * Automatically creates a span for the decorated method, tracks execution time,
+ * and properly handles errors by setting the span status
+ *
+ * @example
+ * ```typescript
+ * class PaymentService {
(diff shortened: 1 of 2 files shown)


apps/api/src/core/services/tracing/tracing.service.ts
```diff
@@ -0,0 +1,182 @@
...
+ *   });
+ *
+ *   return result;
+ * }
+ *
+ * // With span attributes
+ * async function processOrder(orderId: string) {
+ *   return withSpan('order-processing', async () => {
+ *     // Get current span via context
+ *     const span = trace.getSpan(context.active());
```
baktun14, line 148: Where does the context comes from? Can we pass the current span in the function instead?

```
return withSpan('order-processing', async span => {
   span.setAttribute(...);
}
```

baktun14 (approved):

apps/api/src/core/services/tracing/tracing.service.ts
ygrishajev: Context is exported from otel api package. Any improvements are welcome for sure. This pr is functional though and already covers the use case with benchmark implementation from ur pr.

PR #1183, opened 2025-04-09
feat(billing): managed wallet api balances endpoint

baktun14:
https://github.com/akash-network/console/issues/990

### DIFF
--- a/apps/api/src/billing/routes/get-balances/get-balances.router.ts
+++ b/apps/api/src/billing/routes/get-balances/get-balances.router.ts
@@ -24,9 +24,8 @@ const route = createRoute({
     }
   }
 });
-export const getBalancesRouter = new OpenApiHonoHandler();
 
-getBalancesRouter.openapi(route, async function routeGetBalances(c) {
+export const getBalancesRouter = new OpenApiHonoHandler().openapi(route, async function routeGetBalances(c) {
   const query = c.req.valid("query");
   return c.json(await container.resolve(WalletController).getBalances(query.address), 200);
 });
--- a/apps/api/src/billing/routes/get-balances/get-balances.router.ts
+++ b/apps/api/src/billing/routes/get-balances/get-balances.router.ts
@@ -2,7 +2,7 @@ import { createRoute } from "@hono/zod-openapi";
 import { container } from "tsyringe";
 
 import { WalletController } from "@src/billing/controllers/wallet/wallet.controller";
-import { GetBalancesResponseOutputSchema } from "@src/billing/http-schemas/balance.schema";
+import { GetBalancesQuerySchema, GetBalancesResponseOutputSchema } from "@src/billing/http-schemas/balance.schema";
 import { OpenApiHonoHandler } from "@src/core/services/open-api-hono-handler/open-api-hono-handler";
 
 const route = createRoute({
@@ -10,6 +10,9 @@ const route = createRoute({
   path: "/v1/balances",
   summary: "Get user balances",
   tags: ["Wallet"],
+  request: {
+    query: GetBalancesQuerySchema
+  },
   responses: {
     200: {
       description: "Returns user balances",
@@ -24,5 +27,6 @@ const route = createRoute({
 export const getBalancesRouter = new OpenApiHonoHandler();
 
 getBalancesRouter.openapi(route, async function routeGetBalances(c) {
-  return c.json(await container.resolve(WalletController).getBalances(), 200);
+  const query = c.req.valid("query");
+  return c.json(await container.resolve(WalletController).getBalances(query.address), 200);
 });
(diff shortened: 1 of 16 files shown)


apps/api/src/billing/routes/get-balances/get-balances.router.ts
```diff
@@ -0,0 +1,32 @@
...
+      }
+    }
+  }
+});
+export const getBalancesRouter = new OpenApiHonoHandler();
+
+getBalancesRouter.openapi(route, async function routeGetBalances(c) {
+  const query = c.req.valid("query");
+  return c.json(await container.resolve(WalletController).getBalances(query.address), 200);
+});
```
stalniy: suggest(non-blocking): if you write like this :

```ts
export const getBalancesRouter = new OpenApiHonoHandler()
  .openapi(route, async function routeGetBalances(c) {
    const query = c.req.valid("query");
    return c.json(await container.resolve(WalletController).getBalances(query.address), 200);
  });
```

Then `getBalancesRouter` will have a better type inference which potentially we may use in future

apps/api/src/utils/constants.ts
```diff
@@ -15,7 +15,7 @@ export const openApiExampleTransactionHash = "A19F1950D97E576F0D7B591D71A8D0366A
 export const openApiExampleValidatorAddress = "akashvaloper14mt78hz73d9tdwpdvkd59ne9509kxw8yj7qy8f";
 
 export const defaultNodeUrlMapping: { [key: string]: string } = {
```
stalniy: **question(non-blocking):** does it make sense to centralize this? We have `net` package which is autogenerated from akashnetwork repository on every commit which contains this information and probably should be used here instead of manually hardcoding values

apps/api/test/functional/balances.spec.ts
```diff
@@ -0,0 +1,225 @@
...
+      headers: new Headers({
+        "Content-Type": "application/json",
+        "x-api-key": apiKey
+      })
+    });
+
+    expect(response.status).toBe(404);
+  });
+
+  async function setup(options?: SetupOptions) {
```
stalniy, line 106: **praise:** I like this pattern to be spreaded

packages/http-sdk/src/deployment/deployment-http.service.ts
```diff
@@ -108,17 +109,74 @@ type RestAkashDeploymentInfoResponse =
...
+  }
+
+  /**
+   * Load all deployments for an owner with pagination
+   * @param owner Owner address
+   * @param state Optional state filter
+   * @param limit Number of items per page
+   * @returns Array of all deployments
+   */
+  public async loadAllDeployments(owner: string, state?: string, limit: number = 1000): Promise<DeploymentInfo[]> {
```
stalniy, line 176: **suggestion(non-blocking):** returning AsyncIterator would make this method less heavy in terms of data fetching/parsing. And would allow to get of rid of assumption that one user can have max 1000 active deployments. This is reasonable obviously but such payload will take time, memory and probably event loop time for parsing (`JSON.parse` is sync function).

A rough example:

```ts
  public async *loadAllDeployments(owner: string, state?: string, limit: number = 1000): AsyncGenerator<DeploymentInfo[]> {
    let nextPageKey = '';

    do {
      const response: any = this.extractData(
        await this.listDeployments({
          owner,
          state,
          "pagination.limit": limit,
          "pagination.key": nextPageKey
        })
      );
      nextPageKey = response.pagination?.next_key;

      for (const deployment of response.deployments) {
        yield deployment;
      }
    } while (nextPageKey);
  }
```

Then you could use it like this:

```ts
for await (const deployment of this.deploymentHttpService.loadAllDeployments()) {
  console.log(deployment);
}
```

support for async generators is pretty much good https://caniuse.com/mdn-javascript_functions_method_definitions_async_generator_methods

This allows to fetch data in portions with a nice interface, without making assumptions and processing it in at least non-blocking way for other requests

packages/http-sdk/src/utils/pagination.utils.ts
```diff
@@ -0,0 +1,37 @@
...
+  return items.filter(item => item) as T[];
+}
+
+/**
+ * Helper function to check if a URL has query parameters
+ * @param url URL to check
+ * @returns Boolean indicating if the URL has query parameters
+ */
+export function hasQueryParam(url: string): boolean {
+  return /[?&]/gm.test(url);
```
stalniy: **suggestion(non-blocking):** `new URL(url).searchParams.size` can be use to reliably detect this

stalniy (dismissed):

apps/api/src/utils/constants.ts
baktun14: Yea you're right we should update the net repo

packages/http-sdk/src/deployment/deployment-http.service.ts
baktun14: I created an issue to address this: https://github.com/akash-network/console/issues/1200

PR #1193, opened 2025-04-10
feat(notifications): adds alerts module

ygrishajev:


apps/notifications/package.json
```diff
@@ -23,19 +25,26 @@
...
     "@nestjs/config": "^4.0.1",
     "@nestjs/core": "^11.0.1",
     "@nestjs/platform-express": "^11.0.1",
     "@novu/node": "^2.6.6",
+    "@ucast/core": "^1.10.2",
+    "@ucast/js": "^3.0.4",
     "axios": "^1.8.2",
     "cosmjs-types": "^0.9.0",
+    "drizzle-orm": "^0.41.0",
     "jest-mock-extended": "^4.0.0-beta1",
```
stalniy: issue(non-blocking): dev dependency

apps/notifications/src/alert/controllers/chain-events/chain-events.controller.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+        mockBlock,
+      );
+    });
+  });
+
+  async function setup(): Promise<{
+    controller: ChainEventsController;
+    rawAlertsService: MockProxy<RawAlertsService>;
+    deploymentBalanceAlertsService: MockProxy<DeploymentBalanceAlertsService>;
+  }> {
```
stalniy, line 45: **thought(non-blocking):** return type is not needed on private functions. I'd just return module from the function. then you can do  this in test:

```ts
const mySpy = jest.spyOn(module.get(RawAlertsService), 'method')

expect(mySpy).toHaveBeenCalled();
```

OR even this, if service is created with `mock()`:

```ts
expect(module.get(RawAlertsService).method).toHaveBeenCalled();
```

apps/notifications/src/alert/services/deployment-balance-alerts/deployment-balance-alerts.service.ts
```diff
@@ -0,0 +1,108 @@
...
+          await Promise.all(alerts.map(async (alert) => callback(alert)));
+        },
+      });
+    } catch (error) {
+      this.loggerService.error({
+        event: 'ALERT_FAILURE',
+        block,
+        error,
+      });
+      throw error;
```
stalniy, line 54: **thought(non-blocking):** in this case, ideally you should use a different Error or Error with context, like:

```ts
throw createErrorWithContext({
  cause: error,
  event: '...',
  ....
});
```

this is 5xx error I guess, then the only thing you need to do here is to enrich its context. It will be logged by global error handler anyway, right?

apps/notifications/src/alert/services/deployment-balance-alerts/deployment-balance-alerts.service.ts
```diff
@@ -0,0 +1,108 @@
...
+        alert.conditions,
+        balance,
+      );
+      const update: Partial<DeploymentBalanceAlertOutput> = {
+        minBlockHeight: block.height + this.DEPLOYMENT_BALANCE_BLOCKS_THROTTLE,
+      };
+      let message: string | undefined;
+
+      if (isMatching && alert.status === 'normal') {
+        const interpolate = template(alert.template);
```
stalniy: question(non-blocking): is this a user defined template OR it's defined by us? if user defined then it will need additional handling because lodash template will them access to global vars I guess.

apps/notifications/src/alert/services/deployment-balance-alerts/deployment-balance-alerts.service.ts
```diff
@@ -0,0 +1,108 @@
...
+    block: ChainBlockCreatedDto,
+    alert: DeploymentBalanceAlertOutput,
+  ) {
+    try {
+      const balance = await this.deploymentService.getDeploymentBalance(
+        alert.owner,
+        alert.dseq,
+      );
+
+      if (!balance) {
```
stalniy: thought(non-blocking): `balance` may be `0` and if so, then it will disable alert but probably should not

apps/notifications/src/alert/services/deployment/deployment.service.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+import { MockProvider } from '@test/mocks/provider.mock';
+import { mockAkashAddress } from '@test/seeders/akash-address.seeder';
+
+type Config = Pick<GlobalEnvConfig, 'RPC_NODE_ENDPOINT'>;
+
+describe(DeploymentService.name, () => {
+  describe('getDeploymentBalance', () => {
+    it('should return the deployment balance', async () => {
+      const { service } = await setup();
+      jest.spyOn(axios, 'get').mockResolvedValue({
```
stalniy: suggestion(non-blocking): would be nice to use `axios.create` or even pass it via DI. using global instance may cause some issues, especially with hot reloads

apps/notifications/src/alert/services/deployment/deployment.service.ts
```diff
@@ -0,0 +1,67 @@
...
+    private readonly loggerService: LoggerService,
+  ) {
+    this.loggerService.setContext(DeploymentService.name);
+  }
+
+  async getDeploymentBalance(
+    owner: string,
+    dseq: string,
+  ): Promise<{ balance: number } | null> {
+    const { data: result } = await axios.get<any>(
```
stalniy: **nitpick(non-blocking):** instead of `any`, you can use `Record<string, unknown>` here

apps/notifications/src/alert/services/deployment/deployment.service.ts
```diff
@@ -0,0 +1,67 @@
...
+      return null;
+    }
+
+    const balance = result.escrow_account.balance;
+    const funds = result.escrow_account.funds;
+
+    const balanceAmount =
+      balance.denom === 'uakt'
+        ? parseInt(balance.amount)
+        : parseInt(balance.amount) * 1_000_000;
```
stalniy: **issue(?):** usually it's nice to provide the 2nd argument to `parseInt` because it may cause unexpected results -> https://eslint.org/docs/latest/rules/radix

I don't know whether we need to think about loosing precision in this case but there for sure may be rounding issues.

apps/notifications/src/broker/decorators/handler.decorator.ts
```diff
@@ -0,0 +1,9 @@
+import { SetMetadata } from '@nestjs/common';
+
+import type { HandlerConfig } from '../services/pg-boss-handler/pg-boss-handler.service';
+
+export const PG_BOSS_HANDLER = Symbol('PG_BOSS_HANDLER');
+
+export const PgBossHandler = (
```
stalniy: **question(non-blocking):** Can it be just `AlertHandler` to not expose internal implementation details?

apps/notifications/test.js
```diff
@@ -0,0 +1,36 @@
+const { CompoundCondition, FieldCondition } = require("@ucast/core");
```
stalniy: question(blocking): is this committed accidentally?

apps/notifications/test/seeders/deployment-balance-alert.seeder.ts
```diff
@@ -0,0 +1,39 @@
+import { faker } from '@faker-js/faker';
+
+import type { DeploymentBalanceAlertOutput } from '@src/alert/repositories/deployment-balance-alert/deployment-balance-alert.repository';
+
+import { mockAkashAddress } from '@test/seeders/akash-address.seeder';
+
+export const generateDeploymentBalanceAlert = ({
```
stalniy, line 7: **question(non-blocking):** what do you think if we just call it with `build` or `gen` prefix, to make shorter?

stalniy (dismissed):

apps/notifications/src/alert/services/deployment/deployment.service.ts
```diff
@@ -0,0 +1,67 @@
...
+  ) {
+    this.loggerService.setContext(DeploymentService.name);
+  }
+
+  async getDeploymentBalance(
+    owner: string,
+    dseq: string,
+  ): Promise<{ balance: number } | null> {
+    const { data: result } = await axios.get<any>(
+      `${this.configService.getOrThrow('API_NODE_ENDPOINT')}/akash/deployment/v1beta3/deployments/info`,
```
baktun14: question (non-blocking): should we put this in the http sdk package?

apps/notifications/src/alert/services/deployment/deployment.service.ts
baktun14: I don't think it matters in the case of deployment balance alert

apps/notifications/src/alert/services/deployment-balance-alerts/deployment-balance-alerts.service.ts
```diff
@@ -0,0 +1,108 @@
...
+        // TODO: send a relevant notification. e.g. alert suspended for a "reason"
```
baktun14, line 69: question: should we create an issue for this?

apps/notifications/src/alert/services/deployment-balance-alerts/deployment-balance-alerts.service.ts
```diff
@@ -0,0 +1,108 @@
...
+    private readonly alertRepository: DeploymentBalanceAlertRepository,
+    private readonly conditionsMatcher: ConditionsMatcherService,
+    private readonly brokerService: BrokerService,
+    private readonly deploymentService: DeploymentService,
+    private readonly loggerService: LoggerService,
+  ) {
+    this.loggerService.setContext(DeploymentBalanceAlertsService.name);
+  }
+
+  private readonly DEPLOYMENT_BALANCE_BLOCKS_THROTTLE = 10;
```
baktun14, line 31: Ah ok interesting, so we're checking the balance only every 10 blocks or so (every minute). I think this is a fair delay to be notified for the balance alert, but yea if we wanted to do it in real time it would have to be a little different like we discussed.

apps/notifications/src/alert/controllers/chain-events/chain-events.controller.ts
```diff
@@ -0,0 +1,32 @@
...
+  async processBlock(block: ChainBlockCreatedDto) {
+    await this.deploymentBalanceAlertsService.alertFor(block);
+  }
+
+  @PgBossHandler({
+    key: MsgCloseDeployment['$type'],
+    dto: MsgCloseDeploymentDto,
+  })
+  async processDeploymentClosed(event: MsgCloseDeploymentDto) {
+    await this.rawAlertsService.alertFor(event);
```
baktun14, line 30: Ok we won't support all the message types generically like blockspy does, but with the rawAlertsService, we could implement something similar if needed. 👍

baktun14 (dismissed):

apps/notifications/src/alert/controllers/chain-events/chain-events.controller.spec.ts
ygrishajev: idk, how much does this matter really?:)

apps/notifications/src/alert/controllers/chain-events/chain-events.controller.ts
ygrishajev: this actually does support any message. the only difference is it's not subscribed to all of them by default. It's better add stuff when it's needed.

apps/notifications/src/alert/services/deployment-balance-alerts/deployment-balance-alerts.service.ts
ygrishajev: I'm not sure I follow, how different should it be? This is a var that can be changed to match our needs

apps/notifications/src/alert/services/deployment-balance-alerts/deployment-balance-alerts.service.ts
ygrishajev: yea, possibly

apps/notifications/src/alert/services/deployment-balance-alerts/deployment-balance-alerts.service.ts
ygrishajev: no, this is for myself, this whole thing is not finished obviously

apps/notifications/src/alert/services/deployment-balance-alerts/deployment-balance-alerts.service.ts
ygrishajev: in this case balance is not a number, it's an object containing balance prop which is a num. renamed the var to perhaps reduce confusion

stalniy (dismissed):

apps/notifications/src/alert/services/deployment-balance-alerts/deployment-balance-alerts.service.ts
ygrishajev: I think i want to address this in a different PR using ts result as we discussed

apps/notifications/src/alert/services/deployment/deployment.service.ts
ygrishajev: added anyway. i think there's event eslint rule for this

apps/notifications/src/broker/decorators/handler.decorator.ts
ygrishajev: this is not alert handler, this is a message broker handler basically. but yea, can be renamed

apps/notifications/test.js
ygrishajev: yes

apps/notifications/test/seeders/deployment-balance-alert.seeder.ts
ygrishajev: what's wrong with longer?

apps/notifications/src/alert/services/deployment/deployment.service.ts
ygrishajev: this is tmp, I was to add a proper type here. I did actually

apps/notifications/src/alert/services/deployment/deployment.service.ts
ygrishajev: we should and there is such already. but there is an issue with modules resolution. I can't import non commonjs files into this project and I don't want to get distracted into this atm

Issue #671, opened 2025-01-17
Provider uptime checks caching bad connection


Redm4x:
If a provider goes down the uptime checks in the indexer app will flag it as offline. When it comes back up it should be flagged as online, but in some cases the checks will continue to fail even when the provider is back up. Restarting the indexer fixes the issue so it points to a caching issue. The grpc clients are [memoized](https://github.com/akash-network/console/blob/8fcad27b7b09df457228e0f9ef888abc1459aeb0/apps/indexer/src/providers/statusEndpointHandlers/grpc.ts#L102) so that could be related. grpc-js also seems to reuse existing connection even when recreating the client.

Here's 2 provider that had this issue:
- provider.mylibra.org | [akash18gewxdlc5llzgdr75x3ls4w4cte8hgfzsnr27s](https://console.akash.network/providers/akash18gewxdlc5llzgdr75x3ls4w4cte8hgfzsnr27s?network=mainnet)
  - Happened on 2024-11-17 & 2025-01-04, 
- provider.akash-palmito.org | [akash15ksejj7g4su7ljufsg0a8eglvkje94z8qsh68a](https://console.akash.network/providers/akash15ksejj7g4su7ljufsg0a8eglvkje94z8qsh68a?network=mainnet)
  - Happened on 2024-12-02 ([discord thread](https://discord.com/channels/747885925232672829/1111749310325981315/1313181433866616912))
- provider.dcnorse2.ddns.net
  - Happened on 2025-01-23 ([discord threadf](https://discord.com/channels/747885925232672829/1111748920960372876/1331913130049273856))

This ticket is for documenting future occurrences of the bug and ideally try to reproduce it so that we can figure out the root cause and fix it.

jzsfkzm:
@baktun14 added a PR here, I think now we'll clear cache whenever there is a bad connection.

jzsfkzm:
> grpc-js also seems to reuse existing connection even when recreating the client

@baktun14, I think [my PR will cover](https://github.com/akash-network/console/pull/1264) this part, what do you think?

baktun14:
> > grpc-js also seems to reuse existing connection even when recreating the client
> 
> [@baktun14](https://github.com/baktun14), I think [my PR will cover](https://github.com/akash-network/console/pull/1264) this part, what do you think?

Sounds good, thanks! We cant try it out

baktun14:
I will test locally next week and then release a new indexer version if all is good.

PR #1195, opened 2025-04-10
fix: do not cache bad connections

jzsfkzm:
refs #671


stalniy:
@jzsfkzm could you please explain how is this related to provider proxy? indexer doesn't talk to provider proxy to understand provider online status

jzsfkzm:
> @jzsfkzm could you please explain how is this related to provider proxy? indexer doesn't talk to provider proxy to understand provider online status

Indeed, you're right, not related anyhow. For a brief moment it seemed like connected and a great idea to uncache those if failing too.

stalniy (approved):

baktun14 (approved):

PR #1199, opened 2025-04-10
fix(provider): fixed ordering on control node and worker node and cal…

### LINKED ISSUE #1198: Adding nodes does not show nodes in order in form
When adding multiple nodes, it's not currently showing in order.

jigar-arc10:
Fixed ordering and calculations on node selection.


devalpatel67 (approved):
LGTM

PR #1209, opened 2025-04-13
fix(deployment): provider name might be null

### LINKED ISSUE #1197: Cannot access any of my deployments
When I try to open any of my deployments, I get this error:

```
_app-cd5c61f1aa397feb.js:70 TypeError: Cannot read properties of null (reading 'length')
    at i (2703-a528ccf075fcd1c3.js:1:12614)
    at ak (framework-1556cd39c9ff493f.js:1:61841)
    at i (framework-1556cd39c9ff493f.js:1:120359)
    at oD (framework-1556cd39c9ff493f.js:1:100013)
    at framework-1556cd39c9ff493f.js:1:99879
    at oF (framework-1556cd39c9ff493f.js:1:99887)
    at ox (framework-1556cd39c9ff493f.js:1:96638)
    at oP (framework-1556cd39c9ff493f.js:1:97027)
    at r8 (framework-1556cd39c9ff493f.js:1:45751)
    at framework-1556cd39c9ff493f.js:1:94537
```

Image

jzsfkzm:
closes #1197


apps/deploy-web/src/components/providers/ProviderTableRow.tsx
```diff
@@ -58,7 +58,7 @@ export const ProviderListRow: React.FunctionComponent<Props> = ({ provider }) =>
 
   return (
     <TableRow className="provider-list-row cursor-pointer hover:bg-muted-foreground/10 [&>td]:px-2 [&>td]:py-1" onClick={onRowClick}>
-      {provider.isOnline ? (
+      {provider.name ? (
```
baktun14, line 61: question (blocking): What's the reasoning behind using the instead of `isOnline`?

jzsfkzm: @baktun14 checking for `provider.isOnline` is an indirect way of checking if it has a name, [because of this line](https://github.com/akash-network/console/blob/f0dc51392929dd4b6562b9e92b693c05a6c96f28/apps/api/src/utils/map/provider.ts#L14).
I thought we should check the name directly instead, and that will also make TS to understand `name` must have a string value in the following lines. What's your opinion?

baktun14: Ah ok yea I wasn't sure, but sure makes sense because this is to deplay the name anyways.

baktun14 (approved):

PR #1210, opened 2025-04-14
fix: ensure SSR initiated request forward real client ip

stalniy:


baktun14 (dismissed):

packages/network-store/package.json
```diff
@@ -13,6 +13,7 @@
   },
   "dependencies": {
     "axios": "^1.7.2",
+    "jotai": "^2.9.2",
```
baktun14, line 16: question (non-blocking): Do we use it?

baktun14 (dismissed):

packages/network-store/package.json
stalniy: yes, it's used in `NetworkStore` but not specified in its dependencies. eslint error-ed my changes because of that

PR #1227, opened 2025-04-16
chore: adds flags to codecov default

stalniy:
## Why

our project coverage looks strange -> https://app.codecov.io/github/akash-network/console . It has spikes, which is obviously impossible 😄 . I have an assumption that with every PR we currently re-write project coverage, instead of accumulating it. Probably this happens because we are in monorepo and using flags

Screenshot 2025-04-16 at 13 27 02

## What

adds all flags to codecov default project section

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,16 +1,7 @@
-github_checks:
-  annotations: true
-
 coverage:
   status:
     project:
-      default:
-        target: auto
-        branches:
-          - main
-        if_ci_failed: error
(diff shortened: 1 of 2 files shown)


stalniy:
very likely the answer to this is https://docs.codecov.com/docs/carryforward-flags

baktun14 (dismissed):

stalniy:
I added also handling of coverage checks in case tests were not executed

Issue #1042, opened 2025-03-16
Managed wallet API - Get all deployments


baktun14:
- filter state (active/closed)
- pagination

PR #1243, opened 2025-04-18
feat(deployment): managed api all deployments

baktun14:
- https://github.com/akash-network/console/issues/1042
- Add list deployments
- Refactor deployment detail route


apps/api/src/deployment/services/deployment/deployment.service.ts
```diff
@@ -193,4 +193,33 @@ export class DeploymentService {
...
+  public async list(
+    owner: string,
+    { skip, limit }: { skip?: number; limit?: number }
+  ): Promise<{ deployments: GetDeploymentResponse["data"][]; total: number }> {
+    const deployments = await this.deploymentHttpService.loadAllDeployments(owner, "active");
+    const total = deployments.length;
+
+    const paginatedDeployments = skip !== undefined && limit !== undefined ? deployments.slice(skip, skip + limit) : deployments;
+
+    // Fetch all leases in parallel
```
ygrishajev:
```suggestion
```

apps/api/src/deployment/services/deployment/deployment.service.ts
```diff
@@ -193,4 +193,33 @@ export class DeploymentService {
...
+    const leasePromises = paginatedDeployments.map(deployment => this.leaseHttpService.listByOwnerAndDseq(owner, deployment.deployment.deployment_id.dseq));
+    const leaseResults = await Promise.all(leasePromises);
+
+    // Map deployments with their corresponding leases
```
ygrishajev:
```suggestion
```

apps/api/src/deployment/services/deployment/deployment.service.ts
```diff
@@ -193,4 +193,33 @@ export class DeploymentService {
...
   }
+
+  public async list(
+    owner: string,
+    { skip, limit }: { skip?: number; limit?: number }
+  ): Promise<{ deployments: GetDeploymentResponse["data"][]; total: number }> {
+    const deployments = await this.deploymentHttpService.loadAllDeployments(owner, "active");
+    const total = deployments.length;
+
+    const paginatedDeployments = skip !== undefined && limit !== undefined ? deployments.slice(skip, skip + limit) : deployments;
```
ygrishajev: I'm not sure I follow, so we load all of the deployments anyway, why paginating then? let's add params to http service?

apps/api/test/functional/deployments.spec.ts
```diff
@@ -154,20 +138,62 @@ describe("Deployments API", () => {
...
+      const dseq = faker.string.numeric();
+      const deploymentInfo = DeploymentInfoSeeder.create({
+        owner: address,
+        dseq,
+        state
+      });
+
+      deployments.push(deploymentInfo);
+
+      // Mock the individual deployment info endpoint
```
ygrishajev:
```suggestion
```

apps/api/test/functional/deployments.spec.ts
```diff
@@ -154,20 +138,62 @@ describe("Deployments API", () => {
...
+      nock(apiNodeUrl).persist().get(`/akash/deployment/${betaTypeVersion}/deployments/info?id.owner=${address}&id.dseq=${dseq}`).reply(200, deploymentInfo);
+
+      // Mock the leases for this deployment
```
ygrishajev:
```suggestion
      // Mock the leases for this deployment
```

apps/api/test/functional/deployments.spec.ts
```diff
@@ -154,20 +138,62 @@ describe("Deployments API", () => {
...
+      const leases = LeaseSeeder.createMany(2, {
+        owner: address,
+        dseq,
+        state: "active"
+      });
+
+      nock(apiNodeUrl).persist().get(`/akash/market/${betaTypeVersionMarket}/leases/list?filters.owner=${address}&filters.dseq=${dseq}`).reply(200, { leases });
+    }
+
+    // Mock the deployment list endpoint with all deployments
```
ygrishajev:
```suggestion
```

apps/api/src/deployment/services/deployment/deployment.service.ts
stalniy: this also may cause event loop latency issues

apps/api/src/deployment/services/deployment/deployment.service.ts
```diff
@@ -193,4 +193,33 @@ export class DeploymentService {
...
+
+    // Fetch all leases in parallel
+    const leasePromises = paginatedDeployments.map(deployment => this.leaseHttpService.listByOwnerAndDseq(owner, deployment.deployment.deployment_id.dseq));
+    const leaseResults = await Promise.all(leasePromises);
```
stalniy: **suggestion(non-blocking):** this potentially may be up to 100 parallel requests for every list deployment request. Probably you need to use `PromisePool`

stalniy (dismissed):

apps/api/src/deployment/services/deployment/deployment.service.ts
baktun14: You're right, I checked the pagination proto and it support limit and offset I didn't know, I'll fix it.

apps/api/src/deployment/controllers/deployment/deployment.controller.ts
```diff
@@ -86,4 +88,31 @@ export class DeploymentController {
...
   }
+
+  @Protected([{ action: "sign", subject: "UserWallet" }])
+  async list({ skip, limit }: { skip?: number; limit?: number }): Promise<z.infer<typeof ListDeploymentsResponseSchema>> {
+    const { currentUser, ability } = this.authService;
+
+    const userWallet = await this.userWalletRepository.accessibleBy(ability, "sign").findOneByUserId(currentUser.id);
+    assert(userWallet, 404, "UserWallet Not Found");
+    assert(
+      (skip === undefined && limit === undefined) || (skip !== undefined && limit !== undefined),
```
ygrishajev: suggestion(blocking): zod schema can be used for this

apps/api/src/deployment/controllers/deployment/deployment.controller.ts
```diff
@@ -86,4 +88,31 @@ export class DeploymentController {
...
+      400,
+      "Skip and limit must be provided together or not at all"
```
ygrishajev: why can't limit and skip be used independently?

apps/api/src/deployment/controllers/deployment/deployment.controller.ts
ygrishajev: good way to approach pagination is to enforce some reasonable limit. like `1000` by default.

apps/api/src/deployment/services/deployment/deployment.service.ts
```diff
@@ -193,4 +194,33 @@ export class DeploymentService {
       });
     }
   }
+
+  public async list(
+    owner: string,
+    { skip, limit }: { skip?: number; limit?: number }
```
ygrishajev, line 200: if `limit` is enforced it would be a required param here with no need to check below

apps/api/src/deployment/services/deployment/deployment.service.ts
```diff
@@ -193,4 +194,33 @@ export class DeploymentService {
...
+  ): Promise<{ deployments: GetDeploymentResponse["data"][]; total: number; hasMore: boolean }> {
+    const pagination = typeof skip === "number" && typeof limit === "number" ? { offset: skip, limit } : undefined;
+    const deploymentReponse = await this.deploymentHttpService.loadAllDeployments(owner, "active", pagination);
+    const deployments = deploymentReponse.deployments;
+    const total = parseInt(deploymentReponse.pagination.total);
```
ygrishajev: I think we need to add this rule after all https://eslint.org/docs/latest/rules/radix

packages/http-sdk/src/deployment/deployment-http.service.ts
```diff
@@ -141,42 +151,47 @@ export class DeploymentHttpService extends HttpService {
...
-    );
-  }
+  /**
+   * Load deployments for an owner with optional pagination
+   * @param owner Owner address
+   * @param state Optional state filter
+   * @param pagination Optional pagination parameters
+   * @returns Paginated response with deployments
+   */
+  public async loadAllDeployments(owner: string, state?: string, pagination?: PaginationParams): Promise<DeploymentListResponse> {
```
ygrishajev: this is not all deployments now, right? it's paginated

apps/api/src/deployment/services/deployment/deployment.service.ts
```diff
@@ -193,4 +194,33 @@ export class DeploymentService {
...
+
+    const { results: leaseResults } = await PromisePool.withConcurrency(10)
```
ygrishajev: I think the concurrency is safe to be higher

apps/api/src/deployment/services/deployment/deployment.service.ts
baktun14: To enforce the radix?

apps/api/src/deployment/controllers/deployment/deployment.controller.ts
baktun14: I guess we can make independent, I was thinking in terms of pagination it doesn't make sense

apps/api/src/deployment/services/deployment/deployment.service.ts
baktun14: I'll keep both optional but with a default of 1000 for the limit, makes sense?

ygrishajev: Yea, totally

apps/api/src/deployment/controllers/deployment/deployment.controller.ts
```diff
@@ -86,4 +89,28 @@ export class DeploymentController {
...
   }
+
+  @Protected([{ action: "sign", subject: "UserWallet" }])
+  async list({ skip, limit }: { skip?: number; limit?: number }): Promise<z.infer<typeof ListDeploymentsResponseSchema>> {
+    const { currentUser, ability } = this.authService;
+
+    const userWallet = await this.userWalletRepository.accessibleBy(ability, "sign").findOneByUserId(currentUser.id);
+    assert(userWallet, 404, "UserWallet Not Found");
+
+    const validatedParams = ListDeploymentsQuerySchema.parse({ skip, limit });
```
ygrishajev: I actually meant that it should be possible to define query validation schema on the route definition same way as for body or response

baktun14: It's in the route definition, I'm not sure I follow

ygrishajev: Ah, i think i get where the confusion comes from. If schema validates query, the checks wouldn't be needed. But i forgot that we don't have ts strict null checks enabled and because of this zod makes fields optional.

Idk, i'd just cast query or checking as before is also an option. 

https://github.com/colinhacks/zod/issues/121#issuecomment-675671241

ygrishajev: Wait, skip is optional and limit has a default. Scratch my prev comments 😳

ygrishajev (approved):

PR #1250, opened 2025-04-22
test(alert): adds more functional tests

ygrishajev:


apps/notifications/src/alert/services/blockchain-node-http/blockchain-node-http.service.ts
```diff
@@ -1,13 +1,15 @@
...
+import { Axios, defaults } from 'axios';
+import omit from 'lodash/omit';
 
 import { GlobalEnvConfig } from '@src/config/env.config';
 
 @Injectable()
 export class BlockchainNodeHttpService extends Axios {
   constructor(private readonly configService: ConfigService<GlobalEnvConfig>) {
     super({
+      ...omit(defaults, 'headers'),
```
stalniy, line 12: **nitpick**: you can use native js for this:

```ts
const { headers, ...options } = defaults;

super({ ...options, baseURL: ... })
```

apps/notifications/src/alert/services/blockchain-node-http/blockchain-node-http.service.ts
```diff
@@ -1,13 +1,15 @@
 import { Injectable } from '@nestjs/common';
 import { ConfigService } from '@nestjs/config';
-import { Axios } from 'axios';
+import { Axios, defaults } from 'axios';
+import omit from 'lodash/omit';
 
 import { GlobalEnvConfig } from '@src/config/env.config';
 
 @Injectable()
 export class BlockchainNodeHttpService extends Axios {
```
stalniy, line 9: **thought(non-blocking):** I think the naming reveals implementation details. Today it's HttpService but tomorrow it may be GRPC service

stalniy: but in the form it's currently used, it's easier because otherwise you need to encapsulate http calls into methods. Which later you could replace with grpc calls.

Also I don't know how we will mock grpc requests on network layer in that case. Maybe we will need to do it on DI level and create `TestAppModule` which provides mock implementation for blockchainNodeClient

stalniy (approved):

apps/notifications/src/alert/services/blockchain-node-http/blockchain-node-http.service.ts
ygrishajev: yea, i just like it this way. axios does it too anyway

apps/notifications/src/alert/services/blockchain-node-http/blockchain-node-http.service.ts
ygrishajev: this is a base class, child classes are supposed to hide details

baktun14 (approved):

PR #1253, opened 2025-04-22
fix: removes mistakenly added deps

ygrishajev:


ygrishajev:
@blanxii thanks for spotting this!

baktun14 (approved):

PR #1259, opened 2025-04-23
feat(notifications): improves chain polling reliability

ygrishajev:


apps/notifications/src/chain-events/repositories/block-cursor/block-cursor.repository.ts
```diff
@@ -0,0 +1,63 @@
...
+    @InjectDrizzle(DRIZZLE_PROVIDER_TOKEN)
+    private readonly db: NodePgDatabase<typeof schema>,
+  ) {}
+
+  private readonly id = 'latest';
+
+  async getNextBlockForProcessing<T>(
+    cb: (block: number) => Promise<T>,
+  ): Promise<T> {
+    return await this.db.transaction(async (transaction) => {
```
baktun14, line 22: question(blocking): do we want to do a retry inside a db transaction? I think we will end up in the same situation as the start trial

ygrishajev: I don't think so. This is a single record existing specifically to ensure a single worker against a single block. We're going to run a single worker until it's enough and i think it's gonna be enough for quite long.
The only use case when it's going to be blocking is during a deploy and this is when we want it to behave so.

ygrishajev: We had such issue with start trial coz it was blocked by external traffic. This is an internal service with no impact from the outside. Besides the fact it exists specifically to block out other workers spinning up.

baktun14: Ah ok makes sense, but then if we spin up other workers we could potentially stack db connections if they all try to access the blocks at the same time, but I don't think we'll need more workers any time soon so all good.

baktun14 (approved):

apps/notifications/src/chain-events/repositories/block-cursor/block-cursor.repository.ts
ygrishajev: In case we need more workers, we could have more ids than just "latest" and adjust the query. But yea, I don't see a need for this

PR #1284, opened 2025-04-25
feat(alert): implements sufficient notification send interface

ygrishajev:


apps/notifications/src/alert/services/deployment/deployment.service.ts
```diff
@@ -31,40 +35,77 @@ export class DeploymentService {
...
-        ? parseInt(funds.amount, 10)
-        : parseInt(funds.amount, 10) * 1_000_000;
+      const balanceAmount =
+        balance.denom === 'uakt'
+          ? parseInt(balance.amount, 10)
+          : parseInt(balance.amount, 10) * 1_000_000;
+      const fundsAmount =
+        funds.denom === 'uakt'
+          ? parseInt(funds.amount, 10)
+          : parseInt(funds.amount, 10) * 1_000_000;
```
baktun14: question (blocking): if it's not `uakt`, it's probably ibc usdc, but I don't think we have to multiply by `1_000_000` as it's already `uusdc`? I'm not entirely sure

ygrishajev: I haven't checked honestly, took it from [here](https://github.com/akash-network/console/blob/bd79006abff3ee2d06657269ddd0e76d1554f275/apps/indexer/src/monitors/deploymentBalanceMonitor.ts#L51)

baktun14: Can you test or add a test just to make sure please?

ygrishajev: Sure, i was going to after ur 1st comment

ygrishajev: ok, looking on the response it does seem redundant. Perhaps the indexer one needs attention

baktun14 (approved):

PR #1285, opened 2025-04-25
fix(billing): handle insufficient funds error for user and master wallet

### LINKED ISSUE #1107: Differentiate "insufficient funds" error for user and master wallet
Right now "insufficient funds" error during tx or top up is suppressed as non-actionable. However this is not true when master wallet lacks funds. We'd want to instantly know about it to react asap. So it should throw an error.

Acceptance criteria:
1) ~~when error happens on `/v1/tx` a 503 error should be thrown and logged (with relevant level)~~ ✅ 
2) during top up job summariser should count "insufficient funds" errors only when it's user related error (not master wallet)
3) during top up job if we hit this error - stop immediately and log a relevant error
4) alert in grafana should be updated before releasing this

domhhv:
This PR fixes #1107 

It introduces method helpers to ChainErrorService of billing for parsing error messages of the following format:

`Query failed with (6): rpc error: code = Unknown desc = failed to execute message; message index: 1: 1999472uakt is smaller than 5000000uakt: insufficient funds [cosmos/cosmos-sdk@v0.45.16/x/bank/keeper/send.go:186] With gas wanted: '0' and gas used: '145584' : unknown request`

Where `uakt` can also be one of the two denomination values from the `USDC_IBC_DENOMS` config, and `1999472` and `5000000` can be any numbers. We then check the balance of user master wallet and if it's indeed less than the amount required (5000000 in this case) then we throw 503 instead of 400. Otherwise, we throw  400.

This PR also adds unit tests for the ChainErrorService functionality.

### DIFF
--- a/apps/api/src/billing/services/chain-error/chain-error.service.spec.ts
+++ b/apps/api/src/billing/services/chain-error/chain-error.service.spec.ts
@@ -1,0 +1,127 @@
+import type { BalanceHttpService } from "@akashnetwork/http-sdk";
+import type { EncodeObject } from "@cosmjs/proto-signing";
+import { BadRequest, ServiceUnavailable } from "http-errors";
+import type { MockProxy } from "jest-mock-extended";
+import { mock } from "jest-mock-extended";
+
+import type { Wallet } from "@src/billing/lib/wallet/wallet";
+import type { BillingConfigService } from "@src/billing/services/billing-config/billing-config.service";
+import { ChainErrorService } from "./chain-error.service";
+
+const USDC_IBC_DENOMS = {
+  mainnetId: "ibc/170C677610AC31DF0904FFE09CD3B5C657492170E7E52372E48756B71E56F2F1",
+  sandboxId: "ibc/12C6A0C374171B595A0A9E18B83FA09D295FB1F2D8C6DAA3AC28683471752D84"
+};
+
+describe(ChainErrorService.name, () => {
+  describe("toAppError", () => {
+    const encodeMessages: EncodeObject[] = [];
+
+    it("returns the original Error when no clue is found", async () => {
+      const { service } = setup();
+      const err = new Error("just some random failure");
+      const result = await service.toAppError(err, encodeMessages);
+      expect(result).toBe(err);
+    });
+
+    it("returns 503 when master wallet balance is less than required in uakt", async () => {
+      const { service, balanceHttpService } = setup();
+      const denom = "uakt";
+      const err = new Error(`insufficient funds: 10${denom} is smaller than 20${denom}`);
+      balanceHttpService.getBalance.mockResolvedValue({ amount: 5, denom: "uakt" });
+
+      const appErr = await service.toAppError(err, encodeMessages);
+      expect(appErr).toBeInstanceOf(ServiceUnavailable);
+      expect(appErr.message).toBe("Service temporarily unavailable");
+    });
+
+    it("returns 503 when master wallet balance is less than required in mainnet USDC", async () => {
+      const { service, balanceHttpService } = setup();
+      const denom = USDC_IBC_DENOMS.mainnetId;
+      const err = new Error(`insufficient funds: 10${denom} is smaller than 20${denom}`);
+      balanceHttpService.getBalance.mockResolvedValue({ amount: 5, denom: "ibc/12C6A0C374171B595A0A9E18B83FA09D295FB1F2D8C6DAA3AC28683471752D84" });
+
+      const appErr = await service.toAppError(err, encodeMessages);
+      expect(appErr).toBeInstanceOf(ServiceUnavailable);
+      expect(appErr.message).toBe("Service temporarily unavailable");
+    });
+
+    it("returns 503 when master wallet balance is less than required in sandbox USDC", async () => {
+      const { service, balanceHttpService } = setup();
+      const denom = USDC_IBC_DENOMS.sandboxId;
+      const err = new Error(`insufficient funds: 10${denom} is smaller than 20${denom}`);
+      balanceHttpService.getBalance.mockResolvedValue({ amount: 5, denom: "ibc/170C677610AC31DF0904FFE09CD3B5C657492170E7E52372E48756B71E56F2F1" });
+
+      const appErr = await service.toAppError(err, encodeMessages);
+      expect(appErr).toBeInstanceOf(ServiceUnavailable);
+      expect(appErr.message).toBe("Service temporarily unavailable");
+    });
+
+    it("returns 400 when master wallet balance is more than required in uakt", async () => {
+      const { service, balanceHttpService } = setup();
+      const denom = "uakt";
+      const err = new Error(`insufficient funds: 10${denom} is smaller than 20${denom}`);
+      balanceHttpService.getBalance.mockResolvedValue({ amount: 20, denom: "uakt" });
+
+      const appErr = await service.toAppError(err, encodeMessages);
+      expect(appErr).toBeInstanceOf(BadRequest);
+      expect(appErr.message).toBe("Insufficient funds");
+    });
+
+    it("returns 400 when master wallet balance is more than required in mainnet USDC", async () => {
+      const { service, balanceHttpService } = setup();
+      const denom = USDC_IBC_DENOMS.mainnetId;
+      const err = new Error(`insufficient funds: 10${denom} is smaller than 20${denom}`);
+      balanceHttpService.getBalance.mockResolvedValue({ amount: 20, denom: "ibc/12C6A0C374171B595A0A9E18B83FA09D295FB1F2D8C6DAA3AC28683471752D84" });
+
+      const appErr = await service.toAppError(err, encodeMessages);
+      expect(appErr).toBeInstanceOf(BadRequest);
+      expect(appErr.message).toBe("Insufficient funds");
+    });
+
+    it("returns 400 when master wallet balance is more than required in sandbox USDC", async () => {
+      const { service, balanceHttpService } = setup();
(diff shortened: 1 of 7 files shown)


apps/api/src/billing/services/chain-error/chain-error.service.ts
```diff
@@ -54,12 +65,52 @@ export class ChainErrorService {
...
+    const match = message.match(pattern);
+
+    if (!match) {
+      return null;
+    }
+
+    return {
+      availableAmount: parseInt(match[1], 10),
+      requiredAmount: parseInt(match[3], 10),
+      denomination: match[2]
```
ygrishajev:
```suggestion
      denom: match[2]
```
just coz we use this term across the app

apps/api/src/billing/services/chain-error/chain-error.service.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+    balanceHttpService: MockProxy<BalanceHttpService>;
+    billingConfigService: MockProxy<BillingConfigService>;
+    masterWallet: MockProxy<Wallet>;
+    service: ChainErrorService;
+  } {
+    const balanceHttpService = mock<BalanceHttpService>();
+    const billingConfigService = mock<BillingConfigService>();
+    const masterWallet = mock<Wallet>();
+
+    // Stub configuration values
```
ygrishajev:
```suggestion
```

apps/api/src/billing/services/chain-error/chain-error.service.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+import type { BillingConfigService } from "@src/billing/services/billing-config/billing-config.service";
+import { ChainErrorService } from "./chain-error.service";
+
+const USDC_IBC_DENOMS = {
+  mainnetId: "ibc/170C677610AC31DF0904FFE09CD3B5C657492170E7E52372E48756B71E56F2F1",
+  sandboxId: "ibc/12C6A0C374171B595A0A9E18B83FA09D295FB1F2D8C6DAA3AC28683471752D84"
+};
+
+describe(ChainErrorService.name, () => {
+  function setup(): {
```
ygrishajev: let's pls put this in the end as in other tests

apps/api/src/billing/services/chain-error/chain-error.service.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+    const service = new ChainErrorService(balanceHttpService, billingConfigService, masterWallet);
+
+    return { balanceHttpService, billingConfigService, masterWallet, service };
+  }
+
+  describe("parseInsufficientFundsErrorMessage", () => {
+    it("parses uakt correctly", () => {
+      const { service } = setup();
+      const msg = "42uakt is smaller than 100uakt";
+      expect((service as any).parseInsufficientFundsErrorMessage(msg)).toEqual({
```
ygrishajev: could you pls re-write the test to cover public methods only. no need to do casting here. private methods can change more frequently and public tend to keep interface more.

apps/api/src/billing/services/chain-error/chain-error.service.ts
```diff
@@ -54,12 +65,52 @@ export class ChainErrorService {
...
+        code: 503,
+        message: "Service temporarily unavailable"
+      };
+    }
+  }
+
+  private parseInsufficientFundsErrorMessage(message: string): {
+    availableAmount: number;
+    requiredAmount: number;
+    denomination: string;
```
baktun14:
```suggestion
    denom: string;
```

baktun14 (dismissed):

apps/api/src/billing/services/chain-error/chain-error.service.ts
domhhv: SGTM, done

apps/api/src/billing/services/chain-error/chain-error.service.spec.ts
domhhv: Done

apps/api/src/billing/services/chain-error/chain-error.service.spec.ts
domhhv: Done

apps/api/src/billing/services/chain-error/chain-error.service.spec.ts
domhhv: Sure, done

apps/api/src/billing/services/chain-error/chain-error.service.ts
domhhv: Done

apps/api/src/billing/services/chain-error/chain-error.service.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+      expect(result).toBe(err);
+    });
+
+    it("returns 503 when balance < required for uakt", async () => {
+      const { service, balanceHttpService } = setup();
+      const denom = "uakt";
+      const err = new Error(`insufficient funds: 10${denom} is smaller than 20${denom}`);
+      balanceHttpService.getBalance.mockResolvedValue({ amount: 5, denom: "uakt" });
+
+      const appErr = (await service.toAppError(err, encodeMessages)) as any;
```
ygrishajev: pls avoid `any`s. This case could be handled like so:
```suggestion
      import { ServiceUnavailable } from "http-errors";

      const appErr = await service.toAppError(err, encodeMessages);
      
      expect(appErr).toBeInstanceOf(ServiceUnavailable);
      expect(appErr.message).toBe("Service temporarily unavailable");
```

apps/api/src/billing/services/chain-error/chain-error.service.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+    const encodeMessages: EncodeObject[] = [];
+
+    it("returns the original Error when no clue is found", async () => {
+      const { service } = setup();
+      const err = new Error("just some random failure");
+      const result = await service.toAppError(err, encodeMessages);
+      expect(result).toBe(err);
+    });
+
+    it("returns 503 when balance < required for uakt", async () => {
```
ygrishajev: Smth like this would be more clear. Pls update other descriptions too
```suggestion
    it("returns 503 when master wallet balance is less then required required in uakt", async () => {
```

apps/api/src/billing/services/chain-error/chain-error.service.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+      const denom = "uakt";
+      const err = new Error(`insufficient funds: 10${denom} is smaller than 20${denom}`);
+      balanceHttpService.getBalance.mockResolvedValue({ amount: 5, denom: "uakt" });
+
+      const appErr = (await service.toAppError(err, encodeMessages)) as any;
+      expect(appErr.status).toBe(503);
+      expect(appErr.message).toBe("Service temporarily unavailable");
+    });
+
+    it("returns 503 when balance < required for mainnnet IBC-USDC", async () => {
```
ygrishajev:
```suggestion
    it("returns 503 when balance < required for mainnnet USDC", async () => {
```

apps/api/src/billing/services/chain-error/chain-error.service.spec.ts
domhhv: Thanks, updated

apps/api/src/billing/services/chain-error/chain-error.service.spec.ts
domhhv: Good catch, done

apps/api/src/billing/services/chain-error/chain-error.service.spec.ts
domhhv: Done

apps/api/src/billing/services/chain-error/chain-error.service.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+    balanceHttpService: MockProxy<BalanceHttpService>;
+    billingConfigService: MockProxy<BillingConfigService>;
+    masterWallet: MockProxy<Wallet>;
+    service: ChainErrorService;
+  } {
+    const balanceHttpService = mock<BalanceHttpService>();
+    const billingConfigService = mock<BillingConfigService>();
+    const masterWallet = mock<Wallet>();
+
+    (billingConfigService.get as jest.Mock).mockImplementation((key: string) => {
```
ygrishajev, line 115:
```suggestion
    billingConfigService.get.mockImplementation((key: string) => {
```

ygrishajev (dismissed):

packages/net/src/generated/netConfigData.ts
```diff
@@ -1,38 +1,28 @@
...
-    "version": "0.38.0",
-    "apiUrls": [
-      "https://api.sandbox-01.aksh.pw:443"
-    ],
-    "rpcUrls": [
-      "https://rpc.sandbox-01.aksh.pw:443"
-    ]
+  sandbox: {
+    version: "0.38.0",
+    apiUrls: [""],
```
baktun14: Something is wrong here and could the cause of the tests failing

domhhv: Yes... I didn't notice these changes, these were done by a script during commit:

```
> @akashnetwork/net@0.0.2 generate
> ts-node scripts/generate.ts

Generate network configuration: mainnet, sandbox, testnet-02
Network configuration was written to ./packages/net/src/generated/net-config.ts
```

I made another commit and now the file looks to be the same, apart from formatting changes

ygrishajev (approved):

ygrishajev (approved):

Issue #1272, opened 2025-04-24
Refactor provider routes


baktun14:
- [Providers](https://github.com/akash-network/console/tree/main/apps/api/src/routes/v1/providers) [Active leases](https://github.com/akash-network/console/blob/main/apps/api/src/routes/v1/providerActiveLeasesGraphData.ts) 
- [Provider attribute schema](https://github.com/akash-network/console/blob/main/apps/api/src/routes/v1/providerAttributesSchema.ts)
- [Provider dashboard](https://github.com/akash-network/console/blob/main/apps/api/src/routes/v1/providerDashboard.ts) 
- [Provider graph](https://github.com/akash-network/console/blob/main/apps/api/src/routes/v1/providerGraphData.ts) 
- [Provider regions](https://github.com/akash-network/console/blob/main/apps/api/src/routes/v1/providerRegions.ts) 
- [Provider versions](https://github.com/akash-network/console/blob/main/apps/api/src/routes/v1/providerVersions.ts) 
- [Trial providers](https://github.com/akash-network/console/blob/main/apps/api/src/routes/v1/trialProviders.ts)
- [Providers](https://github.com/akash-network/console/tree/main/apps/api/src/routes/v1/providers)

jzsfkzm:
@baktun14 there's only one item left here, [Trial providers](https://github.com/akash-network/console/blob/main/apps/api/src/routes/v1/trialProviders.ts). Also, we changed the [provider list endpoint](https://github.com/akash-network/console/blob/d2edc96dd2e70ffe91fec7bd86f25b658fb85c61/apps/api/src/provider/http-schemas/provider.schema.ts#L7) so it can list trial providers now.

On the other hand, I don't see the old `GET /v1/trial-providers` endpoint being used in frontend, is that right? If so, we should just drop that route, right? And anyone willing to load trial providers later can just add scope to `GET /v1/providers`.

Sorry, bit lost about what's our plan here, but would like to close this one too and I feel like we're near.

baktun14:
> [@baktun14](https://github.com/baktun14) there's only one item left here, [Trial providers](https://github.com/akash-network/console/blob/main/apps/api/src/routes/v1/trialProviders.ts). Also, we changed the [provider list endpoint](https://github.com/akash-network/console/blob/d2edc96dd2e70ffe91fec7bd86f25b658fb85c61/apps/api/src/provider/http-schemas/provider.schema.ts#L7) so it can list trial providers now.
> 
[...]

PR #1288, opened 2025-04-29
feat: can list trial providers, as well as trial/registered providers

jzsfkzm:
refs #1272


apps/api/src/services/db/providerStatusService.ts
baktun14, line 1: question (non-blocking): since you referred the issue to refactor the old routes in the pr, were you planning on creating another pr to refactor this file as a DI service?

apps/api/src/provider/routes/providers/providers.router.ts
```diff
@@ -105,7 +112,25 @@ const route = createRoute({
   }
 });
 
-export default new OpenAPIHono().openapi(route, async c => {
-  const providers = await cacheResponse(60, cacheKeys.getProviderList, getProviderList, true);
+export const providersRouter = new OpenApiHonoHandler();
+
+const getProviderListCacheKey = (scope: "all" | "trial", registered: boolean) => {
```
baktun14: comment (blocking): You can remove the `registered` flag. We're working on a new trial onboarding and it'll be either trial or paid user.

apps/api/src/services/db/providerStatusService.ts
```diff
@@ -96,9 +97,19 @@ export const getProviderList = async () => {
...
-      {
-        model: ProviderAttributeSignature
-      }
+      trial
+        ? {
+            model: ProviderAttributeSignature,
+            required: true,
+            where: {
+              auditor: AUDITOR,
+              key: registered ? TRIAL_REGISTERED_ATTRIBUTE : TRIAL_ATTRIBUTE,
```
baktun14:
```suggestion
              key: TRIAL_ATTRIBUTE,
```

apps/api/src/services/db/providerStatusService.ts
jzsfkzm: I was thinking about it too, but line of dependencies leading right up to files in `apps/api/src/routes/v1` discouraged me. There are functions in this file: getNetworkCapacity(), getProviderList() (the one we just changed), and getProviderDetail(). Moving all 3 out to DI'd services would need changes in routes/v1 files too, which seems like double work, or if we refactored all affected routes/v1 files right away, seemed like a quite large change. How about if we just moved getProviderList() out, and left the other two in place for now? Then files affected would be like this:

**apps/api/src/provider/routes/providers/providers.router.ts** - instead of using getProviderList(), it should use ProviderController, which should use the new service we add
**apps/api/src/services/external/apiNodeService.ts** - here's a function called getAddressDeployments(), which is using getProviderList()
**apps/api/src/billing/services/trial-validation/trial-validation.service.ts** - is using getAddressDeployments(), but is already a singleton service
**apps/api/src/routes/v1/addresses/deployments.ts** - is also using getAddressDeployments(), this route should be refactored to /apps/api/deployment I guess right away? Or maybe this router is not used any more? I see this as a hint for that:
https://github.com/jzsfkzm/console/blob/d718092c064972079cbb409fd3f5dd63023732db/apps/api/src/routers/legacyRouter.ts#L53

baktun14: Ok yea makes sense to me, let's go step by step because changes for this can become quite large

baktun14 (approved):

PR #1291, opened 2025-04-30
refactor: move provider listing to its service

jzsfkzm:
refs #1272


jzsfkzm:
Hey @baktun14, this is the change we discussed in my [prior PR](https://github.com/akash-network/console/pull/1288#pullrequestreview-2807426568), moving getProviderList() to a service. There were circular dependecies and I had no other idea but created a new service for listing deployments by address, let me know if you have a better solution.

baktun14:
> Contributor

I see, what was the circular dependency?

baktun14:
> Hey @baktun14, this is the change we discussed in my [prior PR](https://github.com/akash-network/console/pull/1288#pullrequestreview-2807426568), moving getProviderList() to a service. There were circular dependecies and I had no other idea but created a new service for listing deployments by address, let me know if you have a better solution.

I see, what was the circular dependency?

apps/api/src/deployment/services/deployment-list-by-owner/deployment-list-by-owner.service.ts
```diff
@@ -0,0 +1,76 @@
...
+import { RestAkashDeploymentListResponse } from "@src/types/rest";
+import { RestAkashLeaseListResponse } from "@src/types/rest/akashLeaseListResponse";
+import { apiNodeUrl, betaTypeVersion, betaTypeVersionMarket } from "@src/utils/constants";
+
+@singleton()
+export class DeploymentListByOwnerService {
+  constructor(private readonly providerService: ProviderService) {}
+
+  public async listByOwner(owner: string, skip: number, limit: number, reverseSorting: boolean, filters: { status?: string } = {}) {
+    const response = await axios.get<RestAkashDeploymentListResponse>(`${apiNodeUrl}/akash/deployment/${betaTypeVersion}/deployments/list`, {
```
baktun14: We have the `deploymentHttpService` for this endpoint

apps/api/src/deployment/services/deployment-list-by-owner/deployment-list-by-owner.service.ts
```diff
@@ -0,0 +1,76 @@
...
+        "filters.owner": owner,
+        "filters.state": filters.status,
+        "pagination.offset": skip,
+        "pagination.limit": limit,
+        "pagination.count_total": true,
+        "pagination.reverse": reverseSorting
+      }
+    });
+
+    const leaseResponse = await axios.get<RestAkashLeaseListResponse>(`${apiNodeUrl}/akash/market/${betaTypeVersionMarket}/leases/list`, {
```
baktun14: `listHttpService` here as well

jzsfkzm:
> > Hey @baktun14, this is the change we discussed in my [prior PR](https://github.com/akash-network/console/pull/1288#pullrequestreview-2807426568), moving getProviderList() to a service. There were circular dependecies and I had no other idea but created a new service for listing deployments by address, let me know if you have a better solution.
> 
> I see, what was the circular dependency?

TrialValidationService also relies on this method, loading deployments. If we put the method in DeploymentService, this would be the chain:

`src/billing/services/trial-validation/trial-validation.service.ts`
`src/deployment/services/deployment/deployment.service.ts`
`src/billing/services/managed-signer/managed-signer.service.ts`
`src/billing/services/trial-validation/trial-validation.service.ts` again

apps/api/src/deployment/services/deployment-list-by-owner/deployment-list-by-owner.service.ts
jzsfkzm: Thanks, did it now.

apps/api/src/deployment/services/deployment-list-by-owner/deployment-list-by-owner.service.ts
jzsfkzm: Did this as well.

apps/api/src/billing/services/trial-validation/trial-validation.service.ts
```diff
@@ -6,16 +6,18 @@ import { singleton } from "tsyringe";
...
 const TRIAL_DEPLOYMENT_LIMIT = 5;
 
 @singleton()
 export class TrialValidationService {
+  constructor(private readonly deploymentListByOwnerService: DeploymentListByOwnerService) {}
+
   async validateTrialLimit(decoded: EncodeObject, userWallet: UserWalletOutput) {
     if (userWallet.isTrialing && decoded.typeUrl === "/akash.deployment.v1beta3.MsgCreateDeployment") {
-      const deployments = await getAddressDeployments(userWallet.address, 0, 1, false, {});
+      const deployments = await this.deploymentListByOwnerService.listByOwner(userWallet.address, 0, 1, false, {});
```
baktun14: question: should we not omit the limit arg that we have at 1?

baktun14:
> > > Hey @baktun14, this is the change we discussed in my [prior PR](https://github.com/akash-network/console/pull/1288#pullrequestreview-2807426568), moving getProviderList() to a service. There were circular dependecies and I had no other idea but created a new service for listing deployments by address, let me know if you have a better solution.
> > 
> > 
> > I see, what was the circular dependency?
> 
> TrialValidationService also relies on this method, loading deployments. If we put the method in DeploymentService, this would be the chain:
> 
> `src/billing/services/trial-validation/trial-validation.service.ts` `src/deployment/services/deployment/deployment.service.ts` `src/billing/services/managed-signer/managed-signer.service.ts` `src/billing/services/trial-validation/trial-validation.service.ts` again

Ah I see, perhaps we could have read/write distinct services per entity? The main issue is the `managed-signer.service.ts` and if we split only write service methods into a separate service it should work and the separation would be cleaner. Wdyt @ygrishajev ?

apps/api/src/billing/services/trial-validation/trial-validation.service.ts
jzsfkzm: If I get it right, it's the "count" we load that list for as it is returning total number of records (regardless of pagination). Maybe a dedicated method for this would be nicer though.

jzsfkzm:
> > > > Hey @baktun14, this is the change we discussed in my [prior PR](https://github.com/akash-network/console/pull/1288#pullrequestreview-2807426568), moving getProviderList() to a service. There were circular dependecies and I had no other idea but created a new service for listing deployments by address, let me know if you have a better solution.
> > > 
> > > 
> > > I see, what was the circular dependency?
> > 
> > 
> > TrialValidationService also relies on this method, loading deployments. If we put the method in DeploymentService, this would be the chain:
> > `src/billing/services/trial-validation/trial-validation.service.ts` `src/deployment/services/deployment/deployment.service.ts` `src/billing/services/managed-signer/managed-signer.service.ts` `src/billing/services/trial-validation/trial-validation.service.ts` again
> 
> Ah I see, perhaps we could have read/write distinct services per entity? The main issue is the `managed-signer.service.ts` and if we split only write service methods into a separate service it should work and the separation would be cleaner. Wdyt @ygrishajev ?

That's a great idea! I just did the change, and now we there is a cleaner separation, but there is no circular dependency.

apps/api/src/deployment/http-schemas/deployment.schema.ts
```diff
@@ -127,6 +128,51 @@ export const ListDeploymentsResponseSchema = z.object({
...
+  limit: z.string().openapi({
+    description: "Deployments to return",
+    example: "10",
+    maximum: listByOwnerMaxLimit
+  })
+});
+
+export const ListByOwnerQuerySchema = z.object({
+  status: z.string().optional().openapi({
+    description: "Filter by status", // TODO: Set possible statuses?
```
baktun14: `"active" | "closed"` these are the 2 states

packages/http-sdk/src/lease/lease-http.service.ts
```diff
@@ -54,11 +54,24 @@ export class LeaseHttpService extends HttpService {
...
+
+  public async listByOwner(owner: string): Promise<RestAkashLeaseListResponse> {
+    return this.extractData(
+      await this.get<RestAkashLeaseListResponse>("/akash/market/v1beta4/leases/list", {
+        params: {
+          "filters.owner": owner,
+          "filters.state": "active"
+        }
+      })
+    );
```
baktun14, line 75: These are the same endpoint, but with different filters, we can merge them.

apps/api/src/billing/services/trial-validation/trial-validation.service.ts
baktun14: I'm not entirely sure, can you test with limit `1` when it's supposed to return multiple entries just to make sure?

apps/api/src/deployment/http-schemas/deployment.schema.ts
jzsfkzm: Made this change, thanks @baktun14.

apps/api/src/billing/http-schemas/wallet.schema.ts
```diff
@@ -15,5 +15,45 @@ export const WalletResponseOutputSchema = z.object({
 export const WalletListResponseOutputSchema = z.object({
   data: z.array(WalletOutputSchema)
 });
+
+export const SignTxRequestInputSchema = z.object({
```
ygrishajev: the file is called `wallet.schema` hinting it contains relevant entity schemas. `SignTxRequestInputSchema` and other added here are different requests. Yes, they are using a wallet far under the hood. However I'd suggest to create a different files if you want ti extract them.

apps/api/src/billing/services/trial-validation/trial-validation.service.ts
```diff
@@ -6,16 +6,18 @@ import { singleton } from "tsyringe";
...
 const TRIAL_DEPLOYMENT_LIMIT = 5;
 
 @singleton()
 export class TrialValidationService {
+  constructor(private readonly deploymentReaderService: DeploymentReaderService) {}
+
   async validateTrialLimit(decoded: EncodeObject, userWallet: UserWalletOutput) {
     if (userWallet.isTrialing && decoded.typeUrl === "/akash.deployment.v1beta3.MsgCreateDeployment") {
-      const deployments = await getAddressDeployments(userWallet.address, 0, 1, false, {});
+      const deployments = await this.deploymentReaderService.listByOwner(userWallet.address, 0, 1, false, {});
```
ygrishajev: Please change the interface of this method to an object. Having number/boolean args is impossible to understand.
Compare to: 

```suggestion
      const deployments = await this.deploymentReaderService.listByOwner({ address: userWallet.address, limit: 1 });
```

apps/api/src/deployment/services/deployment-reader/deployment-reader.service.ts
```diff
@@ -0,0 +1,156 @@
...
+    }));
+
+    return {
+      deployments: deploymentsWithLeases,
+      total,
+      hasMore: skip !== undefined && limit !== undefined ? total > skip + limit : false
+    };
+  }
+
+  public async listByOwner(owner: string, skip: number, limit: number, reverseSorting: boolean, filters: { status?: "active" | "closed" } = {}) {
```
ygrishajev: Please change the interface of this method to an object. Having number/boolean args is impossible to understand.

apps/api/src/deployment/services/deployment-reader/deployment-reader.service.ts
```diff
@@ -0,0 +1,156 @@
...
+      deployment: deploymentResponse.deployment,
+      leases: leasesWithStatus.map(({ lease, status }) => ({
+        ...lease,
+        status
+      })),
+      escrow_account: deploymentResponse.escrow_account
+    };
+  }
+
+  public async list(
```
ygrishajev, line 74: I'm haven't looked specifically into this methods and the one below. I see the response types are different. However names are confusing. E.g. this one is `list` and accepts `owner` as an arg, the one below is `listByOwner` - on the outside both doing the same thing.

apps/api/src/deployment/controllers/deployment/deployment.controller.ts
```diff
@@ -110,4 +113,14 @@ export class DeploymentController {
       }
     };
   }
+
+  async listByOwner(
```
ygrishajev: let's pass the whole parsed query as an object as a rule of thumb. It's just easier to extend this later on, less code. Also args look more obvious.

apps/api/src/deployment/http-schemas/deployment.schema.ts
```diff
@@ -127,6 +128,51 @@ export const ListDeploymentsResponseSchema = z.object({
...
+    maximum: listByOwnerMaxLimit
+  })
+});
+
+export const ListByOwnerQuerySchema = z.object({
+  status: z.enum(["active", "closed"]).optional().openapi({
+    description: "Filter by status",
+    example: "closed"
+  }),
+  reverseSorting: z.string().optional().openapi({
```
ygrishajev: I think more common term for this is `order=asc|desc`

apps/api/src/deployment/routes/deployments/deployments.router.ts
```diff
@@ -197,3 +226,28 @@ deploymentsRouter.openapi(listRoute, async function routeListDeployments(c) {
...
+  const skip = parseInt(c.req.valid("param").skip);
+  const limit = Math.min(listByOwnerMaxLimit, parseInt(c.req.valid("param").limit));
+
+  if (isNaN(skip)) {
+    return c.text("Invalid skip.", 400);
+  }
+
+  if (isNaN(limit)) {
+    return c.text("Invalid limit.", 400);
+  }
```
ygrishajev: non-blocking: I think all of the above validations can be performed on a zod schema level

apps/api/src/provider/controllers/provider/provider.controller.ts
```diff
@@ -18,4 +22,13 @@ export class ProviderController {
   async cleanupProviderDeployments(options: ProviderCleanupParams) {
     return await this.providerCleanupService.cleanup(options);
   }
+
+  async getProviderList(scope: ProviderListQuery["scope"]) {
```
ygrishajev, line 26: non-blocking

check this out

```typescript
import { Memoize } from "@src/caching/helpers";

@Memoize(options)
getProviderList()
```

ygrishajev:
> Wdyt @ygrishajev ?

It might be a good idea. It complicates the call tree a bit on the other hand. I don't have hard objections to it. Just remember types can be circularly imported in case that's what's causing the issue.

apps/api/src/provider/controllers/provider/provider.controller.ts
jzsfkzm: Ah, that's nice, I'll keep it in mind! However, won't help in this case, since cache key depends on the scope (trial or not).

apps/api/src/deployment/routes/deployments/deployments.router.ts
jzsfkzm: That's right, I just did.

apps/api/src/deployment/http-schemas/deployment.schema.ts
jzsfkzm: It would be better indeed, but then we'd have this mental burden of having `reverseSorting` on the API side, turning into `order`, used as `order` all around in our router, controller, and service, then again turning into `reverse` when requesting data from akash-api. I mean it would be nicer indeed to have `order`, but then should be changed everywhere, what do you think?

apps/api/src/deployment/controllers/deployment/deployment.controller.ts
jzsfkzm: Done.

apps/api/src/deployment/services/deployment-reader/deployment-reader.service.ts
jzsfkzm: I had a hard time naming these, now it's `listByOwner` and `listByOwnerAndStatus`, but I think we should unify these two methods, and then extract the appropriate type if result for each endpoint. What do you think? Should we? In the same PR?

apps/api/src/deployment/services/deployment-reader/deployment-reader.service.ts
jzsfkzm: Indeed, it's done.

apps/api/src/deployment/controllers/deployment/deployment.controller.ts
```diff
@@ -110,4 +116,14 @@ export class DeploymentController {
...
     };
   }
+
+  async listByOwner(query: {
+    address: string;
+    status?: "active" | "closed";
+    skip?: number;
+    limit?: number;
+    reverseSorting?: boolean;
+  }): Promise<z.infer<typeof ListByOwnerResponseSchema>> {
```
ygrishajev: non-blocking: Would this work?
```suggestion
  query: ListByOwnerQuerySchema & ListByOwnerParamsSchema): Promise<z.infer<typeof ListByOwnerResponseSchema>> {
```

apps/api/src/deployment/services/deployment-reader/deployment-reader.service.ts
```diff
@@ -0,0 +1,168 @@
...
+    }));
+
+    return {
+      deployments: deploymentsWithLeases,
+      total,
+      hasMore: skip !== undefined && limit !== undefined ? total > skip + limit : false
+    };
+  }
+
+  public async listByOwnerAndStatus({
```
ygrishajev: non-blocking: the return type of this method is different then the one above. From the name I'd assume it may be the same. From the code it seems like it's summarising resources somehow. Perhaps it makes sense to reflect it in the name?

I know this is a compound from different sources, ideally we'd want to use smth more unified eventually.

ygrishajev (dismissed):

apps/api/src/deployment/controllers/deployment/deployment.controller.ts
jzsfkzm: Good idea, it works with small adjustments. There were two issues: the type emits address as optional, but we wanted it to be required. Also reverseSorting was a string on the schema level, which I managed to solve with a transform - so now the API still accepts string values, but the query string type emits a boolean.

apps/api/src/deployment/services/deployment-reader/deployment-reader.service.ts
jzsfkzm: I think I found a better name, `listWithResources`, and renamed the chain all along from route to service method name. Is this ok now? I think at a - not much - later point we might want to unify these two.

apps/api/src/deployment/services/deployment-reader/deployment-reader.service.ts
ygrishajev: nah, I don't think it makes sense in the same PR. I think it is better now. I'm totally with you, this one is hard and the source of the issue is a tech debt after all

ygrishajev (approved):

apps/api/src/deployment/controllers/deployment/deployment.controller.ts
ygrishajev: oh, right, I keep forgetting about it. The issue source is [this](https://github.com/colinhacks/zod/issues/121#issuecomment-675671241) and it's hard to refactor.

baktun14 (approved):

packages/http-sdk/src/lease/lease-http.service.ts
baktun14: @jzsfkzm  I'll merge it so you proceed with other tasks, but it would be great if you could merge these functions in another branch! Thanks

PR #1300, opened 2025-05-07
feat(notifications): refactors app structure to separate concerts

ygrishajev:
The purpose of the refactoring is to separate core business logic from outer interfaces which we have several now.

Domain specific modules are now contained in the `modules` dir. These modules should know nothing about outer interfaces like `pg-boss` or `http`.

Various interfaces are contained in the `interfaces` dir, every module in this dir can be treated as an individual entry point.

`main.ts` now implements `process.env.INTERFACE` var which would allow to run a specific module or all of them at once.

Also a basic raw alert create endpoint is implemented


apps/notifications/src/interfaces/rest/interceptors/http-validate/http-validate.interceptor.ts
```diff
@@ -0,0 +1,140 @@
...
+      VALIDATION_META_KEY,
+      handler,
+    );
+
+    if (!options) return next.handle();
+
+    const request = context.switchToHttp().getRequest();
+
+    if (options.body) {
+      const result = options.body.safeParse(request.body);
```
baktun14, line 74: question (non-blocking): what type is this validating? `ZodTypeAny`?

apps/notifications/src/lib/zod-form-ast/zod-form-ast.ts
```diff
@@ -0,0 +1,139 @@
+import type { ZodTypeAny } from 'zod';
+import { ZodFirstPartyTypeKind } from 'zod';
+
+export function toAST(schema: ZodTypeAny): Record<string, any> {
```
baktun14, line 4: question (non-blocking): What does this function do?

baktun14 (dismissed):

apps/notifications/src/interfaces/rest/interceptors/http-validate/http-validate.interceptor.ts
ygrishajev: the type is provided at the controller like so https://github.com/akash-network/console/pull/1300/files#diff-5e8295c91031e3495ac7045129ecb0173edd9c48bc0589c1803250f2299e7267R35

apps/notifications/src/lib/zod-form-ast/zod-form-ast.ts
ygrishajev: oh, forgot to remove it. This is to convert a zod schema to a structure which can be used to dynamically build forms in ui. This one is working as expected but i'd need to actually try building a form to verify

ygrishajev: adding it in here just in case https://gist.github.com/ygrishajev/d8c180ddcc3fb52710460fe62238656b

PR #1312, opened 2025-05-12
feat: jwt provider schema

baktun14:
## Summary by CodeRabbit

- **New Features**
  - Introduced a new `@akashnetwork/jwt` package for Akash provider authentication using JWT tokens, including utilities for token creation, validation, and wallet-based signing.
  - Added comprehensive TypeScript types, validation logic, and schema enforcement for JWT payloads and permissions.
  - Enabled automated code generation and test case fetching for the JWT package.
  - Provided new test suites for JWT token and validator functionality.

- **Chores**
  - Updated project and package metadata (repository, license, author) across multiple packages.
  - Added new pre-commit hooks and lint-staged configurations for the JWT package.
  - Integrated JWT package dependency into the web deployment app.
  - Established automated GitHub Actions workflow for JWT package validation.

- **Documentation**
  - Added configuration and documentation files for TypeScript and Jest in the JWT package.


packages/jwt/src/wallet-utils.ts
```diff
@@ -0,0 +1,45 @@
...
+    address: account.address,
+    signArbitrary: async (signer: string, data: string | Uint8Array, accountIndex: number = 0): Promise<StdSignature> => {
+      const message = typeof data === "string" ? new TextEncoder().encode(data) : data;
+      const hashedMessage = sha256(message);
+      const seed = await fromMnemonic(wallet.mnemonic);
+      const { privkey } = Slip10.derivePath(Slip10Curve.Secp256k1, seed, stringToPath(`${BASE_HD_PATH}${accountIndex}`));
+      const signature = await Secp256k1.createSignature(hashedMessage, privkey);
+      const signatureBytes = new Uint8Array([...signature.r(32), ...signature.s(32)]);
+      const stdSignature = encodeSecp256k1Signature(account.pubkey, signatureBytes);
+
```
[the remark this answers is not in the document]
baktun14: It throws with `Signature must be 64 bytes long. Cosmos SDK uses a 2x32 byte fixed length encoding for the secp256k1 signature integers r and s.` when using `const signatureBytes = signature.toFixedLength();`

packages/jwt/src/jwt-token.ts
```diff
@@ -0,0 +1,118 @@
...
+    // Check not-before time
+    if (payload.nbf && payload.nbf > now) {
+      errors.push("Token is not yet valid (nbf check failed)");
+    }
+
+    return {
+      isValid: errors.length === 0,
+      errors: errors.length > 0 ? errors : undefined
+    };
+  }
```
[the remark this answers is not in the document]
baktun14: Yes give me a reference implementation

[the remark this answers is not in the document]
baktun14: I will leave this verification up to the provider node code

.github/workflows/jwt-validate.yml
```diff
@@ -0,0 +1,29 @@
+name: JWT Package Validation
+
+on:
+  pull_request:
+    branches: ["main"]
+    paths:
+      - "packages/jwt/**"
+      - ".github/workflows/jwt-validate.yml"
```
stalniy, line 8: thought(non-blocking): what do you think if create single workflow for all packages?

package.json
```diff
@@ -39,6 +39,9 @@
     "./packages/net/**/*.ts": [
       "npm run validate:types -w packages/net"
     ],
+    "./packages/jwt/**/*.ts": [
+      "npm run validate:types -w packages/jwt"
```
stalniy, line 43: thought(non-blocking): you can call this from gha, that would be more safe. eventually we need to move all this lint-staged hooks to gha

packages/jwt/package.json
```diff
@@ -0,0 +1,38 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/akash-network/console"
+  },
+  "license": "Apache-2.0",
+  "author": "Akash Network",
+  "main": "src/index.ts",
+  "scripts": {
+    "format": "prettier --write ./*.{ts,json} **/*.{ts,json}",
+    "generate": "ts-node scripts/generate.ts && npm run format",
```
stalniy, line 14: **nitpick:** you can use `node --experimental-strip-types` instead of `ts-node`

packages/jwt/src/jwt-token.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+    // For test cases that should fail, we need to validate the payload first
+    if (testCase.expected.signFail || testCase.expected.verifyFail) {
+      const validationResult = jwtToken.validatePayload(claims);
+      expect(validationResult.isValid).toBe(false);
+
+      if (validationResult.isValid) {
+        throw new Error("Validation should have failed", { cause: testCase });
+      }
+
+      return;
```
stalniy, line 36: thought(blocking): this code is unreachable because if `validationResult.isValid` is `true` `expect` will throw an error. Do I miss something?

stalniy: @coderabbitai thoughts?

packages/jwt/src/jwt-token.ts
```diff
@@ -0,0 +1,118 @@
...
+      nbf: options.nbf,
+      iat: options.iat,
+      jti: options.jti,
+      version: options.version || "v1",
+      leases: options.leases || { access: "full" }
+    };
+
+    const validationResult = this.validatePayload(inputPayload);
+    if (!validationResult.isValid) {
+      throw new Error(`Invalid payload: ${validationResult.errors?.join(", ")}`);
```
stalniy, line 56: thought(non-blocking): I'm just thinking who is interested in this error. Will somebody want to handle it? It seems like a runtime error which should not happen in majority of cases. If nobody wants to catch it and process then we could generate proper TS type which ensures validity of payload at build time and get rid of runtime validation, right? 

BUT if my assumption is wrong, then this `Error` should have a custom type and other validation errors should be passed as `cause`, for example:

```ts
new InvalidJwtPayload({ cause: AggregateError.from(validationResult.errors) })
```

packages/jwt/src/jwt-token.ts
```diff
@@ -0,0 +1,118 @@
...
+    if (parts.length !== 3) {
+      throw new Error("Invalid JWT format");
+    }
+
+    try {
+      const [, payload] = parts;
+      const json = Buffer.from(payload, "base64url").toString("utf8");
+      return JSON.parse(json);
+    } catch (error) {
+      throw new Error("Failed to decode JWT token");
```
stalniy: suggestion(non-blocking): 

```ts
throw new Error("Failed to decode JWT token", { cause: error });
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error#cause

packages/jwt/src/jwt-token.ts
stalniy: ah I see `errors` is a `string[]` maybe it then needs to be a list of error codes, wdyt?

packages/jwt/src/jwt-validator.ts
```diff
@@ -0,0 +1,234 @@
...
+        return result;
+      } else if (typeof token !== "string" && (!token || Object.keys(token).length === 0)) {
+        result.errors.push("Error validating token: Empty payload provided");
+        return result;
+      }
+
+      let payload: Record<string, any>;
+
+      if (typeof token === "string") {
+        const parts = token.split(".");
```
stalniy: **nitpick:** small optimization `token.split('.', 3)`

packages/jwt/src/jwt-validator.ts
```diff
@@ -0,0 +1,234 @@
...
+        if (!header.alg) {
+          result.errors.push("Missing required field in header: alg");
+          return result;
+        }
+      } else {
+        payload = token;
+      }
+
+      // Validate payload with the schema
+      const validate = this.ajv.compile(jwtSchemaData);
```
stalniy: suggestion(non-blocking): this should be cached because `jwtSchemaData` cannot be changed between `validateToken` calls

packages/jwt/src/jwt-validator.ts
```diff
@@ -0,0 +1,234 @@
...
+      // Additional validation for granular access
+      if (payload.leases?.access === "granular") {
+        if (!payload.leases?.permissions) {
+          result.errors.push("Missing required field: permissions");
+          valid = false;
+        } else {
+          // Check for duplicate providers
+          const providers = new Set<string>();
+          for (const perm of payload.leases.permissions) {
+            if (providers.has(perm.provider)) {
```
stalniy, line 118: **question(non-blocking):** did you try to use https://ajv.js.org/packages/ajv-keywords.html#uniqueitemproperties?

packages/jwt/src/jwt-validator.ts
```diff
@@ -0,0 +1,234 @@
...
+                valid = false;
+              }
+            }
+
+            // Check for duplicate scopes within each permission
+            if (perm.scope) {
+              const scopes = new Set<string>();
+              for (const scope of perm.scope) {
+                if (scopes.has(scope)) {
+                  result.errors.push("Duplicate scope in permission");
```
stalniy: suggestion(non-blocking): would be nice to know which scope is duplicated: 

```ts
result.errors.push(`Duplicate scope in permission: ${scope}`);

```

packages/jwt/src/jwt-validator.ts
```diff
@@ -0,0 +1,234 @@
...
+        result.errors.push("Permissions not allowed for full access");
+        valid = false;
+      }
+
+      result.isValid = result.errors.length === 0;
+    } catch (error) {
+      const errorMessage = error instanceof Error ? error.message : String(error);
+      result.errors.push(`Error during JWT validation: ${errorMessage}`);
+      // Optionally log the error for debugging
+      console.error("JWT validation error:", error);
```
stalniy: suggestion(blocking): ideally, this should be done on a calling side and `result.errors` should have enough information for this. If you still want to keep it, I think you need to use our `LoggerInstance` instead because it will be sent to our logging stack

packages/jwt/src/jwt-validator.ts
```diff
@@ -0,0 +1,234 @@
...
+    if (deployment.services) {
+      const services = new Set<string>();
+      for (const service of deployment.services) {
+        if (services.has(service)) {
+          result.errors.push("Duplicate service in deployment");
+          valid = false;
+          break;
+        }
+        services.add(service);
+      }
```
stalniy, line 229: thought(blocking): this code is easily extractable to a function and can be then reused between calls:

```ts
const duplicatedServices = findDuplicated(deployment.services));
duplicatedServices.forEach(service => result.errors.push(`Duplicated service in deployment: ${service}`));
```
Or even this string could be passed as a parameter to findDuplicated, so then you could do:

```ts
result.errors.push(
  ...collectDuplicated(deployment.services, service => `Duplicated service in deployment: ${service}`)
)
```

packages/jwt/src/test/test-utils.ts
```diff
@@ -0,0 +1,46 @@
...
+ */
+export function replaceTemplateValues(testCase: (typeof jwtClaimsTestCases)[0]) {
+  const now = Math.floor(Date.now() / 1000);
+  const issuer = createAkashAddress();
+  const provider = createAkashAddress();
+
+  const claims = { ...testCase.claims } as any;
+  if (claims.iss === "{{.Issuer}}") claims.iss = issuer;
+  if (claims.iat === "{{.IatCurr}}") claims.iat = now;
+  if (claims.iat === "{{.Iat24h}}") claims.iat = now + 86400; // 24 hours from now
```
stalniy: suggestion(non-blocking): create a constant `ONE_DAY` or `ONE_HOUR`. Then you will be able to write

```
now + 24 * ONE_HOUR
// or
now + ONE_DAY
// or
now + 2 * ONE_DAY
```

Which is self-explanatory and doesn't require comment

packages/jwt/src/test/test-utils.ts
```diff
@@ -0,0 +1,46 @@
...
+ * @param testCase - The test case containing template values
+ * @returns The test case with template values replaced
+ */
+export function replaceTemplateValues(testCase: (typeof jwtClaimsTestCases)[0]) {
+  const now = Math.floor(Date.now() / 1000);
+  const issuer = createAkashAddress();
+  const provider = createAkashAddress();
+
+  const claims = { ...testCase.claims } as any;
+  if (claims.iss === "{{.Issuer}}") claims.iss = issuer;
```
stalniy, line 28: suggestion(non-blocking): if `claims` is a pure JSON then you can do it throught `JSON.parse(JSON.stringify(` combo. What will let you replace vars in any object inside claims without knowing its structure

```ts
const knownVars = {
  Issuer: issue,
  // ... other vars
};
const claims = JSON.parse(JSON.stringify(testCase.claims).replace(/"?\{\{\s*\.([\w_]+)\s*\}\}"?/g, (_, varName) => {
  return JSON.stringify(knownVars[varName]);
}))
```

tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+    "strict": true,
+    "esModuleInterop": true,
+    "skipLibCheck": true,
+    "forceConsistentCasingInFileNames": true,
+    "outDir": "dist",
+    "rootDir": ".",
+    "baseUrl": ".",
+    "paths": {
+      "*": ["node_modules/*"]
+    }
```
stalniy: question(non-blocking): is this needed? I think it should work without this path mapping

stalniy (dismissed):

.github/workflows/jwt-validate.yml
baktun14: That would be very useful

tsconfig.json
baktun14: I'm not sure why this tsconfig ended up there, I'll remove it

packages/jwt/src/test/test-utils.ts
baktun14: Good to know

.github/workflows/jwt-validate.yml
baktun14: I'll just leave it as is since it's gonna go into the sdk eventually.

package.json
baktun14: Created this issue if you want to add to it: https://github.com/akash-network/console/issues/1323

packages/jwt/package.json
baktun14: Tried it but throws an error

packages/jwt/src/jwt-token.ts
baktun14: It'll useful for other clients who want to use the jwt authentication with providers. For us internally it should always be handled correctly. What do you mean by a list of error codes?

packages/jwt/src/jwt-validator.ts
baktun14: I think this is an ajv specific feature, I can't find it in the json schema spec. Technically all of this manual validation could be left to the provider go code to validate and we could just validate the schema. wdyt?

packages/jwt/src/jwt-validator.ts
baktun14: Removed it thanks

packages/jwt/src/jwt-token.spec.ts
baktun14: Thsi is to avoid continuing with the createToken

packages/jwt/src/jwt-token.ts
stalniy: But GoLang code will always properly validate it on server side. So, on client side we could rely on build time type system. just a thought

Eventually both client and server side happens in runtime, so eventually client will get the right error back in runtime. client side validation in this case, just decreases load on server.

Though it may be useful, if we create JWT token from some web form. BUT if we do create it from some form, we also need a way to programmatically process it in order to show errors under specific fields. Currently, all collected errors are actually error messages. It's hard to work with error messages programmatically, that's why I suggested to use error code instead of error messages, or maybe even custom error objects like this ->

```ts
interface JwtValidationError {  
  // payload field name
  field?: string;
  // error code, useful for possible i18n support in future and for docs references
  code: string;
  // error message for developer
  message: string;
}
```

and then errors will be of `JwtValidationError[]` type instead of `string[]`

packages/jwt/src/jwt-validator.ts
stalniy: yes, this is non-spec stuff, and I agree that custom stuff could be left for server side validation

packages/jwt/src/jwt-token.ts
stalniy: Eventually this is c

packages/jwt/src/test/test-utils.ts
stalniy: I wanted to point this out because if a Go developer decides to add formatting to the templates - for example, writing `{{ .Issuer }}` with extra spaces around var name - our current solution might break. This makes it somewhat fragile to minor changes.

packages/jwt/src/types.ts
```diff
@@ -0,0 +1,80 @@
...
+  jti?: string;
+  /** Version of the JWT specification (currently fixed at v1). */
+  version: "v1";
+  /** Access control configuration for leases */
+  leases: {
+    /** Access level for the lease: 'full' for unrestricted access to all actions, 'granular' for provider-specific permissions. */
+    access: Access;
+    /** Global list of permitted actions across all owned leases (no duplicates). Applies when access is 'full'. */
+    scope?: Scope[];
+    /** Required if leases.access is 'granular'; defines provider-specific permissions. */
```
stalniy, line 23: suggestion(non-blocking): it's possible to define it with tagged union:

```ts
type Leases = FullAccessToLeases | GranularAccessToLeases

interface FullAccessToLeases {
  access: 'full';
  scope: Scope[];
}

interface GranularAccessToLeases {
  access: 'granular';
  permissions: ...
}
```

This will validate the correct object shape at build time with help of TypeScript

stalniy (approved):

packages/jwt/src/test/test-utils.ts
stalniy: You can also keep it as is, I'd recommend to write at least small utility to prevent this. Like `getTemplateVarName` and then

```ts
if (isTemplateVariable(claims.iss, "Issuer")) claims.iss = ...

function isTemplateVariable(rawValue: unknown, name: string): boolean {
  if (typeof rawValue !== 'string') return false;
  const value = rawValue.trim();
  return value.startsWith('{{') && value.endsWith('}}') && value.slice(2, -2) === name;
}
```

packages/jwt/src/jwt-token.ts
baktun14: For now there's no requirements for jwt in a web form so I'll keep it like this for now so I can implement it in the managed wallet api. It'll only be used from within our api. We'll probably move this package into the sdk, so feel free to implement it in there when we do so.

packages/jwt/src/jwt-validator.ts
baktun14: For the sake of moving forward with the implementation, I'll keep it like this.

packages/jwt/src/types.ts
baktun14: Good to know, I'll keep it like this for the sake of implementing the jwt in the api.

PR #1330, opened 2025-05-16
feat(notifications): implements swagger and code generation

ygrishajev:
## Summary by CodeRabbit

- **New Features**
  - Introduced a code-generated React Query SDK package for notifications, providing strongly typed hooks and utilities for interacting with alerts and contact points APIs.
  - Added new scripts and dependencies to support OpenAPI/Swagger-based SDK and schema generation.
  - Enhanced API validation by integrating DTO classes and improved HTTP validation decorators.
  - Added comprehensive documentation for code generation and SDK usage.

- **Bug Fixes**
  - Removed the unique constraint on the user ID in contact points, allowing multiple contact points per user.

- **Refactor**
  - Replaced custom HTTP validation and bootstrap logic with standardized factory patterns and improved exception handling.
  - Updated controller and bootstrap initialization patterns for better maintainability.

- **Chores**
  - Updated configuration files, build scripts, and ignored files to support new SDK generation and build processes.

- **Tests**
  - Updated and removed tests to align with new validation and bootstrap mechanisms.


stalniy (dismissed):

apps/notifications/src/lib/bootstrap/bootstrapper/bootstrapper.ts
baktun14, line 1: That is quite a path haha

baktun14: Maybe it can be in `/bootstrap` since `bootstrap.ts` got deleted

ygrishajev: the though behind the structure is:
`lib` - contains module like library code, which potentially can be extracted into a dedicated/publishable package

Files inside a module are all in dedicated dirs, as those are files which could be covered with unit tests. I decided not to go with tests for this this one but still keep a common structure.

ygrishajev: another thing, if it's in the root it's kinda assumed it's somehow related to the rest of the files, however it is not

packages/api-schemas/src/notifications.schemas.ts
```diff
@@ -0,0 +1,355 @@
...
+              .object({
+                operator: z.union([z.literal("eq"), z.literal("lt"), z.literal("gt"), z.literal("lte"), z.literal("gte")]),
+                field: z.literal("balance"),
+                value: z.number()
+              })
+              .passthrough()
+          ]),
+          params: z.object({ dseq: z.string(), owner: z.string() }).passthrough()
+        })
+        .passthrough()
```
baktun14: I was looking in `passthrough` and it redirects to `looseObject` https://zod.dev/api?id=zlooseobject. Do we want to have the objects loose and accept unknown keys?

baktun14:
Can you add the same prettier config into the new packages please? Like this one https://github.com/akash-network/console/blob/main/packages/database/.prettierrc.js

packages/api-schemas/src/notifications.schemas.ts
ygrishajev: wait, I'm dropping these generated schemas, there is no need for them at all for now

apps/notifications/src/lib/bootstrap/bootstrapper/bootstrapper.ts
baktun14: Ok I see, makes sense.

Issue #1345, opened 2025-05-20
Console shell returns invalid url


baktun14:
![Image](https://github.com/user-attachments/assets/81924fde-e13c-40c3-b5e0-b77101ee69cf)

stalniy:
I believe this can be closed

PR #1346, opened 2025-05-20
fix(provider): websocket closure bug

baktun14:
https://github.com/akash-network/console/issues/1345

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability of WebSocket messaging to ensure the latest connection details are always used, reducing potential errors during message sending.

### DIFF
--- a/apps/deploy-web/src/hooks/useProviderWebsocket.ts
+++ b/apps/deploy-web/src/hooks/useProviderWebsocket.ts
@@ -1,4 +1,4 @@
-import { useEffect } from "react";
+import { useEffect, useRef } from "react";
 import type { Options as WebsocketOptions } from "react-use-websocket";
 import useWebSocket from "react-use-websocket";
 import type { WebSocketHook as LibWebSocketHook } from "react-use-websocket/dist/lib/types";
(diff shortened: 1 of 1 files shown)


apps/deploy-web/src/hooks/useProviderWebsocket.ts
```diff
@@ -16,6 +16,11 @@ const WS_ERRORS = {
 export function useProviderWebsocket(provider: ProviderInfo | undefined, options: ProviderWebsocketOptions): WebSocketHook {
   const chainNetwork = networkStore.useSelectedNetworkId();
   const { localCert } = useCertificate();
+  const providerRef = useRef<{ owner: string | undefined; hostUri: string | undefined }>({ owner: provider?.owner, hostUri: provider?.hostUri });
+
+  useEffect(() => {
+    providerRef.current = { owner: provider?.owner, hostUri: provider?.hostUri };
+  }, [provider]);
```
stalniy, line 23: **question(non-blocking):** I don't understand how it works. 

We use provider details only when we send a message. `sendJsonMessage` is a function not a hook, meaning React reactivity does not influence it in any way. So, using `useEffect`  and `useRef` has no sense for me unless this is some React magic

@coderabbitai any thoughts on this?

stalniy: but `handleClick` will be re-created on the next re-render and will point to a new provider reference

stalniy: how can it be? if sendJsonMessage changes, then MemoizedChild `sendMessage` prop changes and will force component to re-render

stalniy: What you say doesn't make sense unless somebody makes a mistake, like will use some memoized hooks without specifying all dependencies. But we have eslint which check this, right?

So, this solution is hack-ish, right?

stalniy: Nice... Now I see one more framework which will switch to proper observability pattern in future, knockout.js kind of observables

https://github.com/facebook/react/issues/16956

stalniy (approved):

PR #1444, opened 2025-06-07
refactor: move block prediction endpoints to modules

### LINKED ISSUE #1270: Refactor block routes
- [Blocks](https://github.com/akash-network/console/tree/main/apps/api/src/routes/v1/blocks) 
- [Predicted block height](https://github.com/akash-network/console/blob/main/apps/api/src/routes/v1/predictedBlockDate.ts) 
- [Predicted date height](https://github.com/akash-network/console/blob/main/apps/api/src/routes/v1/predictedDateHeight.ts)

jzsfkzm:
closes #1270

## Summary by CodeRabbit

- **New Features**
  - Introduced new API endpoints to predict the date of a future block by height and to estimate the block height at a future timestamp.
  - Added detailed validation and OpenAPI documentation for block prediction endpoints.
  - Enhanced block retrieval to include transactions and proposer details.
  - Added ability to fetch the latest block for improved predictions.

- **Bug Fixes**
  - Improved error handling for invalid input parameters in block prediction endpoints.

- **Tests**
  - Added comprehensive tests for block prediction endpoints, covering both valid and invalid input scenarios.


stalniy:
@jzsfkzm please make sure that this module is compilable with ts `strict: true`. To do this, add all files of this module to `tsconfig.strict.json` and run `npm run validate:types-strict`

jzsfkzm:
> @jzsfkzm please make sure that this module is compilable with ts `strict: true`. To do this, add all files of this module to `tsconfig.strict.json` and run `npm run validate:types-strict`

Oh, nice, I didn't know this, I just added the module and cleared errors.

baktun14 (dismissed):

apps/api/src/block/http-schemas/block-prediction.schema.ts
```diff
@@ -0,0 +1,37 @@
...
+  predictedHeight: z.number().openapi({ example: 10_000_000 }),
+  date: z.string().openapi({ example: "2024-01-04T18:29:28.000Z" }),
+  blockWindow: z.number().openapi({ example: defaultBlockWindow })
+});
+
+export type GetPredictedBlockDateParams = z.infer<typeof GetPredictedBlockDateParamsSchema>;
+export type GetPredictionQuery = z.infer<typeof GetPredictionQuerySchema>;
+export type GetPredictedBlockDateResponse = z.infer<typeof GetPredictedBlockDateResponseSchema>;
+export type GetPredictedDateHeightParams = z.infer<typeof GetPredictedDateHeightParamsSchema>;
+export type GetPredictedDateHeightResponse = z.infer<typeof GetPredictedDateHeightResponseSchema>;
```
stalniy: **suggestion(non-blocking):** I know that you did as in other places but better to follow cohesion principle and keep related things together. So, `GetPredictedBlockDateParamsSchema` should be written together with its inferred type

apps/api/src/block/routes/block-prediction/block-prediction.router.ts
```diff
@@ -0,0 +1,78 @@
...
+
+export const blockPredictionRouter = new OpenApiHonoHandler();
+
+blockPredictionRouter.openapi(getPredictedBlockDateRoute, async function routeGetPredictedBlockDate(c) {
+  const { height } = c.req.valid("param");
+  const { blockWindow } = c.req.valid("query");
+  const response = await container.resolve(BlockPredictionController).getPredictedBlockDate(height, blockWindow);
+
+  return c.json(response);
+});
```
stalniy: **suggestion(non-blocking):** and here as well, better to keep route description and handler together. Because then I need to `ctrl + click` to see the route description (but then there is no easy way to go back)

stalniy (dismissed):

stalniy:
> > @jzsfkzm please make sure that this module is compilable with ts `strict: true`. To do this, add all files of this module to `tsconfig.strict.json` and run `npm run validate:types-strict`
> 
> Oh, nice, I didn't know this, I just added the module and cleared errors.

Yeah, we are in process of migrating code to strict types. This is the way to do it incrementally in chunks to have a controllable transition

apps/api/src/block/http-schemas/block-prediction.schema.ts
jzsfkzm: Makes sense, done. Will follow this.

apps/api/src/block/routes/block-prediction/block-prediction.router.ts
jzsfkzm: It's done, moved them together, also maybe we should not
```
const someRoute = createRoute({...});
router.openapi(someRoute, async function routeResolver(c) {...});
```

but instead we could just drop the route inline? like this:

```
blockPredictionRouter.openapi(
createRoute({
  method: "get",
  path: "/v1/predicted-date-height/{timestamp}",
  summary: "Get the estimated height of a future date and time.",
  tags: ["Blocks"],
  request: {
    params: GetPredictedDateHeightParamsSchema,
    query: GetPredictionQuerySchema
  },
  responses: {
    200: {
      description: "Returns predicted block height",
      content: {
        "application/json": {
          schema: GetPredictedDateHeightResponseSchema
        }
      }
    },
    400: {
      description: "Invalid timestamp or block window"
    }
  }
}),
async function routeGetPredictedDateHeight(c) {
  const { timestamp } = c.req.valid("param");
  const { blockWindow } = c.req.valid("query");
  const response = await container.resolve(BlockPredictionController).getPredictedDateHeight(timestamp, blockWindow);

  return c.json(response);
}
);

```

What do you think?

stalniy: this makes sense for me, if this thing is not used anywhere else in inference logic

stalniy (dismissed):

stalniy:
@jzsfkzm you have type related issues in tests. In order to see them you need to install dependencies for apps/api only:

```ts
npm ci -w apps/api
```

jzsfkzm:
> @jzsfkzm you have type related issues in tests. In order to see them you need to install dependencies for apps/api only:
> 
> ```ts
> npm ci -w apps/api
> ```

Ah, wasn't aware of the difference, checked and fixed the type issues, thanks.

PR #1448, opened 2025-06-09
refactor: migrates another part of console-api to strict types

stalniy:
## Why

To improve project stability and get early feedback on types incompatibilities

## What

Migrates another part of console api files to strict types

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved type safety and null checks across various controllers and services to prevent runtime errors and ensure consistent API responses.
  - Adjusted validation schemas to make certain fields optional and corrected method chaining for default values.
  - Fixed property naming in pricing calculation schema for clarity.
  - Added checks to prevent errors when user or environment data is missing.

- **Refactor**
  - Updated method signatures to reflect accurate return types, especially for methods that may return null.
  - Enhanced scheduler and task definitions for stricter type enforcement and safer task handling.
  - Refined error handling logic in transaction and block data processing.

- **Chores**
  - Expanded TypeScript strict mode coverage to additional source files and directories.

### DIFF
--- a/apps/api/src/block/http-schemas/block.schema.ts
+++ b/apps/api/src/block/http-schemas/block.schema.ts
@@ -3,7 +3,7 @@ import { z } from "@hono/zod-openapi";
 const defaultLimit = 20;
 
 export const ListBlocksQuerySchema = z.object({
-  limit: z.coerce.number().min(1).max(100).default(defaultLimit).optional().openapi({
+  limit: z.coerce.number().min(1).max(100).optional().default(defaultLimit).openapi({
     type: "number",
     minimum: 1,
     maximum: 100,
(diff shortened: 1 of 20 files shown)


apps/api/src/block/http-schemas/block.schema.ts
```diff
@@ -3,7 +3,7 @@ import { z } from "@hono/zod-openapi";
 const defaultLimit = 20;
 
 export const ListBlocksQuerySchema = z.object({
-  limit: z.coerce.number().min(1).max(100).default(defaultLimit).optional().openapi({
+  limit: z.coerce.number().min(1).max(100).optional().default(defaultLimit).openapi({
```
stalniy, line 6: very likely `.optional()` is redundant in this sequence, because the input is optional and the final value is not optional (but `.default()` must go after `.optional()`)

baktun14 (approved):

PR #1460, opened 2025-06-10
refactor: migrates billing module from console api to strict types

stalniy:
## Summary by CodeRabbit

- **New Features**
	- Stripe webhook endpoint now returns a clear error message if the Stripe signature header is missing.

- **Bug Fixes**
	- Improved validation to ensure required wallet addresses and user IDs are present before processing certain actions.
	- Enhanced error handling and type safety across billing, wallet, and provider services to prevent runtime errors.

- **Refactor**
	- Streamlined Stripe customer creation and checkout session logic for improved reliability.
	- Updated wallet and schema definitions to allow nullable user IDs and addresses.
	- Adjusted configuration sourcing for Stripe services to use centralized settings.
	- Refined transaction execution and batch processing for better robustness.
	- Reworked wallet data handling to explicitly manage trial status.

- **Chores**
	- Enabled strict type checking for all billing-related files to improve code quality.

### DIFF
--- a/apps/api/src/billing/repositories/user-wallet/user-wallet.repository.ts
+++ b/apps/api/src/billing/repositories/user-wallet/user-wallet.repository.ts
@@ -1,5 +1,4 @@
 import { and, count, eq, inArray, lte } from "drizzle-orm";
-import pick from "lodash/pick";
 import { singleton } from "tsyringe";
 
 import { ApiPgDatabase, ApiPgTables, InjectPg, InjectPgTable } from "@src/core/providers";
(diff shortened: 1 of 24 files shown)


ygrishajev (dismissed):

apps/api/src/billing/repositories/user-wallet/user-wallet.repository.ts
```diff
@@ -108,6 +113,12 @@ export class UserWalletRepository extends BaseRepository<ApiPgTables["UserWallet
   }
 
   toPublic(output: UserWalletOutput): UserWalletPublicOutput {
-    return pick(output, ["id", "userId", "address", "creditAmount", "isTrialing"]);
+    return {
```
baktun14, line 116: question(non-blocking): pick isn't strict friendly?

baktun14 (dismissed):

apps/api/src/billing/repositories/user-wallet/user-wallet.repository.ts
stalniy: In this particular case, I needed  `isTrialing` to be always a `boolean` but `UserWalletOutput.isTrialing` is nullable for some reason. So, we have 3 (true, false, null) states for `isTrialing` but on API response level we have only 2 (true and false)

But to be honest, I don't understand the purpose of using `pick` in era of object de-structuring. It could be written as this:

```ts
function toPublic({ id, userId, address, creditAmount, isTrialing }: UserWalletOutput): UserWalletPublicOutput {
  return { id, userId, address, creditAmount, isTrialing }
}
```

So, I don't see the value which `pick` can bring to modern JS

PR #1468, opened 2025-06-11
refactor: move /v1/addresses/* to modules

### LINKED ISSUE #1267: Refactor addresses routes
[Addresses](https://github.com/akash-network/console/tree/main/apps/api/src/routes/v1/addresses)

jzsfkzm:
closes #1267

## Summary by CodeRabbit

- **New Features**
  - Introduced a new Address API with endpoints for detailed address info and paginated transactions.
  - Enhanced data schemas to provide comprehensive address assets, delegations, rewards, commissions, redelegations, and transactions.
  - Added integration with Cosmos blockchain services for real-time balance and staking data.
  - Improved validator data retrieval including avatar URLs for enriched delegation info.
- **Bug Fixes**
  - Strengthened input validation for address and transaction parameters with improved error responses.
- **Tests**
  - Added comprehensive functional tests covering address details, transactions, and deployments with validation scenarios.
- **Chores**
  - Refactored seeder utilities for streamlined test data creation and database insertion.
  - Removed deprecated address and transaction route handlers and related types for cleaner codebase.


apps/api/src/address/services/address/address.service.ts
```diff
@@ -0,0 +1,166 @@
...
+    const [balancesResponse, delegationsResponse, rewardsResponse, redelegationsResponse, latestTransactions] = await Promise.all([
+      this.cosmosHttpService.getBankBalancesByAddress(address),
+      this.cosmosHttpService.getStakingDelegationsByAddress(address),
+      this.cosmosHttpService.getDistributionDelegatorsRewardsByAddress(address),
+      this.cosmosHttpService.getStakingDelegatorsRedelegationsByAddress(address),
+      this.transactionService.getTransactionsByAddress({ address, skip: 0, limit: 5 })
+    ]);
+
+    const validatorsFromDb = await Validator.findAll();
+    const validatorFromDb = validatorsFromDb.find(v => v.accountAddress === address);
```
stalniy: **question(blocking):** Why don't you do it via SQL where?

```ts
await Validator.findAll({ where: { accountAddress } });
```

Also I see that we have `ValidatorRepository` in apps/api/src/validator/repositories/validator/validator.repository.ts, any reason why you can't use it?

**P.S.:** eventually we want to replace sequelize with drizzle, so would be nice to keep all sequelize in repositories

apps/api/test/seeders/address-refererence.seeder.ts
```diff
@@ -0,0 +1,18 @@
+import { AddressReference } from "@akashnetwork/database/dbSchemas/base";
+import { faker } from "@faker-js/faker";
+import type { CreationAttributes } from "sequelize";
+
+export class AddressReferenceSeeder {
```
stalniy: **thought(non-blocking):** we have a discussion [here](https://github.com/akash-network/console/discussions/914) which suggests to use functions for seeders instead of classes because classes doesn't bring any benefits in this case. FYI

apps/api/test/seeders/address-refererence.seeder.ts
```diff
@@ -0,0 +1,18 @@
...
+      transactionId: input.transactionId || faker.string.uuid(),
+      messageId: input.messageId || faker.string.uuid(),
+      address: input.address || faker.string.hexadecimal({ length: 64 }),
+      type: input.type || faker.string.alphanumeric({ length: 10 })
+    };
+  }
+
+  static async createInDatabase(input: Partial<CreationAttributes<AddressReference>> = {}): Promise<AddressReference> {
+    return await AddressReference.create(AddressReferenceSeeder.create(input));
+  }
```
stalniy: **thought(non-blocking):** this pattern doesn't work well when we need to create entity together with its relations

apps/api/test/seeders/akash-message.seeder.ts
```diff
@@ -29,12 +16,23 @@ export class MessageSeeder {
...
       },
       input
     );
   }
+
+  static async createInDatabase(input: Partial<CreationAttributes<AkashMessage>> = {}): Promise<AkashMessage> {
+    try {
+      return await AkashMessage.create(AkashMessageSeeder.create(input));
+    } catch (error) {
+      console.error(error);
```
stalniy: **question(non-blocking):** is this to improve visibility of the error in tests logs?

stalniy:
In general, I would advice starting refactoring from adding functional tests to routes which you want to refactor (1 PR). And in another PR, you refactor code and do not change tests. Then everybody understands that the functionality is working as expected.

Also reviewing tests is kind of easier because it doesn't affect production. So as far as you have enough coverage, PR with tests will be approved quicker. It's also brings benefits for the follow up PR because we will need to review less and eventually it will land quicker to the codebase

apps/api/test/seeders/akash-message.seeder.ts
jzsfkzm: Oftentimes when building test data I find myself trying and failing to create some entity in the db, where creation fails because there is a mandatory link to some other entity I did not create beforehand. Then I found myself adding this try-catch block and removing it when committing my changes, multiple times. Then I was like it's ok to have this.
What do you think?

apps/api/test/seeders/address-refererence.seeder.ts
jzsfkzm: True that, we need to precreate a deployment and a deploymentGroup before creating a Lease, for example. But is that a problem?

baktun14:
> In general, I would advice starting refactoring from adding functional tests to routes which you want to refactor (1 PR). And in another PR, you refactor code and do not change tests. Then everybody understands that the functionality is working as expected.
> 
> Also reviewing tests is kind of easier because it doesn't affect production. So as far as you have enough coverage, PR with tests will be approved quicker. It's also brings benefits for the follow up PR because we will need to review less and eventually it will land quicker to the codebase

Agreed, but the goal of this task is to move from the old architecture to the new module one. So to avoid making these changes too cumbersome, it's okay if tests are not added right away. Although, yes if there's changes to the logic, better to add tests now.

jzsfkzm:
> > In general, I would advice starting refactoring from adding functional tests to routes which you want to refactor (1 PR). And in another PR, you refactor code and do not change tests. Then everybody understands that the functionality is working as expected.
> > Also reviewing tests is kind of easier because it doesn't affect production. So as far as you have enough coverage, PR with tests will be approved quicker. It's also brings benefits for the follow up PR because we will need to review less and eventually it will land quicker to the codebase
> 
> Agreed, but the goal of this task is to move from the old architecture to the new module one. So to avoid making these changes too cumbersome, it's okay if tests are not added right away. Although, yes if there's changes to the logic, better to add tests now.

My method of operation for these tasks was to
- read, and try and understand what the code is doing
- create a series of tests, not just happy paths but error cases too
- refactor the code and make sure tests are still green

So, the suggestion from @stalniy perfectly fits this, apparently. I'll create the upcoming PRs in this manner.

baktun14:
> > > In general, I would advice starting refactoring from adding functional tests to routes which you want to refactor (1 PR). And in another PR, you refactor code and do not change tests. Then everybody understands that the functionality is working as expected.
> > > Also reviewing tests is kind of easier because it doesn't affect production. So as far as you have enough coverage, PR with tests will be approved quicker. It's also brings benefits for the follow up PR because we will need to review less and eventually it will land quicker to the codebase
> > 
> > 
> > Agreed, but the goal of this task is to move from the old architecture to the new module one. So to avoid making these changes too cumbersome, it's okay if tests are not added right away. Although, yes if there's changes to the logic, better to add tests now.
> 
> My method of operation for these tasks was to
> 
> * read, and try and understand what the code is doing
> * create a series of tests, not just happy paths but error cases too
> * refactor the code and make sure tests are still green
> 
> So, the suggestion from @stalniy perfectly fits this, apparently. I'll create the upcoming PRs in this manner.

Sounds good!

apps/api/test/seeders/akash-message.seeder.ts
stalniy: It's fine for now, I just have a feeling that this is hacky solution to something what doesn't display error properly: jest? too much logs during test execution? 

That's why I'm looking for a reason. Maybe it makes sense to fix this from different side

stalniy:
@jzsfkzm my comment about splitting refactoring into 2 pieces (tests and then actual refactoring) was more driven by desire to simplify and speed up code review process. 

If it's possible to write functional tests that can survive refactoring without changes (minor changes), then if you submit them alone, we will very likely approve that quickly. Because after all tests doesn't change behavior of the app, so it's safe to skim through such PR and later we can pay more attention to the actual refactoring. I guess this should reduce cognitive load, amount of comments in PR, time to merge.

We can try and see how it goes. Thanks for taking this into account for the future PRs

apps/api/src/address/services/address/address.service.ts
jzsfkzm: Right, changed this, resolving.

apps/api/test/seeders/akash-message.seeder.ts
jzsfkzm: Right, removed this try-catch, as it's indeed hacky.

baktun14 (dismissed):

jzsfkzm:
@baktun14 pushed one more change so tests are not breaking now (added a line cleaning a nock in the gpu tests), can you re-approve please?

Issue #1348, opened 2025-05-21
Usage and Analytics


baktun14:
- [x] Api endpoint that returns daily spent in uakt, uusdc and total usd
Here's the sql query to fetch the daily spent per address:
```
WITH current_block AS (
    SELECT MAX(height) as height 
    FROM block
),
daily_leases AS (
    SELECT 
        d.date,
        l.owner,
        l.denom,
        l.price,
        l."createdHeight",
        CASE 
            WHEN l.id IS NULL THEN 0
            ELSE LEAST(d."lastBlockHeightYet", COALESCE(l."closedHeight", l."predictedClosedHeight")) - 
                 GREATEST(d."firstBlockHeight", l."createdHeight")
        END as blocks_in_day,
        d."aktPrice"
    FROM day d
    LEFT JOIN lease l ON 
        l.owner = 'akash18andxgtd6r08zzfpcdqg9pdr6smks7gv76tyt6' AND -- INPUT ADDRESS HERE
        l."createdHeight" < d."lastBlockHeightYet" AND 
        COALESCE(l."closedHeight", l."predictedClosedHeight") > d."firstBlockHeight"
),
daily_costs AS (
    SELECT 
        date,
        ROUND(CAST(SUM(CASE 
            WHEN denom = 'uakt' 
            THEN blocks_in_day * price / 1000000.0
            ELSE 0 
        END) as numeric), 2) as daily_akt_spent,
        ROUND(CAST(SUM(CASE 
            WHEN denom = 'uusdc' 
            THEN blocks_in_day * price / 1000000.0
            ELSE 0 
        END) as numeric), 2) as daily_usdc_spent,
        ROUND(CAST(SUM(CASE 
            WHEN denom = 'uakt' 
            THEN blocks_in_day * price * "aktPrice" / 1000000.0
            WHEN denom = 'uusdc' 
            THEN blocks_in_day * price / 1000000.0
            ELSE 0 
        END) as numeric), 2) as daily_usd_spent
    FROM daily_leases
    GROUP BY date
)
SELECT 
    d.date::DATE AS "Date",
    COUNT(dl.owner) AS "Active Leases",
    COALESCE(dc.daily_akt_spent, 0) as "Daily AKT Spent",
    COALESCE(SUM(dc.daily_akt_spent) OVER (ORDER BY d.date), 0) as "Total AKT Spent",
    COALESCE(dc.daily_usdc_spent, 0) as "Daily USDC Spent",
[...]

PR #1483, opened 2025-06-13
feat(billing): add usage endpoint

domhhv:
Updates #1348 

## Summary by CodeRabbit

- **New Features**
  - Introduced API endpoints to retrieve detailed usage history and aggregated usage statistics for wallet addresses, including daily and cumulative spending in AKT, USDC, and USD.
  - Added input validation for addresses and date ranges, with clear error responses for invalid queries.
  - Added a new method to count distinct leases by owner for enhanced usage statistics.

- **Bug Fixes**
  - Improved address generation in test data to ensure valid Bech32-encoded addresses.

- **Tests**
  - Added comprehensive unit and functional tests for usage history and statistics endpoints, including edge cases and error handling.
  - Introduced new data seeders to support robust test scenarios.


apps/api/src/billing/repositories/usage/usage.repository.ts
```diff
@@ -0,0 +1,111 @@
...
+
+interface BillingUsageRawResult {
+  Date: string;
+  "Active Leases": number;
+  "Daily AKT Spent": number;
+  "Total AKT Spent": number;
+  "Daily USDC Spent": number;
+  "Total USDC Spent": number;
+  "Daily USD Spent": number;
+  "Total USD Spent": number;
```
ygrishajev: Non-blocking: this notation looks weird as internal query. I'd suggest using smth more usual like snake/camel case.

apps/api/src/billing/services/usage/usage.service.spec.ts
```diff
@@ -0,0 +1,282 @@
...
+import { UsageService } from "./usage.service";
+
+import { AkashAddressSeeder } from "@test/seeders/akash-address.seeder";
+
+describe(UsageService.name, () => {
+  let service: UsageService;
+  let usageRepository: jest.Mocked<UsageRepository>;
+  let leaseRepository: jest.Mocked<LeaseRepository>;
+
+  beforeEach(() => {
```
ygrishajev: Pls use setup function approach, avoid jest hooks. This is smth we follow for all cases.

apps/api/src/billing/services/usage/usage.service.spec.ts
```diff
@@ -0,0 +1,282 @@
...
+
+    service = new UsageService(usageRepository, leaseRepository);
+  });
+
+  describe("getHistory", () => {
+    const address = AkashAddressSeeder.create();
+    const startDate = "2024-01-01";
+    const endDate = "2024-01-31";
+
+    const mockUsageData = [
```
ygrishajev: This looks like a case for some seeder with faker.

apps/api/src/billing/services/usage/usage.service.spec.ts
```diff
@@ -0,0 +1,242 @@
...
+        dailyAktSpent: 2.1,
+        totalAktSpent: 10.8,
+        dailyUsdcSpent: 1.0,
+        totalUsdcSpent: 4.75,
+        dailyUsdSpent: 3.1,
+        totalUsdSpent: 15.55
+      })
+    ];
+
+    const usageRepository = {
```
ygrishajev: There are utilities to avoid this type casting and easy mocking

https://github.com/akash-network/console/blob/main/apps/api/test/services/stub.ts

jest-mock-extended https://github.com/akash-network/console/blob/main/apps/api/src/healthz/services/healthz/healthz.service.spec.ts

domhhv: Thanks, used `stub` utility

apps/api/src/billing/services/usage/usage.service.spec.ts
domhhv: Created and used `BillingUsageSeeder` in such cases

apps/api/src/billing/services/usage/usage.service.spec.ts
domhhv: Done

apps/api/src/billing/repositories/usage/usage.repository.ts
domhhv: +1
Used camelCase

apps/api/src/billing/controllers/usage/usage.controller.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+      averageLeasesPerDay: 1.5
+    };
+
+    service.getHistoryStats.mockResolvedValue(output);
+
+    const result = await controller.getHistoryStats(address, startDate, endDate);
+
+    expect(service.getHistoryStats).toHaveBeenCalledWith(address, startDate, endDate);
+    expect(result).toEqual(output);
+  });
```
stalniy: thought(non-blocking): based on the implementation, I see that controller just forwards the call to underlying service. So, there is no logic to test and I would not add UNIT tests like this at all because they give very little value.

Each time we refactor controller, we also will need to refactor this test. Even in case of minor changes, like adding additional field to returned result.

So, I'd recommend not to do this in future. Controllers are the best covered with functional tests, then we have higher confidence that this thing work. And it will make our lives easier, if some day we will decide to migrate to Nestjs

apps/api/src/billing/http-schemas/usage.schema.ts
```diff
@@ -0,0 +1,113 @@
...
+    startDate: z.string().date().optional().openapi({
+      description: "Start date (YYYY-MM-DD). Defaults to 30 days before endDate",
+      example: "2024-01-01"
+    }),
+    endDate: z
+      .string()
+      .date()
+      .default(() => new Date().toISOString().split("T")[0])
+      .openapi({
+        description: "End date (YYYY-MM-DD). Defaults to today",
```
stalniy: **suggestion(non-blocking):** I'd note that "today" is by UTC 23:59:59

apps/api/src/billing/http-schemas/usage.schema.ts
```diff
@@ -0,0 +1,113 @@
...
+    }
+  )
+  .refine(
+    data => {
+      return isValidBech32Address(data.address, "akash");
+    },
+    {
+      message: "Invalid Akash address format"
+    }
+  );
```
stalniy: question(non-blocking): is this possible to connect this with corresponding field?
 
Like this:
```ts
address: z.string().min(1).openapi({
      description: "The wallet address to get billing and usage data for",
      example: "akash18andxgtd6r08zzfpcdqg9pdr6smks7gv76tyt6"
    }).refine(
    data => {
      return isValidBech32Address(data.address, "akash");
    },
    {
      message: "Invalid Akash address format"
    }
  )
```

apps/api/src/billing/services/usage/usage.service.ts
```diff
@@ -0,0 +1,44 @@
...
+    private readonly leaseRepository: LeaseRepository
+  ) {}
+
+  async getHistory(address: string, startDate: string, endDate: string): Promise<UsageHistoryResponse> {
+    return await this.usageRepository.getHistory(address, startDate, endDate);
+  }
+
+  async getHistoryStats(address: string, startDate: string, endDate: string): Promise<UsageHistoryStats> {
+    const historyData = await this.usageRepository.getHistory(address, startDate, endDate);
+    const totalLeases = await this.leaseRepository.countByOwner(address);
```
stalniy: suggestion(non-blocking): it seems this can be done in parallel

apps/api/test/seeders/akash-address.seeder.ts
```diff
@@ -1,7 +1,14 @@
...
 export class AkashAddressSeeder {
   static create(): string {
-    return `akash${faker.string.alphanumeric({ length: 39 })}`;
+    const addressData = new Uint8Array(20);
+
+    for (let i = 0; i < 20; i++) {
+      addressData[i] = faker.number.int({ min: 0, max: 255 });
+    }
+
+    return toBech32("akash", addressData);
```
stalniy, line 12: **praise:** 👍 thanks

apps/api/src/billing/services/usage/usage.service.spec.ts
```diff
@@ -0,0 +1,243 @@
...
+
+        expect(result.totalSpent).toBe(427.22);
+        expect(result.averagePerDay).toBe(213.61);
+        expect(result.totalLeases).toBe(1000);
+        expect(result.averageLeasesPerDay).toBe(500.0);
+      });
+    });
+  });
+
+  function setup(usageData?: BillingUsageRawResult[], totalLeases?: number) {
```
stalniy: suggestion(non-blocking): better to have single `input` parameter and inline its type:

```ts
function setup(input?: {
  usageData?: BillingUsageRawResult[], 
  totalLeases?: number
}) {}
```

apps/api/src/billing/services/usage/usage.service.spec.ts
```diff
@@ -0,0 +1,243 @@
...
+    });
+
+    const leaseRepository = stub<LeaseRepository>({
+      countByOwner: jest.fn()
+    });
+
+    const service = new UsageService(usageRepository, leaseRepository);
+
+    usageRepository.getHistory.mockResolvedValue(mockUsageData);
+    leaseRepository.countByOwner.mockResolvedValue(leasesCount);
```
stalniy: suggestion(non-blocking): Please inline this in `stub` creation:

```ts
    const usageRepository = stub<UsageRepository>({
      getHistory: jest.fn().moceResolvedValue(mockUsageData)
    });
```

apps/api/src/billing/services/usage/usage.service.spec.ts
```diff
@@ -0,0 +1,243 @@
...
+        dailyAktSpent: 2.1,
+        totalAktSpent: 10.8,
+        dailyUsdcSpent: 1.0,
+        totalUsdcSpent: 4.75,
+        dailyUsdSpent: 3.1,
+        totalUsdSpent: 15.55
+      })
+    ];
+
+    const usageRepository = stub<UsageRepository>({
```
stalniy: suggestion(non-blocking): I think `stub` is an early helper which was added before `jest-mock-extended`. As we now have jest-mock-extended, we can use its `mock` helper

apps/api/test/seeders/day.seeder.ts
```diff
@@ -17,6 +17,11 @@ export class DaySeeder {
 
   static async createInDatabase(overrides: Partial<CreationAttributes<Day>> = {}): Promise<Day> {
     const seed = DaySeeder.create(overrides);
-    return await Day.create(seed);
+    try {
+      return await Day.create(seed);
+    } catch (error) {
+      console.error("Error creating Day in database:", error);
```
stalniy: question(non-blocking): why do you need this?

apps/api/src/billing/repositories/usage/usage.repository.ts
```diff
@@ -0,0 +1,111 @@
...
+export class UsageRepository {
+  async getHistory(address: string, startDate: string, endDate: string) {
+    const query = `
+      WITH date_range AS (
+        SELECT generate_series(
+          :startDate::date,
+          :endDate::date,
+          '1 day'::interval
+        )::date as date
+      ),
```
stalniy, line 27: issue(blocking): this SQL query contains the main logic responsible to calculating and gathering statistic, but there are no tests that covers it.

It's important to ensure this query properly work, otherwise it will be hard to change it

stalniy (changes requested):

apps/api/src/billing/http-schemas/usage.schema.ts
stalniy: Actually we have a schema for akash address -> `AkashAddressSchema`: https://github.com/akash-network/console/blob/main/apps/api/src/utils/schema.ts#L5

apps/api/src/billing/controllers/usage/usage.controller.spec.ts
domhhv: I think so too. I removed it for now.

apps/api/src/billing/http-schemas/usage.schema.ts
domhhv: Added to description

apps/api/src/billing/http-schemas/usage.schema.ts
domhhv: Nice one - reused

apps/api/src/billing/services/usage/usage.service.spec.ts
domhhv: Makes sense, refactored

apps/api/src/billing/services/usage/usage.service.spec.ts
domhhv: Done

apps/api/src/billing/services/usage/usage.service.spec.ts
domhhv: Done

apps/api/src/billing/services/usage/usage.service.ts
domhhv: Right, it surely can - used Promise.all

apps/api/test/seeders/akash-address.seeder.ts
domhhv: Thank you!

apps/api/test/seeders/day.seeder.ts
domhhv: Ouch, we don't. This was some debugging I added when my functional tests were failing and then totally forgot to remove. Thanks for the catch!

apps/api/src/billing/repositories/usage/usage.repository.ts
domhhv: Sounds good, I've created usage.repository.spec.ts which covers absent and present leases with various denominations. I tried to base it on an existing spec - provider.repository.spec.ts - which appears to be the only spec for a repository and is quite recent. Both specs use a straightforward setup function and an afterEach hook. I'm open to any suggestions if the new one could be clearer or more comprehensive.

apps/api/src/billing/repositories/usage/usage.repository.spec.ts
```diff
@@ -0,0 +1,327 @@
...
+    let createdDeploymentIds: string[] = [];
+    let createdDeploymentGroupIds: string[] = [];
+    let createdProviderAddresses: string[] = [];
+
+    afterEach(async () => {
+      await chainDb.models.lease.destroy({
+        where: {
+          id: { [Op.in]: createdLeaseIds }
+        }
+      });
```
stalniy, line 34: **suggestion(non-blocking):** maybe for time being, we could create 

```ts
function createEntityTracker<T extends Model>(options: { 
  create: (attrs: CreationAttributes<T>) => Promise<T>, 
  destroy: (keys: unknown[]) => Promise<void> 
}) {
  const primaryKeys = new Set<unknown>()
  return {
    async create<T>(attrs: CreationAttributes<T>): Promise<T> {
       const entity = await options.create(attrs);
       primaryKeys.add(getSomehowPrimaryKey(entity));
       return entity;
    },
    async destroy() {
      await options.destroy(primaryKeys);
    }
  }
}
```

stalniy: ok, never mind. Need to check libraries like fishery and configure long term solution for this

stalniy (approved):

apps/api/src/billing/controllers/usage/usage.controller.ts
```diff
@@ -0,0 +1,17 @@
+import { singleton } from "tsyringe";
+
+import { UsageHistoryResponse, UsageHistoryStats } from "@src/billing/http-schemas/usage.schema";
+import { UsageService } from "@src/billing/services/usage/usage.service";
+
+@singleton()
+export class UsageController {
+  constructor(private readonly usageService: UsageService) {}
+
+  async getHistory(address: string, startDate: string, endDate: string): Promise<UsageHistoryResponse> {
```
baktun14, line 10: We should memoize these endpoints to something like few seconds, maybe 15 seconds

PR #1494, opened 2025-06-16
feat(auth): uses user aware feature flagging in console web

ygrishajev:
## Summary by CodeRabbit

- **New Features**
  - Feature flag evaluations now consider individual user context, enabling more personalized feature rollouts.
- **Refactor**
  - Adjusted provider hierarchy to wrap wallet-related providers with feature flag logic, improving context accuracy.
  - Enhanced feature flag services to accept richer context objects, supporting more granular flag checks.
  - Improved server-side feature flag checks by verifying user authentication before evaluation.
- **Bug Fixes**
  - Ensured feature flag checks are only performed for authenticated users and return appropriate responses for unauthenticated access.
- **Tests**
  - Added and updated tests to verify user context is correctly passed to feature flag providers and services.

### DIFF
--- a/apps/deploy-web/src/context/FlagProvider/FlagProvider.tsx
+++ b/apps/deploy-web/src/context/FlagProvider/FlagProvider.tsx
@@ -1,7 +1,21 @@
 import { FlagProvider as FlagProviderOriginal } from "@unleash/nextjs";
 
 import { browserEnvConfig } from "@src/config/browser-env.config";
+import { useUser } from "@src/hooks/useUser";
+import type { FCWithChildren } from "@src/types/component";
 
 const DummyFlagProvider: typeof FlagProviderOriginal = props => <>{props.children}</>;
 
-export const FlagProvider = browserEnvConfig.NEXT_PUBLIC_UNLEASH_ENABLE_ALL ? DummyFlagProvider : FlagProviderOriginal;
+const COMPONENTS = {
+  FlagProvider: FlagProviderOriginal,
+  useUser
+};
+
+export type Props = { components?: typeof COMPONENTS };
+
+export const UserAwareFlagProvider: FCWithChildren<Props> = ({ children, components: c = COMPONENTS }) => {
+  const user = c.useUser();
+  return <c.FlagProvider config={{ context: { userId: user?.id } }}>{children}</c.FlagProvider>;
+};
+
+export const FlagProvider = browserEnvConfig.NEXT_PUBLIC_UNLEASH_ENABLE_ALL ? DummyFlagProvider : UserAwareFlagProvider;
--- a/apps/deploy-web/src/services/feature-flag/feature-flag.service.spec.ts
+++ b/apps/deploy-web/src/services/feature-flag/feature-flag.service.spec.ts
@@ -18,13 +18,15 @@ describe(FeatureFlagService.name, () => {
     it("evaluates flag and returns true", async () => {
       const { service, unleash, flagsClient } = setup();
       flagsClient.isEnabled.mockReturnValue(true);
+      const flag = "feature-x";
+      const context = { sessionId: "abc123" };
 
-      const result = await service.getFlag("feature-x", "abc123");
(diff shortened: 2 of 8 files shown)


apps/deploy-web/src/services/feature-flag/feature-flag.service.spec.ts
```diff
@@ -1,25 +1,24 @@
...
-import type { MockProxy } from "jest-mock-extended";
 import { mock } from "jest-mock-extended";
 import type { GetServerSidePropsContext } from "next";
 import type { UnleashClient } from "unleash-proxy-client";
 
 import type { ServerEnvConfig } from "@src/config/env-config.schema";
 import { FeatureFlagService } from "./feature-flag.service";
 
-describe(FeatureFlagService.name, () => {
+describe("FeatureFlagService.name", () => {
```
stalniy: Probably typo

```suggestion
describe(FeatureFlagService.name, () => {
```

apps/deploy-web/src/services/feature-flag/feature-flag.service.ts
```diff
@@ -30,17 +33,28 @@ export class FeatureFlagService {
...
+  async isEnabledForCtx(name: string, ctx: GetServerSidePropsContext, options?: { registeredOnly?: boolean; userId?: string }): Promise<boolean> {
     if (this.config.NEXT_PUBLIC_UNLEASH_ENABLE_ALL) return true;
 
-    const sessionId = this.extractSessionId(ctx);
+    const userId = options?.userId || (await this.extractUserId(ctx));
 
-    return await this.getFlag(name, sessionId);
+    if (!userId && options?.registeredOnly) {
+      return false;
+    }
```
stalniy: thought(non-blocking): I think this is out of flags service responsibility because feature by itself should be protected by authentication mechanism. If feature toggle is only for registered users, unregistered users won't be able to bypass authentication system and as a result will never interact with registeredOnly feature flag.

So, here we have a mix of authentication logic and feature toggle logic

apps/deploy-web/src/services/feature-flag/index.ts
```diff
@@ -1,6 +1,7 @@
+import { getSession } from "@auth0/nextjs-auth0";
 import * as unleashModule from "@unleash/nextjs";
 
 import { serverEnvConfig } from "@src/config/server-env.config";
 import { FeatureFlagService } from "@src/services/feature-flag/feature-flag.service";
 
-export const featureFlagService = new FeatureFlagService(unleashModule, serverEnvConfig);
+export const featureFlagService = new FeatureFlagService(unleashModule, getSession, serverEnvConfig);
```
stalniy: thought(non-blocking): would be nice to mark ServerSide only services with some comment to make it explicit and easy to understand thanks to IDE hints:

```suggestion
/**
 * Server side only service
 */
export const featureFlagService = new FeatureFlagService(unleashModule, getSession, serverEnvConfig);
```

stalniy (dismissed):

apps/deploy-web/src/context/FlagProvider/FlagProvider.tsx
```diff
@@ -1,7 +1,21 @@
...
 
 const DummyFlagProvider: typeof FlagProviderOriginal = props => <>{props.children}</>;
 
-export const FlagProvider = browserEnvConfig.NEXT_PUBLIC_UNLEASH_ENABLE_ALL ? DummyFlagProvider : FlagProviderOriginal;
+const COMPONENTS = {
+  FlagProvider: FlagProviderOriginal,
+  useUser
+};
+
+export type Props = { components: typeof COMPONENTS };
```
stalniy: **issue**: `components` should be optional

apps/deploy-web/src/services/feature-flag/feature-flag.service.ts
ygrishajev: yea, I agreed

stalniy (approved):

PR #1498, opened 2025-06-17
refactor: move /templates endpoints to a module

### LINKED ISSUE #1273: Refactor templates routes
[Templates](https://github.com/akash-network/console/tree/main/apps/api/src/routes/v1/templates)

jzsfkzm:
closes #1273

## Summary by CodeRabbit

* **New Features**
  * Introduced new API endpoints under `/v1` to retrieve deployment templates, including full lists, filtered lists, and individual templates by ID.
  * Added detailed template metadata with categories, summaries, deployment instructions, and configuration options.
  * Implemented caching and concurrency controls to optimize template data fetching from multiple GitHub repositories.
  * Provided OpenAPI documentation and schema validation for all template-related endpoints.

* **Refactor**
  * Modularized template logic into dedicated services, controllers, and routers for better maintainability and scalability.
  * Replaced previous template fetching and caching mechanisms with a new service-based architecture.

* **Bug Fixes**
  * Enhanced error handling for template retrieval, returning proper 404 responses when templates are not found.

* **Tests**
  * Added extensive functional and cache tests covering template API endpoints and cache file generation.

* **Chores**
  * Extended TypeScript configuration to include new template source files.
  * Added utility functions for safe JSON parsing and markdown processing.


apps/api/src/template/services/template-gallery/template-gallery.service.ts
```diff
@@ -0,0 +1,106 @@
...
+import { injectable } from "tsyringe";
+
+import { Memoize } from "@src/caching/helpers";
+import type { Category, FinalCategory, Template } from "@src/template/types/template";
+import { dataFolderPath } from "@src/utils/constants";
+import { REPOSITORIES, TemplateFetcherService } from "../template-fetcher/template-fetcher.service";
+import { TemplateProcessorService } from "../template-processor/template-processor.service";
+
+const generatingTasks: Record<string, Promise<Category[]> | null> = {};
+let lastServedData: FinalCategory[] | null = null;
```
stalniy: **suggestion(blocking):** better to move it under service properties. In real app, this service will be a singleton, so eventually we will have the same results BUT additionally we will have improvement on testability side

apps/api/test/functional/templates.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+
+const fakeHeaders = {
+  headers: {
+    "x-ratelimit-remaining": 100
+  }
+};
+
+env.GITHUB_PAT = "ghp_1234567890";
+
+jest.mock("@octokit/rest", () => {
```
stalniy, line 17: suggestion(non-blocking): avoid `jest.mock` usage in DI based systems. If you still need to use`jest.mock` in DI system, it means that some service doesn't properly encapsulate its dependencies

apps/api/test/functional/templates.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+              ...fakeHeaders
+            });
+          })
+        }
+      }
+    }))
+  };
+});
+
+describe("Templates API", () => {
```
stalniy, line 39: thought(non-blocking): in this case testing API doesn't bring much value. The core functionality here is handled by:

```ts
TemplateGalleryService
TemplateFetcherService,
TemplateProcessorService
```

And we should always concentrate our attention on what entity does the most work because then we can apply Pareto principle and cover in details only important things which has higher complexity or higher likelihood of being broken

stalniy (changes requested):

apps/api/test/functional/templates.spec.ts
jzsfkzm: My intention was to avoid real Octokit requests here.
Actual content on GitHub is subject to change for which it is hard to write `expect` lines. Also it is a rate limited API, and tests are eating that capacity quickly. I think faking an external resource is ok, isn't it? Can you suggest another way @stalniy?

apps/api/src/template/services/template-gallery/template-gallery.service.ts
jzsfkzm: Thanks, moved it under.

apps/api/test/functional/templates.spec.ts
jzsfkzm: Well, yeah, current tests are inducing the whole flow, but maybe we should not do that. What would be the right way of doing thing here?

Now that I think about it more, here's an idea:
- We should keep template data in PG tables. API should simply read that data. Easy to test.
- The script reading the data from GitHub should write data to PG instead of local files. Also easy to test.

What do you think @stalniy?

apps/api/test/functional/templates.spec.ts
stalniy: faking external resource is the right way to test the service. The actual issue is how we do it. `jest.mock` is a hacky way to do it. You can provide a mock for github service via DI container.

To simplify its definition, you can pass it as an optional dependency with default value which equal to `Octocit` instance. You can take a look at this example with feature flags -> https://github.com/akash-network/console/blob/main/apps/api/src/core/services/feature-flags/feature-flags.service.ts#L19

apps/api/test/functional/templates.spec.ts
stalniy: I think we should not do it at least for now because I don't see a usecase. We prebuild templates at docker build time and they don't change often. There is [npm script](https://github.com/akash-network/console/blob/main/apps/api/package.json#L14) which is called by [npm run build](https://github.com/akash-network/console/blob/main/apps/api/package.json#L13).

You can read some of the ideas about testing in our [Testing strategy](https://github.com/akash-network/console/discussions/1130).

apps/api/scripts/buildAkashTemplatesCache.ts
```diff
@@ -9,7 +9,7 @@ if (!githubPAT) {
   process.exit(1);
 }
 
-getTemplateGallery({ githubPAT, dataFolderPath: "./dist/.data" }).catch(err => {
+getTemplateGallery().catch(err => {
```
stalniy: **issue:** the build is failing because it tries to prebuild templates and cannot because it requires full application configuration to be provided. But here we just need `GITHUB_PAT`. That's why this function accepted `githubPAT` parameter

apps/api/test/functional/templates.spec.ts
jzsfkzm: > I think we should not do it at least for now because I don't see a usecase. We prebuild templates at docker build time and they don't change often. There is [npm script](https://github.com/akash-network/console/blob/main/apps/api/package.json#L14) which is called by [npm run build](https://github.com/akash-network/console/blob/main/apps/api/package.json#L13).
> 
> You can read some of the ideas about testing in our [Testing strategy](https://github.com/akash-network/console/discussions/1130).

In this case my suggestion for separating the code still stands. If we have the cache files pregenerated at build time, I think the router-controller-service structure should just be responsible for reading those files (as if it was a DB), and should not know anything about pregenerating files. And the code writing those files should not be moved to modules at all maybe? It could remain the same as it ever was. So full application configuration is not needed there. Now I wish we discussed this earlier.

apps/api/src/services/external/templates/template-fetcher.service.ts
```diff
@@ -0,0 +1,341 @@
...
+  "akash-linuxserver": {
+    repoOwner: "cryptoandcoffee",
+    repoName: "akash-linuxserver",
+    mainBranch: "main"
+  }
+};
+
+export class TemplateFetcherService {
+  private githubRequestsRemaining: string | null | undefined = null;
+  private _octokit: Octokit | undefined = undefined;
```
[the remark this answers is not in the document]
jzsfkzm: We use it as a string, so I think it's good as a string.

apps/api/src/services/external/templates/template-fetcher.service.ts
```diff
@@ -0,0 +1,341 @@
...
+
+  private async processTemplateSource(
+    templateSource: TemplateSource,
+    directoryItems: GithubDirectoryItem[],
+    options: { includeConfigJson?: boolean }
+  ): Promise<any> {
+    try {
+      if (templateSource.path.startsWith("http:") || templateSource.path.startsWith("https:")) {
+        throw "Absolute URL";
+      }
```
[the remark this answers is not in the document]
jzsfkzm: Throwing may break the whole process, let's not do that.

apps/api/test/functional/templates.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+      expect(response.status).toBe(404);
+    });
+  });
+
+  function setup() {
+    const cacheDir = path.join(__dirname, "..", "mocks/templates/cache");
+    const templatesDir = path.join(__dirname, "..", "..", "dist/.data/templates");
+    fs.cpSync(cacheDir, templatesDir, { recursive: true });
+  }
+});
```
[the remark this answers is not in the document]
jzsfkzm: Setup is not creating any state to return here, and there's no need to pass params either. It's just a simple function copying some files, app is coming from an import, already available on the root level.

apps/api/test/functional/templates.spec.ts
jzsfkzm: Moved the whole logic outside of tsyringe world so cache generation still works in build time with minimal config. But then it's either Octokit itself or the getOctokit function we should mock and the mock would be the same in both cases (a fake Octokit instance we currently have in the branch). I think what we have now is what we can achieve here, or do I misunderstand and can it still be done in a better way?

apps/api/test/functional/templates.spec.ts
jzsfkzm: Discussed in Slack, not separating for now so it doesn't need a rebuild/redeploy round on changes in the source repos.

stalniy (approved):

apps/api/test/functional/templates.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+  });
+
+  const expectCategory = (result: GetTemplatesFullResponse, expectedCategory: string, expectedTemplateIds: string[]) => {
+    const category = result.find((c: { title: string }) => c.title === expectedCategory);
+    expect(category.templates).toHaveLength(expectedTemplateIds.length);
+    const templateIds = category.templates.map(({ id }) => id);
+    templateIds.forEach(templateId => {
+      expect(expectedTemplateIds).toContain(templateId);
+    });
+  };
```
stalniy, line 51: suggestion(non-blocking): introducing reusable assertion function is bad idea because when test fails, test runner will display you part of the stack trace from reusable function which won't be informative/helpful.

Better to rewrite this into shorter form or create a custom expect matcher. For example:

```ts
expect(findCategory(result, expectedCategory)?.templates).toHaveLength(5);
expect(findCategory(result, expectedCategory)?.templates.map(t => t.id)).toEqual(expect.arrayContaining(expectedTemplateIds));
```

stalniy (approved):

PR #1512, opened 2025-06-18
feat(alert): improves deployment alerts UI

ygrishajev:
https://github.com/user-attachments/assets/ef83c435-0434-4be2-807a-807f25349fa8

## Summary by CodeRabbit

- **New Features**
  - Added a badge indicator to the "Alerts" tab in deployment details when there are unsaved changes to alert settings.

- **Refactor**
  - Unified alert configuration into a single form with validation and streamlined submission in the deployment alerts section.
  - Simplified alert subcomponents to rely on a shared form context for state and validation, improving consistency and maintainability.
  - Updated notification channel selection to support disabled and error states, with improved accessibility and styling.
  - Refactored deployment alerts container to handle denomination and price-aware conversions for alert thresholds, enhancing data consistency.
  - Modified notification channels guard to use a function-as-children pattern for better data passing.

- **Bug Fixes**
  - Improved error handling and user feedback for alert configuration actions.

- **Tests**
  - Expanded and improved test coverage for deployment alerts and notification channel selection, including realistic data and new scenarios.
  - Refined tests to simulate user interactions and validate form behavior more accurately.

- **Chores**
  - Removed unused components including the deployment balance container.
  - Added new test data builders for alerts.


apps/deploy-web/src/components/deployments/DeploymentBalanceAlert/DeploymentBalanceAlert.tsx
```diff
@@ -1,153 +1,73 @@
...
-            </div>
-            <div className="space-y-3">
-              <LoadingButton type="submit" loading={isLoading}>
-                {initialValues ? "Update" : "Create"}
-              </LoadingButton>
-            </div>
-          </form>
-        </Form>
+            )}
+          />
```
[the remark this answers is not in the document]
ygrishajev: it is validated by zod schema in the parent

apps/deploy-web/src/components/alerts/DeploymentAlertsContainer/DeploymentAlertsContainer.spec.tsx
```diff
@@ -62,17 +66,152 @@ describe("DeploymentAlertsContainer", () => {
...
+        deploymentClosed: {
+          enabled: true,
+          notificationChannelId: buildNotificationChannel().id
+        }
+      }
+    };
+
+    child.upsert(input);
+
+    await waitFor(() => {
```
stalniy: suggestion(non-blocking): there proper to do it is to use `act` from testing library. Normally we don't use because it's hidden by testing-library in all its functions (i.e., `fireEvent.click`, etc). So, each time we do state change triggers not with help of testing library, we need to wrap that operation in `act` call.

So, it should be:

```ts

    await act(() => child.upsert(input));
    expect(requestFn).toHaveBeenCalledWith(
        expect.objectContaining({
          method: "post",
          url: "/v1/deployment-alerts/{dseq}"
        }),
     //....
```

The solution with `waitFor` works because `waitFor` retries operation until it reach max attempts or timeout but this is conceptionaly incorrect.

If you wrap `child.upsert` with `act`, then `waitFor` succeeds from the 1st try and as a result  is redundant

apps/deploy-web/src/components/alerts/NotificationChannelSelectForm/NotificationChannelSelect.spec.tsx
```diff
@@ -5,36 +5,83 @@ import type { FCWithChildren } from "@src/types/component";
...
-        <NotificationChannelSelectView data={[]} isFetched={true} />
-      </Wrapper>
-    );
+  it("renders select trigger when data is provided", () => {
+    setup();
 
-    expect(screen.getByText("Notification Channel")).toBeInTheDocument();
-    expect(screen.getByText("Select notification channel")).toBeInTheDocument();
+    expect(screen.queryByText("Notification Channel")).toBeInTheDocument();
+    expect(screen.getByTestId("notification-channel-select-trigger")).toBeInTheDocument();
```
stalniy: suggestion(non-blocking): you don't need test-id here, you can rely on label:

```ts
    expect(screen.getByLabelText(/Notification Channel/)).toBeInTheDocument();
```

The first 2 methods to use for finding form fields in testing-library are:
* `getByLabelText`
* `getByPlaceholderText`

And only if for some reason this is not possible, we should add testid

apps/deploy-web/src/components/deployments/DeploymentAlerts/DeploymentAlerts.tsx
```diff
@@ -1,61 +1,159 @@
...
+        });
+  }, [data?.alerts, maxBalanceThreshold, notificationChannels]);
+
+  const form = useForm({
+    defaultValues: providedValues,
+    reValidateMode: "onSubmit",
+    resolver: zodResolver(strictSchema)
+  });
+
+  const [hasChanges, setHasChanges] = useState(false);
```
stalniy, line 105: **thought(non-blocking):** probably could be computed in `useMemo`

stalniy (dismissed):

apps/deploy-web/src/components/deployments/DeploymentAlerts/DeploymentAlerts.tsx
ygrishajev: Nope, it would cause an infinite state update loop with the parent.

Issue #1505, opened 2025-06-17
Refactor stripe controller to not re-query user


ygrishajev:
https://github.com/akash-network/console/blob/85c046b1f7286b6c5fea41251712b3e89f413163/apps/api/src/billing/controllers/stripe/stripe.controller.ts

User is already queried by an auth interceptor and is present in `authService.currentUser`

Issue #1504, opened 2025-06-17
Properly handle invalid coupon error


ygrishajev:
This error results in 500 response code. Instead it should be 400.

```json
{
  "level": "error",
  "time": "2025-06-17T07:54:38.256Z",
  "pid": 1,
  "hostname": "console-api-mainnet-6bfd97b4bb-5j8v9",
  "context": "ErrorHandler",
  "traceId": "6af93ab88add7e83d0b9a26105cef0bf",
  "spanId": "e416a862ddf7ae78",
  "traceFlags": 1,
  "msg": "Error: No valid promotion code or coupon found with the provided code\n    at StripeService.applyCoupon (/app/apps/api/dist/server.js:27619:15)\n    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)\n    at async StripeController.applyCoupon (/app/apps/api/dist/server.js:163821:24)\n    at async applyCoupon (/app/apps/api/dist/server.js:600806:22)\n    at async dispatch (/app/apps/api/dist/server.js:67832:17)\n    at async /app/apps/api/dist/server.js:69257:5\n    at async mw (/app/apps/api/dist/server.js:638863:26)\n    at async dispatch (/app/apps/api/dist/server.js:67832:17)\n    at async /app/apps/api/dist/server.js:3612:9\n    at async dispatch (/app/apps/api/dist/server.js:67832:17)"
}
```

Issue #1486, opened 2025-06-14
Improve credit card error handling


baktun14:
- credit card declined
- "The Cash App Pay customer request expired"
- display custom error message on console
- log to amplitude

```
Error: Your card was declined.
    at StripeError.generate (/app/apps/api/dist/server.js:129118:20)
    at res.toJSON.then.Error_js_1.StripeAPIError.message (/app/apps/api/dist/server.js:659463:54)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
```

stalniy:
@baktun14 is this related https://github.com/akash-network/console/issues/1504 ?

baktun14:
> [@baktun14](https://github.com/baktun14) is this related [#1504](https://github.com/akash-network/console/issues/1504) ?

yes

PR #1569, opened 2025-06-26
fix(billing): improve stripe error handling

baktun14:
- https://github.com/akash-network/console/issues/1486
- https://github.com/akash-network/console/issues/1504
- https://github.com/akash-network/console/issues/1505
- Based on the [stripe error handling docs](https://docs.stripe.com/api/errors/handling)

## Summary by CodeRabbit

* **New Features**
  * Centralized and improved error handling for Stripe payments and coupons, providing user-friendly error messages and suggested actions.
  * Inline error alerts and snackbars now display detailed feedback and allow users to clear errors without reloading the payment page.
  * Introduced a new service to standardize Stripe error interpretation, retry logic, and error classification.

* **Bug Fixes**
  * Enhanced reliability in payment and coupon application flows by handling Stripe errors more robustly.

* **Refactor**
  * Streamlined user context handling in billing APIs and removed redundant user lookups.
  * Removed coupon support from payment confirmation requests and related APIs.
  * Improved API error responses with consistent error codes and types for better client handling.

* **Tests**
  * Added comprehensive tests for Stripe error handling and coupon application scenarios, including utility functions.
  * Added tests validating error handling utilities in the web client.

* **Chores**
  * Removed an unused dependency from the JWT package.

### DIFF
--- a/apps/api/src/billing/controllers/stripe/stripe.controller.ts
+++ b/apps/api/src/billing/controllers/stripe/stripe.controller.ts
@@ -6,15 +6,15 @@ import { AuthService, Protected } from "@src/auth/services/auth.service";
 import type { StripePricesOutputResponse } from "@src/billing";
 import { ConfirmPaymentRequest, Discount, Transaction } from "@src/billing/http-schemas/stripe.schema";
 import { StripeService } from "@src/billing/services/stripe/stripe.service";
+import { StripeErrorService } from "@src/billing/services/stripe-error/stripe-error.service";
 import { Semaphore } from "@src/core/lib/semaphore.decorator";
-import { UserRepository } from "@src/user/repositories";
 
 @singleton()
 export class StripeController {
   constructor(
     private readonly stripe: StripeService,
-    private readonly userRepository: UserRepository,
-    private readonly authService: AuthService
+    private readonly authService: AuthService,
+    private readonly stripeErrorService: StripeErrorService
   ) {}
 
   @Protected([{ action: "read", subject: "StripePayment" }])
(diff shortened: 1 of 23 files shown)


apps/api/src/billing/controllers/stripe/stripe.controller.ts
```diff
@@ -24,96 +24,105 @@ export class StripeController {
...
 
   @Protected([{ action: "read", subject: "StripePayment" }])
   async getCustomerDiscounts(): Promise<{ data: { discounts: Discount[] } }> {
-    const userId = this.authService.currentUser.userId;
-    const user = await this.userRepository.findOneBy({ userId });
+    const { currentUser } = this.authService;
 
-    assert(user, 404, "User not found");
-    assert(user.stripeCustomerId, 400, "User does not have a Stripe customer ID");
+    assert(currentUser.stripeCustomerId, 400, "User does not have a Stripe customer ID");
```
stalniy: **thought(non-blocking):** this error message says what the issue is but I do not understand how to fix it (from client perspective)

apps/api/src/billing/services/stripe-error/stripe-error.service.ts
```diff
@@ -0,0 +1,340 @@
+import createError, { HttpError } from "http-errors";
+import Stripe from "stripe";
+import { singleton } from "tsyringe";
+
+@singleton()
+export class StripeErrorService {
+  private readonly COUPON_ERRORS = {
+    "No valid promotion code or coupon found with the provided code": {
```
stalniy, line 8: question(non-blocking): why don't we use error codes in our business logic?

apps/deploy-web/src/queries/usePaymentQueries.spec.tsx
```diff
@@ -146,6 +146,27 @@ describe("usePaymentQueries", () => {
...
+        }
+      };
+      (useServices().stripe.applyCoupon as jest.Mock).mockResolvedValue(mockErrorResponse);
+      const { result } = setupQuery(() => usePaymentMutations());
+
+      await act(async () => {
+        const response = await result.current.applyCoupon.mutateAsync({ coupon: "INVALID" });
+        expect(response.error).toBeDefined();
+        expect(response.error?.message).toBe("No valid promotion code or coupon found with the provided code");
+      });
```
stalniy:
```suggestion
      const response = await act(async () => result.current.applyCoupon.mutateAsync({ coupon: "INVALID" }));
      expect(response.error?.message).toBe("No valid promotion code or coupon found with the provided code");
```

apps/deploy-web/src/utils/stripeErrorHandler.ts
```diff
@@ -0,0 +1,135 @@
...
+}
+
+/**
+ * Handle coupon-specific errors
+ */
+export function handleCouponError(response: { error?: { message: string } }): StripeErrorInfo {
+  if (response.error) {
+    const errorMsg = response.error.message.toLowerCase();
+
+    if (errorMsg.includes("no valid promotion code")) {
```
stalniy: question(blocking): is it possible to reorganize code, so that errors are checked by their codes, not by partial string inclusion?

The issue with the implemented approach is that it may be fragile, somebody change the string and this will cause cascading changes. On the other hand, error codes are extremely unlikely to change.

apps/api/src/billing/controllers/stripe/stripe.controller.ts
baktun14: Perhaps this should be a 500 since it's probably a bug on our side

apps/deploy-web/src/utils/stripeErrorHandler.ts
baktun14: Good point, I'll make the changes

PR #1637, opened 2025-07-06
feat: add calendar and date-range-picker ui components

domhhv:
Needed for #1348 

Installs and uses `react-day-picker` package, promoted by shadcn: https://ui.shadcn.com/docs/components/calendar

Screenshot 2025-07-08 at 00 20 45
Screenshot 2025-07-08 at 00 21 07
Screenshot 2025-07-08 at 00 21 49
Screenshot 2025-07-08 at 00 22 02
Screenshot 2025-07-08 at 00 24 36

## Summary by CodeRabbit

* **New Features**
  * Introduced a customizable Calendar component with enhanced styling and navigation icons.
  * Added a responsive Date Range Picker supporting presets, month selection, and validation for date constraints.
  * Added a useMediaQuery hook to detect viewport changes for responsive design.

* **Chores**
  * Updated dependencies to include date-fns and react-day-picker.
  * Made the Date Range Picker component publicly available via the central UI components export.


packages/ui/components/custom/date-range-picker.tsx
```diff
@@ -0,0 +1,411 @@
...
+  maxDate?: Date;
+  maxRangeInDays?: number;
+  disableFuture?: boolean;
+  showPresets?: boolean;
+}
+
+export function DateRangePicker({
+  className,
+  date,
+  onDateChange,
```
ygrishajev: non-blocking: is there anything else to change? perhaps it's better to follow common input approach and just call it
```suggestion
  onChange,
```

packages/ui/components/custom/date-range-picker.tsx
```diff
@@ -0,0 +1,411 @@
...
+
+  const isMobile = useMediaQuery("(max-width: 768px)");
+
+  const today = startOfDay(new Date());
+  const effectiveMaxDate = disableFuture ? today : maxDate;
+  const effectiveMinDate = minDate;
+
+  const months = Array.from({ length: 12 }, (_, i) => format(new Date(2024, i, 1), "MMMM"));
+
+  const presets = [
```
ygrishajev: `useMemo`?

packages/ui/components/custom/date-range-picker.tsx
```diff
@@ -0,0 +1,411 @@
...
+    {
+      label: "Last year",
+      getValue: () => ({
+        from: subYears(today, 1),
+        to: today
+      })
+    }
+  ];
+
+  const validateDateRange = (range: DateRange | undefined): string | null => {
```
ygrishajev: `useCallback`

packages/ui/components/custom/date-range-picker.tsx
```diff
@@ -0,0 +1,443 @@
...
+
+    return null;
+  };
+
+  const handleDateSelect = (range: DateRange | undefined) => {
+    const error = validateDateRange(range);
+    if (error) {
+      console.warn(error);
+      return;
+    }
```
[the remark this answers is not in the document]
ygrishajev: this actually makes sense, @domhhv is there a reason this is like so? Perhaps `onError` callback would make sense so the client decides what to do with it?

packages/ui/components/custom/date-range-picker.tsx
domhhv: Only date changes so far, so onChange looks more suitable. Done.

packages/ui/components/custom/date-range-picker.tsx
domhhv: Done

packages/ui/components/custom/date-range-picker.tsx
domhhv: Done

packages/ui/components/custom/date-range-picker.tsx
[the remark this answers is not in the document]
domhhv: Right, `onError` sounds good. Done.

packages/ui/components/custom/date-range-picker.tsx
```diff
@@ -0,0 +1,504 @@
+"use client";
+
+import * as React from "react";
+import type { DateRange } from "react-day-picker";
+import { UI } from "react-day-picker";
+import { differenceInDays, format, isAfter, isBefore, isSameDay, startOfToday, subDays, subMonths, subYears } from "date-fns";
+import { CalendarIcon, ChevronDown, ChevronLeft, ChevronRight, ChevronUp, X } from "lucide-react";
```
baktun14: blocker: please use icons from `iconoir-react`. Same for `Calendar`

domhhv: Done

packages/ui/components/custom/date-range-picker.tsx
```diff
@@ -0,0 +1,515 @@
...
+    const rangedMaxDate = addDays(selectedRange.to, remainingRangeInDays);
+
+    if (maxDate) {
+      return min([maxDate, rangedMaxDate]);
+    }
+
+    return rangedMaxDate;
+  }, [maxDate, selectedRange?.to, remainingRangeInDays]);
+
+  const presets = React.useMemo(
```
ygrishajev, line 130: does this need to be filtered if `maxRangeInDays` and `minDate`/`maxDate` is set up?

packages/ui/components/custom/date-range-picker.tsx
```diff
@@ -0,0 +1,534 @@
...
+
+      return {
+        isDisabled:
+          !!(minDate && isBefore(from, minDate)) ||
+          !!(maxDate && isAfter(to, maxDate)) ||
+          !!(maxRangeInDays && differenceInDays(to, from) + 1 > maxRangeInDays),
+        isSelected: selectedRange?.from && selectedRange.to && isSameDay(selectedRange.from, from) && isSameDay(selectedRange.to, to)
+      };
+    },
+    [selectedRange]
```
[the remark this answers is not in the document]
ygrishajev: @domhhv is this intended?

packages/ui/components/custom/date-range-picker.tsx
```diff
@@ -0,0 +1,515 @@
...
+        getValue: () => ({
+          from: subYears(today, 1),
+          to: today
+        })
+      }
+    ],
+    [today]
+  );
+
+  const handlePresetSelect = React.useCallback(
```
ygrishajev: methods should state what they do, not when they're called. e.g.
```suggestion
  const selectRangeAndClosePresets = React.useCallback(
```
or anything better you may come up with

packages/ui/components/custom/date-range-picker.tsx
domhhv: Currently, it's not filtered out from the UI, but when the presets render with `.map()`, I call `getPresetMetadata` function below for each item to determine if it should be disabled for selection (both mobile and desktop JSX). There, I use `maxRangeInDays`, `minDate` and `maxDate`.

packages/ui/components/custom/date-range-picker.tsx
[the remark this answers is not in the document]
domhhv: Not intended, those dependencies are indeed missing (added now)

packages/ui/components/custom/date-range-picker.tsx
domhhv: Renamed all functions in this component

packages/ui/components/custom/date-range-picker.tsx
ygrishajev: it seems like that could be a `useMemo` value instead of a function. is this right?

```ts
{
        label: "Last 7 days",
        getValue: () => ({
          from: subDays(today, 6),
          to: today
        }),
        isDisabled: true,
        isSelected: false
      }
```

stalniy (dismissed):

packages/ui/components/custom/date-range-picker.tsx
domhhv: ✅ Destructured `getPresetMetadata()` call inside the memoized array

ygrishajev (approved):

PR #1638, opened 2025-07-07
feat: allow filtering stripe charges list api by date created and endingBefore cursor

domhhv:
Needed for #1348 

## Summary by CodeRabbit

* **New Features**
  * Enhanced transaction retrieval with support for date range filtering and backward pagination.
  * Added new pagination fields to responses, including both next and previous page cursors.
  * API and SDK now support filtering transactions by start and end dates.

* **Bug Fixes**
  * Improved validation to ensure date ranges are valid and do not exceed 366 days.

* **Tests**
  * Expanded and added comprehensive tests for transaction pagination, date filtering, and key generation logic.

* **Documentation**
  * Updated API schema descriptions for new query parameters and validation rules.


apps/deploy-web/src/queries/queryKeys.spec.ts
```diff
@@ -0,0 +1,412 @@
+import { QueryKeys } from "./queryKeys";
```
stalniy, line 1: question(blocking): could you please explain the value for QueryKeys to be covered with tests?

apps/api/src/billing/http-schemas/stripe.schema.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+
+if (!ZodType.prototype.openapi) {
+  ZodType.prototype.openapi = function () {
+    return this;
+  };
+}
+
+import { CustomerTransactionsQuerySchema } from "./stripe.schema";
+
+describe("CustomerTransactionsQuerySchema", () => {
```
stalniy, line 13: question(blocking): Could you help me understand the value these tests bring? From what I see, we’re testing the schema definitions, which seems to mainly verify that Zod works as intended. Since this is already covered by the library itself, I’m wondering if we could trust Zod here and focus our tests on higher-level validation logic instead?

baktun14 (dismissed):

apps/api/src/billing/http-schemas/stripe.schema.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+  const ONE_DAY_MS = 1000 * 60 * 60 * 24;
```
ygrishajev: non-blocking: check this out https://github.com/date-fns/date-fns/blob/main/src/constants/index.ts

apps/api/src/billing/http-schemas/stripe.schema.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+    const endDate = new Date(startDate.getTime() + ONE_DAY_MS * 367);
+    expect(() =>
+      schema.parse({
+        startDate: startDate.toISOString(),
+        endDate: endDate.toISOString()
+      })
+    ).toThrow("Date range cannot exceed 366 days and startDate must be before endDate");
+  });
+
+  function setup() {
```
ygrishajev: i don't think setup function is needed in this case :)

apps/api/src/billing/http-schemas/stripe.schema.spec.ts
domhhv: Thanks, utilized

apps/api/src/billing/http-schemas/stripe.schema.spec.ts
domhhv: Yeah, removed

ygrishajev (dismissed):

apps/api/src/billing/http-schemas/stripe.schema.ts
```diff
@@ -127,20 +127,50 @@ export const CustomerTransactionsResponseSchema = z.object({
...
+  .refine(
+    data => {
+      if (!data.startDate || !data.endDate) {
+        return true;
+      }
+
+      const start = new Date(data.startDate);
+      const end = new Date(data.endDate);
+
+      const daysDiff = Math.floor((end.getTime() - start.getTime()) / (24 * 60 * 60 * 1000));
```
stalniy: suggestion(non-blocking): since we have date-fns, we should utilize it - https://date-fns.org/v2.30.0/docs/differenceInDays

apps/api/src/billing/services/stripe/stripe.service.spec.ts
```diff
@@ -7,6 +8,7 @@ import { StripeService } from "./stripe.service";
 
 import { create as StripeSeederCreate } from "@test/seeders/stripe.seeder";
 import { UserSeeder } from "@test/seeders/user.seeder";
+import { stub } from "@test/services/stub";
```
stalniy, line 11: suggestion(non-blocking): please use `jest-mock-extended` in future PRs, we need to deprecate `stub`

stalniy (dismissed):

apps/deploy-web/src/queries/queryKeys.spec.ts
domhhv: The CI failed because of missing codecov... Specifically, for the changes made in `getPaymentTransactionsKey` method. I think this must be the correct link where it complained about missing coverage: https://app.codecov.io/gh/akash-network/console/commit/47bba3b8515de15470402ba92810e7629f3d6d01

So I had to fix it by covering `getPaymentTransactionsKey`, and while I was at it, I decided to vibe-cover everything. Of course, basically only `getPaymentTransactionsKey` needs to be covered, so I could just remove everything else.

Screenshot 2025-07-14 at 19 25 00

apps/api/src/billing/http-schemas/stripe.schema.spec.ts
domhhv: The CI failed because of missing codecov for the schema changes as well... I think this time it's this link to failed codecov report: https://app.codecov.io/gh/akash-network/console/commit/2729819834d00a72a24d4724a8a7deacc495c522

I also think that there's little value in such tests, as the schema is mainly and already tested in the service, controller, or somewhere else. But the CI wouldn't let me go green otherwise.

Screenshot 2025-07-14 at 19 27 24

apps/deploy-web/src/queries/queryKeys.spec.ts
domhhv: https://github.com/akash-network/console/runs/45465446448

apps/api/src/billing/http-schemas/stripe.schema.spec.ts
domhhv: https://github.com/akash-network/console/runs/45472433899

apps/api/src/billing/services/stripe/stripe.service.spec.ts
domhhv: Ah, sure, good to know

apps/api/src/billing/http-schemas/stripe.schema.ts
domhhv: 👍 Done

apps/api/src/billing/http-schemas/stripe.schema.spec.ts
domhhv: I removed tests that didn't cover .refine logic from this schema specs

apps/deploy-web/src/queries/queryKeys.spec.ts
domhhv: I removed a big chunk of unnecessary tests from these queryKeys specs and left only those related to the changes that codecov complained about

ygrishajev (approved):

PR #1658, opened 2025-07-10
feat(billing): apply coupon to balance

baktun14:
## Summary by CodeRabbit

* **New Features**
  * When applying a coupon, users now see the amount added to their balance if the coupon is valid and of a fixed amount.
  * The payment page displays a success message with the credited amount after a successful coupon claim.

* **Bug Fixes**
  * Improved error handling for unsupported coupon types, ensuring only fixed amount coupons can be applied.
  * Added rollback logic to revert coupon application if wallet top-up fails.

* **Tests**
  * Expanded test coverage for coupon application, including validation for fixed amount and percentage-based coupons.
  * Added tests for new coupon-related error messages.

* **Chores**
  * Introduced coding rules to enforce TypeScript best practices and discourage deprecated methods and unnecessary comments.
  * Removed deprecated error handling methods from Stripe error service.


apps/api/src/billing/services/stripe/stripe.service.ts
```diff
@@ -200,36 +200,82 @@ export class StripeService extends Stripe {
...
+      [updateField]: updateId
+    });
+
+    const amountToAdd = coupon.amount_off; // amount_off is already in cents
+
+    if (amountToAdd > 0) {
+      await this.refillService.topUpWallet(amountToAdd, currentUser.id);
+    }
+
+    // Then, remove the coupon from the customer
```
stalniy: question(non-blocking): do you need transaction here or try/finally?

apps/api/src/billing/services/stripe/stripe.service.ts
```diff
@@ -200,36 +200,82 @@ export class StripeService extends Stripe {
...
+        updateField: "coupon",
+        updateId: matchingCoupon.id
       });
-      return matchingCoupon;
     }
 
     throw new Error("No valid promotion code or coupon found with the provided code");
   }
 
+  private async applyCouponOrPromotionCode({
```
stalniy, line 236: question(blocking): how do you prevent double coupon application? What if this endpoint is called concurrently few times?

baktun14: Good point I'll add prevention for that

apps/api/src/billing/services/stripe/stripe.service.ts
baktun14: Yea definitely, will do.

apps/api/src/billing/services/stripe/stripe.service.ts
```diff
@@ -257,23 +260,49 @@ export class StripeService extends Stripe {
...
+            couponId: updateId,
+            error: error instanceof Error ? error.message : String(error)
+          });
+        } catch (rollbackError) {
+          logger.error({
+            event: "COUPON_APPLICATION_ROLLBACK_FAILED",
+            userId: currentUser.id,
+            couponId: updateId,
+            originalError: error instanceof Error ? error.message : String(error),
+            rollbackError: rollbackError instanceof Error ? rollbackError.message : String(rollbackError)
```
stalniy: suggestion(blocking): `LoggerService` supports special payload structure `{ error }`, when it sees such structure, it will grab many details from passed error (nested call stack trace including causes, code, response, etc). So, never pass `error.message` to the logger if you have an actual Error instance

stalniy:
```suggestion
            error: new AggregateError([error, rollbackError])
```

apps/api/src/billing/controllers/stripe/stripe.controller.ts
```diff
@@ -75,17 +75,19 @@ export class StripeController {
...
   async applyCoupon(
-    couponId: string
+    params: ApplyCouponRequest["data"]
   ): Promise<{ data: { coupon: Stripe.Coupon | Stripe.PromotionCode | null; amountAdded?: number; error?: { message: string } } }> {
     const { currentUser } = this.authService;
 
     assert(currentUser.stripeCustomerId, 500, "Payment account not properly configured. Please contact support.");
-    assert(couponId, 400, "Coupon ID is required");
+    assert(params.couponId, 400, "Coupon ID is required");
+    assert(params.userId, 400, "User ID is required");
```
stalniy, line 87: question(non-blocking): don't we enforce this with zod on router level?

baktun14: We do, perhaps this is unnecessary

baktun14: I'll leave it there, it doesn't hurt to explicitly validate it in the controller

PR #1659, opened 2025-07-10
fix: do not list suppressed alerts

### LINKED ISSUE #1547: Disabled Close Deployment alerts are not being cleaned up
Steps to reproduce:
1) create a deployment
2) enable Close Deployment alert
3) disable Close Deployment alert
4) close deployment
5) visit alerts page

EXR: the alert should not be in the list
ACR: the alert is in the list

This is happening due to how alerts are being processed. Alerts are fetched, checked and if found - suppressed. However alerts suppressed by system aren't being picked up. Potentially we may need a different way of suppressing such alerts. E.g. create a deployment closed handler and suppress all alerts having dseq and owner in the params - basically suppress all kinds of deployment alerts at once.

jzsfkzm:
closes #1547

## Summary by CodeRabbit

* **Bug Fixes**
  * Suppressed alerts are now correctly excluded from the alerts list.

* **Tests**
  * Added a test to verify that suppressed alerts do not appear in the alerts list.
  * Improved test setup to allow custom alert parameters and ensure a clean test environment.

* **New Features**
  * Introduced a safer mock data generator to improve test reliability.

* **Enhancements**
  * Improved alert filtering logic for more consistent results.
  * Made alert list query parameters publicly accessible for better integration.

### DIFF
--- a/apps/notifications/src/modules/alert/repositories/alert/alert.repository.ts
+++ b/apps/notifications/src/modules/alert/repositories/alert/alert.repository.ts
@@ -30,7 +30,7 @@ export type AlertOutput = GeneralAlertOutput | DeploymentBalanceAlertOutput;
 
 export type AlertType = AlertOutput["type"];
 
-const isNotSuppressed = sql`(${schema.Alert.params}->>'suppressedBySystem' IS NULL OR ${schema.Alert.params}->>'suppressedBySystem' != 'true')`;
+const isNotSuppressed = sql`(${schema.Alert.params} @> '{"suppressedBySystem": null}' OR NOT (${schema.Alert.params} @> '{"suppressedBySystem": true}'))`;
 
 export type AlertInputTypeMap = {
   DEPLOYMENT_BALANCE: DeploymentBalanceAlertInput;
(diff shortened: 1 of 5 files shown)


apps/notifications/src/modules/alert/repositories/alert/alert.repository.ts
```diff
@@ -30,6 +30,8 @@ export type AlertOutput = ChainMessageAlertOutput | DeploymentBalanceAlertOutput
 
 export type AlertType = AlertOutput["type"];
 
+const isNotSuppressed = sql`(${schema.Alert.params}->>'suppressedBySystem' IS NULL OR ${schema.Alert.params}->>'suppressedBySystem' != 'true')`;
```
stalniy: thought(non-blocking): queries with `->>` cannot use indexes but queries that use JSON match operator (`@>`) can

stalniy (dismissed):

apps/notifications/src/modules/alert/repositories/alert/alert.repository.ts
jzsfkzm: @stalniy dseq and type also search with `->>` (conditionally, though). Maybe we can change all three to `@>` so [this index](https://github.com/jzsfkzm/console/blob/594fb2d731c593353cabc745e3ff87b344eeb899/apps/notifications/src/modules/alert/model-schemas/alert.schema.ts#L39) would be used. Should we?

ygrishajev:
@jzsfkzm @stalniy just spotted this PR. I think the notification service is done at this point. Not sure this PR is needed anymore. Any objections if I close it?

jzsfkzm:
Not me to tell, for me it just seemed something worth changing (some alerts are getting through despite the user tried to suppress those).

PR #1677, opened 2025-07-14
fix: make userAgent 500 max chars long

### LINKED ISSUE #1520: PostgressError: value too long for type character varying
```
PostgresError: value too long for type character varying(255)
    at ErrorResponse (/app/apps/api/dist/server.js:613834:26)
    at handle (/app/apps/api/dist/server.js:613520:6)
    at Socket.data (/app/apps/api/dist/server.js:613361:9)
    at Socket.emit (node:events:518:28)
    at addChunk (node:internal/streams/readable:561:12)
    at readableAddChunkPushByteMode (node:internal/streams/readable:512:3)
    at Readable.push (node:internal/streams/readable:392:5)
    at TCP.onStreamRead (node:internal/stream_base_commons:189:23)
    at TCP.callbackTrampoline (node:internal/async_hooks:130:17)
```

jzsfkzm:
@stalniy I suspect userAgent field to be short (varchar 255). Probably a varchar 500 or 1000 would be enough too, but text will nail it forever.

closes #1520

## Summary by CodeRabbit

* **New Features**
  * Expanded database schema with new tables for user wallets, checkout sessions, user settings, deployment settings, and API keys, supporting enhanced user and deployment management.

* **Improvements**
  * Increased the allowed length for user agent information from 255 to 500 characters.
  * User agent data is now automatically truncated to fit the new maximum length, ensuring consistent data storage.

* **Chores**
  * Updated database schema snapshots and migration history for better tracking of changes.

### DIFF
--- a/apps/api/src/services/db/userDataService.ts
+++ b/apps/api/src/services/db/userDataService.ts
@@ -6,6 +6,7 @@ import { container } from "tsyringe";
 
 import { UserWalletRepository } from "@src/billing/repositories";
 import { AnalyticsService } from "@src/core/services/analytics/analytics.service";
+import { userAgentMaxLength } from "@src/user/model-schemas/user/user.schema";
 
 const logger = LoggerService.forContext("UserDataService");
 
@@ -91,6 +92,7 @@ export async function getSettingsOrInit({
   fingerprint
 }: UserInput) {
   let userSettings: UserSetting | null = null;
+  const truncatedUserAgent = userAgent?.substring(0, userAgentMaxLength);
   let isAnonymous = false;
 
   if (anonymousUserId) {
@@ -104,7 +106,7 @@ export async function getSettingsOrInit({
           stripeCustomerId: null,
           subscribedToNewsletter,
           lastIp: ip,
-          lastUserAgent: userAgent,
+          lastUserAgent: truncatedUserAgent,
           lastFingerprint: fingerprint
         },
         { where: { id: anonymousUserId, userId: null }, returning: ["*"] }
@@ -147,7 +149,7 @@ export async function getSettingsOrInit({
       stripeCustomerId: null,
       subscribedToNewsletter,
       lastIp: ip,
-      lastUserAgent: userAgent,
+      lastUserAgent: truncatedUserAgent,
       lastFingerprint: fingerprint
     });
     analyticsService.track(userSettings.id, "user_registered");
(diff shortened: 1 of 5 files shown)


apps/api/src/user/model-schemas/user/user.schema.ts
```diff
@@ -20,7 +20,7 @@ export const Users = pgTable("userSetting", {
   githubUsername: varchar("githubUsername", { length: 255 }),
   lastActiveAt: timestamp("last_active_at").defaultNow(),
   lastIp: varchar("last_ip", { length: 255 }),
-  lastUserAgent: varchar("last_user_agent", { length: 255 }),
+  lastUserAgent: text("last_user_agent"),
```
stalniy: suggestion(blocking): I think that having a limit on this value is important. This is user provided value and if we don't set some validation on it, we can save some junk, potentially big junk depending on reverse proxy total headers limits. So, let's put big enough but not unlimited. l think 500 is enough to understand what is inside

stalniy: the value is thrown by postgresql, to get rid of it we can either slice string on app level or cast on postgres level. It’s cleaner to handle truncation directly at the application level, since we anyway need to know the length for both cases

stalniy (changes requested):

apps/api/src/user/model-schemas/user/user.schema.ts
jzsfkzm: @stalniy That's right, I didn't think about the possibly big junk. Ok, limited it to 500 chars now, and truncating incoming data in the service.

apps/api/src/services/db/userDataService.ts
```diff
@@ -91,6 +92,7 @@ export async function getSettingsOrInit({
   fingerprint
 }: UserInput) {
   let userSettings: UserSetting | null = null;
+  const truncatedUserAgent = userAgent?.substring(0, userAgentMaxLength);
```
stalniy, line 95: question: what do you think about moving it on repository level?

jzsfkzm: There's no repository for UserSettings at the moment, just the model. We can change the UserSetting model though, adding a setter, what do you think @stalniy?
```
  /**
   * The last user agent of the user setting
   */
  @Column({
    field: "last_user_agent",
    set(value: string | undefined) {
      if (value) {
        this.setDataValue("lastUserAgent", value.substring(0, userAgentMaxLength));
      }
    }
  })
  lastUserAgent?: string;
```

stalniy: Ah, I see. OK let it be like this for now!

stalniy (approved):

baktun14 (approved):

PR #1684, opened 2025-07-15
fix: fix wallet creation error on stripe webhook

### LINKED ISSUE #1662: API error on stripe webhook
## What

```
PostgresError: duplicate key value violates unique constraint \"user_wallets_user_id_unique\" (code: 23505)\

"httpRequest": {
    "requestMethod": "POST",
    "requestUrl": "http://console-api.akash.network/v1/stripe-webhook",
    "status": 500,
    "referrer": "about:client",
    "protocol": "https",
    "duration": "7947.732ms",
    "userAgent": "Stripe/1.0 (+https://stripe.com/docs/webhooks)",
    "remoteIp": "54.187.216.72"
  },
```

jzsfkzm:
I think multiple webhook calls incoming at the same time can cause a race condition:
- first one is not finding the wallet, creating it
- second one is not finding it either, trying to create it, but it's been already created by the first one in the meantime => unique index error

closes #1662

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive tests for wallet top-up scenarios, including handling of existing wallets, creation of new wallets, and concurrent top-up requests.
* **Refactor**
  * Improved wallet top-up process to ensure reliable wallet creation and management during concurrent requests, enhancing overall service stability.

### DIFF
--- a/apps/api/src/billing/services/refill/refill.service.ts
+++ b/apps/api/src/billing/services/refill/refill.service.ts
@@ -6,6 +6,7 @@ import { BillingConfig, InjectBillingConfig } from "@src/billing/providers";
 import { UserWalletOutput, UserWalletRepository } from "@src/billing/repositories";
 import { ManagedUserWalletService, WalletInitializerService } from "@src/billing/services";
 import { BalancesService } from "@src/billing/services/balances/balances.service";
+import { Semaphore } from "@src/core/lib/semaphore.decorator";
 import { AnalyticsService } from "@src/core/services/analytics/analytics.service";
 
 @singleton()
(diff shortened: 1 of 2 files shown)


baktun14 (dismissed):

apps/api/src/billing/services/refill/refill.service.ts
```diff
@@ -53,14 +54,8 @@ export class RefillService {
...
-    let userWallet = await this.userWalletRepository.findOneBy({ userId });
-    let currentLimit: number = 0;
-
-    if (userWallet) {
-      currentLimit = await this.balancesService.retrieveDeploymentLimit(userWallet);
-    } else {
-      userWallet = await this.walletInitializerService.initialize(userId);
-    }
+    const userWallet = await this.getOrCreateUserWallet(userId);
+    const currentLimit = (await this.balancesService.retrieveDeploymentLimit(userWallet)) || 0;
```
ygrishajev: non-blocking: it looks like this function is not expected to return a non-number value
```typescript
async retrieveDeploymentLimit(userWallet: Pick<UserWalletOutput, "address">): Promise<number> 
```
```suggestion
    const currentLimit = (await this.balancesService.retrieveDeploymentLimit(userWallet));
```

ygrishajev (dismissed):

apps/api/src/billing/services/refill/refill.service.ts
jzsfkzm: True that, silly me, didn't mock the method where it should've been.

PR #1707, opened 2025-07-20
feat(billing): add stripe charges table list to usage ui

domhhv:
Updates #1348 

## Summary by CodeRabbit

* **New Features**
  * Introduced a dedicated Billing page with paginated billing history, date range filtering, CSV export, and interactive tab navigation between Billing and Usage views.
  * Added a new menu item "Billing & Usage" in the account dropdown for streamlined navigation.
  * Implemented CSV export utilities for both billing and usage data.

* **Enhancements**
  * Improved pagination controls with a reusable page size selector and clearer disabled states.
  * Updated tabbed layout for easier switching between Billing and Usage sections.

* **Bug Fixes**
  * Ensured billing transaction date filters use ISO 8601 formatting for improved accuracy.

* **Tests**
  * Added comprehensive test suites for BillingContainer and BillingView components to verify rendering, data handling, and user interactions.

* **Chores**
  * Refactored utility functions for CSV field sanitization and downloading to improve code reuse and maintainability.


apps/deploy-web/src/components/billing-usage/BillingView/BillingView.tsx
```diff
@@ -0,0 +1,355 @@
...
+        >
+          {capitalizeFirstLetter(info.getValue())}
+        </div>
+      )
+    }),
+    columnHelper.display({
+      id: "receipt",
+      header: "Receipt",
+      cell: info => (
+        <TooltipProvider>
```
baktun14: We already have a Tooltip provider at the root in `_app.tsx`

baktun14: Also, we have a component called `CustomTooltip` that you can reuse for less boilerplate

apps/deploy-web/src/components/billing-usage/BillingView/BillingView.tsx
```diff
@@ -0,0 +1,355 @@
...
+                        {flexRender(cell.column.columnDef.cell, cell.getContext())}
+                      </TableCell>
+                    ))}
+                  </TableRow>
+                ))}
+              </TableBody>
+            </Table>
+          </div>
+
+          <div className="flex flex-col justify-start gap-2 pt-2 sm:flex-row sm:items-center sm:gap-0 sm:pt-6">
```
baktun14: suggestion(blocking): we have a component for this `CustomPagination`

apps/deploy-web/src/pages/billing/index.tsx
```diff
@@ -0,0 +1,11 @@
+import { BillingPage } from "@src/components/billing-usage/BillingPage";
+import { RegisteredUsersOnly } from "@src/hoc/registered-users-only/registered-users-only.hoc";
+import { defineServerSideProps } from "@src/lib/nextjs/defineServerSideProps/defineServerSideProps";
+import { isAuthenticated, isFeatureEnabled } from "@src/lib/nextjs/pageGuards/pageGuards";
+
+export default RegisteredUsersOnly(BillingPage);
+
+export const getServerSideProps = defineServerSideProps({
+  route: "/billing",
+  if: async ctx => (await isAuthenticated(ctx)) && (await isFeatureEnabled("billing_usage", ctx))
```
baktun14, line 10: we have a HOC for this called `withCustomPageAuthRequired`, perhaps we can modify it accept some feature list to check for as well to check for `isFeatureEnabled`

baktun14 (changes requested):
Please address the comments, thanks!

apps/deploy-web/src/components/billing-usage/BillingView/BillingView.tsx
domhhv: `CustomPagination` is difficult to re-use in this case without changing it. It requires `totalPageCount` prop to be passed, and it in turn passes it to `Paginator` component -> `generatePaginationLinks` function, both of which also require it. Initially, I didn't want to change so many shared components, and they seem to rely quite heavily on total pages count.

Unfortunately, it's not possible to get the `total_count` with Stripe API anymore: https://docs.stripe.com/changelog/basil/2025-03-31/deprecate-total-count-expansion. I get TypeScript errors when I try to expand `total_count` field. The don't seem to expose total pages count either.

I could suggest extracting this JSX into a pagination component that doesn't need total. Or I could try changing the implementation of those 3 components/functions.

baktun14: Ah ok I see, yea I guess the best would be to extract re-usable components from the CustomPagination

apps/deploy-web/src/pages/billing/index.tsx
ygrishajev: imo this approach is more verbose an transparent. let's discuss pls

baktun14: Let's keep like this for now and we'll address this in another pr

apps/deploy-web/src/components/billing-usage/BillingView/BillingView.tsx
domhhv: Sounds good, so here's what I re-used:

- `PaginationSizeSelector`, new common component, now re-used here in `CustomPagination`, lives in `CustomPagination.tsx` file, LMK if a better location is suitable
The next ones are shadcn originals from `packages/ui/pagination.tsx`:
- `PaginationContent`
- `PaginationEllipsis`
- `PaginationItem`
- `PaginationLink`
- `PaginationNext`
- `PaginationPrevious`

apps/deploy-web/src/components/billing-usage/BillingView/BillingView.tsx
domhhv: Thanks! `CustomTooltip` is def a saver

apps/deploy-web/src/components/billing-usage/BillingContainer/BillingContainer.tsx
```diff
@@ -0,0 +1,134 @@
...
+  data: Charge[];
+  hasMore: boolean;
+  hasPrevious: boolean;
+  isFetching: boolean;
+  isError: boolean;
+  error: Error | null;
+  onPaginationChange: (state: PaginationState) => void;
+  pagination: PaginationState;
+  totalCount: number;
+  dateRange: { from: Date | undefined; to?: Date };
```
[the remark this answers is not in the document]
ygrishajev: non-blocking: @domhhv definition seems inconsistent indeed. I guess this would be the same
```suggestion
  dateRange: { from?: Date; to?: Date };
```

baktun14 (approved):

ygrishajev (approved):

apps/deploy-web/src/components/billing-usage/BillingContainer/BillingContainer.tsx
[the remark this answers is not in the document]
domhhv: It does, but this definition comes from the `react-day-picker` itself. The type is imported [here](https://github.com/akash-network/console/blob/main/packages/ui/components/custom/date-range-picker.tsx#L4) and swapping `from` to optional results in this error in BillingView:

```
TS2322: Type
{
  from?: Date | undefined;
  to?: Date | undefined;
}
is not assignable to type DateRange
Property from is optional in type
{
  from?: Date | undefined;
  to?: Date | undefined;
}
but required in type DateRange
date-range-picker.tsx(19, 3): The expected type comes from property date which is declared here on type IntrinsicAttributes & DateRangePickerProps
```

It seems that an optional property, one that might be non-existent on an object, is different from enforcing to have the property, but allowing `undefined` value for it.

In other words, I made the type like this just to conveniently match the library's interface in the end.

PR #1708, opened 2025-07-20
feat(billing): add date filtering to usage ui

domhhv:
Updates https://github.com/akash-network/console/issues/1348

## Summary by CodeRabbit

* **New Features**
  * Added date range filtering to usage data, allowing users to select custom date ranges for viewing usage statistics.
  * Introduced a date range picker in the usage view for easier selection of start and end dates.

* **Bug Fixes**
  * Improved date calculations for filtering by leveraging more accurate date utilities.

* **Tests**
  * Added tests to verify correct behavior when changing the date range filter in usage components.


apps/deploy-web/src/utils/dateUtils.ts
```diff
@@ -54,3 +42,17 @@ export function getPrettyTime(timeMs: number): string {
     return `${Math.floor(timeMs / 1_000 / 60 / 60)}h ${roundDecimal(timeMs / 1_000 / 60, 2) % 60}m`;
   }
 }
+
+export function createDateRange(input: { from?: Date; to?: Date } = {}): { from: Date; to: Date } {
```
ygrishajev, line 46: just for discussion at this point:
- why `to` defaults to `now`? why not `from`? 
- why default range is 30 days?
- perhaps it makes sense to make the type smth like this for more consistency

```suggestion
export function createDateRange(input: [date1?: Date, date2: Date]): { from: Date; to: Date } {
```

baktun14 (approved):

apps/deploy-web/src/utils/dateUtils.ts
domhhv: Regarding the first two questions, those values (range, to and from) stemmed from the way Usage schema refinement was implemented: https://github.com/akash-network/console/blob/main/apps/api/src/billing/http-schemas/usage.schema.ts

Since it was a new function and I could pick any values as default ones, I decided to go with those which are already made default for the endpoint this function is being added for.

I can change them to something else, of course, as well as change the input type

PR #1718, opened 2025-07-22
feat(log-collector): improve stream resilience with Promise.allSettled

ygrishajev:
- Use Promise.allSettled() instead of Promise.all() for individual pod failure handling
- Improve error messages with specific failed pod names
- Add stream result logging for better debugging

Individual pod failures no longer kill other streams, but process still restarts on errors.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling during log collection to provide clearer messages when log streams from pods fail.

* **Tests**
  * Updated test cases to reflect the new error messaging for failed log stream scenarios.

### DIFF
--- a/apps/log-collector/src/services/k8s-log-collector/k8s-log-collector.service.ts
+++ b/apps/log-collector/src/services/k8s-log-collector/k8s-log-collector.service.ts
@@ -1,4 +1,5 @@
 import { CoreV1Api, KubeConfig, Log, V1Pod } from "@kubernetes/client-node";
+import partition from "lodash/partition";
 import { PassThrough } from "stream";
 import { singleton } from "tsyringe";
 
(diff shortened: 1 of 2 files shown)


baktun14 (approved):

apps/log-collector/src/services/k8s-log-collector/k8s-log-collector.service.ts
```diff
@@ -140,7 +141,28 @@ export class K8sLogCollectorService {
...
+      message: "Log stream results"
+    });
+
+    if (failed > 0) {
+      const failedPods = rejected.map(result => {
+        const originalIndex = results.indexOf(result);
+        return pods[originalIndex].podName;
+      });
+
+      throw new Error(`Log streams failed for pods: ${failedPods.join(", ")}`);
```
stalniy, line 164: suggestion(non-blocking): you can use AggregateError to create a single error from multiple ones, if `rejected` has some useful errors inside

ygrishajev: ok, i got smth in progress so I'll address this later for simplicity

PR #1726, opened 2025-07-22
feat: adds safe node packages installation

### LINKED ISSUE #1549: Ignore postinstall and preinstall scripts
## Why

To prevent malicious software execution, we need to forbid postinstall/preinstall scripts. We can do this with `npm --ignore-scripts` but some trusted packages need this to install binaries

## What

We need to update:
1. GHA `npm ci` -> https://github.com/akash-network/console/blob/main/.github/actions/setup-app-deps/action.yml#L30
2. Docker files -> https://github.com/akash-network/console/blob/main/packages/docker/Dockerfile.node https://github.com/akash-network/console/blob/main/packages/docker/Dockerfile.nextjs
3. GHA -> https://github.com/akash-network/console/blob/main/.github/workflows/local-packages-validate.yml

Additionally we need to create script that runs postinstall/preinstall scripts of trusted packages of trusted versions. somethins like this:

```ts
const { execSync } = require("child_process");

const trustedPackages = [
  { path: "node_modules/trusted-package-a", version: '1.0.0' }
  { path: "node_modules/trusted-package-b", version: '1.0.0' }
];

for (const pkg of trustedPackages) {
  const pkgJson = require(`${pkg.path}/package.json`);
  
  if (package.version !== pkgPath.version) {
    console.error('Skips hooks execution for required trusted package ${pkg.path} because of version mismatch (current: ${pkgJson.version}, trusted: ${pkg.version})')
    continue;
  }

  if (pkgJson.scripts && pkgJson.scripts.postinstall) {
    console.log(`Running postinstall for ${pkg.path}@${pkg.version}`);
    execSync("npm run postinstall", { cwd: pkg.path, stdio: "inherit" });
  }
}
```

stalniy:
## Why

To prevent potential security attacks from a package by using pre/postinstall hooks. Additionally we will know for what packages (and which versions!) we need to run pre/postinstall script. Usually package authors use that hooks to compile a package written in another language or to download some binaries specific to the machine which uses this dependency

Closes #1549 

## What

Adds `safe-install` npm script which is a wrapper around `npm ci --ignore-scripts`. Additionally, it reads `trustedDependencies` array from package.json and runs for every package `npm run install` if its version matches trusted version. Otherwise package is ignored.

The script doesn't fail but prints a log of errors. It's done intentionally because sometimes we may want to just ignore it and not fail CI. Anyway, if that package does something important, then CI will fail on later stage

`trustedDependencies` is not an object as other `*Depndencies` because we may want to specify multiple packages with the same name but different versions. And **yes**, it's important to ensure that package specific version and what its pre/postinstall script does. We don't have those much and will not update them often.

## Summary by CodeRabbit

* **Chores**
  * Updated dependency installation across workflows and Dockerfiles to use a new custom script that selectively runs trusted post-install scripts.
  * Added a new npm script for safe dependency installation to improve security and reliability.
  * Introduced a list of trusted dependency versions for better control and transparency.

### DIFF
--- a/.github/actions/setup-app-deps/action.yml
+++ b/.github/actions/setup-app-deps/action.yml
@@ -37,7 +37,7 @@ runs:
         APP_UNDER_TEST: ${{ inputs.app }}
       run: |
         if [[ "$APP_UNDER_TEST" == "all" ]]; then
-          npm ci
+          npm run safe-install
         else
-          npm ci -w "apps/$APP_UNDER_TEST"
+          npm run safe-install -- -w "apps/$APP_UNDER_TEST"
         fi
--- a/.github/workflows/create-pre-release-pr.yml
+++ b/.github/workflows/create-pre-release-pr.yml
@@ -76,7 +76,7 @@ jobs:
 
       - name: Install Dependencies
         if: steps.deps-cache.outputs.cache-hit != 'true'
-        run: npm ci -w packages/releaser
+        run: npm run safe-install -- -w packages/releaser
 
       - name: Generate Release Changes
         run: |
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "lint:fix": "npm run lint -- --fix",
     "prepare": "husky || true",
     "pretty": "prettier --write \"./**/*.{js,jsx,mjs,cjs,ts,tsx,json}\"",
+    "safe-install": "script/safe-deps-install.sh",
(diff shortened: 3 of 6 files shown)


ygrishajev (commented):
I can't spot the `safe-deps-install.sh` file. Where is it?

stalniy:
> I can't spot the `safe-deps-install.sh` file. Where is it?

Nice catch 😃

ygrishajev (approved):

baktun14 (approved):

Issue #1701, opened 2025-07-18
Security description in console-api open API documentation


stalniy:
## Why

We need to mark methods which require authentication, so then autogenerated client can accept that parameter

## What

hono open API configuration allows to define security schemas -> https://hono.dev/examples/hono-openapi#add-security-definitions
We support 2:
1. Bearer token
2. api key x-api-key in header

We need to add this in our documentation and then update all routes to include this information in definition

stalniy:
@jzsfkzm would you like to work on this task?

jzsfkzm:
@stalniy sure, assign it to me please.

stalniy:
@jzsfkzm could you please also think on how to enforce people to specify `security` field? The quick one out of my head is eslint rule or somehow alter typescript type to make that property required.

Without enforcement, this property will be omitted and we will end up fixing this again

PR #1733, opened 2025-07-24
refactor: stop extending Axios in HttpService, expect an instance instead

### LINKED ISSUE #1424: Refactor http services
## Why

Currently all `@akashnetwork/http-sdk` services extend AxiosInstance. It breaks encapsulation and makes it harder to configure app because every service requires configuration of the same interceptors and configurations. Take a look at https://github.com/akash-network/console/blob/main/apps/deploy-web/src/services/http-factory/http-factory.service.ts#L16

Additionally switching to 1 or few axios instances would make it easier to configure axios to use fetch adapter -> #1423 

## What

Refactor http services to NOT extend axios instance and instead utilize composition:

Having one (or few) axios instance we could configure this once and pass it to every service. For example this code:

```ts
const apiConfig = { baseURL: config.BASE_API_MAINNET_URL };
  return createContainer({
    user: () =>
      withInterceptors(new UserHttpService(apiConfig), {
        request: [config.globalRequestMiddleware, authService.withAnonymousUserHeader],
        response: [
          response => {
            if (response.config.url?.startsWith("/v1/anonymous-users") && response.config.method === "post" && response.status === 200) {
              analyticsService.track("anonymous_user_created", { category: "user", label: "Anonymous User Created" });
            }
            return response;
          }
        ]
      }),
```

would become this:

```ts
  const mainnetAdapter = axios.create({ baseURL: config.BASE_API_MAINNET_URL });
  mainnetAdapter.intercetors.use([config.globalRequestMiddleware, authService.withAnonymousUserHeader], response => {
    if (response.config.url?.startsWith("/v1/anonymous-users") && response.config.method === "post" && response.status === 200) {
      analyticsService.track("anonymous_user_created", { category: "user", label: "Anonymous User Created" });
    }
    return response;
  });
  return createContainer({
    user: () => new UserHttpService(mainnetAdapter),
    stripe: () => new StripeService(mainnetAdapter),
  })
```

jzsfkzm:
closes #1424

## Summary by CodeRabbit

* **Refactor**
  * Unified all HTTP service classes to use pre-configured Axios instances instead of configuration objects.
  * Centralized Axios instance creation and interceptor management across backend and frontend services.
  * Updated service constructors and dependency injection to accept Axios instances, simplifying service instantiation and configuration.
  * Improved analytics tracking for specific API responses by consolidating response interceptors.
  * No changes to user-facing features or interfaces.

### DIFF
--- a/apps/api/src/chain/services/block-http/block-http.service.spec.ts
+++ b/apps/api/src/chain/services/block-http/block-http.service.spec.ts
@@ -2,23 +2,26 @@ import "@test/mocks/logger-service.mock";
 
 import { BlockHttpService as BlockHttpServiceCommon } from "@akashnetwork/http-sdk";
 import { faker } from "@faker-js/faker";
+import axios from "axios";
 
 import { BlockHttpService } from "./block-http.service";
 
 describe(BlockHttpService.name, () => {
-  let service: BlockHttpService;
-  let blockHttpService: BlockHttpServiceCommon;
-
-  beforeEach(() => {
-    blockHttpService = new BlockHttpServiceCommon();
-    service = new BlockHttpService(blockHttpService);
-  });
-
   it("should get current height", async () => {
-    const height = faker.number.int({ min: 1000000, max: 10000000 });
-    jest.spyOn(blockHttpService, "getCurrentHeight").mockResolvedValue(height);
-    const result = await service.getCurrentHeight();
+    const { height, blockHttpService } = setup();
+
+    const result = await blockHttpService.getCurrentHeight();
 
     expect(result).toBe(height);
   });
+
+  const setup = () => {
+    const blockHttpServiceCommon = new BlockHttpServiceCommon(axios.create());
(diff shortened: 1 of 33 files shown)


stalniy:
WoW @jzsfkzm it's a challenging task. Don't push it so hard  though but I appreciate that you started it. Because `axiosRetry` doesn't work with this custom axios classes and throws runtime error

Though it's something we need to do but there is also an alternative way. We need to fix first https://github.com/akash-network/console/issues/1701 and then we could generate client from swagger doc. And most of this self-written stuff will be just removed and chain sdk related stuff will be hopefully replaced with chain sdk library. So, for now probably it makes sense to refactor only chain sdk related http classes

stalniy:
@jzsfkzm I will close this since amount of conflicts is huge and I refactored some services which used axiosRetry. There is a new lib which we started using recently -> https://www.npmjs.com/package/cockatiel . It supports circuit breaker additionally to retry logic

PR #1740, opened 2025-07-25
feat(log-collector): implements fs log streaming and delivery with fluent-bit

ygrishajev:
## Summary by CodeRabbit

- **New Features**
  - File-based per-namespace/pod logging with rotation, resume-from-last-timestamp, and Fluent Bit forwarding to external providers.

- **Documentation**
  - README updated for file-based pipeline, rotation/config env vars, Datadog forwarding examples, and deployment/storage guidance.

- **Chores**
  - Container runtime and startup revised: Debian-based image, non-root user, init/entrypoint starts Fluent Bit + app; added ignore files; removed embedded Datadog client dependency.

- **Tests**
  - Expanded unit tests for discovery, collectors, file destination, and error aggregation.


baktun14 (dismissed):

apps/log-collector/src/services/file-destination/file-destination.service.ts
```diff
@@ -0,0 +1,303 @@
...
+   *
+   * Reads the entire file content, splits it into lines, filters out empty
+   * lines, and returns the last non-empty line or null if no content exists.
+   *
+   * @param filePath - The absolute path of the file to read
+   * @returns Promise that resolves to the last non-empty line or null
+   * @throws Error if file reading fails
+   */
+  private async readLastLineFromFile(filePath: string): Promise<string | null> {
+    const fileContent = await this.fs.promises.readFile(filePath, "utf8");
```
stalniy: **issue(blocking):** this reads the whole file into memory, what potentially may crash nodejs process with out of memory error. Better to read file line by line. You can use stream and readline to achieve this

https://geshan.com.np/blog/2021/10/nodejs-read-file-line-by-line/

apps/log-collector/src/services/file-destination/file-destination.service.ts
```diff
@@ -0,0 +1,303 @@
...
+      if (isRotating) return; // Prevent concurrent rotations
+
+      try {
+        const stats = await this.fs.promises.stat(filePath);
+
+        if (stats.size >= this.configService.get("LOG_MAX_FILE_SIZE_BYTES")) {
+          isRotating = true;
+
+          try {
+            await this.rotateLogFile(filePath);
```
stalniy: **issue(blocking):** I think there is a possibility for this to be called 2 times and cause concurrency issues. For example, if `fs.promises.stat` takes longer than 100 ms to finish or even loop will be overloaded then `setInterval` will call its callback 2 times. So, the 2nd call will pass through `if(isRotating)` because the previous call hasn't reached the part where `isRotating = true` and then the 1st call will start rotation while the 2nd will check file stat and when it finishes and if it gets the stat of the same file as the 1st call, it will call `this.rotateLogFile`

Don't know whether it can be cause real issues but something to consider

stalniy: anyway, better to use recursive `setTimeout` because then such issues are not possible because of the implementation. The next `setTimeout` won't be called until `fs.stat` is resolved

stalniy (changes requested):

stalniy:
Those issues which I pointed in the changes better to address in a separate PR because this one is big by itself.

stalniy (dismissed):

apps/log-collector/src/services/pod-logs-collector/pod-logs-collector.service.ts
```diff
@@ -0,0 +1,251 @@
...
+        const outputContent = outputLines.join("\n");
+        writeStream.write(outputContent + "\n");
+      }
+
+      isFirstChunk = false;
+    });
+
+    if (remainingBuffer.length > 0) {
+      writeStream.write(remainingBuffer);
+    }
```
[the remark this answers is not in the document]
ygrishajev: @coderabbitai is this still an issue?

apps/log-collector/src/services/pod-discovery/pod-discovery.service.spec.ts
```diff
@@ -0,0 +1,354 @@
...
+    const { podDiscoveryService, k8sClient } = setup({
+      KUBERNETES_NAMESPACE_OVERRIDE: overrideNamespace
+    });
+
+    k8sClient.listNamespacedPod.mockResolvedValue({ items: [] });
+
+    await podDiscoveryService.discoverPodsInNamespace();
+
+    expect(k8sClient.listNamespacedPod).toHaveBeenCalledWith({ namespace: overrideNamespace });
+  });
```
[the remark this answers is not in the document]
ygrishajev: no, this is correct:

```ts
    listNamespacedPod(param: CoreV1ApiListNamespacedPodRequest, options?: ConfigurationOptions): Promise<V1PodList>;

export interface CoreV1ApiListNamespacedPodRequest {
    /**
     * object name and auth scope, such as for teams and projects
     * Defaults to: undefined
     * @type string
     * @memberof CoreV1ApilistNamespacedPod
     */
    namespace: string;
//.....
}
```

apps/log-collector/src/services/pod-logs-collector/pod-logs-collector.service.ts
```diff
@@ -0,0 +1,250 @@
...
+        message: "Starting log collection for container"
+      });
+
+      const k8sPromise = this.startKubernetesLogStream(containerName, logStream, lastTimestamp);
+
+      await Promise.all([writePromise, k8sPromise]);
+    });
+
+    await this.errorHandlerService.aggregateConcurrentResults(collectionPromises, "container log collection");
+  }
```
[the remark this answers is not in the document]
ygrishajev: this is a long running process handled in the root entry point. It is expected to be pending while worker if functional and it is expected to fail the process if needed

stalniy (dismissed):

apps/log-collector/src/services/file-destination/file-destination.service.ts
```diff
@@ -0,0 +1,441 @@
...
+            if (isEnded) break;
+
+            const stats = await this.fs.promises.stat(filePath);
+            if (stats.size >= maxFileSize) {
+              try {
+                stableStream.unpipe(currentWriteStream!);
+                currentWriteStream!.end();
+
+                await new Promise<void>((resolve, reject) => {
+                  currentWriteStream!.once("close", resolve);
```
stalniy: **nitpick:** `events` module has `once` helper which does exactly what you did manually -> https://nodejs.org/api/events.html#eventsonceemitter-name-options

PR #1766, opened 2025-07-31
feat(auth): implements managed wallet API JWT auth

### LINKED ISSUE #1038: Managed wallet api - Implement JWT authentication
## Why

Instead of asking for customers to create their own certificate, we need to implement the new JWT authentication for providers and remove certificates from endpoints.

## Solution

Implement as per the JWT signing as per specs here https://github.com/akash-network/akash-api/pull/197

- [x] New package that generates a json object from the JSON schema in akash-api, like the net package
  - Class to generate and sign a valid ES256K with secp256k1 curve jwt token
- [ ] Refactor managed wallet api to use jwt instead of certificates

jzsfkzm:
closes #1038

## Summary by CodeRabbit

- New Features
  - JWT-based provider authentication and a Provider JWT token service for provider calls.
  - API key creation now returns the visible token; reads/updates return a hidden view.
  - Added sample NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY env var.

- Refactor
  - Deployment and lease flows moved to a wallet-driven model (derive wallet from authenticated user); certificates are now optional.
  - Provider interactions unified with JWT/MTLS auth, repository-backed lookups, retries, and improved error reporting.

- Chores
  - Added runtime dependencies for JWT support and retry utilities.


baktun14:
This is great, thanks @jzsfkzm.

A few things missing to consider this issue closed:
- Make sure the [lease flow](https://github.com/akash-network/console/blob/main/apps/api/examples/lease-flow.ts) example is updated and doesn't use certificates
- Right now, only the lease creation/update returns the lease status because they're the only endpoints that received the certificates. I wanted to add it to the get deployment list/details but it wasn't realistic with passing certs as query params or doing post query for get

baktun14:
Also, I haven't done any benchmarks on jwt generation, but if it takes some time to generate a new one every time, we could keep a cache per address for like a few minutes.

apps/api/src/deployment/http-schemas/deployment.schema.ts
```diff
@@ -101,11 +101,7 @@ export const DepositDeploymentResponseSchema = z.object({
 
 export const UpdateDeploymentRequestSchema = z.object({
   data: z.object({
-    sdl: z.string(),
-    certificate: z.object({
-      certPem: z.string(),
-      keyPem: z.string()
-    })
```
stalniy, line 108: @baktun14 this is an issue because managed wallet API is consider "public" and we introduce a breaking change. This will break automations of our customers who relies on managed wallet API

stalniy: Ideally, managed wallet API should be separated under a different namespace to make it more obvious that this API is public and cannot be just changed

apps/api/src/provider/services/jwt-token/jwt-token.service.ts
```diff
@@ -0,0 +1,32 @@
...
+
+const JWT_TOKEN_TTL_IN_SECONDS = 30;
+
+@singleton()
+export class JwtTokenService {
+  constructor(@InjectBillingConfig() private readonly config: BillingConfig) {}
+
+  async generateJwtToken(walletId: number) {
+    const wallet = new Wallet(this.config.MASTER_WALLET_MNEMONIC, walletId);
+    const akashWallet = await createSignArbitraryAkashWallet(await wallet.getInstance());
```
stalniy: suggestion(non-blocking): would be nice to cache it

apps/api/src/provider/services/jwt-token/jwt-token.service.ts
```diff
@@ -0,0 +1,32 @@
...
+    const now = Math.floor(Date.now() / 1000);
+
+    const token = await jwtToken.createToken({
+      iss: akashWallet.address,
+      exp: now + JWT_TOKEN_TTL_IN_SECONDS,
+      nbf: now,
+      iat: now,
+      jti: uuid.v4(),
+      version: "v1",
+      leases: { access: "full" }
```
stalniy: **question(blocking):** is it possible to generate per deployment access token? or what is the reason to make a full access token?

apps/api/src/provider/services/provider/provider.service.ts
```diff
@@ -29,12 +29,13 @@ export class ProviderService {
...
     private readonly auditorsService: AuditorService,
-    @InjectBillingConfig() private readonly config: BillingConfig
+    @InjectBillingConfig() private readonly config: BillingConfig,
+    private readonly jwtTokenService: JwtTokenService
   ) {
     this.chainNetwork = this.config.NETWORK as SupportedChainNetworks;
   }
 
-  async sendManifest(provider: string, dseq: string, manifest: string, options: { certPem: string; keyPem: string }) {
+  async sendManifest(provider: string, dseq: string, manifest: string, walletId: number) {
```
stalniy: suggestion(non-blocking): would be nice to convert all these parameters into an object

apps/api/src/provider/services/provider/provider.service.ts
```diff
@@ -89,8 +92,10 @@ export class ProviderService {
 
     const response = await this.providerProxy.fetchProviderUrl<LeaseStatusResponse>(`/lease/${dseq}/${gseq}/${oseq}/status`, {
       method: "GET",
-      certPem: options.certPem,
-      keyPem: options.keyPem,
+      headers: {
+        Authorization: `Bearer ${await this.jwtTokenService.generateJwtToken(walletId)}`,
+        "Content-Type": "application/json"
```
stalniy: **question(blocking):** provider proxy doesn't support JWT authentication. Do I miss something?

stalniy: with JWT, you don't need to use provider-proxy, you can directly talk to provider API

apps/api/src/deployment/http-schemas/deployment.schema.ts
baktun14: We only "soft released" to some select customers and warned them that the api will change in the future, so this should be fine. But yes in the future we should release a different version `/v2` for example of these endpoints.

apps/api/src/provider/services/provider/provider.service.ts
jzsfkzm: Did it.

apps/api/src/provider/services/jwt-token/jwt-token.service.ts
jzsfkzm: Uhm, a full access token is indeed not right. My knowledge about this is kind of limited, but tried to scope it to the provider now, is that what you would like to see?

apps/api/src/provider/services/jwt-token/jwt-token.service.ts
jzsfkzm: Memoized the intermediary steps for 5 minutes, but we cannot cache the whole token long, or the token TTL would break. Or we can raise the token TTL to the same time window. What do you think?

apps/api/src/provider/services/provider/provider.service.ts
jzsfkzm: Changed it and now it's communicating right with the provider, would you check?

apps/api/src/deployment/http-schemas/deployment.schema.ts
jzsfkzm: So we should still have it working in v1 with cert, and v2 with JWT token? In this change? I can do that, just please tell me if we need that.

baktun14: Actually, can we just make sure that the api endpoint is callable with or without the certificates? That way the endpoints will continue to work without any changes and just ignore the certs in the endpoint. I think this is just making sure zod doesn't throw an error if additional properties are sent to the body of the endpoints that were expecting a cert previously.

stalniy: JWT will be generated under the hood, so from usage perspective even if we change this API and remove certificates from payload, customer's scripts should still work. They will send certificates but we will ignore them. Just need to ensure that hono doesn't throw an error if there are additional keys sent

apps/api/src/provider/services/jwt-token/jwt-token.service.ts
stalniy: My comment was related to akashWallet caching only, so what you did is fine 👍

apps/api/src/provider/services/jwt-token/jwt-token.service.ts
stalniy: I created [a PR](https://github.com/akash-network/console/pull/1794) to improve types in JWT package, so it's easier to "guess" how it should look.

So, in this case (you are creating a provider lease, right?) probably the best permissions are:
```
{
   leases: { 
      access: "granular",
      permissions: [
        { provider: "...", access: "scoped", scope: ["send-manifest"] }
      ]
   }
}
```

Ideally this service should accept `JWTTokenPayload['leases']` parameter to generate a token we need in particular case

apps/api/src/provider/services/jwt-token/jwt-token.service.ts
```diff
@@ -0,0 +1,54 @@
+import { createSignArbitraryAkashWallet, JwtToken } from "@akashnetwork/jwt";
+import { minutesToSeconds } from "date-fns";
+import { singleton } from "tsyringe";
+import * as uuid from "uuid";
+
+import { Wallet } from "@src/billing/lib/wallet/wallet";
+import { BillingConfig, InjectBillingConfig } from "@src/billing/providers";
+import { Memoize } from "@src/caching/helpers";
+
+const JWT_TOKEN_TTL_IN_SECONDS = 30;
```
stalniy, line 12: **suggestion(non-blocking):** after merging changes to jwt, I'd like you to allow to pass this option into `JwtTokenService` with default to 30 secs or even less

apps/api/src/provider/services/jwt-token/jwt-token.service.ts
```diff
@@ -0,0 +1,54 @@
...
+  async generateJwtToken({ walletId, provider }: { walletId: number; provider: string }) {
+    const { jwtToken, address } = await this.getJwtToken(walletId.toString());
+    const now = Math.floor(Date.now() / 1000);
+
+    const token = await jwtToken.createToken({
+      iss: address,
+      exp: now + JWT_TOKEN_TTL_IN_SECONDS,
+      nbf: now,
+      iat: now,
+      jti: uuid.v4(),
```
stalniy, line 43: **question(non-blocking):** what do you think about using token prefixes as github does?

packages/http-sdk/src/provider/provider-http.service.ts
```diff
@@ -11,4 +11,32 @@ export class ProviderHttpService extends HttpService {
   async getProvider(address: string): Promise<GetProviderResponse> {
     return this.extractData(await this.get<GetProviderResponse>(`/akash/provider/v1beta3/providers/${address}`));
   }
+
+  async sendManifest({ hostUri, dseq, jsonStr, jwtToken }: { hostUri: string; dseq: string; jsonStr: string; jwtToken: string }) {
```
stalniy: **suggestion(blocking):**  Please rename `jsonStr` to `manifest`

jzsfkzm: Done.

apps/api/src/provider/services/jwt-token/jwt-token.service.ts
jzsfkzm: Do you mean prefixing the `jti` field? Should I add a param for that to the method? What would be the use of that? Should it have a default?

apps/api/src/provider/services/jwt-token/jwt-token.service.ts
jzsfkzm: Made it configurable. I'd keep the 30s default, considering Rabbit's suggestion [about network reasons](https://github.com/akash-network/console/pull/1766#discussion_r2246438544). What do you think?

apps/api/src/provider/services/jwt-token/jwt-token.service.ts
jzsfkzm: Thanks, made changes here, now it's scoped.

apps/api/src/deployment/http-schemas/deployment.schema.ts
jzsfkzm: Right, I just changed tests now, we're checking endpoints are working with or without cert provided.

baktun14: Sweet thanks

apps/api/src/provider/services/jwt-token/jwt-token.service.ts
jzsfkzm: @stalniy rebased my PR to main so now the branch is using these altered types.

baktun14:
@jzsfkzm We're close to merging this one, the testnet providers will be updated soon to support jwt.

jzsfkzm:
> @jzsfkzm We're close to merging this one, the testnet providers will be updated soon to support jwt.

Yay! 🥹 I'm keeping this one rebased to main.

apps/api/src/provider/services/jwt-token/jwt-token.service.ts
```diff
@@ -0,0 +1,60 @@
...
+};
+
+type GenerateJwtTokenParams = {
+  walletId: number;
+  leases: JwtTokenPayload["leases"];
+  ttl?: number;
+};
+
+@singleton()
+export class JwtTokenService {
```
ygrishajev: I think this name can cause quite some confusion with console-api server auth. Would it make sense to rename it to something like 
```suggestion
export class ProviderJwtTokenService {
```

apps/api/src/provider/services/jwt-token/jwt-token.service.ts
```diff
@@ -0,0 +1,60 @@
...
+      version: "v1",
+      leases
+    });
+
+    return token;
+  }
+
+  @Memoize({ ttlInSeconds: minutesToSeconds(5) })
+  private async getJwtToken(walletId: string): Promise<JwtTokenWithAddress> {
+    const wallet = new Wallet(this.config.MASTER_WALLET_MNEMONIC, Number(walletId));
```
ygrishajev: looks like it is casted to string in the calling method and back to number in here. I guess it makes sense to update the arg to number and remove casting, right?

apps/api/src/deployment/services/deployment-reader/deployment-reader.service.ts
```diff
@@ -91,12 +83,23 @@ export class DeploymentReaderService {
...
+    const wallet = await this.getWalletByAddress(owner);
+    const leaseStatuses = await Promise.all(
+      leaseResults.map(async ({ leases }) => {
+        return await Promise.all(
+          leases.map(async ({ lease }) => {
+            return await this.providerService.getLeaseStatus(lease.lease_id.provider, lease.lease_id.dseq, lease.lease_id.gseq, lease.lease_id.oseq, wallet.id);
+          })
+        );
+      })
+    );
```
ygrishajev: looks like it would make sense to put it inside PromisePool callback above, it does some http calls

apps/api/src/deployment/services/deployment-reader/deployment-reader.service.ts
```diff
@@ -285,4 +288,13 @@ export class DeploymentReaderService {
       other: deploymentData
     };
   }
+
+  private async getWalletByAddress(address: string) {
+    const wallet = await this.userWalletRepository.findOneBy({ address });
+    if (!wallet) {
+      throw new Error(`Wallet not found for address: ${address}`);
```
ygrishajev: how is this error handled? looks like it'll cause 500 on the responses, which would cause some error alerts from monitoring.

apps/api/src/provider/services/provider/provider.service.ts
```diff
@@ -47,55 +42,65 @@ export class ProviderService {
...
-        const providerError = err instanceof AxiosError && err.response?.data;
-        assert(!providerError?.toLowerCase()?.includes("invalid manifest"), 400, err?.response?.data);
 
-        throw new Error(providerError || err);
+        const providerError = error instanceof AxiosError && error.response?.data;
+        if (typeof providerError === "string") {
+          assert(!providerError.toLowerCase().includes("invalid manifest"), 400, providerError);
+        }
+
+        throw new Error(providerError || error);
```
ygrishajev: non-blocking: there is a nice utility we use in other places for retries `exponential-backoff`. Just leaving the comment for more visibility for everyone, I know it's not your code @jzsfkzm

apps/api/src/provider/services/provider/provider.service.ts
```diff
@@ -47,55 +42,65 @@ export class ProviderService {
...
+            provider: providerIdentity.owner,
+            scope: ["send-manifest"]
+          })
         });
+        const result = await this.providerHttpService.sendManifest({ hostUri: providerIdentity.hostUri, dseq, manifest, jwtToken });
 
-        if (result) return result;
-      } catch (err: any) {
-        if (err.message?.includes("no lease for deployment") && i < this.MANIFEST_SEND_MAX_RETRIES) {
+        if (result) {
```
ygrishajev: non-blocking: if there is no result, is it `undefined`? perhaps it should be returned unconditionally?

apps/api/src/provider/services/jwt-token/jwt-token.service.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+
+import * as WalletModule from "@src/billing/lib/wallet/wallet";
+import type { BillingConfig } from "@src/billing/providers";
+import { JwtTokenService } from "./jwt-token.service";
+
+import { createAkashAddress } from "@test/seeders";
+
+describe("JwtTokenService", () => {
+  const mockMnemonic = "test mnemonic phrase for testing purposes only";
+  const mockAddress = "akash1testaddress123456789";
```
ygrishajev: https://github.com/akash-network/console/blob/main/apps/api/test/seeders/akash-address.seeder.ts

apps/api/src/provider/services/jwt-token/jwt-token.service.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+  describe("generateJwtToken", () => {
+    it("should generate a JWT token successfully", async () => {
+      const { jwtTokenService, mockWalletId, mockWallet, leases } = setup();
+
+      const now = Math.floor(Date.now() / 1000);
+      jest.spyOn(Date, "now").mockReturnValue(now * 1000);
+
+      const result = JSON.parse(await jwtTokenService.generateJwtToken({ walletId: mockWalletId, leases }));
+
+      expect(result.leases).toEqual(leases);
```
ygrishajev: non-blocking: `leases` name made me wonder and go checking how it is related to jwt

apps/api/src/provider/services/jwt-token/jwt-token.service.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+
+    const mockWallet = mock<WalletModule.Wallet>();
+    const mockDirectSecp256k1HdWallet = mock<DirectSecp256k1HdWallet>();
+    mockDirectSecp256k1HdWallet.getAccounts.mockResolvedValue([{ address: mockAddress, pubkey: new Uint8Array([1, 2, 3]), algo: "secp256k1" }]);
+    mockWallet.getInstance.mockResolvedValue(mockDirectSecp256k1HdWallet);
+
+    const mockBillingConfig = mock<BillingConfig>();
+    mockBillingConfig.MASTER_WALLET_MNEMONIC = mockMnemonic;
+
+    jest.spyOn(WalletModule, "Wallet").mockImplementation(() => mockWallet);
```
ygrishajev: non-blocking. I don't think it makes sense to rework this. However just for the sake of discussion.

We generally prefer to avoid jest global mocking. This is not the one but the sense is the same. When something like this is needed most probably there is something wrong with dependencies design. Like in this case I'm thinking it would make sense to create a factory wired into DI. Then it would be easier to provide stubs same as we do with other injected deps.

apps/api/src/provider/services/jwt-token/jwt-token.service.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+
+    const mockAkashWallet = {
+      address: mockAddress,
+      pubkey: new Uint8Array([1, 2, 3]),
+      signArbitrary: jest.fn().mockResolvedValue({ signature: "test-signature" })
+    };
+
+    jest.spyOn(JwtModule, "createSignArbitraryAkashWallet").mockResolvedValue(mockAkashWallet);
```
ygrishajev: non-blocking. This is similar to my prev comment about deps/DI. This whole lib can be provided as an optional dep, so in real app it would be provided via DI and in tests we'd be able to pass whatever stubs we want.

apps/api/src/provider/services/jwt-token/jwt-token.service.ts
jzsfkzm: Problem is with Memoize, it works well with string args, but not numbers. Maybe we should patch Memoize to try and toString() each args, what do you think?
https://github.com/jzsfkzm/console/blob/a1501f1d245bf435fadd52103bf765ffade09545/apps/api/src/caching/helpers.ts#L30-L31

apps/api/src/deployment/services/deployment-reader/deployment-reader.service.ts
jzsfkzm: What do your prefer in this case? Should we throw 404 here?

apps/api/src/provider/services/jwt-token/jwt-token.service.ts
ygrishajev: Ah, i see now! 
Perhaps let's just also allow numbers there? It's going to be stringified with a join anyway

apps/api/src/deployment/services/deployment-reader/deployment-reader.service.ts
ygrishajev: It would be the easiest i guess

apps/api/src/provider/services/jwt-token/jwt-token.service.spec.ts
jzsfkzm: What would be a better name here? Type is `JwtTokenPayload["leases"]`, that's how it ended up to be `leases`.

ygrishajev: Well, i just shared my experience on reading this line;)

Perhaps leasesPermissions/scope/rules/etc would do. This is a non-blocking comment though

apps/api/test/mocks/config-service.mock.ts
```diff
@@ -1,13 +1,25 @@
...
-    return super.get(key);
-  }
-}
+/**
+ * Creates a mock for a ConfigService subclass.
+ * You can pass a partial map of config keys to values; `get(key)` will return those.
+ * If a key isn't provided, the mock will throw (fail fast in tests).
+ */
+export const mockConfigService = <T extends ConfigService<any, any>>(values: Partial<ConfigOf<T>> = {}): MockProxy<T> => {
+  const svc = mock<T>();
```
stalniy, line 17: **suggestion(non-blocking):** it's just possible to do this:

```ts
mock<ConfigService>({
  get: (key) =>({
    PROP: '...'
  }[key])
})
```

baktun14 (dismissed):
👌 Very nice!

apps/api/src/provider/services/provider/provider.service.ts
```diff
@@ -26,131 +27,101 @@ export class ProviderService {
...
-        const providerError = err instanceof AxiosError && err.response?.data;
-        assert(!providerError?.toLowerCase()?.includes("invalid manifest"), 400, err?.response?.data);
+
+        const providerError = err instanceof AxiosError && (err.response?.data?.message || err.response?.data);
+        if (typeof providerError === "string") {
+          assert(!providerError.toLowerCase().includes("invalid manifest"), 400, providerError);
+          assert(!providerError.toLowerCase().includes("unauthorized access"), 401, providerError);
+        }
 
         throw new Error(providerError || err);
```
stalniy: **suggestion(non-blocking):** throw original error since there is no reason to wrap it into additional one. 
```suggestion
        throw err;
```

apps/api/src/provider/services/provider-jwt-token/provider-jwt-token.service.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+    jest.spyOn(WalletModule, "Wallet").mockImplementation(() => mockWallet);
+
+    const mockAkashWallet = {
+      address: mockAddress,
+      pubkey: new Uint8Array([1, 2, 3]),
+      signArbitrary: jest.fn().mockResolvedValue({ signature: "test-signature" })
+    };
+
+    jest.spyOn(JwtModule, "createSignArbitraryAkashWallet").mockResolvedValue(mockAkashWallet);
+    jest.spyOn(JwtModule, "JwtToken").mockImplementation(
```
stalniy: **issue(non-blocking):** mocking modules may become a blocker to switching to ES modules in future

apps/api/src/deployment/services/deployment-writer/deployment-writer.service.ts
```diff
@@ -29,10 +31,12 @@ export class DeploymentWriterService {
     private readonly sdlService: SdlService,
     private readonly billingConfig: BillingConfigService,
     private readonly providerService: ProviderService,
-    private readonly deploymentReaderService: DeploymentReaderService
+    private readonly deploymentReaderService: DeploymentReaderService,
+    private readonly walletReaderService: WalletReaderService
```
stalniy, line 35: **issue(blocking):** walletReaderService depends on AuthService which depends on ExecutionContextService. This will break pg boss workers

stalniy: we probably somehow need to mark those services which depend on ExecutionContext or agree on the boundaries. For example, that ExecutionContext cannot be used in application services, only in controllers

stalniy: We could also run pgboss workers in ExecutionContext but we have an issue with improper types. https://github.com/akash-network/console/blob/main/apps/api/src/auth/services/auth.service.ts#L17 the same for Ability. Eventually we still may face hard to debug/resolve runtime issue because of this

apps/api/src/deployment/services/deployment-reader/deployment-reader.service.ts
```diff
@@ -78,10 +81,14 @@ export class DeploymentReaderService {
     };
   }
 
-  public async list(
-    owner: string,
-    { skip, limit }: { skip?: number; limit?: number }
-  ): Promise<{ deployments: GetDeploymentResponse["data"][]; total: number; hasMore: boolean }> {
+  public async list({
```
stalniy, line 82: **issue(non-blocking):** it's non-obvious that it lists deployments for current user

apps/api/src/deployment/services/deployment-reader/deployment-reader.service.ts
```diff
@@ -20,14 +22,22 @@ export class DeploymentReaderService {
     private readonly providerService: ProviderService,
     private readonly deploymentHttpService: DeploymentHttpService,
     private readonly leaseHttpService: LeaseHttpService,
-    private readonly messageService: MessageService
+    private readonly messageService: MessageService,
+    private readonly userWalletRepository: UserWalletRepository,
+    private readonly walletReaderService: WalletReaderService
```
stalniy, line 25: **issue(blocking):** this is potentially the same issue as in DeploymentWriterService: https://github.com/akash-network/console/pull/1766#discussion_r2396438335

apps/api/src/caching/helpers.ts
```diff
@@ -27,7 +27,7 @@ export const Memoize = (options?: MemoizeOptions) => (target: object, propertyNa
     const argsKey =
       args.length > 0
         ? `${cacheKey}#${args
-            .map(arg => (typeof arg === "string" ? arg : null))
+            .map(arg => (["string", "number"].includes(typeof arg) ? String(arg) : null))
```
stalniy, line 30: **suggestion(non-blocking):** this should also include boolean, bigint, null, Date. and there should be possibility to pass a function to generate custom cacheKey. Automatic detection may not work in all cases, it can even cause issues in cases somebody passes an object of options which significantly change the return value but its value is not taken into account in cacheKey

stalniy (changes requested):

apps/api/src/caching/helpers.ts
ygrishajev: I totally agree, however this change was done by @jzsfkzm to only fix this case for the purpose of this PR. I'd prefer to not proceed with inessential changes as PR is already grew a bit

stalniy (dismissed):
До

baktun14 (dismissed):

apps/api/src/billing/lib/wallet/wallet.ts
```diff
@@ -46,4 +50,11 @@ export class Wallet implements OfflineDirectSigner {
   async getMnemonic() {
     return (await this.instanceAsPromised).mnemonic;
   }
+
+  async getInstance() {
+    return await this.instanceAsPromised;
+  }
```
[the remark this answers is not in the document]
stalniy: @ygrishajev this is valuable advice. Basically `Wallet` can implement signArbitrary interface and you could pass the whole wallet instance to JwtToken or `Wallet` could have `getSignArbitraryMethod` that wraps createSignArbitrary.

but this is non-blocking and can be done later

PR #1790, opened 2025-08-06
feat(billing): implement charges csv export

### LINKED ISSUE #1788: Usage -csv only exports current view
1. Billing & Usage - > Usage Tab
2. Select a longer time range and export CSV
3. It only exports the currently viewed page because of the pagination. nNeed to programmatically loop through all the pages or create a dedicated endpoint to get the export.

domhhv:
Closes #1788 

## Summary by CodeRabbit

- **New Features**
  - Streamed, timezone-aware Stripe transactions CSV export endpoint with date-based filename and user-friendly message when payments aren’t configured.
- **Refactor**
  - Centralized date-range validation; date-range APIs now require both start/end dates. UI exposes an Export action, surfaces errorMessage strings, and downloads CSV as a Blob.
- **Tests**
  - Added tests for CSV streaming (pagination, empty results, payment-method edge cases), schema validations, DOM download utility, and UI export flows.
- **Chores**
  - Added CSV stringifier dependency and test seeders/types.


anilmurty (dismissed):

apps/api/src/billing/services/stripe/stripe.service.ts
```diff
@@ -413,6 +413,65 @@ export class StripeService extends Stripe {
     };
   }
 
+  async exportTransactionsCsv(customerId: string, options: { startDate: string; endDate: string }): Promise<string> {
```
stalniy: **issue(blocking):** I don't know how big csv files can be but it looks like you are loading a lot of data into the memory in order to generate it and the final CSV file is also stored in memory. So, if multiple people will call this endpoint nodejs process can die with Out of memory exception.

Instead of loading all of this into memory, you need to use streaming and load only a chunk of data, transform it into csv and send to the client. So, you have few options:

1. If customer transactions are fully available on frontend, then you can paginate over it on frontend and build CSV on frontend. I used [papaparse](https://www.papaparse.com/) on the prev project.
2. if customer transactions are not available on frontend or csv format needs to be accessed on API, you can use [csv-stringify](https://www.npmjs.com/package/csv-stringify) library (don't use its `/sync` API!). You can create async generator which goes through the data, then pipe it to csv-stringify and pipe that stream to nodejs response (Hono supports this, also has own streaming helpers).

Image

stalniy: `myWritableStream` in your case will be stream created using csv-stringify library

apps/api/src/billing/services/stripe/stripe.service.ts
```diff
@@ -413,6 +413,65 @@ export class StripeService extends Stripe {
...
+        startDate: options.startDate,
+        endDate: options.endDate
+      });
+
+      allTransactions.push(...batch.transactions);
+      hasMore = batch.hasMore;
+      startingAfter = batch.nextPage || undefined;
+
+      if (allTransactions.length > 50000) {
+        throw new Error("Transaction export limit exceeded. Please use a smaller date range.");
```
stalniy: **question(blocking):** wouldn't this result in 500 response error?

stalniy: it'd be less important if you used streaming. We could just have a limit on data range, if we need this because of some limitations

stalniy (changes requested):

apps/api/src/billing/services/stripe/stripe.service.ts
baktun14: Limit on date range makes sense

baktun14 (commented):
Looks good! Just need to address the potentially memory issues

apps/api/src/billing/services/stripe/stripe.service.ts
domhhv: I implemented the 2nd option as the transactions count can't be determined on neither side: Stripe [recently stopped](https://docs.stripe.com/changelog/basil/2025-03-31/deprecate-total-count-expansion) exposing `total_count` in its API.

I used csv-stringify library with streaming to avoid OOM issues.

apps/api/src/billing/services/stripe/stripe.service.ts
domhhv: I removed this error since I implemented streaming. There's already a date range limitation on the query schema for this route.

apps/api/src/billing/routes/stripe-transactions/stripe-transactions.router.ts
```diff
@@ -102,15 +102,31 @@ stripeTransactionsRouter.openapi(getCustomerTransactionsRoute, async function ge
...
+      try {
+        for await (const chunk of csvStream) {
+          controller.enqueue(new TextEncoder().encode(chunk));
+        }
+        controller.close();
+      } catch (error) {
+        controller.error(error);
+      }
+    }
+  });
```
stalniy: @coderabbitai can this be replaced with `Readable.from` or `Readable.toWeb`?

apps/api/src/billing/services/stripe/stripe.service.spec.ts
```diff
@@ -404,40 +391,42 @@ describe(StripeService.name, () => {
...
 
-      const expectedCsv = [
-        '"Transaction ID","Date","Amount","Currency","Status","Payment Method","Card Brand","Card Last 4","Description","Receipt URL"',
-        '"ch_123","2022-01-01","10.00","USD","succeeded","card","visa","4242","Test payment","https://receipt.url/123"',
-        '"ch_456","2022-01-02","25.00","EUR","succeeded","card","mastercard","8888","",""'
-      ].join("\n");
+      const chunks: string[] = [];
+      for await (const chunk of csvStream) {
+        chunks.push(chunk);
+      }
```
stalniy:
```suggestion
      const chunks = await Array.fromAsync(csvStream)
```

apps/api/src/billing/services/stripe/stripe.service.ts
```diff
@@ -427,49 +461,58 @@ export class StripeService extends Stripe {
...
-        transaction.description || "",
-        transaction.receiptUrl || ""
-      ];
-    });
+    return value;
+  }
 
-    const csvContent = [headers, ...rows].map(row => row.map(field => `"${String(field).replace(/"/g, '""')}"`).join(",")).join("\n");
+  private transformTransactionForCsv(transaction: Transaction) {
+    const date = new Date(transaction.created * 1000).toISOString().split("T")[0];
```
stalniy: **question(non-blocking):** is it expected for the api caller that this date will be in UTC?

apps/deploy-web/src/components/billing-usage/BillingContainer/BillingContainer.tsx
```diff
@@ -119,12 +123,21 @@ export const BillingContainer: React.FC<BillingContainerProps> = ({ children, de
     setDateRange(createDateRange(range));
   };
 
+  const exportCsv = () => {
+    if (startDate && endDate) {
+      exportMutation.mutate({ startDate, endDate });
+    } else {
+      setError(new Error("Please select a valid date range before exporting."));
```
stalniy: **suggestion(non-blocking): ** this is expected error and using `Error` in this case doesn't have much sense. You could instead use `{ code: "DATE_RANGE_UNKNOWN" }`. Then it's easy to check what error has been passed

apps/deploy-web/src/queries/usePaymentQueries.ts
```diff
@@ -106,3 +113,27 @@ export const usePaymentMutations = () => {
     removePaymentMethod
   };
 };
+
+export const useExportTransactionsCsvMutation = () => {
+  const { stripe } = useServices();
+
+  return useMutation<Blob, Error, ExportTransactionsCsvParams>({
```
stalniy: **question(non-blocking):** this doesn't sound as a mutation. We just query data in CSV format. But does it actually mutate the application state?

apps/api/src/billing/routes/stripe-transactions/stripe-transactions.router.ts
stalniy: @domhhv please update the code to use built-in nodejs helpers

domhhv: @coderabbitai After applying the code you suggested, I'm seeing this TypeScript error:

```
TS2345: Argument of type ReadableStream<any> is not assignable to parameter of type BodyInit | null | undefined
Type import("stream/web").ReadableStream<any> is not assignable to type ReadableStream<any>
Types of property pipeThrough are incompatible.
Type '<T>(transform: import("stream/web").ReadableWritablePair<T, any>, options?: import("stream/web").StreamPipeOptions | undefined) => import("stream/web").ReadableStream<T>' is not assignable to type '<T>(transform: ReadableWritablePair<T, any>, options?: StreamPipeOptions | undefined) => ReadableStream<T>'.
Types of parameters transform and transform are incompatible.
Type ReadableWritablePair<T, any> is not assignable to type ReadableWritablePair<T | undefined, any>
Types of property readable are incompatible.
Type ReadableStream<T> is missing the following properties from type ReadableStream<T | undefined>: values, [Symbol.asyncIterator
```

It seems that there are two different ReadableStream types in this environment:

1. Node.js Web Streams API (import("stream/web").ReadableStream) - what Readable.toWeb() returns
2. Web API ReadableStream - what the Response constructor expects

Can this issue be addressed while utilizing the Node.js built-ins?

apps/api/src/billing/services/stripe/stripe.service.ts
domhhv: I think this is an important note. The user could call this endpoint from any part of the globe by picking the date range on the picker (in local timezone). The endpoint receives both startDate and endDate with timezone data. I should use it to properly filter the transactions.

apps/deploy-web/src/queries/usePaymentQueries.ts
domhhv: `useQuery` makes indeed more sense here

apps/deploy-web/src/components/billing-usage/BillingContainer/BillingContainer.tsx
domhhv: I'm only passing this error down to BillingView where I extract the message. I think it makes more sense to just setErrorMessage instead and pass the string message down.

apps/api/src/billing/services/stripe/stripe.service.spec.ts
domhhv: I will need to add `"lib": ["esnext"]` to `apps/api/tsconfig.json` so that TypeScript recognises more modern `Array.fromAsync()` here

apps/api/src/billing/routes/stripe-transactions/stripe-transactions.router.ts
domhhv: @coderabbitai

apps/deploy-web/src/components/billing-usage/BillingContainer/BillingContainer.spec.tsx
```diff
@@ -6,7 +6,7 @@ import type { PaginationState } from "@tanstack/react-table";
 import type { AxiosError } from "axios";
 import { mock } from "jest-mock-extended";
 
-import type { usePaymentTransactionsQuery } from "@src/queries";
+import type { useExportTransactionsCsvMutation, usePaymentTransactionsQuery } from "@src/queries";
```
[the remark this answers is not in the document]
domhhv: Our rule `@typescript-eslint/consistent-type-imports` won't allow me to avoid type-only imports here

apps/deploy-web/src/queries/usePaymentQueries.ts
domhhv: `useQuery` is used instead of `useMutation` now

apps/deploy-web/src/components/billing-usage/BillingContainer/BillingContainer.tsx
domhhv: `setErrorMessage<string>` is used instead of `setError<Error>` now

apps/api/src/billing/services/stripe/stripe.service.ts
domhhv: I'm now extracting the timezone offset from UTC inside `*createTransactionGenerator()`, passing it to `transformTransactionForCsv()` here and re-applying to the date it when returning a transaction item. This should provide expected dates in the report.

Transactions are already properly filtered on the Stripe side since I'm passing timestamps of start and end dates that I'm getting from `Date.getTime()`, which is UTC-based, inside `getCustomerTransactions()`

apps/api/src/billing/services/stripe/stripe.service.spec.ts
domhhv: Extended `lib` in the api's tsconfig, now `Array.fromAsync()` can used in this app

apps/api/src/billing/routes/stripe-transactions/stripe-transactions.router.ts
domhhv: @stalniy Applied Node.js stream built-ins here

baktun14 (dismissed):

apps/api/src/billing/services/stripe/stripe.service.ts
stalniy: You could add `timezone` as a required parameter to controller. So, when frontend calls it, it will need to specify timezone. because all dates which are sent over the wire are in UTC. `toISOString()` converts date to ISO8601 in UTC timezone

apps/api/src/billing/services/stripe/stripe.service.spec.ts
stalniy: just check that it builds correctly. I added this to some apps before

apps/api/src/billing/services/stripe/stripe.service.spec.ts
```diff
@@ -364,6 +366,180 @@ describe(StripeService.name, () => {
...
+        endDate: "2022-01-31T23:59:59Z"
+      });
+
+      const chunks = await Array.fromAsync(csvStream);
+
+      const fullCsv = chunks.join("");
+      expect(fullCsv).toContain("No transactions found for the specified date range");
+    });
+
+    it("handles transactions with null payment methods", async () => {
```
stalniy, line 509: **issue(non-blocking):** please add test that validates the behavior when stripe transactions stream fails in the middle. For example, the 1 transactions page is returned but the 2nd fails. You can do this in the next PR

stalniy: https://github.com/akash-network/console/issues/1830

apps/api/test/seeders/stripe-transaction.seeder.ts
```diff
@@ -0,0 +1,31 @@
+import { faker } from "@faker-js/faker";
+
+import type { Transaction } from "@src/billing/http-schemas/stripe.schema";
+
+import { generatePaymentMethod } from "@test/seeders/payment-method.seeder";
+
+export class StripeTransactionSeeder {
+  static create({
```
stalniy, line 8: suggestion(non-blocking): please read this https://github.com/akash-network/console/discussions/914

apps/deploy-web/src/components/billing-usage/BillingContainer/BillingContainer.tsx
```diff
@@ -45,9 +48,14 @@ export const BillingContainer: React.FC<BillingContainerProps> = ({ children, de
...
-  const [error, setError] = React.useState<Error | null>(null);
+  const [errorMessage, setErrorMessage] = React.useState("");
 
-  const { from, to } = dateRange;
+  const { from: startDate, to: endDate } = dateRange;
+
+  const exportCsvQuery = D.useExportTransactionsCsvQuery({
+    startDate,
+    endDate
+  });
```
stalniy: question(non-blocking): do we even need to make it via a hook? This could be done via service call on button.

apps/deploy-web/src/queries/usePaymentQueries.ts
```diff
@@ -106,3 +113,15 @@ export const usePaymentMutations = () => {
     removePaymentMethod
   };
 };
+
+export const useExportTransactionsCsvQuery = (options: ExportTransactionsCsvParams): UseQueryResult<Blob, Error> => {
+  const { stripe } = useServices();
+
+  return useQuery({
```
stalniy: **thought(non-blocking):** I don't see the reason to make it via react-query. This is just one time query which is sent on button click. We should not store the result anywhere in the local cache of react-query. We just need to directly call `stripe.exportTransactionsCsv` in corresponding component

apps/deploy-web/src/components/billing-usage/BillingContainer/BillingContainer.tsx
```diff
@@ -58,16 +66,20 @@ export const BillingContainer: React.FC<BillingContainerProps> = ({ children, de
...
 
   React.useEffect(() => {
     if (axios.isAxiosError(queryError)) {
-      setError(new Error(queryError.response?.data.message || "An error occurred while fetching payment transactions."));
+      setErrorMessage(queryError.response?.data.message || "An error occurred while fetching payment transactions.");
     }
   }, [queryError]);
 
+  React.useEffect(() => {
+    exportCsvQuery.data?.text().then(downloadCsv);
```
stalniy: **issue(non-blocking):** you need to change `downloadCsv` to support `Blob`. `exportCsvQuery.data` should be a `Blob` instance, so you don't need to read it and then re-create again.

Reading the whole file, into memory issues in the browser

Image

stalniy: well, I think it's a blocking issue but I'm fine to fix in the next PR because this is in review for too long

stalniy (dismissed):

stalniy:
@domhhv I think that the most important comment here is about timezone. Let's fix it and then we can merge and address the rest of the tech issues in the next PR

apps/api/src/billing/services/stripe/stripe.service.spec.ts
domhhv: Got it, will do

apps/api/test/seeders/stripe-transaction.seeder.ts
domhhv: Thanks! I'll prefer function seeders going forward and see to refactor this one

apps/deploy-web/src/components/billing-usage/BillingContainer/BillingContainer.tsx
domhhv: Makes sense, the blob can be passed to `downloadCsv` directly. It's only used in BillingContainer and UsageView currently. I'll make sure to refactor it.

It is accepting string right now because previously, the CSV content was constructed as a string in the browser in both cases, but now that I'll be coming as blob from the backend, this refactoring is spot-on.

apps/deploy-web/src/components/billing-usage/BillingContainer/BillingContainer.tsx
domhhv: Yes, it's simpler that way, noted

apps/deploy-web/src/queries/usePaymentQueries.ts
domhhv: +1

apps/api/src/billing/services/stripe/stripe.service.ts
domhhv: Required timezone on the query schema side with custom refinement + passed it down from the frontend 👍

domhhv (commented):
@stalniy Sounds good to me, I addressed the timezone issue and left everything else for a follow-up PR

apps/api/src/billing/services/stripe/stripe.service.ts
```diff
@@ -439,20 +444,24 @@ export class StripeService extends Stripe {
...
     const csvStream = sourceStream.pipe(csvStringifier);
 
-    for await (const chunk of csvStream) {
-      yield chunk.toString();
+    try {
+      for await (const chunk of csvStream) {
+        yield typeof chunk === "string" ? chunk : (chunk as Buffer).toString("utf8");
+      }
+    } catch (err) {
+      logger.error({ event: "CSV_STREAM_ERROR", error: err instanceof Error ? err.message : String(err) });
```
stalniy: **nitpick:** logger supports error unpacking, so you just need to pass `error` inside and it will handle the rest


```suggestion
      logger.error({ event: "CSV_STREAM_ERROR", error: err });
```

apps/api/src/billing/controllers/stripe/stripe.controller.ts
```diff
@@ -152,9 +152,11 @@ export class StripeController {
   }
 
   @Protected([{ action: "read", subject: "StripePayment" }])
-  async exportTransactionsCsvStream(options: { startDate: string; endDate: string }): Promise<AsyncIterable<string>> {
+  async exportTransactionsCsvStream(options: { startDate: string; endDate: string; timezone: string }): Promise<AsyncIterable<string>> {
     const { currentUser } = this.authService;
 
+    console.log("HERE, received timezone:", options.timezone);
```
stalniy: leftover

apps/api/src/billing/services/stripe/stripe.service.ts
```diff
@@ -498,17 +507,15 @@ export class StripeService extends Stripe {
...
 
-  private transformTransactionForCsv(transaction: Transaction, utcHoursOffset: number) {
+  private transformTransactionForCsv(transaction: Transaction, timeZone: string) {
     const amount = (transaction.amount / 100).toFixed(2);
-    const date = new Date(transaction.created * 1000);
-
-    if (utcHoursOffset !== 0) {
-      date.setHours(date.getHours() + utcHoursOffset);
-    }
+    const date = new Date(transaction.created * 1000).toLocaleDateString("en-CA", {
```
stalniy: **question(non-blocking):** why do we show date without time?

apps/api/src/billing/http-schemas/stripe.schema.ts
```diff
@@ -152,6 +152,15 @@ const dateRangeCheck = (data: { startDate?: string; endDate?: string }) => {
 
 const dateRangeErrorMessage = "Date range cannot exceed 366 days and startDate must be before endDate";
 
+const isValidIanaTimeZone = (tz: string) => {
+  try {
+    new Intl.DateTimeFormat("en-CA", { timeZone: tz });
+    return true;
+  } catch {
+    return false;
+  }
```
stalniy: **nitpick:** you can use `Intl.supportedValuesOf("timeZone").includes(tz);`

stalniy (dismissed):

apps/api/src/billing/controllers/stripe/stripe.controller.ts
domhhv: Thanks for the catch, removed

apps/api/src/billing/http-schemas/stripe.schema.ts
domhhv: Thanks, done

apps/api/src/billing/services/stripe/stripe.service.ts
domhhv: Done

apps/api/src/billing/services/stripe/stripe.service.ts
domhhv: No reason, as the requirements for the CSV export format weren't clearly defined. I changed `.toLocaleDateString()` to `.toLocaleString()` here.

apps/deploy-web/src/components/billing-usage/BillingContainer/BillingContainer.tsx
domhhv: I refactored `downloadCsv` to accept `Blob` instead of `string` and updated `UsageView` and `BillingContainer`.

apps/deploy-web/src/queries/usePaymentQueries.ts
domhhv: I removed this query to rely on a direct call instead

apps/deploy-web/src/queries/queryKeys.ts
```diff
@@ -104,4 +104,18 @@ export class QueryKeys {
...
+    if (options.startDate) {
+      key.push("start_date", options.startDate.toISOString());
+    }
+
+    if (options.endDate) {
+      key.push("end_date", options.endDate.toISOString());
+    }
+
+    return key;
+  };
```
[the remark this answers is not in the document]
domhhv: We only need timezone for calling transactions CSV export endpoint

apps/deploy-web/src/components/billing-usage/BillingContainer/BillingContainer.tsx
```diff
@@ -58,16 +67,20 @@ export const BillingContainer: React.FC<BillingContainerProps> = ({ children, de
     limit: pagination.pageSize,
     startingAfter: currentCursors.startingAfter,
     endingBefore: currentCursors.endingBefore,
-    startDate: from,
-    endDate: to
+    startDate,
+    endDate
```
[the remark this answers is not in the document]
domhhv: We only need timezone for transactions CSV export endpoint

apps/api/src/billing/services/stripe/stripe.service.spec.ts
domhhv: This is tabled still

apps/deploy-web/src/components/billing-usage/BillingContainer/BillingContainer.tsx
domhhv: I removed this query to rely on a direct call instead

apps/api/src/billing/controllers/stripe/stripe.controller.ts
```diff
@@ -150,4 +150,13 @@ export class StripeController {
     const response = await this.stripe.getCustomerTransactions(currentUser.stripeCustomerId, options);
     return { data: response };
   }
+
+  @Protected([{ action: "read", subject: "StripePayment" }])
+  async exportTransactionsCsvStream(options: { startDate: string; endDate: string; timezone: string }): Promise<AsyncIterable<string>> {
+    const { currentUser } = this.authService;
+
+    assert(currentUser.stripeCustomerId, 500, "Payment account not properly configured. Please contact support.");
```
ygrishajev:
```suggestion
    assert(currentUser.stripeCustomerId, 403, "Payments are not configured. Please start with a trial first");
```

stalniy (dismissed):

apps/api/src/billing/controllers/stripe/stripe.controller.ts
domhhv: This assertion was taken from `this.getCustomerTransactions` above, which is now replaced with gracefully returning empty data there. I'm wondering if something similar should be done here or if assertion is better suited still?

```
    if (!currentUser.stripeCustomerId) {
        return 'No transactions found';
    }
```

domhhv: I used this placeholder iterator:

```typescript
    if (!currentUser.stripeCustomerId) {
      return (async function* () {
        yield "Payments are not configured. Please start with a trial first";
      })();
    }
```

ygrishajev (dismissed):

apps/api/src/billing/controllers/stripe/stripe.controller.ts
```diff
@@ -150,4 +150,17 @@ export class StripeController {
...
     return { data: response };
   }
+
+  @Protected([{ action: "read", subject: "StripePayment" }])
+  async exportTransactionsCsvStream(options: { startDate: string; endDate: string; timezone: string }): Promise<AsyncIterable<string>> {
+    const { currentUser } = this.authService;
+
+    if (!currentUser.stripeCustomerId) {
+      return (async function* () {
+        yield "Payments are not configured. Please start with a trial first";
```
ygrishajev: I'm not sure i follow. this should be an error. basically this call is forbidden 403 as user gotta do other stuff. we can also use 428 Precondition Required

domhhv: @ygrishajev I used this assertion that you proposed earlier:

```
assert(currentUser.stripeCustomerId, 403, "Payments are not configured. Please start with a trial first");
```

ygrishajev (approved):

apps/api/src/billing/controllers/stripe/stripe.controller.ts
```diff
@@ -150,4 +150,13 @@ export class StripeController {
...
   }
+
+  @Protected([{ action: "read", subject: "StripePayment" }])
+  async exportTransactionsCsvStream(options: { startDate: string; endDate: string; timezone: string }): Promise<AsyncIterable<string>> {
+    const { currentUser } = this.authService;
+
+    assert(currentUser.stripeCustomerId, 403, "Payments are not configured. Please start with a trial first");
+
+    return this.stripe.exportTransactionsCsvStream(currentUser.stripeCustomerId, options);
+  }
```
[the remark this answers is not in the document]
ygrishajev: @domhhv this is a valid comment

apps/api/src/billing/services/stripe/stripe.service.spec.ts
```diff
@@ -364,6 +366,184 @@ describe(StripeService.name, () => {
...
+      });
+
+      const chunks = await Array.fromAsync(csvStream);
+
+      const fullCsv = chunks.join("");
+
+      expect(fullCsv).toContain("ch_123");
+      expect(fullCsv).toContain("No payment method");
+      expect(fullCsv).toMatch(/ch_123,"[^"]*",[^,]*,[^,]*,[^,]*,,,,No payment method,/);
+    });
```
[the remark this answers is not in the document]
ygrishajev: @domhhv is this so?

Issue #1735, opened 2025-07-24
Implement onboarding email notifications


baktun14:
- [x] When a trial is initiated to welcome the user with a personalized email - Warm welcome to explore the possibilities
- [x] When the first trial deployment is created - Personalized congratulations and incentivize non-trial conversion
- [x] When a trial user has a deployment that is about to close with the 24h cutoff (1 hour before) - Reminder to incentivize non-trial conversion
- [x] When a trial user has a deployment that closed because of the 24h cutoff - Reminder to incentivize non-trial conversion
- [x] When a trial user is 1 week into the trial - Reminder to explore the platform and use the credits - Reminder to incentivize non-trial conversion
- [x] Whtn a trial user is 1 week away from the end of the 30 days trial - Reminder to incentivize non-trial conversion to keep the credits
- [x] When the trial ends - Ask for feedback and try to convert
- [x] 1 week after the trial ended - Retargeting for feedback and try to convert

ygrishajev:
### Regarding email notifications 📯

So the entry point to the handler is [here](https://github.com/akash-network/console/blob/main/apps/notifications/src/interfaces/notifications-events/handlers/notification/notification.handler.ts). You may find its type there and it should be pretty easy to follow the call stack down to Novu API.

How it can be sent is [here](https://github.com/akash-network/console/blob/main/apps/notifications/src/interfaces/alert-events/handlers/chain-events/chain-events.handler.ts#L32).

---

I see a couple of options of how this can be implemented:

1. **Another simple endpoint on [rest interface](https://github.com/akash-network/console/tree/main/apps/notifications/src/interfaces/rest/controllers)**  
   In this case it would just pass it to the event bus and respond with 200. The delivery responsibility is on the notifications from now on. Swagger would be updated and SDK auto-generated which would be easy to use in other services.  
   Another benefit of it is HTTP is easier to observe and it’s already set up.

2. **Create SDKs manually or create code gen depending on [broker service](https://github.com/akash-network/console/blob/main/apps/notifications/src/infrastructure/broker/services/broker/broker.service.ts)** and send events directly to the event bus.  
   I can see some issues with this approach: harder observability and another infrastructural dependency are some of them.

---

Regarding the interface of the endpoint/event.  
Right now it accepts `notificationChannelId` which might not be as useful on the public interface.  
We might want to make an interface like this:

```ts
{ notificationChannelId: string } | { userId: string; notificationChannelId?: string }
```
[...]

PR #1796, opened 2025-08-11
refactor: moves user registraction logic under user controller

stalniy:
## Why

To make logic more transparent and add tests on db level to cover different edge cases. Ref: #1735

## Summary by CodeRabbit

* **New Features**
  * Added POST /v1/register-user and GET /v1/user/me (included in OpenAPI) and a public User schema.

* **Refactor**
  * Auth callback now registers/refreshes users and merges profile data.
  * Frontend onboarding uses a centralized storage key and injected URL service; logout now uses in-app logout that clears tokens and redirects.

* **Tests**
  * New comprehensive tests for user registration, wallet/anonymous flows, and auth interceptor behavior.

* **Documentation**
  * Deprecated old token info route in favor of new endpoints.

### DIFF
--- a/apps/api/src/core/repositories/base.repository.ts
+++ b/apps/api/src/core/repositories/base.repository.ts
@@ -3,6 +3,7 @@ import type { DBQueryConfig } from "drizzle-orm";
 import { and, eq, inArray, isNull, sql } from "drizzle-orm";
 import type { PgTableWithColumns } from "drizzle-orm/pg-core";
 import type { SQL } from "drizzle-orm/sql/sql";
+import { PostgresError } from "postgres";
 
 import type { ApiPgDatabase, ApiPgTables, TxService } from "@src/core";
 import { DrizzleAbility } from "@src/lib/drizzle-ability/drizzle-ability";
@@ -221,3 +222,8 @@ type TableName<T extends PgTableWithColumns<any>> = T extends PgTableWithColumns
 type TableNameInSchema<T extends PgTableWithColumns<any>> = {
   [K in keyof TablesOnly<ApiPgTables> as TableName<ApiPgTables[K]>]: K;
 }[TableName<T>];
+
+const UNIQUE_VIOLATION_CODE = "23505";
+export function isUniqueViolation(error: unknown): error is PostgresError {
+  return error instanceof PostgresError && error.code === UNIQUE_VIOLATION_CODE;
+}
--- a/apps/api/src/routers/userRouter.ts
+++ b/apps/api/src/routers/userRouter.ts
@@ -39,6 +39,9 @@ userOptionalRouter.get("/byUsername/:username", async c => {
   return c.json(user);
 });
 
+/**
+ * @deprecated Use /v1/register-user instead
+ */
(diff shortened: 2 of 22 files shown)


apps/api/src/user/services/user/user.service.ts
```diff
@@ -0,0 +1,149 @@
...
+  let baseUsername = wantedUsername.replace(/[^\w-]+/g, "");
+
+  if (baseUsername.length < 3) {
+    baseUsername = baseUsername.padEnd(10, "anonymous");
+  } else if (baseUsername.length > 40) {
+    baseUsername = baseUsername.slice(0, 40);
+  }
+
+  return baseUsername + randomInt(1000, 9999).toString();
+}
```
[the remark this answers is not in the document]
stalniy: sensible defaults may cause username unique index conflicts

apps/api/src/core/repositories/base.repository.ts
```diff
@@ -221,3 +222,7 @@ type TableName<T extends PgTableWithColumns<any>> = T extends PgTableWithColumns
 type TableNameInSchema<T extends PgTableWithColumns<any>> = {
   [K in keyof TablesOnly<ApiPgTables> as TableName<ApiPgTables[K]>]: K;
 }[TableName<T>];
+
+export function isUniqueViolation(error: unknown): error is PostgresError {
+  return error instanceof PostgresError && error.code === "23505";
```
baktun14: suggestion(non-blocking): use a const for the code

apps/api/src/user/routes/register-user/register-user.router.ts
```diff
@@ -0,0 +1,52 @@
...
+export type RegisterUserInput = z.infer<typeof registerUserInputSchema>;
+
+const registerUserResponseSchema = z.object({
+  data: UserSchema
+});
+export type RegisterUserResponse = z.infer<typeof registerUserResponseSchema>;
+
+const route = createRoute({
+  method: "post",
+  path: "/v1/registerUser",
```
baktun14: suggestion(non-blocking): we use lower case dash for new routes
```suggestion
  path: "/v1/register-user", // or /v1/register might be simpler
```

apps/api/src/routers/userRouter.ts
```diff
@@ -39,6 +39,9 @@ userOptionalRouter.get("/byUsername/:username", async c => {
   return c.json(user);
 });
 
+/**
+ * @deprecated Use /v1/registerUser instead
```
baktun14: reminder(non-blocking): in case you change the route

baktun14 (dismissed):

apps/api/src/core/repositories/base.repository.ts
stalniy: do we have a const for this? or do you want me to create one? I thought that this method is small and does exactly what it does, checks whether the passed in error is a unique violation error from postgres.

baktun14: Create one, but I guess the name of the function tells what it does

PR #1808, opened 2025-08-14
feat: adds creation of default notification channel on user login/registration

stalniy:
## Why

#1735 

## Summary by CodeRabbit

- New Features
  - Automatically creates a default email notification channel after user registration; failures are logged but don’t block registration.
  - Added API endpoint: POST /v1/notification-channels/default (returns 204).

- SDK
  - React Query SDK: createDefaultChannel mutation and updated schemas/types to include isDefault.

- Database
  - Added is_default column and unique partial index to enforce one default channel per user.

- Tests
  - New tests covering default-channel creation and lookup.

- Chores
  - Added new dependencies and updated TypeScript module setting.


apps/notifications/src/modules/notifications/repositories/notification-channel/notification-channel.repository.ts
```diff
@@ -81,23 +81,46 @@ export class NotificationChannelRepository {
...
+      .values({
+        ...this.toInput(input),
+        isDefault: true
+      })
+      .onConflictDoNothing({
+        target: [schema.NotificationChannel.userId, schema.NotificationChannel.isDefault],
+        where: eq(schema.NotificationChannel.isDefault, true)
+      });
+  }
+
```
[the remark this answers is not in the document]
stalniy: ability check is enforced on controller level, repository level should not enforce ability checks as it doesn't know about the user and can be used in different contexts

apps/api/src/user/services/user/user.service.ts
```diff
@@ -66,6 +68,29 @@ export class UserService {
     this.logger.info({ event, id: user.id, userId: user.userId });
     this.analyticsService.track(user.id, "user_registered");
 
+    const result = await this.notificationsApi.v1
+      .createDefaultChannel({
```
baktun14, line 72: This is cool, are all the parameters typed as well?

apps/notifications/src/interfaces/rest/controllers/notification-channel/notification-channel.controller.ts
```diff
@@ -71,6 +72,23 @@ export class NotificationChannelController {
     });
   }
 
+  @Post("default")
+  @HttpCode(204)
+  @ValidateHttp({
+    204: {
+      schema: createZodDto(z.void()),
+      description: "Creates the default notification channel only if it doesn't exist. If it does exist, it returns the existing channel."
```
baktun14: question(non-blocking): is it supposed to return the channel?

baktun14 (dismissed):

apps/notifications/src/interfaces/rest/controllers/notification-channel/notification-channel.controller.ts
stalniy: OK, it did before but then after few changes, I decided that there is no point

apps/api/src/user/services/user/user.service.ts
stalniy: yes but I don't like that it requires us to specify where which parameter to put, what reveals underlying transport requirements. SDK should hide transport under abstraction and pick parameters from aggregated object by itself

apps/notifications/src/modules/notifications/repositories/notification-channel/notification-channel.repository.ts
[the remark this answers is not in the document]
stalniy: it's a nice catch!

PR #1818, opened 2025-08-15
feat: adds notification on start trial

stalniy:
## Why

#1735 

## Summary by CodeRabbit

* **New Features**
  * Sends "Start Trial" notification for relevant users when the anonymous-free-trial flag is disabled.
  * Adds a notifications job endpoint to create/schedule notification jobs and a resilient NotificationService with retries and default-channel fallback.
  * SDK/clients expose a createNotification operation.

* **Documentation**
  * API docs and SDK updated with POST /v1/jobs/notification, new schemas, and simplified default-channel contract.

* **Bug Fixes**
  * Error responses include extra exception fields for richer diagnostics.

* **Tests**
  * New and updated tests covering notification flows, jobs controller, feature-flag wallet behavior, and user-service integrations.

### DIFF
--- a/apps/api/src/billing/services/wallet-initializer/wallet-initializer.service.ts
+++ b/apps/api/src/billing/services/wallet-initializer/wallet-initializer.service.ts
@@ -2,6 +2,10 @@ import { singleton } from "tsyringe";
 
 import { AuthService } from "@src/auth/services/auth.service";
 import { UserWalletInput, UserWalletPublicOutput, UserWalletRepository } from "@src/billing/repositories";
+import { FeatureFlags } from "@src/core/services/feature-flags/feature-flags";
+import { FeatureFlagsService } from "@src/core/services/feature-flags/feature-flags.service";
+import { NotificationService } from "@src/notifications/services/notification/notification.service";
+import { startTrialNotification } from "@src/notifications/services/notification-templates/start-trial-notification";
 import { ManagedUserWalletService } from "../managed-user-wallet/managed-user-wallet.service";
 
 @singleton()
@@ -9,7 +13,9 @@ export class WalletInitializerService {
   constructor(
     private readonly walletManager: ManagedUserWalletService,
     private readonly userWalletRepository: UserWalletRepository,
-    private readonly authService: AuthService
+    private readonly authService: AuthService,
+    private readonly notificationService: NotificationService,
+    private readonly featureFlagsService: FeatureFlagsService
   ) {}
 
   async initializeAndGrantTrialLimits(userId: UserWalletInput["userId"]): Promise<UserWalletPublicOutput> {
@@ -32,6 +38,10 @@ export class WalletInitializerService {
     } catch (error) {
       await this.userWalletRepository.deleteById(userWallet.id);
       throw error;
+    }
+
+    if (!this.featureFlagsService.isEnabled(FeatureFlags.ANONYMOUS_FREE_TRIAL) && this.authService.currentUser?.email) {
+      await this.notificationService.createNotification(startTrialNotification(this.authService.currentUser));
     }
 
     return this.userWalletRepository.toPublic(userWallet);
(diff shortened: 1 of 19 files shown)


apps/api/src/billing/services/wallet-initializer/wallet-initializer.service.ts
```diff
@@ -2,14 +2,20 @@ import { singleton } from "tsyringe";
 
 import { AuthService } from "@src/auth/services/auth.service";
 import { UserWalletInput, UserWalletPublicOutput, UserWalletRepository } from "@src/billing/repositories";
+import { FeatureFlags } from "@src/core/services/feature-flags/feature-flags";
+import { FeatureFlagsService } from "@src/core/services/feature-flags/feature-flags.service";
+import { NotificationService } from "@src/notifications/services/notification/notification.service";
+import { startTrial } from "@src/notifications/services/notification-templates/start-trial";
```
baktun14: suggestion(non-blocking): perhaps we can rename this to `startTrialTemplate` to be more specific about what it does

apps/notifications/src/interfaces/rest/controllers/jobs/jobs.controller.ts
```diff
@@ -0,0 +1,76 @@
...
+          code: "NOTIFICATION_CHANNEL_NOT_FOUND"
+        })
+      );
+    }
+
+    const startAfter = job.startAfter ? new Date(job.startAfter) : undefined;
+    const publishOptions: PublishOptions = {
+      id: job.notificationId,
+      startAfter,
+      expireInSeconds: DEFAULT_NOTIFICATION_EXPIRATION_IN_SECONDS
```
baktun14, line 60: question(non-blocking): if the job is not processed within 24h it expires?

baktun14 (dismissed):

apps/notifications/src/interfaces/rest/controllers/jobs/jobs.controller.ts
stalniy: yes, so if notification was not processed during 24h from the moment of creation, I think it's too late to send it. The default value in pgboss is  15 minutes

stalniy: we can reduce this time if you think it's too big

apps/api/src/billing/services/wallet-initializer/wallet-initializer.service.ts
```diff
@@ -29,6 +35,10 @@ export class WalletInitializerService {
         },
         { returning: true }
       );
+
+      if (!this.featureFlagsService.isEnabled(FeatureFlags.ANONYMOUS_FREE_TRIAL) && this.authService.currentUser.email) {
```
ygrishajev: why sending a notification depends on anonymous free trial? shouldn't it only depend on the email set?

apps/api/src/notifications/services/notification/notification.service.ts
```diff
@@ -0,0 +1,76 @@
...
+        .catch(error => ({ error, response: null }));
+
+      if (!result.error) return;
+
+      if (result.error.code === "NOTIFICATION_CHANNEL_NOT_FOUND" && user.email) {
+        await this.createDefaultChannel(user);
+      }
+
+      throw new Error("Failed to create notification", { cause: result.error });
+    }, DEFAULT_BACKOFF_OPTIONS);
```
ygrishajev: there is a `retry` callback on `backOff` options containing an error and attempt number, which should be a bit more handy then handling via catch.

image

ygrishajev (dismissed):

apps/api/src/notifications/services/notification/notification.service.ts
stalniy: I saw it but this functions do not throw, they return response, so if there is 502 error response, it doesn't throw and return an object, as a result `backOff` won't retry it. But we need to retry this operation

`retry` option receives an error, that's why I assumed I need to throw in order to enable retry mechanism

apps/api/src/billing/services/wallet-initializer/wallet-initializer.service.ts
stalniy: well, it's part of a new onboarding flow. That's why I check this flag here. I also double checked with @baktun14 about this and he recommended to rely on feature flag

PR #1823, opened 2025-08-18
feat(network): rpc proxy for server apps

baktun14:
## Summary by CodeRabbit

- New Features
  - Optional proxy support for API and RPC with new environment flags and default proxy endpoints; services and web app can route via proxy when enabled.
  - Centralized, runtime-configurable chain URL resolution and feature-flag toggle for proxy usage.

- Bug Fixes
  - Safer null checks and early guards in deployment and block-height lookups to prevent runtime errors.

- Chores
  - Removed deprecated env keys and updated configs/tests to use proxy-aware URLs; expanded mainnet RPC endpoints.


apps/api/src/address/services/address/address.service.ts
```diff
@@ -14,16 +14,16 @@ const logger = LoggerService.forContext("AddressService");
 export class AddressService {
   constructor(
     private readonly transactionService: TransactionService,
-    private readonly cosmosHttpService: CosmosHttpService,
+    private readonly cosmosHttpServiceWrapper: CosmosHttpServiceWrapper,
```
stalniy: suggestion(non-blocking): I'd not chnage the name of property, only typ

stalniy: we dont care whether its wrapper or not in this class.

stalniy: actually after a bit of thinking, you could just register a different class via DI and not change consumers at all if eventual interface remains the same

apps/api/src/core/services/http-service-wrapper/http-service-wrapper.ts
```diff
@@ -0,0 +1,69 @@
...
+import { singleton } from "tsyringe";
+
+import { ChainConfigService } from "../chain-config/chain-config.service";
+
+@singleton()
+export class BalanceHttpServiceWrapper extends BalanceHttpService {
+  constructor(chainConfigService: ChainConfigService) {
+    super({ baseURL: chainConfigService.getBaseAPIUrl() });
+  }
+}
```
stalniy:
```ts
container.register(BalanceHttpService, { 
  useFactory: c => new BalanceHttpService({ baseURL: c.resolve(ChainConfigService).getBaseAPIUrl() }) 
});
```

apps/api/src/address/services/address/address.service.ts
baktun14: I can do that yes

apps/api/src/core/services/http-service-wrapper/http-service-wrapper.ts
baktun14: Ah that is cleaner indeed, I'll refactor this back.

apps/provider-proxy/src/services/feature-flags/feature-flags.service.ts
```diff
@@ -0,0 +1,18 @@
...
+      case "use_proxy_urls":
+        return process.env.USE_PROXY_URLS === "true";
+      default:
+        return false;
+    }
+  }
+
+  async isEnabledAsync(featureFlag: FeatureFlagValue): Promise<boolean> {
+    return this.isEnabled(featureFlag);
+  }
```
stalniy: **question(non-blocking):** could you please explain the purpose of having `isEnabledAsync`?

baktun14: Probably some leftovers when I was trying to implement with unleash, I'll remove it

apps/api/src/billing/lib/batch-signing-client/batch-signing-client.service.ts
```diff
@@ -63,6 +64,7 @@ export class BatchSigningClientService {
     private readonly wallet: Wallet,
     private readonly registry: Registry,
     private readonly connectWithSigner: ConnectWithSignerFn,
+    private readonly chainNetworkConfigService: ChainNetworkConfigService,
```
ygrishajev: the idea behind modular design and config per module is that every module declares what it needs. E.g. if several modules need rpc url both would declare it in their config. This imo allows to loose coupling between modules and make it more transparent what every module depends on.

apps/api/src/core/services/chain-network-config/chain-network-config.service.ts
```diff
@@ -0,0 +1,34 @@
+import { createNetConfig, type SupportedChainNetworks } from "@akashnetwork/net";
+import { singleton } from "tsyringe";
+
+import { env } from "../../../utils/env";
+import { FeatureFlagsService } from "../feature-flags/feature-flags.service";
+
+@singleton()
+export class ChainNetworkConfigService {
```
ygrishajev: Why haven't you use same pattern for config service as used already in almost all modules? there is even a config already https://github.com/akash-network/console/blob/main/apps/api/src/chain/services/chain-config/chain-config.service.ts

apps/api/env/.env.functional.test
```diff
@@ -38,3 +38,6 @@ AMPLITUDE_SAMPLING=1
 NOTIFICATIONS_API_BASE_URL=http://localhost:3081
 
 FEATURE_FLAGS_ENABLE_ALL=true
+
+PROXY_API_URL=https://api.sandbox-01.aksh.pw:443
+PROXY_RPC_URL=https://rpc.sandbox-01.aksh.pw:443
```
ygrishajev, line 43: non-blocking: i don't thing any client cares if this url is a proxy or not. it would still treat it as a API/RPC endpoint, right? perhaps renaming it might not make much sense.

apps/api/src/core/providers/http-sdk.provider.ts
```diff
@@ -13,20 +13,33 @@ import {
...
+});
+container.register(LeaseHttpService, {
+  useFactory: c => new LeaseHttpService({ baseURL: c.resolve(ChainNetworkConfigService).getBaseAPIUrl() })
+});
+container.register(ProviderHttpService, {
+  useFactory: c => new ProviderHttpService({ baseURL: c.resolve(ChainNetworkConfigService).getBaseAPIUrl() })
+});
+container.register(CosmosHttpService, {
+  useFactory: c => new CosmosHttpService({ baseURL: c.resolve(ChainNetworkConfigService).getBaseAPIUrl() })
+});
```
ygrishajev: i don't think i understand this change. this is a repetitive code, could still iterate over every service in array

apps/api/src/billing/lib/batch-signing-client/batch-signing-client.service.ts
```diff
@@ -81,9 +83,12 @@ export class BatchSigningClientService {
   }
 
   private async initClient() {
+    const rpcUrl = this.chainNetworkConfigService.getBaseRpcUrl();
+    if (!rpcUrl) throw new Error("RPC url for the batch signing client is not set.");
```
stalniy, line 87: **suggestion(non-blocking):** would be nice to integrate this exception into `getBaseRpcUrl` method. So, it either returns value or throws error

apps/api/src/core/services/chain-network-config/chain-network-config.service.ts
```diff
@@ -0,0 +1,34 @@
...
+  private readonly netConfig: ReturnType<typeof createNetConfig>;
+
+  constructor(private readonly featureFlagsService: FeatureFlagsService) {
+    this.netConfig = createNetConfig({
+      useProxyUrls: this.featureFlagsService.isEnabled("use_proxy_urls"),
+      proxyApiUrl: env.PROXY_API_URL,
+      proxyRpcUrl: env.PROXY_RPC_URL
+    });
```
stalniy: **issue(blocking):** this won't re-evaluate when feature flag changes. take a look how this is done in https://github.com/akash-network/console/blob/main/apps/api/src/auth/services/ability/ability.service.ts#L47

Even though this won't work in this case because baseApiUrl value is then copied into baseURL of other services. What means that when feature flag changes, we need to re-create those services OR restart application. Potentially we could call `container.clearInstances()` but I don't know what other side effects this can cause.

So, either we need to restart applications when feature flag changes or find another way

apps/notifications/src/modules/chain/services/rpc-url-resolver/rpc-url-resolver.service.ts
```diff
@@ -0,0 +1,25 @@
...
+import { FeatureFlagsService } from "../feature-flags/feature-flags.service";
+
+@Injectable()
+export class RpcUrlResolverService {
+  constructor(
+    private readonly configService: ConfigService<ChainEventsConfig>,
+    private readonly featureFlagsService: FeatureFlagsService
+  ) {}
+
+  async getRpcUrl(): Promise<string> {
```
ygrishajev, line 15: ConfigService has a `set` method. There is also a way to listen for ff changes as done in api already https://github.com/akash-network/console/blob/main/apps/api/src/core/services/feature-flags/feature-flags.service.ts#L48

This means the listener can be done like so

```ts
@Injectable()
export class RpcUrlWatcherService implements OnModuleInit {
  constructor(
    private readonly featureFlagsService: FeatureFlagsService,
    private readonly configService: ConfigService<ChainEventsConfig>
  ) {}

  onModuleInit() {
    this.featureFlagsService.onChanged((...args) => {
      this.configService.set("chain.RPC_NODE_ENDPOINT", some value);
    });
  }
}
```

this way the rest of the app would still use same config service without hustle.

ygrishajev: I also think when it gets to the config in other apps similar approach can be taken as it's much easier to utilise config service directly from many perspectives

apps/api/src/core/services/chain-network-config/chain-network-config.service.ts
stalniy: basically instead of providing this to different services, we need to call `getBaseApiUrl` or `getBaseRpcUrl` on every http call inside http-sdk. or we need to write a proxy/wrapper around every service that uses it and re-create instance. It seems to be too complex to do it with feature flags.

apps/api/src/core/services/chain-network-config/chain-network-config.service.ts
```diff
@@ -0,0 +1,34 @@
...
+import { env } from "../../../utils/env";
+import { FeatureFlagsService } from "../feature-flags/feature-flags.service";
+
+@singleton()
+export class ChainNetworkConfigService {
+  private readonly netConfig: ReturnType<typeof createNetConfig>;
+
+  constructor(private readonly featureFlagsService: FeatureFlagsService) {
+    this.netConfig = createNetConfig({
+      useProxyUrls: this.featureFlagsService.isEnabled("use_proxy_urls"),
```
stalniy:
```suggestion
      useProxyUrls: this.featureFlagsService.isEnabled(FeatureFlags.USE_PROXY_URLS),
```

apps/api/src/deployment/services/deployment-reader/deployment-reader.service.ts
```diff
@@ -31,7 +31,7 @@ export class DeploymentReaderService {
     const deploymentResponse = await this.deploymentHttpService.findByOwnerAndDseq(owner, dseq);
 
     if ("code" in deploymentResponse) {
-      assert(!deploymentResponse.message?.toLowerCase().includes("deployment not found"), 404, "Deployment not found");
+      assert(!deploymentResponse.message?.toLowerCase()?.includes("deployment not found"), 404, "Deployment not found");
```
stalniy: **thought(non-blocking):** I don't think that adding `?` after `toLowerCase()` solves something because if message exists, then `toLowerCase` won't remove `undefined` and if message is undefined `toLowerCase` won't be called

stalniy (changes requested):

apps/api/src/billing/lib/batch-signing-client/batch-signing-client.service.ts
baktun14: Isn't this a core config that should be re-used by all modules? I don't see the advantage to declaring multiple services that will return the rpc for each module

apps/api/src/core/services/chain-network-config/chain-network-config.service.ts
baktun14: I initially merged it with this service, but it was causing issues because it's dependent on the feature flag service.

apps/api/env/.env.functional.test
baktun14: Yea sure, it's currently only used in the tests for nock

apps/api/src/billing/lib/batch-signing-client/batch-signing-client.service.ts
ygrishajev: i don't think there should be this method at all, config values can be updated on ff change

apps/api/env/.env.functional.test
baktun14: These are actually for the rpc proxy specifically

apps/notifications/src/modules/chain/services/rpc-url-resolver/rpc-url-resolver.service.ts
baktun14: I switched it to env var because it was causing too many potential issues live updating the api/rpc urls with many dependencies relying on it. We could clear the instances but that would have unintented consequences.

PR #1826, opened 2025-08-20
fix: billing usage page guard flicker

### LINKED ISSUE #1822: Billing usage pages flicker with 404 on first load
When /usage page first loads or is refocused, 404 page shows up for moment. This is probably happening inside [this guard HOC](https://github.com/akash-network/console/blob/main/apps/deploy-web/src/hoc/guard/guard.hoc.tsx) and should be fixable by exposing loading states from the hooks composed in the [Billing](https://github.com/akash-network/console/blob/main/apps/deploy-web/src/pages/billing/index.tsx#L8) and [Usage](https://github.com/akash-network/console/blob/main/apps/deploy-web/src/pages/usage/index.tsx#L8) page files.

https://github.com/user-attachments/assets/17300c0c-a7e2-4ff8-8179-e9cb4f2b3cc3

domhhv:
Fixes #1822

## Summary by CodeRabbit

- **New Features**
  - Guarded pages now show a loading state while account status is resolved, giving clearer feedback during sign-in checks.
  - Various hooks and checks now expose loading states so the UI can indicate when user identity is still being determined.

- **Bug Fixes**
  - Fixed inconsistent account-reading across alerts, sidebar, onboarding, payments, API keys and email channel components, improving gating and visibility once signed in.


baktun14 (dismissed):

apps/deploy-web/src/hoc/guard/guard.hoc.tsx
```diff
@@ -1,8 +1,18 @@
...
+  isLoading: boolean;
+};
+
+export const Guard = <P extends object>(Component: React.ComponentType<P>, useCheck: UseCheck, FallbackComponent = FourOhFour) => {
   const WithGuard = (props: P) => {
-    const canVisit = useCheck();
+    const { canVisit, isLoading } = useCheck();
+
+    if (isLoading) {
+      return <Loading text="Please wait..." />;
```
baktun14, line 14: question(non-blocking): Is there a way we can customize the text in different pages?

apps/deploy-web/src/hooks/useUser.ts
```diff
@@ -5,14 +5,22 @@ import { useStoredAnonymousUser } from "@src/hooks/useStoredAnonymousUser";
...
+  const { user: registeredUser, isLoading: isLoadingRegisteredUser } = useCustomUser();
+  const { user: anonymousUser, isLoading: isLoadingAnonymousUser } = useStoredAnonymousUser();
+  const user = useMemo(() => registeredUser || anonymousUser || {}, [registeredUser, anonymousUser]);
+  const isLoading = useMemo(() => isLoadingRegisteredUser || isLoadingAnonymousUser, [isLoadingRegisteredUser, isLoadingAnonymousUser]);
 
-  return user;
+  return {
+    ...user,
+    isLoading
+  };
```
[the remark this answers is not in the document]
ygrishajev: @domhhv not sure empty object for a user makes sense. if there is no user it should be undefined or null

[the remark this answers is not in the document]
stalniy: I'd also separate isLoading from userProfile data. This would look and behave consistent with other places

apps/deploy-web/src/components/user/UserProviders/UserProviders.spec.tsx
```diff
@@ -53,8 +53,13 @@ describe(UserProviders.name, () => {
...
 
-    expect(userTracker.track).toHaveBeenCalledWith(undefined);
-    expect(userTracker.track).toHaveBeenCalledWith(anonymousUser);
+    expect(userTracker.track).toHaveBeenCalledWith({
+      isLoading: true
+    });
+    expect(userTracker.track).toHaveBeenCalledWith({
+      ...anonymousUser,
+      isLoading: false
+    });
```
stalniy: **issue(blocking):** `userTracker.track` should not receive an object without `.id` because then it will behave in the way that user exists but without properties. it's either user or falsy value

domhhv: I included `user` as a property in the return value of `useUser()`, so now `userTracker.track` here receives either undefined or full user value

apps/deploy-web/src/hoc/guard/guard.hoc.tsx
domhhv: I think we could pass text as an argument to the `Guard()` call, after optional `FallbackComponent` param or by combining the two into an `options` param

apps/deploy-web/src/hooks/useUser.ts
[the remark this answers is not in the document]
domhhv: I made `user` and `isLoading` two separate properties returned from `useUser()`

ygrishajev (approved):

PR #1827, opened 2025-08-20
feat: adds more notifications on start trial event

stalniy:
## Why

ref #1735

## Summary by CodeRabbit

- **New Features**
  - Background job queue and domain-event publishing with automated trial notifications (start, reminders, end, post-trial).
  - New notification templates and a CLI daemon command to drain job queues.

- **Improvements**
  - Parallelized startup, single graceful shutdown, and initializer wiring.
  - Idempotent wallet get-or-create flow and event-driven trial signaling.
  - Feature-flag client auto-initializes at startup.

- **Chores**
  - Added runtime dependencies for job queueing and conditional notification logic.

- **Tests**
  - New tests for job queue, notification handler, and trial flows.


baktun14 (dismissed):

apps/api/src/console.ts
```diff
@@ -121,10 +130,25 @@ async function executeCliHandler(name: string, handler: () => Promise<unknown>)
...
+
+const shutdown = once(async () => {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const { closeConnections } = require("./core/providers/postgres.provider");
+
+  await Promise.all([closeConnections(), chainDb.close(), container.dispose()]);
+});
+process.on("SIGTERM", shutdown);
+process.on("SIGINT", shutdown);
 program.parse();
```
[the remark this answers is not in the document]
stalniy: it has `dispose`, then it goes over all instances and call `dispose` method on them

apps/api/src/billing/repositories/user-wallet/user-wallet.repository.ts
```diff
@@ -46,6 +46,31 @@ export class UserWalletRepository extends BaseRepository<ApiPgTables["UserWallet
...
+      .returning();
+
+    if (newWallet) {
+      return {
+        wallet: this.toOutput(newWallet),
+        isNew: true
+      };
+    }
+
+    // race condition, wallet was created by another call
```
ygrishajev, line 69: what should one do about it?

apps/api/src/core/services/job-queue/job-queue.service.ts
```diff
@@ -0,0 +1,152 @@
...
+    });
+
+    return await this.pgBoss.send({
+      name: job.name,
+      data: { ...job.data, version: job.version },
+      options
+    });
+  }
+
+  async drain(options: ProcessOptions = {}): Promise<void> {
```
ygrishajev: The name drain feels a bit misleading - I’d expect it to just empty the queue once and stop, not spin up long-running workers to process jobs. Something like startWorkers or consumeJobs might make the intent clearer.

ygrishajev (dismissed):
💪

apps/api/src/billing/repositories/user-wallet/user-wallet.repository.ts
stalniy: nothing, it's just a note why I call find user again because it's not self-explanatory

apps/api/src/core/services/job-queue/job-queue.service.ts
stalniy: @coderabbitai could you suggest the best name here? I also don't like `drain` maybe `listenForJobs` or `waitForJobs`?

apps/api/src/billing/repositories/user-wallet/user-wallet.repository.ts
ygrishajev: ah, i missed the the point of it generally at first

ygrishajev (approved):

apps/api/src/console.ts
```diff
@@ -97,21 +102,25 @@ program
     });
   });
 
+program
+  .command("drain-job-queues")
```
ygrishajev, line 106: does the command also need to be re-named?

PR #1833, opened 2025-08-22
feat: adds api background-jobs server setup

stalniy:
## Why

ref #1735

## Summary by CodeRabbit

- **New Features**
  - Added a dedicated background-jobs server and option to run REST + background processes.
  - New CLI commands for maintenance (refill-wallets, top-up-deployments, cleanup tasks, gpu-pricing-bot, stale-anonymous-user cleanup).

- **Refactor**
  - Centralized startup/bootstrap with a reusable server starter and init hooks.
  - Job-queue now supports lifecycle checks and per-job processing for improved reliability.
  - Notification templates accept optional vars.

- **Chores**
  - Updated Docker ports and local/test configs (notifications -> 3090; API exposes REST + background-jobs).


apps/api/src/rest-app.ts
```diff
@@ -89,8 +86,6 @@ appHono.use(
   })
 );
 
-const { PORT = 3080 } = process.env;
-
 const scheduler = new Scheduler({
```
baktun14, line 89: The scheduler isn't used in api, we can safely removed it if you feel like pg-boss can take the responsibility or scheduling tasks

apps/api/src/server.ts
```diff
@@ -0,0 +1,60 @@
...
+  const port = parseInt(PORT, 10) || 3080;
+
+  if (INTERFACE === "all") {
+    const boostrapList = SUPPORTED_INTERFACES.map((interfaceName, index) => bootstrapInChildProcess({ PORT: String(port + index), INTERFACE: interfaceName }));
+    await Promise.all(boostrapList);
+    return;
+  }
+
+  let appModule: { bootstrap: (port: number) => Promise<void> };
+  switch (INTERFACE) {
```
baktun14, line 21: Nice!

baktun14 (dismissed):
🚀

apps/api/src/rest-app.ts
stalniy: I decided to remove in the next PRs

baktun14 (approved):

apps/api/src/app/providers/jobs.provider.ts
```diff
@@ -10,8 +10,12 @@ container.register(APP_INITIALIZER, {
   useValue: {
     async [ON_APP_START]() {
       const jobQueueManager = container.resolve(JobQueueService);
-      await jobQueueManager.registerHandlers([container.resolve(TrialStartedHandler), container.resolve(NotificationHandler)]);
-      await jobQueueManager.start();
+      await jobQueueManager.setup();
+      await jobQueueManager.registerHandlers([
+        // keep new lines
```
ygrishajev, line 15: non-blocking. isn't there a way to configure prettier for this?

ygrishajev (approved):
amazing ❤️ 💪

apps/api/src/app/providers/jobs.provider.ts
stalniy: No to my knowledge, prettier likes people to struggle. Their philosophy is either accept what it does or don't use it 😄  because formatting is not important, it just must be universal to eliminate stylistic holy wars. We could play with line width but it will break other things.

Refs:
* https://github.com/prettier/prettier-vscode/issues/352
* https://github.com/prettier/prettier/issues/8507
* https://github.com/prettier/prettier/issues/4424

We just need to switch to https://eslint.style/ at some point. I've already used it in chain sdk project, as a replacement to prettier.

PR #1875, opened 2025-08-29
test: cover stripe transaction stream failing in the middle

### LINKED ISSUE #1830: Add test for stripe transactions failure
## Why

We need to ensure that stream failure will be properly propagated and handled on application level

## What

https://github.com/akash-network/console/pull/1790/files#r2290426484

domhhv:
plus miscellaneous leftover fixes related to Billing Usage feature

Closes #1830

## Summary by CodeRabbit

- Bug Fixes
  - Transaction CSV export is more resilient: if an error occurs while fetching pages, the download continues, preserves headers and available transactions, and includes a clear error entry in the file instead of failing outright.

- UI/UX
  - Loading screens now display a clean spinner without the “Please wait...” text for a less distracting experience.


apps/deploy-web/src/hoc/guard/guard.hoc.tsx
```diff
@@ -11,7 +11,7 @@ export const Guard = <P extends object>(Component: React.ComponentType<P>, useCh
     const { canVisit, isLoading } = useCheck();
 
     if (isLoading) {
-      return <Loading text="Please wait..." />;
+      return <Loading text="" />;
```
baktun14, line 14: question: How does it look like? Should we display text?

baktun14 (approved):

apps/deploy-web/src/hoc/guard/guard.hoc.tsx
ygrishajev: i think it's just a spinner

domhhv: Yes, it's just a spinner. Currently, when first loaded, the billing [page](https://console-beta.akash.network/billing) on beta displays an empty spinner followed by a spinner with text below, both appear momentarily. So I decided to remove the text to avoid that flickering.

PR #1877, opened 2025-08-30
fix: anonymous user reassignment

stalniy:
## Why

this bug was introduced during refactoring in https://github.com/akash-network/console/pull/1796. The thing is that callback url is called in browser, via redirect and in this case obviously our api won't receive additional headers.

## Summary by CodeRabbit

- **New Features**
  - Added a reusable Sign Up button and migrated signup actions to a centralized signup flow.

- **Bug Fixes**
  - Improved anonymous/trial cookie/header handling and cleanup; added environment-aware Secure cookie behavior.
  - More consistent token propagation for API requests via centralized interceptors.

- **Chores**
  - Server env schema now includes NODE_ENV.
  - Centralized feature-flag type introduced and reused across the app.

- **Tests**
  - Added/updated unit tests for signup UI, auth flows, and token interceptors.


apps/deploy-web/src/services/auth/auth/auth.service.spec.ts
baktun14, line 1: question: was it intentional to move it down a folder level into another `/auth/auth` folder?

apps/deploy-web/src/services/auth/auth/auth.service.ts
```diff
@@ -0,0 +1,63 @@
...
+    private readonly location = window.location,
+    private readonly localStorage = window.localStorage
+  ) {}
+
+  async signup(options?: { returnTo?: string }): Promise<void> {
+    const params = new URLSearchParams();
+    if (options?.returnTo) params.append("returnTo", options.returnTo);
+
+    // send http request to store anonymous user token in a cookie
+    // this is is needed only when anonymous free trial is enabled
```
baktun14, line 20:
```suggestion
    // this is only needed when anonymous free trial is enabled
```

baktun14: Do we need to check for the feature flag?

apps/deploy-web/src/services/auth/auth/auth.service.ts
```diff
@@ -0,0 +1,63 @@
...
+    });
+    const queryParams = params.toString();
+    // redirect user to the same url because it's impossible to read Location header in browser
+    this.location.assign(this.urlService.signup() + (queryParams ? `?${queryParams}` : ""));
+  }
+
+  logout() {
+    this.localStorage.removeItem(ANONYMOUS_USER_TOKEN_KEY);
+    this.localStorage.removeItem(ONBOARDING_STEP_KEY);
+    this.location.assign(this.urlService.logout());
```
baktun14, line 34: question: is it the same as `window.location.href = this.urlService.logout();`?

baktun14: I just remember that we had to change to `window.location.href = this.urlService.logout();` to make sure auth0 properly clears its cookies so it doesn't auto reconnect when you click signin/signup

apps/deploy-web/src/services/auth/auth/auth.service.ts
stalniy: It requires reconfiguration of unleash on frontend and I just don’t want to do it in this PR. It seems like too much for such a simple thing

apps/deploy-web/src/services/auth/auth/auth.service.spec.ts
stalniy: Ah, yes. It was intentional because auth service should live auth folder with own tests. 

other files in auth folder needs to be moved out (bit bucket, GitHub, etc)

apps/deploy-web/src/services/auth/auth/auth.service.ts
stalniy: Yes, they do the same thing. 

 If cookie is not removed then user wouldn’t be logged out, right?

stalniy: https://developer.mozilla.org/en-US/docs/Web/API/Location/assign

https://github.com/user-attachments/assets/76b480f7-45fd-43fa-beff-e3faa4a518c7

baktun14: Ok thanks!

apps/deploy-web/src/services/auth/auth/auth.service.ts
baktun14: Ok I see makes sense

apps/deploy-web/src/services/auth/auth/auth.service.spec.ts
baktun14: Okay sounds good

baktun14 (dismissed):

apps/deploy-web/src/pages/api/auth/signup.ts
```diff
@@ -3,11 +3,27 @@ import { handleLogin } from "@auth0/nextjs-auth0";
...
+      if (token) {
+        const lifetime = 5 * 60; // 5 minutes
+        const isSecure = services.config.NODE_ENV === "production";
+        res.setHeader(
+          "Set-Cookie",
+          `${ANONYMOUS_HEADER_COOKIE_NAME}=${token.replace(/^Bearer\s+/i, "")}; HttpOnly; ${isSecure ? "Secure;" : ""} SameSite=Lax; Path=/api/auth/callback; Max-Age=${lifetime}`
+        );
+        res.status(204).end();
+        return;
+      }
```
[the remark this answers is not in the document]
stalniy: can I just encode value of the header?

PR #1878, opened 2025-08-31
feat(billing): attempt to fix account sequence errors with rpc proxy

baktun14:
## Summary by CodeRabbit

* New Features
  * Configurable REST API endpoint via new environment variable, used for API base URL resolution.
  * Optional flag to always fetch fresh account info during batch signing for improved consistency.
* Changes
  * Increased default wallet batching interval to 3000 ms.
* Tests
  * Updated functional and unit test environments to include the new REST API endpoint variable.
* Chores
  * Added an additional Akash RPC endpoint to improve connectivity.
  * Aligned environment schema and URL sourcing with the new configuration.


apps/api/src/billing/lib/batch-signing-client/batch-signing-client.service.ts
```diff
@@ -288,4 +296,23 @@ export class BatchSigningClientService {
   private async simulate(messages: readonly EncodeObject[], memo: string) {
     return (await this.clientAsPromised).simulate(await this.getCachedFirstAddress(), messages, memo);
   }
+
+  private async getFreshAccountInfo(): Promise<ShortAccountInfo> {
+    const client = await this.clientAsPromised;
+    const address = await this.getCachedFirstAddress();
+    const account = await client.getAccount(address);
```
stalniy, line 303: **thought(non-blocking):** I guess that at this moment we dont know which rpc will be used by rpc proxy. if it's a different one than the one we will use to send tx, then there is still likelihood that they won't be in sync. 

But we need to ensure that `getAccount` and `sign` uses the same rpc proxy

stalniy (approved):

apps/api/src/billing/lib/batch-signing-client/batch-signing-client.service.ts
baktun14: Yea you're right, I was just thinking about this this morning and realized this probably won't work for that reason.

apps/api/src/billing/lib/batch-signing-client/batch-signing-client.service.ts
```diff
@@ -288,4 +296,23 @@ export class BatchSigningClientService {
   private async simulate(messages: readonly EncodeObject[], memo: string) {
     return (await this.clientAsPromised).simulate(await this.getCachedFirstAddress(), messages, memo);
   }
+
+  private async getFreshAccountInfo(): Promise<ShortAccountInfo> {
```
ygrishajev, line 300: seems like this function duplicates the implementation of `getCachedAccountInfo`. if so, it needs to be used there, right?

baktun14: I'll convert this pr to draft while we wait for updates from core team. I did this one thinking we could just get the new account info before every batch butt it doens't matter since every call to the rpc will result to a random rpc under the hood as of now.

Issue #1863, opened 2025-08-28
Cannot verify email in new onboarding flow


stalniy:
## Steps

1. Signup as non verified user
2. Start trial
3. Go to verification email step
4. Verify email
5. Go back to console, refresh page

## Actual result

The page says that email is still not verified

## Expected result

Email must be verified and deploy-web should allow to move to the next step

## Additional details

This happens because `email_verified` is not synched with our database `emailVerified` field. This happens because of the recent changes https://github.com/akash-network/console/pull/1796 where I moved user update login in callback handler from oauth provider. As a result if user do not re-login, `emailVerified` field in db won't be updated.

Ideally, we need to sync this field right after verification, maybe by redirecting user from email to our API. Though this is not easy because auth0 doesn't support callback for email verification. Potentially we could use "Redirect To" field. 

Another solution is to use our own endpoint to verify user email. When user clicks on the link, our API will be called, we will verify user on auth0 side (by generating verification ticket), update emailVerified field and then redirect him whenever we want. This requires template customization.

The easiest possible solution is to sync this field on requesting information about current user but I don't like this approach because we won't know whether user is verified (in our db) until he fetches information about himself. Though it may work as a short-term fix.

Some refs:
* https://community.auth0.com/t/setting-an-email-verification-server-callback/7537/2
* https://community.auth0.com/t/how-to-get-user-id-at-redirect-to-url-at-email-templates-of-verify-email-using-link/125167/3
* https://auth0.com/docs/customize/email/email-templates

baktun14:
@jzsfkzm Can you take this?

baktun14:
1. update auth0 email template redirect
2. implement custom success email verification page that receives email
3. api endpoint that checks the email verification with auth0 service `this.managementClient.usersByEmail.getByEmail({ email: "john@example.com" });`
4. update database if user is verified
5. redirect to console home page if success?

PR #1880, opened 2025-09-01
fix: sync email_verified from auth0 on demand

jzsfkzm:
refs #1863

## Summary by CodeRabbit

* **New Features**
  * Added "Verify Email" API endpoint, website verify-email page with success/failure feedback and auto-continue, client SDK verifyEmail(email) method, and a timed AutoButton UI component.

* **Security**
  * Verification enforces that only the signed-in user may verify their own email; verification status is synced from the identity provider.

* **Integration**
  * Added identity provider integration to fetch remote email verification status and sync with local user records.

* **Tests**
  * New unit and functional tests covering verify-email behaviors and edge cases.

### DIFF
--- a/apps/api/src/auth/controllers/auth/auth.controller.ts
+++ b/apps/api/src/auth/controllers/auth/auth.controller.ts
@@ -1,14 +1,17 @@
 import { singleton } from "tsyringe";
 
 import type { SendVerificationEmailRequestInput } from "@src/auth";
+import { VerifyEmailRequest } from "@src/auth/http-schemas/verify-email.schema";
 import { AuthService, Protected } from "@src/auth/services/auth.service";
 import { Auth0Service } from "@src/auth/services/auth0/auth0.service";
+import { UserService } from "@src/user/services/user/user.service";
 
 @singleton()
 export class AuthController {
   constructor(
     private readonly authService: AuthService,
-    private readonly auth0: Auth0Service
+    private readonly auth0: Auth0Service,
+    private readonly userService: UserService
   ) {}
 
   @Protected()
@@ -19,5 +22,17 @@ export class AuthController {
     if (currentUser?.userId) {
       await this.auth0.sendVerificationEmail(currentUser.userId);
     }
+  }
+
+  @Protected()
+  async syncEmailVerified({ data: { email } }: VerifyEmailRequest) {
+    this.authService.throwUnlessCan("verify-email", "User", { email });
+    const { emailVerified } = await this.userService.syncEmailVerified({ email });
+
+    return {
+      data: {
+        emailVerified
+      }
+    };
   }
 }
--- a/apps/api/src/auth/services/auth0/auth0.service.spec.ts
+++ b/apps/api/src/auth/services/auth0/auth0.service.spec.ts
@@ -123,18 +123,11 @@ describe(Auth0Service.name, () => {
         });
       }
 
-      const auth0Service = new (class extends Auth0Service {
-        constructor() {
-          super(mockAuthConfig);
-          // Override the managementClient property
-          Object.assign(this, { managementClient: mockManagementClient });
-        }
-      })();
+      const auth0Service = new Auth0Service(mockManagementClient);
 
       return {
         auth0Service,
         mockGetByEmail,
-        mockAuthConfig,
         mockManagementClient
       };
     }
--- a/apps/api/src/auth/services/auth0/auth0.service.spec.ts
+++ b/apps/api/src/auth/services/auth0/auth0.service.spec.ts
@@ -5,18 +5,16 @@ import { mock } from "jest-mock-extended";
 import type { AuthConfigService } from "@src/auth/services/auth-config/auth-config.service";
 import { Auth0Service } from "./auth0.service";
 
+import { Auth0UserSeeder } from "@test/seeders";
+
 describe(Auth0Service.name, () => {
   describe("getUserByEmail", () => {
     it("should return user when user is found", async () => {
       const email = faker.internet.email();
-      const mockUser: Partial<GetUsers200ResponseOneOfInner> = {
-        user_id: faker.string.uuid(),
+      const mockUser: Partial<GetUsers200ResponseOneOfInner> = Auth0UserSeeder.create({
         email: email,
-        email_verified: true,
-        name: faker.person.fullName(),
-        created_at: faker.date.past().toISOString(),
-        updated_at: faker.date.recent().toISOString()
-      };
+        email_verified: true
+      });
 
       const { auth0Service, mockGetByEmail } = setup({
         mockUsers: [mockUser as GetUsers200ResponseOneOfInner]
@@ -44,22 +42,14 @@ describe(Auth0Service.name, () => {
     it("should return first user when multiple users are found", async () => {
       const email = faker.internet.email();
       const mockUsers: Partial<GetUsers200ResponseOneOfInner>[] = [
-        {
-          user_id: faker.string.uuid(),
-          email: email,
-          email_verified: true,
-          name: faker.person.fullName(),
-          created_at: faker.date.past().toISOString(),
-          updated_at: faker.date.recent().toISOString()
-        },
-        {
-          user_id: faker.string.uuid(),
-          email: email,
-          email_verified: false,
-          name: faker.person.fullName(),
-          created_at: faker.date.past().toISOString(),
-          updated_at: faker.date.recent().toISOString()
-        }
+        Auth0UserSeeder.create({
+          email,
+          email_verified: true
+        }),
+        Auth0UserSeeder.create({
+          email,
+          email_verified: false
+        })
       ];
 
       const { auth0Service, mockGetByEmail } = setup({
--- a/apps/api/src/auth/services/auth0/auth0.service.spec.ts
+++ b/apps/api/src/auth/services/auth0/auth0.service.spec.ts
@@ -72,18 +72,6 @@ describe(Auth0Service.name, () => {
       expect(mockGetByEmail).toHaveBeenCalledWith({ email });
     });
 
-    it("should propagate errors from Auth0 API", async () => {
-      const email = faker.internet.email();
-      const error = new Error("Auth0 API error");
-
-      const { auth0Service, mockGetByEmail } = setup({
-        mockError: error
-      });
-
-      await expect(auth0Service.getUserByEmail(email)).rejects.toThrow("Auth0 API error");
-      expect(mockGetByEmail).toHaveBeenCalledWith({ email });
-    });
-
     it("should handle empty email string", async () => {
       const email = "";
 
--- a/apps/api/src/auth/services/auth0/auth0.service.spec.ts
+++ b/apps/api/src/auth/services/auth0/auth0.service.spec.ts
@@ -1,0 +1,142 @@
+import { faker } from "@faker-js/faker";
+import type { GetUsers200ResponseOneOfInner, ManagementClient } from "auth0";
+import { mock } from "jest-mock-extended";
+
+import type { AuthConfigService } from "@src/auth/services/auth-config/auth-config.service";
+import { Auth0Service } from "./auth0.service";
+
+describe(Auth0Service.name, () => {
+  describe("getUserByEmail", () => {
+    it("should return user when user is found", async () => {
+      const email = faker.internet.email();
+      const mockUser: Partial<GetUsers200ResponseOneOfInner> = {
+        user_id: faker.string.uuid(),
+        email: email,
+        email_verified: true,
+        name: faker.person.fullName(),
+        created_at: faker.date.past().toISOString(),
+        updated_at: faker.date.recent().toISOString()
+      };
+
+      const { auth0Service, mockGetByEmail } = setup({
+        mockUsers: [mockUser as GetUsers200ResponseOneOfInner]
+      });
+
+      const result = await auth0Service.getUserByEmail(email);
+
+      expect(result).toEqual(mockUser);
+      expect(mockGetByEmail).toHaveBeenCalledWith({ email });
+    });
+
+    it("should return null when no user is found", async () => {
+      const email = faker.internet.email();
+
+      const { auth0Service, mockGetByEmail } = setup({
+        mockUsers: []
+      });
+
+      const result = await auth0Service.getUserByEmail(email);
+
+      expect(result).toBeNull();
+      expect(mockGetByEmail).toHaveBeenCalledWith({ email });
+    });
+
+    it("should return first user when multiple users are found", async () => {
+      const email = faker.internet.email();
+      const mockUsers: Partial<GetUsers200ResponseOneOfInner>[] = [
+        {
+          user_id: faker.string.uuid(),
+          email: email,
+          email_verified: true,
+          name: faker.person.fullName(),
+          created_at: faker.date.past().toISOString(),
+          updated_at: faker.date.recent().toISOString()
+        },
+        {
+          user_id: faker.string.uuid(),
+          email: email,
+          email_verified: false,
+          name: faker.person.fullName(),
+          created_at: faker.date.past().toISOString(),
+          updated_at: faker.date.recent().toISOString()
+        }
+      ];
+
+      const { auth0Service, mockGetByEmail } = setup({
+        mockUsers: mockUsers as GetUsers200ResponseOneOfInner[]
+      });
+
+      const result = await auth0Service.getUserByEmail(email);
+
+      expect(result).toEqual(mockUsers[0]);
+      expect(mockGetByEmail).toHaveBeenCalledWith({ email });
+    });
+
+    it("should propagate errors from Auth0 API", async () => {
+      const email = faker.internet.email();
+      const error = new Error("Auth0 API error");
+
+      const { auth0Service, mockGetByEmail } = setup({
+        mockError: error
+      });
+
+      await expect(auth0Service.getUserByEmail(email)).rejects.toThrow("Auth0 API error");
+      expect(mockGetByEmail).toHaveBeenCalledWith({ email });
+    });
+
+    it("should handle empty email string", async () => {
+      const email = "";
+
+      const { auth0Service, mockGetByEmail } = setup({
+        mockUsers: []
+      });
+
+      const result = await auth0Service.getUserByEmail(email);
+
+      expect(result).toBeNull();
+      expect(mockGetByEmail).toHaveBeenCalledWith({ email: "" });
+    });
+
+    function setup(input: { mockUsers?: GetUsers200ResponseOneOfInner[]; mockError?: Error }) {
+      const mockAuthConfig = mock<AuthConfigService>();
+      mockAuthConfig.get.mockImplementation((key: string) => {
+        const config = {
+          AUTH0_M2M_DOMAIN: "test-domain.auth0.com",
+          AUTH0_M2M_CLIENT_ID: "test-client-id",
+          AUTH0_M2M_SECRET: "test-client-secret"
+        };
+        return config[key as keyof typeof config];
+      });
+
+      const mockGetByEmail = jest.fn();
+      const mockManagementClient = {
+        usersByEmail: {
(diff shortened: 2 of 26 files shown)


apps/api/src/auth/routes/email-verified/email-verified.router.ts
```diff
@@ -0,0 +1,34 @@
...
+
+const EmailVerifiedQuerySchema = z.object({
+  email: z.string()
+});
+
+export type EmailVerifiedQuery = z.infer<typeof EmailVerifiedQuerySchema>;
+
+const route = createRoute({
+  method: "get",
+  path: "/v1/email-verified",
```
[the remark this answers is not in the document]
jzsfkzm: Auth0 can send the user's browser to a "redirect to" URL after verifying email. This will be that URL, that's why it's a GET endpoint.

apps/api/src/auth/routes/email-verified/email-verified.router.ts
```diff
@@ -0,0 +1,34 @@
+import { createRoute } from "@hono/zod-openapi";
+import { container } from "tsyringe";
+import { z } from "zod";
+
+import { AuthController } from "@src/auth/controllers/auth/auth.controller";
+import { OpenApiHonoHandler } from "@src/core/services/open-api-hono-handler/open-api-hono-handler";
+
+export const emailVerifiedRouter = new OpenApiHonoHandler();
```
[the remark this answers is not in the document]
jzsfkzm: Moved schema to a schema file, no circular references any more.

apps/api/src/auth/routes/email-verified/email-verified.router.ts
```diff
@@ -0,0 +1,34 @@
...
+    301: {
+      description: "Redirects to the front page"
+    }
+  }
+});
+
+emailVerifiedRouter.get(route.path, async function emailVerified(c) {
+  await container.resolve(AuthController).syncEmailVerified({ email: c.req.query("email") as string });
+  return c.redirect("/", 302);
+});
```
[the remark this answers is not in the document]
jzsfkzm: I'd rather keep `emailVerifiedRouter.get`, so the endpoint won't land in Swagger docs.

apps/api/src/auth/routes/email-verified/email-verified.router.ts
```diff
@@ -0,0 +1,28 @@
...
+const route = createRoute({
+  method: "get",
+  path: "/v1/email-verified",
+  summary: "Checks if the email is verified",
+  tags: ["Users"],
+  request: {
+    query: EmailVerifiedQuerySchema
+  },
+  responses: {
+    301: {
```
baktun14:
```suggestion
    302: {
```

apps/api/src/services/db/userDataService.ts
```diff
@@ -221,3 +221,16 @@ export async function subscribeToNewsletter(userId: string) {
     }
   );
 }
+
+export async function setEmailVerified(userId: string, emailVerified: boolean) {
```
baktun14: Let's move this in the `UserService`

baktun14:
Next step will be to implement the page and then we can update the email template?

apps/api/test/functional/auth0.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+import { UserRepository } from "@src/user/repositories/user/user.repository";
+
+import { UserSeeder } from "@test/seeders/user.seeder";
+
+jest.setTimeout(30000);
+
+describe("Auth0 Email Verification", () => {
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
```
ygrishajev: one of the point of using `setup` function is that jest hooks aren't needed as every test case gets it's own clean state

apps/api/src/auth/controllers/auth/auth.controller.ts
```diff
@@ -20,4 +24,27 @@ export class AuthController {
       await this.auth0.sendVerificationEmail(currentUser.userId);
     }
   }
+
+  @Protected()
+  async syncEmailVerified({ data: { email } }: VerifyEmailRequest) {
+    const { currentUser } = this.authService;
+
+    const targetEmail = email.trim().toLowerCase();
+    const currentEmail = currentUser.email?.trim().toLowerCase();
```
ygrishajev: this duplicates zod schema transfomations

apps/api/src/auth/controllers/auth/auth.controller.ts
```diff
@@ -20,4 +24,27 @@ export class AuthController {
...
+    assert(currentUser.userId && currentEmail === targetEmail, 401);
+
+    const auth0User = await this.auth0.getUserByEmail(targetEmail);
+    assert(auth0User, 404);
+
+    await this.userService.upsertUser({
+      userId: currentUser.userId,
+      emailVerified: auth0User.email_verified
+    });
```
ygrishajev: I think that it's a bit too much logic for a controller. It seems a bit more appropriate to implement a certain method inside user service (`syncEmailVerified`), it would be more re-usable

apps/api/src/auth/controllers/auth/auth.controller.ts
```diff
@@ -20,4 +24,27 @@ export class AuthController {
...
     }
   }
+
+  @Protected()
+  async syncEmailVerified({ data: { email } }: VerifyEmailRequest) {
+    const { currentUser } = this.authService;
+
+    const targetEmail = email.trim().toLowerCase();
+    const currentEmail = currentUser.email?.trim().toLowerCase();
+    assert(currentUser.userId && currentEmail === targetEmail, 401);
```
ygrishajev: 403 code here seems more appropriate

ygrishajev: This looks like a permission job here. `REGULAR_*` roles can be updated to contain permissions like

```
{ action: ["read", "update"], subject: "User", conditions: { $or: [{ id: "${user.id}" }, { email: "${user.email}" }] } },
```
or
```
{ action: "verify-email", subject: "User", conditions: { email: "${user.email}" } },
```

and then checked against permissions:
```
ability.can("update", subject("User", { email })
ability.can("verify-email", subject("User", { email })
```

we are already planning admin dashboard where straight permission check can be problematic.

apps/api/src/auth/controllers/auth/auth.controller.ts
jzsfkzm: Oh, indeed, turning email into targetEmail was unnecessary as we already transform that input with the schema.

ygrishajev (commented):
thanks for the changes @jzsfkzm! i think it's essential to update the type here so it's not misused accidentally https://github.com/akash-network/console/pull/1880/files#diff-98e087b6c1c3c2f322904604e83ce6cb3632af0db4438fc354d08945d9428d37R55 - should also have optional email

apps/api/test/functional/verify-email.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+    });
+
+    async function setup(input: { emailVerified: boolean }) {
+      const walletTestingService = new WalletTestingService(app);
+      const { user, token } = await walletTestingService.createRegisteredUser({
+        email_verified: input.emailVerified
+      });
+
+      const auth0Service = container.resolve(Auth0Service);
+      jest.spyOn(auth0Service, "getUserByEmail").mockImplementation(async (email: string) => {
```
ygrishajev: I'm thinking this should be mocked via request interception as this is a functional? wdyt @baktun14 @stalniy ?

apps/api/test/functional/verify-email.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+
+      expect(response.status).toBe(200);
+      const result = await response.json();
+      expect(result).toMatchObject({
+        data: {
+          emailVerified: true
+        }
+      });
+
+      const userSetting = await UserSetting.findOne({ where: { userId: user.userId } });
```
ygrishajev: as this is a functional test, I'm thinking this should be using endpoints.

```ts
const currentUserResponse = await app.request("/v1/user/me", {
  method: "GET",
  headers: {
    authorization: `Bearer ${token}`,
    "content-type": "application/json"
  }
});
const body = await currentUserResponse.json();
expect(body.data.emailVerified).toBe(true);
```

apps/api/test/functional/verify-email.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+      });
+
+      const userAuthTokenService = container.resolve(UserAuthTokenService);
+      jest.spyOn(userAuthTokenService, "getValidUserId").mockResolvedValue(user.userId);
+
+      const userRepository = container.resolve(UserRepository);
+      jest.spyOn(userRepository, "findById").mockResolvedValue(user);
+      jest.spyOn(userRepository, "findByUserId").mockResolvedValue(user);
+
+      jest.spyOn(UserSetting, "update");
```
ygrishajev: I don't think any mocks above are needed (besides 3rd party auth0). App should function normally without tweaks

apps/deploy-web/src/pages/user/verify-email/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+        <h3 className="text-2xl">
+          {isVerifying && "Just a moment while we finish verifying your email."}
+          {isVerified === true && "Your email was verified. You can continue using the application."}
+          {isVerified === false && "Your email was not verified. Please try again."}
+        </h3>
+
+        {isVerified === true && (
+          <div className="pt-6">
+            <Link href={UrlService.home()} className={cn(buttonVariants({ variant: "default" }), "inline-flex items-center")}>
+              Continue
```
ygrishajev: shouldn't we just navigate and show some success message?

packages/http-sdk/src/auth/auth-http.types.ts
```diff
@@ -0,0 +1,9 @@
+import { z } from "zod";
+
+export const VerifyEmailResponseSchema = z.object({
```
ygrishajev, line 3: why is schema needed here? perhaps just type would be enough?

apps/deploy-web/src/pages/user/verify-email/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+    if (email) {
+      verifyEmail(email);
+    }
+  });
+
+  return (
+    <Layout isLoading={isVerifying}>
+      <NextSeo title="Verifying your email" />
+
+      <div className="mt-10 text-center">
```
baktun14: We can add the Akash Console logo and big icon (check or X or spinner) to make it prettier and more interactive.

apps/deploy-web/src/pages/user/verify-email/index.tsx
baktun14: I agree, we could add a little coutndown "5 seconds until being redirected" and keep the continue button to fast forward

jzsfkzm:
> thanks for the changes @jzsfkzm! i think it's essential to update the type here so it's not misused accidentally https://github.com/akash-network/console/pull/1880/files#diff-98e087b6c1c3c2f322904604e83ce6cb3632af0db4438fc354d08945d9428d37R55 - should also have optional email

Hm, that's right, types should be the same, but shouldn't it rather be accepting the same UserOutput we're passing when using this method?
```diff
-  getAbilityFor(role: Role, user: { userId: string | null }) {
+  getAbilityFor(role: Role, user: UserOutput) {
```

ygrishajev:
> Hm, that's right, types should be the same, but shouldn't it rather be accepting the same UserOutput we're passing when using this method?

this can work too

apps/api/test/functional/verify-email.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+
+        return Promise.resolve(null);
+      });
+
+      const userAuthTokenService = container.resolve(UserAuthTokenService);
+      jest.spyOn(userAuthTokenService, "getValidUserId").mockResolvedValue(user.userId);
+
+      const userRepository = container.resolve(UserRepository);
+      jest.spyOn(userRepository, "findById").mockResolvedValue(user);
+      jest.spyOn(userRepository, "findByUserId").mockResolvedValue(user);
```
ygrishajev: none of these mocks (except 3rd part auth0) is really needed i think. it should be possible to create a user and it would work without them

apps/api/src/auth/controllers/auth/auth.controller.ts
```diff
@@ -20,4 +23,16 @@ export class AuthController {
       await this.auth0.sendVerificationEmail(currentUser.userId);
     }
   }
+
+  @Protected()
```
baktun14, line 27:
```suggestion
  @Protected([{ action: "verify-email", subject: "User" }])
```

baktun14: Instead of the `this.authService.throwUnlessCan("verify-email", "User", { email });`

jzsfkzm: But it is not the same, is it?

`@Protected([{ action: "verify-email", subject: "User" }])` means the user acting can `verify-email` some `User` in general. We don't check if they can `verify-email` the User in request (identified by the email address we send).

On the other hand, `this.authService.throwUnlessCan("verify-email", "User", { email })` would be ok only for the user with the same email, and superusers (who have `manage/all`).

What do you think?

baktun14: Ahh I see you're right nvm

apps/deploy-web/src/pages/user/verify-email/index.tsx
jzsfkzm: Added such a timeout, now it will redirect automatically.

apps/api/src/auth/services/auth0/auth0.service.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+        });
+      }
+
+      const auth0Service = new (class extends Auth0Service {
+        constructor() {
+          super(mockAuthConfig);
+          // Override the managementClient property
+          Object.assign(this, { managementClient: mockManagementClient });
+        }
+      })();
```
ygrishajev: this is the reason to extract ManagementClient to a dedicated provider and inject it into the auth0service

apps/api/src/auth/services/auth0/auth0.service.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+    it("should propagate errors from Auth0 API", async () => {
+      const email = faker.internet.email();
+      const error = new Error("Auth0 API error");
+
+      const { auth0Service, mockGetByEmail } = setup({
+        mockError: error
+      });
+
+      await expect(auth0Service.getUserByEmail(email)).rejects.toThrow("Auth0 API error");
+      expect(mockGetByEmail).toHaveBeenCalledWith({ email });
```
ygrishajev: the code doesn't handle errors and thus should not test them

apps/api/test/functional/verify-email.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+  });
+
+  async function setup(input: { emailVerified: boolean }) {
+    const walletTestingService = new WalletTestingService(app);
+    const { user, token } = await walletTestingService.createRegisteredUser({
+      email_verified: input.emailVerified
+    });
+
+    const auth0Service = container.resolve(Auth0Service);
+    jest.spyOn(auth0Service, "getUserByEmail").mockImplementation(async (email: string) => {
```
ygrishajev: should be mocked via request interception or on sdk level (ManagementClient)

apps/api/src/auth/services/auth0/auth0.service.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+    it("should return user when user is found", async () => {
+      const email = faker.internet.email();
+      const mockUser: Partial<GetUsers200ResponseOneOfInner> = {
+        user_id: faker.string.uuid(),
+        email: email,
+        email_verified: true,
+        name: faker.person.fullName(),
+        created_at: faker.date.past().toISOString(),
+        updated_at: faker.date.recent().toISOString()
+      };
```
ygrishajev: this is prob reused in here and in functional tests, should be extracted to a seeder.

apps/api/src/auth/services/auth0/auth0.service.spec.ts
jzsfkzm: Add a provider for ManagementClient.

apps/api/src/auth/services/auth0/auth0.service.spec.ts
jzsfkzm: Indeed, this class is opaque for errors, dropped this test.

apps/api/test/functional/verify-email.spec.ts
jzsfkzm: Started mocking on the ManagementClient level. I chose this as we don't really know (and I think we're fine with not knowing) about how it works on the HTTP level.

apps/api/src/auth/services/auth0/auth0.service.spec.ts
jzsfkzm: That's right, added a seeder for Auth0 users.

apps/api/src/core/providers/auth.provider.ts
```diff
@@ -0,0 +1,26 @@
...
+        clientSecret: authConfig.get("AUTH0_M2M_SECRET")
+      });
+    }
+
+    return ManagementClientFactory.instance;
+  }
+}
+
+container.register(ManagementClient, {
+  useFactory: () => ManagementClientFactory.getInstance()
```
ygrishajev: this is marked a singleton higher so the container should make sure this would be a singleton.

basically this should work:

```typescript
import { ManagementClient } from "auth0";

import { AuthConfigService } from "@src/auth/services/auth-config/auth-config.service";

export const AUTH0_MANAGEMENT_CLIENT = "AUTH0_MANAGEMENT_CLIENT";

container.register(AUTH0_MANAGEMENT_CLIENT, {
  useFactory: c => {
    const config = c.resolve(AuthConfigService);
    return new ManagementClient({
      domain: config.get("AUTH0_M2M_DOMAIN"),
      clientId: config.get("AUTH0_M2M_CLIENT_ID"),
      clientSecret: config.get("AUTH0_M2M_SECRET")
    });
  }
});


class Foo () {
   constructor(@inject(AUTH0_MANAGEMENT_CLIENT) private readonly managementClient: ManagementClient) {}
}
```

or 



```typescript
import { ManagementClient } from "auth0";

import { AuthConfigService } from "@src/auth/services/auth-config/auth-config.service";

container.register(ManagementClient, {
  useFactory: c => {
    const config = c.resolve(AuthConfigService);
    return new ManagementClient({
      domain: config.get("AUTH0_M2M_DOMAIN"),
      clientId: config.get("AUTH0_M2M_CLIENT_ID"),
      clientSecret: config.get("AUTH0_M2M_SECRET")
    });
  }
});

// above needs to be imported somewhere in the root of the app

class Foo () {
   constructor(private readonly managementClient: ManagementClient) {}
}
```

jzsfkzm: Simplified it a lot now, but the instance injected still has to be a singleton - otherwise the instance used in production and the one we get from DI in the test will be different instances, so when we mock the latter, the prior will just go unmocked, and cause errors, and tests will break this way.

ygrishajev (approved):

apps/api/src/core/providers/auth.provider.ts
```diff
@@ -0,0 +1,22 @@
+import { ManagementClient } from "auth0";
+import { container } from "tsyringe";
+
+import { AuthConfigService } from "@src/auth/services/auth-config/auth-config.service";
+
+let managementClientInstance: ManagementClient | null = null;
```
ygrishajev, line 6: non-blocking: why is this important to make it a singleton? 

if it is needed indeed, found this helper https://github.com/Microsoft/tsyringe?tab=readme-ov-file#instancecachingfactory

PR #1888, opened 2025-09-04
refactor: changes LeaseHttpService to accept http client

stalniy:
## Why

Default http client is retriable. It makes it easier to make this service a retry on network or server error. ref #1424 

## Summary by CodeRabbit

* **Refactor**
  * Standardized HTTP client usage across multiple backend services for improved consistency and maintainability.
  * Consolidated service registration to reduce duplication and simplify configuration.
  * Updated lease and related services to accept a shared HTTP client instance rather than creating their own.

* **Impact**
  * No UI or functional changes for end-users.
  * Developers integrating the SDK may need minor adjustments to service instantiation.

### DIFF
--- a/apps/notifications/src/modules/alert/providers/http-sdk.provider.ts
+++ b/apps/notifications/src/modules/alert/providers/http-sdk.provider.ts
@@ -1,26 +1,26 @@
+import type { HttpClient } from "@akashnetwork/http-sdk";
 import { createHttpClient, DeploymentHttpService, LeaseHttpService } from "@akashnetwork/http-sdk";
-import type { Provider } from "@nestjs/common";
+import type { InjectionToken, Provider } from "@nestjs/common";
 import { ConfigService } from "@nestjs/config";
 
 import type { AlertConfig } from "@src/modules/alert/config";
 
+export const CHAIN_API_HTTP_CLIENT_TOKEN: InjectionToken<HttpClient> = Symbol("CHAIN_API_HTTP_CLIENT");
+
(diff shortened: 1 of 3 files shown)


apps/notifications/src/modules/alert/providers/http-sdk.provider.ts
```diff
@@ -1,26 +1,26 @@
...
 import { ConfigService } from "@nestjs/config";
 
 import type { AlertConfig } from "@src/modules/alert/config";
 
+export const CHAIN_API_HTTP_CLIENT_TOKEN: InjectionToken<HttpClient> = Symbol("CHAIN_API_HTTP_CLIENT");
+
 export const HTTP_SDK_PROVIDERS: Provider[] = [
   {
-    provide: DeploymentHttpService,
+    provide: CHAIN_API_HTTP_CLIENT_TOKEN,
```
ygrishajev, line 12: non-blocking: why not using `HttpClient` as token?

ygrishajev (approved):

baktun14 (approved):

apps/notifications/src/modules/alert/providers/http-sdk.provider.ts
stalniy: HttpClient is a type. But also there may be many different http clients (chain api, internal services api, etc). That's why I think it makes to be explicit about what type of http client is used.

stalniy: According to soli**D**, we must code everything around interfaces and `InjectionToken`s but this not always makes sense

PR #1890, opened 2025-09-04
fix(billing): lease flow test

baktun14:
## Summary by CodeRabbit

- Documentation
  - Updated API documentation base URL to use the origin only (version path removed) while keeping docs version as v1. This aligns server listings and may improve client generation and tooling compatibility.

- Chores
  - Added a script to start a mock OAuth2 server to streamline testing workflows.

- Tests
  - Adjusted functional tests to clear balances via the updated balance service, improving test reliability around balance checks.

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -31,6 +31,7 @@
     "test:functional": "jest --selectProjects functional",
     "test:functional:cov": "npm run test:functional -- --coverage",
     "test:functional:watch": "npm run test:functional -- --watch",
+    "test:oauth2-server": "dc up:dev -d mock-oauth2-server",
     "test:unit": "jest --selectProjects unit",
     "test:unit:cov": "jest --selectProjects unit --coverage",
     "test:unit:watch": "jest --selectProjects unit --watch",
--- a/apps/api/src/core/services/openapi-docs/openapi-docs.service.ts
+++ b/apps/api/src/core/services/openapi-docs/openapi-docs.service.ts
@@ -10,7 +10,7 @@ export class OpenApiDocsService {
     const version = "v1";
     const docs = {
       openapi: "3.0.0",
-      servers: [{ url: `${env.SERVER_ORIGIN}/${version}` }],
+      servers: [{ url: `${env.SERVER_ORIGIN}` }],
       info: {
         title: "Akash Network Console API",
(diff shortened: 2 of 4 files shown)


apps/api/src/core/services/openapi-docs/openapi-docs.service.ts
```diff
@@ -10,7 +10,7 @@ export class OpenApiDocsService {
     const version = "v1";
     const docs = {
       openapi: "3.0.0",
-      servers: [{ url: `${env.SERVER_ORIGIN}/${version}` }],
+      servers: [{ url: env.SERVER_ORIGIN }],
```
stalniy, line 13: question: what is the purpose of this change?

apps/api/package.json
```diff
@@ -31,6 +31,7 @@
     "test:functional": "jest --selectProjects functional",
     "test:functional:cov": "npm run test:functional -- --coverage",
     "test:functional:watch": "npm run test:functional -- --watch",
+    "test:oauth2-server": "dc up:dev -d mock-oauth2-server",
```
stalniy, line 34: question: could you please explain this as well? Is this just a convenience?

stalniy (approved):

apps/api/package.json
baktun14: Yes just convinience for running the tests locally as it's needed for functional tests. How do you run them?

apps/api/src/core/services/openapi-docs/openapi-docs.service.ts
baktun14: In the swagger docs the version is added twice because we already append it in the route definition. Some users reported the issue.

image

stalniy: Ah! Nice catch!

apps/api/package.json
stalniy: I’m using docker for development so this thing is always up but we have another hook for ci which you could potentially use to start db and oauth server - https://github.com/akash-network/console/blob/main/apps/api/package.json#L28

baktun14: Ah ok yea that's what I thought, I don't use docker for the database because it was too slow for the indexer, so hence why I added this command to just run the oauth server.

Issue #1886, opened 2025-09-03
Implement test charge for free trial


baktun14:
- When customer adds a credit card payment in the trial, do a test charge with automatic refund to validate the card
- Only when the card has been verified that we allow the free trial activation

PR #1897, opened 2025-09-15
feat(billing): add database schema and core types for payment method validation

baktun14:
https://github.com/akash-network/console/issues/1886

- Add database migration for payment method changes
- Add new payment method types and schemas
- Update payment method repository with validation support
- Update payment method seeder for testing

## Summary by CodeRabbit

- New Features
  - Payment methods gain a validation status; methods can be marked validated and listed per user.
  - CSV exports now include Card Brand and Card Last 4 columns.

- Bug Fixes
  - Improved handling of duplicate trial-account detection across payment methods (now considers multiple candidates).

- Tests
  - Test data and seeders updated to use richer Stripe-like payment method shapes and cover validation, duplicate checks, and CSV exports.


apps/api/src/billing/services/stripe/stripe.service.spec.ts
```diff
@@ -896,19 +896,22 @@ describe(StripeService.name, () => {
...
-      });
+      paymentMethodRepository.findOthersTrialingByFingerprint.mockResolvedValue([
+        {
+          id: "existing_pm",
+          userId: "other_user",
+          fingerprint: "fp_123",
+          paymentMethodId: "pm_existing",
+          createdAt: new Date(),
+          updatedAt: new Date(),
+          isValidated: true
```
ygrishajev, line 916: non-blocking: this should be a seeder, which should also be used in other places.

ygrishajev: oh, there is a seeder! should it be used then?

baktun14: Will fix in the next pr

apps/api/src/billing/types/payment-method.type.ts
```diff
@@ -0,0 +1,30 @@
+export interface PaymentMethod {
```
ygrishajev: non-blocking: why does this type exist? why doesn't it rely on Stripe types?

apps/api/src/billing/services/stripe/stripe.service.spec.ts
ygrishajev: i see, the existing seeder is for stripe payload, i guess it gets back to initial question in the thread then

apps/api/src/billing/types/payment-method.type.ts
baktun14: Yea I'll remove it. It's used in changes that are implemented in the other branches, but using the stripe type works better.

ygrishajev (approved):

PR #1898, opened 2025-09-15
feat(billing): test charge for free trial

baktun14:
https://github.com/akash-network/console/issues/1886

## Summary by CodeRabbit

- **New Features**
  - Wallet-aware 3DS flows for payments and trial setup; endpoints return 202 when authentication is required.
  - New endpoint to finalize post-3DS validation; confirm payment now returns structured results.
  - Payment methods include a validated flag; removals now return 204; payment-method listing uses user-aware lookup.

- **Documentation**
  - OpenAPI updated to document 3DS flows, 200/202 responses, validation payloads, and payment-method details.

- **Bug Fixes**
  - Improved Stripe error mappings and clearer payment error handling.

- **Tests**
  - Expanded 3DS/validation test coverage and new test helpers/seeders.

### DIFF
--- a/apps/api/src/billing/controllers/stripe/stripe.controller.ts
+++ b/apps/api/src/billing/controllers/stripe/stripe.controller.ts
@@ -6,8 +6,9 @@ import type { infer as ZodInfer } from "zod";
 import { AuthService, Protected } from "@src/auth/services/auth.service";
 import type { StripePricesOutputResponse } from "@src/billing";
 import { CustomerTransactionsCsvExportQuerySchema } from "@src/billing/http-schemas/stripe.schema";
-import { ApplyCouponRequest, ConfirmPaymentRequest, Discount, Transaction } from "@src/billing/http-schemas/stripe.schema";
-import { PaymentMethod, StripeService } from "@src/billing/services/stripe/stripe.service";
+import { ApplyCouponRequest, ConfirmPaymentRequest, ConfirmPaymentResponse, Discount, Transaction } from "@src/billing/http-schemas/stripe.schema";
+import { UserWalletRepository } from "@src/billing/repositories";
+import { StripeService } from "@src/billing/services/stripe/stripe.service";
 import { StripeErrorService } from "@src/billing/services/stripe-error/stripe-error.service";
 import { Semaphore } from "@src/core/lib/semaphore.decorator";
 
@@ -16,7 +17,8 @@ export class StripeController {
   constructor(
     private readonly stripe: StripeService,
     private readonly authService: AuthService,
-    private readonly stripeErrorService: StripeErrorService
+    private readonly stripeErrorService: StripeErrorService,
+    private readonly userWalletRepository: UserWalletRepository
   ) {}
 
   @Protected([{ action: "read", subject: "StripePayment" }])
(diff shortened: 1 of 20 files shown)


apps/api/src/billing/controllers/stripe/stripe.controller.ts
```diff
@@ -59,15 +61,32 @@ export class StripeController {
...
+            requiresAction: true,
+            clientSecret: result.clientSecret,
+            paymentIntentId: result.paymentIntentId
+          }
+        };
+      }
+
+      // If payment was not successful and it's not a 3D Secure case, throw an error
+      if (!result.success) {
+        throw new Error("Payment not successful");
```
stalniy, line 86: **issue(non-blocking):** this will result in 500 error but I support that this usually means some issue with the card, no?

apps/api/src/billing/services/stripe/stripe.service.ts
```diff
@@ -190,7 +197,27 @@ export class StripeService extends Stripe {
...
+        return {
+          success: false,
+          paymentIntentId: paymentIntent.id,
+          requiresAction: true,
+          clientSecret: paymentIntent.client_secret || undefined
+        };
+
+      case "requires_payment_method":
+        // Card was declined
+        throw new Error("Payment method was declined. Please try a different card.");
```
stalniy, line 216: **question(non-blocking):** is this expected to be a 500 error?

apps/api/src/billing/services/stripe/stripe.service.ts
```diff
@@ -590,35 +617,220 @@ export class StripeService extends Stripe {
...
-      city: string | null;
-      country: string | null;
-      line1: string | null;
-      line2: string | null;
-      postal_code: string | null;
-      state: string | null;
-    } | null;
-    email?: string | null;
-    name?: string | null;
-    phone?: string | null;
```
stalniy, line 622: **info(non-blocking):** I added this to better hide dependency on the stripe and expose only what we use on type level. Then it's easier to mock it in tests as well

apps/api/src/billing/services/stripe/stripe.service.ts
```diff
@@ -590,35 +617,220 @@ export class StripeService extends Stripe {
...
+              event: "PAYMENT_METHOD_ALREADY_VALIDATED",
+              customerId: params.customer,
+              userId: user.id,
+              paymentMethodId: params.payment_method
+            });
+            return { success: true, paymentIntentId: "already_validated" };
+          }
+        }
+
+        throw new Error("Duplicate card validation request detected. Please wait before retrying.");
```
stalniy, line 686: **question(non-blocking):** this results in 500 error I guess, is this expected?

apps/api/src/billing/services/stripe/stripe.service.ts
```diff
@@ -590,35 +617,220 @@ export class StripeService extends Stripe {
...
+          status: paymentIntent.status
+        });
+      }
+    } catch (error) {
+      logger.error({
+        event: "FAILED_TO_CHECK_PAYMENT_INTENT_AFTER_3DS",
+        customerId,
+        paymentMethodId,
+        paymentIntentId,
+        error: error instanceof Error ? error.message : String(error)
```
stalniy, line 775: **nitpick(blocking):** just pass an `error` to logger, it automatically collects full stack trace and request specific details, so no need to do it in client code

apps/api/test/seeders/user-test.seeder.ts
```diff
@@ -0,0 +1,10 @@
+import { TEST_CONSTANTS } from "./stripe-test-data.seeder";
+import { UserSeeder } from "./user.seeder";
+
+export function createTestUser(overrides: Partial<{ id: string; stripeCustomerId: string | null }> = {}) {
```
stalniy, line 4: **question(non-blocking):** what is the difference between `createTestUser` and `UserSeeder.create` in terms of when to use what?

stalniy (approved):

apps/api/src/billing/controllers/stripe/stripe.controller.ts
baktun14: It's handled by the stripe error service and throws a 402

apps/api/src/billing/services/stripe/stripe.service.ts
baktun14: Same thing, handled in the try catch stripe error service of the controller

apps/api/src/billing/services/stripe/stripe.service.ts
baktun14: Ah I see, perhaps we can make custom types using the stripe types and pick only the fields we want. For now I'll leave it like that.

apps/api/src/billing/services/stripe/stripe.service.ts
baktun14: Same thing, stripe-error.service

apps/api/test/seeders/user-test.seeder.ts
baktun14: Not much indeed, I'll remove it in my next pr.

PR #1904, opened 2025-09-15
fix(billing): prevent 0 amount payments

baktun14:
## Summary by CodeRabbit

- New Features
  - Enforces a $20 minimum payment before discounts in API and web checkout.
  - Centralized minimum-amount constant for consistent validation.

- Bug Fixes
  - Rejects non-positive amounts with clear messaging.
  - Ensures final amount after discounts is at least $1.
  - API validation now enforces the $20 minimum and blocks negative/zero amounts with user-friendly errors.

- Tests
  - Updated test fixtures to use nested card details; no production behavior changes.

### DIFF
--- a/apps/api/src/billing/services/stripe/stripe.service.ts
+++ b/apps/api/src/billing/services/stripe/stripe.service.ts
@@ -26,9 +26,6 @@ interface StripePrices {
   isCustom: boolean;
   currency: string;
 }
-
-const MINIMUM_PAYMENT_AMOUNT = 20;
-
 @singleton()
 export class StripeService extends Stripe {
   constructor(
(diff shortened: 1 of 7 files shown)


apps/api/src/billing/services/stripe/stripe.service.ts
```diff
@@ -144,6 +146,11 @@ export class StripeService extends Stripe {
     metadata?: Record<string, string>;
   }): Promise<{ success: boolean; paymentIntentId?: string }> {
     const discounts = await this.getCustomerDiscounts(params.customer);
+
+    if (!discounts.length && params.amount < MINIMUM_PAYMENT_AMOUNT) {
+      throw new Error(`Minimum payment amount is $${MINIMUM_PAYMENT_AMOUNT} (before any discounts)`);
```
ygrishajev: I'm not sure I follow, why not putting a limit of 20 in the zod schema instead of this check?

baktun14: I was planning on removing this logic of discounts in the payment intent entirely in another pr. I'll move it to zod and remove this check for now..

ygrishajev (approved):

PR #1935, opened 2025-09-17
feat: make Unleash session id visible for backend

### LINKED ISSUE #1867: Properly propagate unleash sessionId between frontend and backend
## Why

This will let unleash reliably evaluate feature flag state for browser session between requests. The same session id must be used by frontend and backend and it should be provided to `isEnabled` method in context parameter. Otherwise unleash gradual rollout won't work reliably for session. For request feature may be enabled and for another it will be disabled

## What

Propagate unleash sessionId context parameter between frontend and backend

jzsfkzm:
closes #1867

## Summary by CodeRabbit

- New Features
  - Feature flags now consider session-specific context via the unleash-session-id cookie, enabling more precise and consistent rollouts.
  - The web proxy now forwards unleash-session-id alongside cf_clearance, preserving session-aware behavior across proxied requests.
  - No UI or public API changes; no user action required.

- Tests
  - Added tests verifying sessionId propagation and cookie-based parsing in feature flag evaluation.

### DIFF
--- a/apps/api/src/core/services/feature-flags/feature-flags.service.ts
+++ b/apps/api/src/core/services/feature-flags/feature-flags.service.ts
@@ -32,20 +32,15 @@ export class FeatureFlagsService implements Disposable, AppInitializer {
 
     assert(this.client, "Feature flags service was not initialized. Call initialize() method first.");
 
-    const clientInfo = this.executionContext.get("HTTP_CONTEXT")?.get("clientInfo");
     const currentUser = this.executionContext.get("CURRENT_USER");
     const sessionId = this.extractSessionId();
 
     return this.client.isEnabled(featureFlag, {
       currentTime: new Date(),
-      remoteAddress: clientInfo?.ip,
       userId: currentUser?.id,
       sessionId,
       environment: this.configService.get("DEPLOYMENT_ENV"),
       properties: {
-        userAgent: clientInfo?.userAgent,
-        fingerprint: clientInfo?.fingerprint,
-        nodeEnv: this.configService.get("NODE_ENV"),
         chainNetwork: this.configService.get("NETWORK")
       }
     });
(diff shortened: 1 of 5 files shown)


apps/api/src/core/services/feature-flags/feature-flags.service.ts
```diff
@@ -32,11 +34,13 @@ export class FeatureFlagsService implements Disposable, AppInitializer {
 
     const clientInfo = this.executionContext.get("HTTP_CONTEXT")?.get("clientInfo");
     const currentUser = this.executionContext.get("CURRENT_USER");
+    const sessionId = this.extractSessionId();
 
     return this.client.isEnabled(featureFlag, {
       currentTime: new Date(),
       remoteAddress: clientInfo?.ip,
```
stalniy: **question(blocking):** do we need to keep frontend and backend context to be the same?

I assume this is required to ensure correct feature toggling during gradual rollout. If this assumption is correct, then we need to remove everything what is not available on frontend, like userAgent, ip, fingerprint and nodeEnv

jzsfkzm: Not sure if we want to remove those. Properties available in both ends (frontend and backend) can be used in Unleash rules for consistent flags, that is true. In other words, an Unleash rule relying on a frontend only/backend only property will not lead to a consistent flag.
Question is if we will ever want a flag to be relying on a backend property. In that case we would want to make backend properties be opaque (usable on frontend) too.

Anyway, if we don't use those now, let's ditch those, for the sake of simplicity. We can always re-introduce them. Agreed, @stalniy ? Can you check if there are no Unleash rules relying on backend properties at the moment?

stalniy: we don't use any of that for sure. The main use cases are:
- enable by audience
- enable by session
- gradual rollout -> from what I read in past, it depends on what is passed to the context and it's important for backend and frontend context to be in sync in order to have consistent toggle for single session. Otherwise we may end up with backend flag being out of sync with frontend flag. Also right now we don't use FF on other services (e.g., notifications, indexer, provider-proxy) which may not have some information or it may differ, so it's just easier to sync all of them having less properties for now

Initially those properties were added by me but I didn't know the consequences, So I believe we can safely remove them.

stalniy: I've just looked into the doc. And want to note that I was wrong. https://docs.getunleash.io/reference/stickiness

stalniy (approved):

PR #1943, opened 2025-09-19
fix: ensure anonymous user does not cause sentry errors

stalniy:
## Why

currently getOrCreateAnonymousUser function result is cached without taking into account whether it's a success or error. as a result `useAnonymousUserQuery` hook tries to re-trigger its side-effect and always falls into error state (until the next page refresh in 1h). This generates too many errors from a single user in sentry and burns our sentry quota 

## Summary by CodeRabbit

- New Features
  - Smarter retry scheduling when rate-limited while creating or fetching an anonymous user.
- Bug Fixes
  - Prevents duplicate network requests and race conditions for anonymous user creation/fetch.
  - More reliable loading and error states, with clearer recovery after failures.
- Refactor
  - Streamlined anonymous user request flow for greater resilience and consistency.
- Tests
  - Added comprehensive test coverage for anonymous user lifecycle, retries, and error handling.
- Chores
  - Cleaned up internal imports and paths with no impact on public APIs or runtime behavior.


packages/http-sdk/src/user-http/user-http.service.ts
```diff
@@ -25,13 +24,31 @@ export type UserCreateResponse = {
...
+
   constructor(config?: AxiosRequestConfig) {
     super(config);
-    this.getOrCreateAnonymousUser = memoize(this.getOrCreateAnonymousUser.bind(this));
   }
 
-  async getOrCreateAnonymousUser(id?: string): Promise<UserCreateResponse | ApiOutput<UserOutput>> {
-    return await (id ? this.getAnonymousUser(id) : this.createAnonymousUser());
+  getOrCreateAnonymousUser(id?: string): Promise<UserCreateResponse | ApiOutput<UserOutput>> {
+    const key = id ?? String(this.userCreateId++);
```
stalniy, line 35: right now `getOrCreateAnonymousUser` is bound to its caller context. it expects that application may have only 1 anonymous user which is true from deploy-web perspective but false from http-sdk perspective which potentially may need to create multiple anonymous users. 

This is the reason why I create a custom key, in case user is not provided

packages/http-sdk/src/user-http/user-http.service.ts
```diff
@@ -25,13 +24,31 @@ export type UserCreateResponse = {
...
+      this.anonymousUsersPromise[key] = promise
+        .then(response => {
+          if (!id) {
+            // do not cache create user response
+            delete this.anonymousUsersPromise[key];
+          }
+          return response;
+        })
+        .catch(error => {
+          delete this.anonymousUsersPromise[key];
```
stalniy, line 47: we must not cache rejected promises otherwise retry won't be possible

apps/deploy-web/src/queries/useAnonymousUserQuery/useAnonymousUserQuery.ts
```diff
@@ -0,0 +1,71 @@
...
+        if (get(userStateAtom).isLoading) return;
+        try {
+          set(userStateAtom, { isLoading: true });
+          const { data: fetched, ...rest } = await userService.getOrCreateAnonymousUser(id);
+          const token = "token" in rest ? rest.token : undefined;
+          set(userStateAtom, { user: fetched, token, isLoading: false });
+        } catch (error: any) {
+          const retryAfterInSeconds =
+            isHttpError(error) && error.response?.status === 429 ? error.response.data.retryAfter || DEFAULT_RETRY_AFTER_SECONDS : 10_000;
+          set(userStateAtom, { isLoading: false, error, retryAfter: new Date(Date.now() + retryAfterInSeconds * 1000) });
```
stalniy, line 48: we need retryAfter in case of any error because if we don't do this, hook will stuck in infinite loop due to condition in `useWhen`. change to `userStateAtom` atom will trigger re-render of the component which will call `useWhen` which condition evaluates to `true` and triggers side-effect again.

I also don't want it to stop retrying in case of error

baktun14 (dismissed):

Issue #1934, opened 2025-09-17
Failing login via sign up link


ygrishajev:
Steps to reproduce:
1) go to https://console.akash.network/
2) sign up via email/pass flow
3) log out
4) click sign up again
5) in the auth screen go to login
6) log back in

EXR: console api `POST /v1/register-user` is gracefully handling a callback as user is already created
ACR: console api `POST /v1/register-user` fails with duplicate 500 error

stalniy:
The issue happens in case we create new anonymous user for existing one. So:
1. Register a new user
2. Logout, clear localStorage, refresh the page
3. Try to login with previously registered user

The issue happens because [at line here](https://github.com/akash-network/console/blob/main/apps/api/src/user/services/user/user.service.ts#L48) we check whether there is anonymous user Id in payload, in this case we have one, then it tries to update this anonymous user data with data which received from auth0 and it fails because user with such id already exists (it was created on step #1).

```
update "userSetting"
set
  "userId" = $1,
  "email" = $2,
  "emailVerified" = $3,
  "subscribedToNewsletter" = $4,
  "last_ip" = $5,
  "last_user_agent" = $6,
  "last_fingerprint" = $7
where
  (
    "userSetting"."id" = $8
    and "userSetting"."userId" is null
  )
returning
  "id",
  "userId",
  "username",
  "email",
  "emailVerified",
  "stripeCustomerId",
  "bio",
  "subscribedToNewsletter",
  "youtubeUsername",
  "twitterUsername",
  "githubUsername",
  "last_active_at",
  "last_ip",
  "last_user_agent",
  "last_fingerprint",
  "created_at" -- params: ["52381fcb-c25a-4ae7-8490-35dd977053c3", "Marisa_Crona@hotmail.com", true, false, "::ffff:192.168.97.5", "node", "0a55bde02b5fada44d95c9f8aceefb40e5ca664e372036b6e72865162fbf795a", "de0d403b-1481-4ae9-99a9-9140fa44bdc4"]
    context: "POSTGRES"
    orm: "drizzle"
[11:56:47.487] ERROR (48): PostgresError: duplicate key value violates unique constraint "userSetting_userId_unique" (code: 23505)
    at ErrorResponse (webpack://akashnetwork/node_modules/postgres/cjs/src/connection.js:788:1)
    at handle (webpack://akashnetwork/node_modules/postgres/cjs/src/connection.js:474:1)
[...]

PR #1951, opened 2025-09-23
fix: ensure user can be signed in with a different anonymous user id but still be resolved to proper user

jzsfkzm:
refs #1934

## Summary by CodeRabbit

* **Bug Fixes**
  * Safer anonymous-session linking: when a conflict with an existing account occurs during anonymous linking, the linking step is now safely aborted for that conflict type and does not raise an error, preventing unintended account linking or profile changes.

* **Tests**
  * Added a test to validate behavior when registering with different anonymous sessions ensures existing user and wallet linkage remain unchanged.

### DIFF
--- a/apps/api/src/user/services/user/user.service.spec.ts
+++ b/apps/api/src/user/services/user/user.service.spec.ts
@@ -237,7 +237,7 @@ describe(UserService.name, () => {
       expect(walletAfter?.userId).toBe(existingUser.id);
     });
 
-    it("does not throw error when user was already created under different anonymous user", async () => {
+    it("returns existing user (unique by userId) when user is registered under different anonymous user", async () => {
       const { service } = setup();
 
       const anonymousUserSeed = {
@@ -245,6 +245,7 @@ describe(UserService.name, () => {
         subscribedToNewsletter: false
       };
       const anonymousUser = await container.resolve(UserRepository).create(anonymousUserSeed);
+      const userWallet = await container.resolve(UserWalletRepository).create({ userId: anonymousUser.id });
 
       const input: RegisterUserInput = {
         userId: faker.string.uuid(),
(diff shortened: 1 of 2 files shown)


apps/api/src/user/services/user/user.service.ts
```diff
@@ -46,6 +46,9 @@ export class UserService {
     };
 
     if (data.anonymousUserId) {
+      const existingUser = await this.userRepository.findOneBy({ userId: userDetails.userId });
+      assert(!existingUser || existingUser.id === data.anonymousUserId, 403);
```
stalniy: this is actually a valid case. It works like this due to automatic anonymous user creation. Imagine, you visit akash for the first time, created/signed up user, did you stuff and the logged out. on logout, system creates a new anonymous user for you (even if not it will create the next time you open akash console).

Then you are trying to sign up and system passes new anonymousUserId (not the one with which you signed up). As a result it tries to update new anonymous user with details from auth0 and it fails on auth0 userId key because we already have such user in a system.

So, what you should is to catch the error after `this.userRepository.updateBy`, checks whether it's unique violation error and whether it's on user_id constraint. If so, then just ignore it and return `null` in `catch` handler

stalniy (changes requested):

apps/api/src/user/services/user/user.service.spec.ts
```diff
@@ -237,6 +237,32 @@ describe(UserService.name, () => {
...
+        emailVerified: faker.datatype.boolean(),
+        subscribedToNewsletter: faker.datatype.boolean(),
+        ip: faker.internet.ipv4(),
+        userAgent: faker.string.alphanumeric(32),
+        fingerprint: faker.string.alphanumeric(16)
+      };
+      await service.registerUser(input);
+
+      const otherAnonymousUser = await container.resolve(UserRepository).create(anonymousUserSeed);
+      await service.registerUser({ ...input, anonymousUserId: otherAnonymousUser.id });
```
stalniy: **suggestion:** it's also good to check that:
1. resulting `user.id` does not equal to `otherAnonymousUser.id`
2. that wallet of existing user has not been overwritten to the wallet of `otherAnonymousUser.id` (in case he started trial). basically that due to this operation wallet is not lost

apps/api/src/user/services/user/user.service.ts
```diff
@@ -46,16 +46,21 @@ export class UserService {
...
+          },
+          userDetails,
+          { returning: true }
+        );
+        isAnonymous = !!user;
+      } catch (error) {
+        if (!isUniqueViolation(error) || !error.constraint_name?.includes("userSetting_userId_unique")) {
+          throw error;
+        }
+      }
```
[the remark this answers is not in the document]
stalniy: the name is correct. the thing is that userSetttings table has been created by sequelize and then migrate to drizzle

stalniy (changes requested):

stalniy (approved):

Issue #1947, opened 2025-09-23
Deployment list and details in offline mode on deploy-web


stalniy:
## What

- [x] expose API to the frontend to fetch deployment list when blockchain node is down (take it from indexer)
- [x] expose API to fetch deployment details for dseq (take it from indexer)
- [x] expose API to fetch leases from indexer
- [x] integrate into deploy-web (depends on #1950)

## Additional

the data structure is here -> https://github.com/akash-network/console/blob/main/apps/deploy-web/src/utils/deploymentDetailUtils.ts#L25

Consider to mimic url and data structure of blockchain node to reduce code changes.

PR #1958, opened 2025-09-24
feat(deployment): add fallback deployment list endpoint

baktun14:
https://github.com/akash-network/console/issues/1947

## Summary by CodeRabbit

* **New Features**
  * Added a database-backed fallback endpoint to list deployments with filtering (owner, state) and pagination (offset/limit, key-based, reverse). Responses include deployment details, groups, and escrow information.
* **Improvements**
  * Standardized request/response validation for the fallback deployments list.
  * Resource storage in responses now supports any number of entries (was fixed-size).
* **Tests**
  * Added functional tests covering response structure, filtering, and pagination (offset, key-based, reverse, totals).

### DIFF
--- a/apps/api/src/deployment/http-schemas/deployment-rpc.schema.ts
+++ b/apps/api/src/deployment/http-schemas/deployment-rpc.schema.ts
@@ -1,6 +1,5 @@
 import { z } from "zod";
 
-// Fallback API schemas for database deployment list
 export const FallbackDeploymentListQuerySchema = z.object({
   "filters.owner": z.string().optional(),
   "filters.state": z.enum(["active", "closed"]).optional(),
--- a/apps/api/src/deployment/http-schemas/deployment-rpc.schema.ts
+++ b/apps/api/src/deployment/http-schemas/deployment-rpc.schema.ts
@@ -1,0 +1,152 @@
+import { z } from "zod";
+
+// Fallback API schemas for database deployment list
+export const FallbackDeploymentListQuerySchema = z.object({
+  "filters.owner": z.string().optional(),
+  "filters.state": z.enum(["active", "closed"]).optional(),
+  "pagination.offset": z.coerce.number().optional(),
+  "pagination.limit": z.coerce.number().optional(),
+  "pagination.key": z.string().optional(),
+  "pagination.count_total": z.coerce.boolean().optional(),
+  "pagination.reverse": z.coerce.boolean().optional()
+});
+
+const DeploymentIdSchema = z.object({
+  owner: z.string(),
+  dseq: z.string()
+});
+
+const GroupIdSchema = z.object({
+  owner: z.string(),
+  dseq: z.string(),
+  gseq: z.number()
+});
+
+const SignedBySchema = z.object({
+  all_of: z.array(z.string()),
+  any_of: z.array(z.string())
+});
+
+const AttributeSchema = z.object({
+  key: z.string(),
+  value: z.string()
+});
+
+const RequirementsSchema = z.object({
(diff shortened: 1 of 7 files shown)


apps/api/src/deployment/services/deployment-reader/database-deployment-reader.service.ts
```diff
@@ -0,0 +1,180 @@
...
+  state?: "active" | "closed";
+  skip?: number;
+  limit?: number;
+  key?: string;
+  countTotal?: boolean;
+  reverse?: boolean;
+}
+
+// For the values that are not stored in the database
+export const FALLBACK_VALUE = "FALLBACK";
```
stalniy: **thought(non-blocking):** maybe we should include something more self descriptive. Maybe "<unknown value>"

stalniy (dismissed):

apps/api/src/deployment/http-schemas/deployment-rpc.schema.ts
```diff
@@ -0,0 +1,152 @@
+import { z } from "zod";
+
+// Fallback API schemas for database deployment list
```
ygrishajev: non-blocking. the comment duplicates the name of the var ;)

apps/api/src/deployment/routes/deployments/deployments.router.ts
```diff
@@ -214,6 +216,27 @@ const getByOwnerAndDseqRoute = createRoute({
   }
 });
 
+// Fallback route that matches the node API signature
```
ygrishajev: non-blocking. this comment can be removed and the summary can be extended

apps/api/src/deployment/services/deployment-reader/db-deployment-reader.service.ts
```diff
@@ -0,0 +1,180 @@
...
+      whereClause.closedHeight = null;
+    } else if (state === "closed") {
+      whereClause.closedHeight = { [Op.ne]: null };
+    }
+
+    // Handle pagination similar to HTTP service
+    // If key is provided, use it as offset (simplified key-based pagination)
+    const offset = key ? parseInt(key, 10) || 0 : skip;
+
+    const queryOptions: Record<string, unknown> = {
```
ygrishajev, line 39: this method contains quite some repo logic. there is even a repo for deployments https://github.com/akash-network/console/blob/main/apps/api/src/deployment/repositories/deployment/deployment.repository.ts

perhaps it's non-blocking, however i don't think not following common patterns would have saved any time

apps/api/test/functional/deployment-fallback.spec.ts
```diff
@@ -0,0 +1,316 @@
...
+
+import { createAkashBlock, createDay, createDeployment, createDeploymentGroup, createDeploymentGroupResource, createTransaction } from "@test/seeders";
+import { createAkashAddress } from "@test/seeders/akash-address.seeder";
+
+describe("Deployment Fallback API", () => {
+  const now = new Date();
+  const date = now.toISOString().split("T")[0];
+
+  afterAll(async () => {
+    await closeConnections();
```
ygrishajev: this is not needed https://github.com/akash-network/console/blob/main/apps/api/test/setup-functional-tests.ts#L39

apps/api/src/deployment/services/deployment-reader/db-deployment-reader.service.ts
baktun14: I didn't want to mix the code replicating the rpc node over the place, but if you think this belongs in the repository I'll move it in my next branch for the deployment detail.

PR #1968, opened 2025-09-25
feat: control maintenance banner with a feature flag

### LINKED ISSUE #1949: Feature flag to turn on/off warning about future blockchain upgrade
## What

Add FF flag (or add a simple endpoint) to warn users about upcoming blockchain upgrade (unavailability). Currently it's done via an env variable in deploy-web -> https://github.com/akash-network/console/blob/main/apps/deploy-web/env/.env.production#L27

**AC**:
- [ ] possible to enable this banner dynamically, without redeployment
- [ ] possible to change message and date without redeployment

jzsfkzm:
closes #1949

## Summary by CodeRabbit

- New Features
  - Maintenance banner is now driven by a runtime feature flag; message, date, and enabled state are served dynamically without redeploys.
  - Banner visibility and initial open state reflect the current flag and its payload.

- Chores
  - Removed legacy environment-based maintenance banner variables and updated environment samples and config schema accordingly.


stalniy:
@jzsfkzm we don't want to only enable or disable it, we also want to change the message without redeployment. For example, we will need to change date. Sorry this was not specified anywhere in the task but maintanance related variables should be controlled dynamically. 

FFs in unleash has variants, later this variants can be retried on client side. I thought to use variants to store message and date

stalniy:
Awesome!

stalniy (approved):

Issue #1967, opened 2025-09-25
Refresh balance and session when addings funds


baktun14:
## Why 
Currently when a user adds credits with a credit card, there's a delay between the transaction initiated and when the balance is updated in the webhook. Also, if it's the first transaction after registering in the free trial, the user has to refresh the browser in order to have the balance updated and the free status updated.

https://github.com/akash-network/console/blob/main/apps/deploy-web/src/pages/payment.tsx#L88
https://github.com/akash-network/console/blob/main/apps/api/src/billing/services/stripe-webhook/stripe-webhook.service.ts#L41

## What
Poll for the balance to be updated in the frontend until it's updated and call `await checkSession();` to update the user session for the free trial status update

baktun14:
cc @anilmurty

PR #1975, opened 2025-09-26
feat(billing): implement polling for balance update after payment

baktun14:
https://github.com/akash-network/console/issues/1967

## Summary by CodeRabbit

- **New Features**
  - Adds a payment polling provider to verify credited funds after payment; payment page treats polling as part of processing and shows loading/timeout/success notifications.

- **Tests**
  - Comprehensive tests for polling lifecycle, timeouts, success/error flows, analytics, and cleanup.

- **Chores**
  - Added test seed helpers and a new analytics event for trial completion.

- **Quality**
  - Centralized cache key usage and surfaced an additional wallet-fetching status for consistency.


apps/deploy-web/src/context/PaymentPollingProvider/PaymentPollingProvider.tsx
```diff
@@ -0,0 +1,189 @@
...
+      startTimeRef.current = Date.now();
+
+      const loadingSnackbarKey = enqueueSnackbar(<d.Snackbar title="Processing payment..." subTitle="Please wait while we update your balance" showLoading />, {
+        variant: "info",
+        autoHideDuration: null,
+        persist: true
+      });
+      loadingSnackbarKeyRef.current = loadingSnackbarKey;
+
+      pollingRef.current = setInterval(async () => {
```
stalniy: **suggeation(blocking):** `setInterval` is bad for polling. The issue is that `setInterval` doesn't wait for internal function finish. if `refetchBalaance` & `refetchManagedWallet`, take more time then POLLING_INTERVAL_MS, then it will call them again.

Better to use recursive `setTimeout`. This approach may take more time for 1 polling attempt (which we don't care much about) but it ensures that another attempt won't happen until the current one is finished

apps/deploy-web/tests/seeders/analytics.ts
```diff
@@ -0,0 +1,10 @@
+import type { AnalyticsService as RealAnalyticsService } from "@src/services/analytics/analytics.service";
+
+export interface AnalyticsService {
+  track: jest.MockedFunction<RealAnalyticsService["track"]>;
+}
+
+export const buildAnalyticsService = (overrides: Partial<AnalyticsService> = {}): AnalyticsService => ({
+  track: jest.fn() as jest.MockedFunction<RealAnalyticsService["track"]>,
+  ...overrides
+});
```
stalniy: **suggestion(non-blocking):** you don't need this, just use `mock<RealAnalyticsService>()` from `jest-mock-extended`

apps/deploy-web/tests/seeders/snackbar.ts
```diff
@@ -0,0 +1,23 @@
...
+export const buildSnackbarService = (overrides: Partial<SnackbarService> = {}): SnackbarService => ({
+  enqueueSnackbar: jest.fn() as jest.MockedFunction<
+    (
+      message: ReactNode,
+      options?: { variant?: VariantType; autoHideDuration?: number; persist?: boolean; action?: (key: SnackbarKey) => ReactNode }
+    ) => SnackbarKey
+  >,
+  closeSnackbar: jest.fn() as jest.MockedFunction<(key?: SnackbarKey) => void>,
+  ...overrides
+});
```
stalniy: **suggestion(non-blocking):** can be replaced with 

```ts
import { ProviderContext as SnackbarService } from "notistack";

const snackbarService = mock<SnackbarService>();
```

apps/deploy-web/src/context/PaymentPollingProvider/PaymentPollingProvider.tsx
```diff
@@ -0,0 +1,189 @@
...
+        const elapsed = Date.now() - (startTimeRef.current || 0);
+
+        if (elapsed >= MAX_POLLING_DURATION_MS) {
```
stalniy: **thought(non-blocking):** I assume it's possible to count only attempts which is a bit simple. `MAX_ATTEMPTS = MAX_POLLING_DURATION_MS / POLLING_INTERVAL_MS` and then just `attempt++`

stalniy (changes requested):

apps/deploy-web/src/context/PaymentPollingProvider/PaymentPollingProvider.tsx
```diff
@@ -0,0 +1,189 @@
...
+}
+
+const PaymentPollingContext = createContext<PaymentPollingContextType | null>(null);
+
+export interface PaymentPollingProviderProps {
+  children: React.ReactNode;
+  dependencies?: typeof DEPENDENCIES;
+}
+
+export const PaymentPollingProvider: React.FC<PaymentPollingProviderProps> = ({ children, dependencies: d = DEPENDENCIES }) => {
```
stalniy, line 46: **thought(non-blocking):** I think we need to stop adding providers that cause re-rendering. Any change in state of this provider causes re-rendering of the whole application. Probably we need to mostly use providers for passing down values that do not change (e.g., config, services) or values that change rarely and do the rest via jotai store which allows to re-render only atom affected places, not the whole app

I will prepare a doc in Discussions

stalniy (dismissed):

Issue #1917, opened 2025-09-16
Integrate ChainSDK library into indexer


anilmurty:
- Update with new types with the chain-sdk
- https://github.com/akash-network/AEP/blob/main/spec/aep-75/README.md
- Transaction event attribute key/values are not encoded anymore https://github.com/akash-network/console/blob/19a51805fcb17dee3dd925c2c2e7d79b67bdb17e/apps/indexer/src/chain/chainSync.ts#L247-L248

PR #1983, opened 2025-09-30
feat(analytics): chain sdk next indexer

baktun14:
https://github.com/akash-network/console/issues/1917


apps/indexer/src/indexers/validatorIndexer.ts
```diff
@@ -169,4 +173,42 @@ export class ValidatorIndexer extends Indexer {
...
+  }
+
+  /**
+   * Converts commission rate from Cosmos SDK format to decimal format
+   * Handles both integer format (18 decimal places) and decimal format
+   */
+  private convertCommissionRate(rate: string): number {
+    const parsedRate = parseFloat(rate);
+    return parsedRate > 1 ? parsedRate / 1e18 : parsedRate;
+  }
```
stalniy, line 213: **issue(non-blocking):** here we will loose precision and will have values like `61.000000000000000001`. Money should be either represented as string or as a BigInt in minor units

packages/database/chainDefinitions.ts
```diff
@@ -85,10 +85,10 @@ export const chainDefinitions: { [key: string]: ChainDef } = {
   },
   akashTestnet: {
     code: "akash-testnet",
-    rpcNodes: ["https://rpc.testnet-02.aksh.pw:443", "https://akash-testnet-rpc.cosmonautstakes.com:443"],
+    rpcNodes: ["https://testnetrpc.akashnet.net:443"],
```
stalniy: **question(non-blocking):** does it make sense to use @akashnetwork/net package here to get  rpc node for testnet?

stalniy: at least for now we could have less amount of hardcoded values

apps/indexer/package.json
```diff
@@ -28,6 +28,7 @@
   },
   "dependencies": {
     "@akashnetwork/akash-api": "^1.3.0",
+    "@akashnetwork/chain-sdk": "^1.0.0-alpha.0",
```
stalniy: **suggestion(blocking):** need to update the version, just to ensure it's not forgotten

apps/indexer/src/indexers/validatorIndexer.ts
baktun14: Yea true, we handle it better in other places

packages/database/chainDefinitions.ts
baktun14: Sure

PR #1991, opened 2025-10-02
feat(analytics): add amplitude to notifications for emails

baktun14:
## Summary by CodeRabbit

- **New Features**
  - Added Amplitude analytics for notifications; service reports "email_sent" and "email_failed". Email sender now tracks "email_sent" and logs analytics errors. Deterministic, configurable per-user sampling with stricter validation.

- **Tests**
  - New/updated unit tests for sampling, event reporting, email-sender analytics integration, and analytics error handling.

- **Chores**
  - Added analytics and hashing dependencies and new environment variables for analytics configuration.


apps/notifications/src/modules/notifications/services/email-sender/email-sender.service.ts
```diff
@@ -17,30 +18,47 @@ type EmailSendOptions = {
...
-    });
+      });
+
+      this.analyticsService.track(userId, "email_sent", {
+        recipient_count: addresses.length,
+        subject,
+        workflow_id: this.configService.getOrThrow("notifications.NOVU_MAILER_WORKFLOW_ID")
+      });
+    } catch (error) {
+      this.analyticsService.track(userId, "email_failed", {
```
stalniy: **thought(non-blocking):** not sure whether it makes sense to log this error here because if it fails, pg-boss will retry it. We should receive alert (but first we need to configure pg-boss observability on notifications). Every event to amplitude costs us money, and if we send this on retries may not be a good use of amplitude quota

stalniy (dismissed):

apps/notifications/src/modules/notifications/services/email-sender/email-sender.service.ts
stalniy: @coderabbitai thoughts?

baktun14: Ok I will remove it

PR #1994, opened 2025-10-02
feat: can multi-select deployments when pressing shift

### LINKED ISSUE #1982: Add multi select for deployment lists
When there is a need to close multiple sequential deployments user has to click every row's checkbox. It would help if there was a possibility to select items between the selected one and another item in the list when SHIFT is hold.

jzsfkzm:
closes #1982

## Summary by CodeRabbit

* **New Features**
  * Added multi-select capability with Shift+click range selection support for deployments.

* **Refactor**
  * Improved selection state management to provide more consistent and efficient multi-item selection handling across the deployments interface.

### DIFF
--- a/apps/deploy-web/src/hooks/useListSelection/useListSelection.spec.ts
+++ b/apps/deploy-web/src/hooks/useListSelection/useListSelection.spec.ts
@@ -7,24 +7,48 @@ import { act, renderHook, waitFor } from "@testing-library/react";
 
 const expectSelectedItems = (actualIds: number[], expectedIds: number[]) => {
   expect(actualIds.length).toBe(expectedIds.length);
-  expect(expectedIds.every(id => actualIds.includes(id))).toBe(true);
+  expect(actualIds).toEqual(expect.arrayContaining(expectedIds));
+};
+
+type Click = {
+  id: number;
+  isShiftPressed: boolean;
+};
+
+type TestConfig = {
+  clicks: Click[];
+  expectedItems: number[];
+};
+
+const testSelection = async (config: TestConfig) => {
+  const result = setup();
+
+  config.clicks.forEach(({ id, isShiftPressed }) => {
+    act(() => {
+      result.current.selectItem({ id, isShiftPressed });
+    });
+  });
+
+  await waitFor(() => {
+    expectSelectedItems(result.current.selectedItemIds, config.expectedItems);
+  });
 };
 
 describe(useListSelection.name, () => {
   it("should not explode for an empty list", () => {
-    const hook = setup({ ids: [] });
-
-    expect(hook.selectedItemIds).toEqual([]);
+    const result = setup({ ids: [] });
+    expect(result.current.selectedItemIds).toEqual([]);
   });
 
-  [
-    {
-      name: "can set a single item as selected",
+  it("can set a single item as selected", async () => {
+    await testSelection({
       clicks: [{ id: 2, isShiftPressed: false }],
       expectedItems: [2]
-    },
-    {
-      name: "can select multiple items, toggling back and forth",
+    });
+  });
+
+  it("can select multiple items, toggling back and forth", async () => {
+    await testSelection({
       clicks: [
         { id: 2, isShiftPressed: false },
         { id: 3, isShiftPressed: false },
@@ -32,66 +56,82 @@ describe(useListSelection.name, () => {
         { id: 3, isShiftPressed: false }
       ],
       expectedItems: [2, 4]
-    },
-    {
-      name: "can shift-select first item",
+    });
+  });
+
+  it("can shift-select first item", async () => {
+    await testSelection({
       clicks: [{ id: 2, isShiftPressed: true }],
       expectedItems: [2]
-    },
-    {
-      name: "can shift-select items down",
+    });
+  });
+
+  it("can shift-select items down", async () => {
+    await testSelection({
       clicks: [
         { id: 2, isShiftPressed: false },
         { id: 4, isShiftPressed: true }
       ],
       expectedItems: [2, 3, 4]
-    },
-    {
-      name: "can shift-select items up",
+    });
+  });
+
+  it("can shift-select items up", async () => {
+    await testSelection({
       clicks: [
         { id: 4, isShiftPressed: false },
         { id: 2, isShiftPressed: true }
       ],
       expectedItems: [2, 3, 4]
-    },
-    {
-      name: "can shift-select down and then up",
+    });
+  });
+
+  it("can shift-select down and then up", async () => {
+    await testSelection({
       clicks: [
         { id: 2, isShiftPressed: false },
         { id: 4, isShiftPressed: true },
         { id: 0, isShiftPressed: true }
       ],
       expectedItems: [0, 1, 2, 3, 4]
-    },
-    {
-      name: "can shift-select up and then down",
+    });
+  });
+
+  it("can shift-select up and then down", async () => {
+    await testSelection({
       clicks: [
         { id: 2, isShiftPressed: false },
         { id: 0, isShiftPressed: true },
         { id: 4, isShiftPressed: true }
       ],
       expectedItems: [0, 1, 2, 3, 4]
-    },
-    {
-      name: "can deselect the whole range up",
+    });
+  });
+
+  it("can deselect the whole range up", async () => {
+    await testSelection({
       clicks: [
         { id: 2, isShiftPressed: false },
         { id: 4, isShiftPressed: true },
         { id: 2, isShiftPressed: true }
       ],
       expectedItems: []
-    },
-    {
-      name: "can deselect the whole range down",
+    });
+  });
+
+  it("can deselect the whole range down", async () => {
+    await testSelection({
       clicks: [
         { id: 4, isShiftPressed: false },
         { id: 2, isShiftPressed: true },
         { id: 4, isShiftPressed: true }
       ],
       expectedItems: []
-    },
-    {
-      name: "can mark even more items down",
+    });
+  });
+
+  it("can mark even more items down", async () => {
+    await testSelection({
       clicks: [
         { id: 2, isShiftPressed: false },
         { id: 4, isShiftPressed: true },
@@ -99,9 +139,11 @@ describe(useListSelection.name, () => {
         { id: 6, isShiftPressed: true }
       ],
       expectedItems: [0, 1, 2, 3, 4, 5, 6]
-    },
-    {
-      name: "can mark even more items up",
+    });
+  });
+
+  it("can mark even more items up", async () => {
+    await testSelection({
       clicks: [
         { id: 4, isShiftPressed: false },
         { id: 6, isShiftPressed: true },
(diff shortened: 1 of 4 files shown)


ygrishajev:
hey @jzsfkzm I've tested this and the logic feels a bit weird when it comes to some edge cases. I've asked ChatGPT to get me some common flow instructions below to make it more clear. wdut?
@coderabbitai feel free to post your feedback to the below.


# Row Multiselect (Shift) — Logic

## Overview
When **Shift** is pressed, clicking a row selects a **continuous range** between the **first selected row** (the _anchor_) and the **currently clicked row**.

This lets users select multiple **consecutive rows** in one action.

---

## Behavior

1. **First Shift-click**
   - If no selection exists, select the clicked row.
   - That row becomes the **anchor**.

2. **Next Shift-click**
   - Select all rows between the **anchor** and the clicked row (inclusive).
   - Forms a **continuous block**.

3. **Further Shift-clicks**
   - The **anchor stays the same**.
   - Each new click redefines the range between the anchor and clicked row.
   - The previous range is replaced.

4. **Normal Click (no Shift)**
   - Clears any existing selection.
   - Selects only the clicked row.
   - Sets a new **anchor**.

---

## Examples

### Example 1 — Simple Downward Selection
| Action | Result |
|:--|:--|
| Click Row 2 | Selects Row 2 (anchor = 2) |
| Shift + Click Row 5 | Selects Rows 2–5 |

**Outcome:** A continuous block from 2 → 5.

---

### Example 2 — Reversing Direction
| Action | Result |
|:--|:--|
| Click Row 5 | Selects Row 5 (anchor = 5) |
| Shift + Click Row 2 | Selects Rows 2–5 |

**Outcome:** Works the same upward — direction doesn’t matter.

---

### Example 3 — Expanding, Then Contracting
| Action | Result |
|:--|:--|
| Click Row 3 | Selects Row 3 (anchor = 3) |
| Shift + Click Row 7 | Selects Rows 3–7 |
| Shift + Click Row 5 | Updates to Rows 3–5 |

**Outcome:** The range shrinks or grows depending on the new target, but always remains continuous from the same anchor.

---

### Example 4 — New Anchor After Normal Click
| Action | Result |
|:--|:--|
| Click Row 4 | Selects Row 4 (anchor = 4) |
| Shift + Click Row 8 | Selects Rows 4–8 |
| Click Row 2 | Selection resets (anchor = 2) |
| Shift + Click Row 5 | Selects Rows 2–5 |

**Outcome:** A plain click resets both selection and anchor.

---

### Example 5 — Jumping Around
| Action | Result |
|:--|:--|
| Click Row 3 | Selects Row 3 (anchor = 3) |
| Shift + Click Row 6 | Selects Rows 3–6 |
| Shift + Click Row 1 | Selects Rows 1–3 |

**Outcome:** Each Shift-click recalculates the range from the same anchor (Row 3).

---

### Example 6 — First Click is Shifted
| Action | Result |
|:--|:--|
| Shift + Click Row 5 (no anchor yet) | Selects Row 5 only (anchor = 5) |
| Shift + Click Row 8 | Selects Rows 5–8 |

**Outcome:** If Shift-click happens first, it behaves as a normal click for the initial anchor.

---

## Key Points
- **Shift** means “select everything between the anchor and this row.”  
- The **anchor** stays fixed until a normal click occurs.  
- Each Shift-click **replaces**, not extends, the previous selection range.  
- Works equally well when selecting upward or downward.

ygrishajev:
@coderabbitai what do you think of the spec provided generally? Would you consider this to be a common approach?

jzsfkzm:
> hey @jzsfkzm I've tested this and the logic feels a bit weird when it comes to some edge cases. I've asked ChatGPT to get me some common flow instructions below to make it more clear. wdut? @coderabbitai feel free to post your feedback to the below.
> 

@ygrishajev I see why you said it's weird. Re-selecting the same interval, or an overlapping interval indeed was behaving in unexpected ways. I changed it a bit, now I think it's better, will you check too?
Thanks for these examples, I also clicked around in Gmail as well as in Docker Desktop's lists. I think those can be relevant for our case, since there are checkboxes on these UIs - unlike in Finder.

apps/deploy-web/src/hooks/useListSelection/useListSelection.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+      name: "can mark even more items up",
+      clicks: [
+        { id: 4, isShiftPressed: false },
+        { id: 6, isShiftPressed: true },
+        { id: 2, isShiftPressed: true },
+        { id: 0, isShiftPressed: true }
+      ],
+      expectedItems: [0, 1, 2, 3, 4, 5, 6]
+    }
+  ].forEach(({ name, clicks, expectedItems }) => {
```
ygrishajev: this is an awesome test! 💪 🔥

apps/deploy-web/src/hooks/useListSelection/useListSelection.ts
```diff
@@ -0,0 +1,91 @@
...
+        setSelectedItemIds(prev => prev.filter(x => !lastIntervalSelectionIds.includes(x)));
+        setLastIntervalSelectionIds([]);
+        return;
+      }
+
+      setSelectedItemIds(prev => uniq([...prev, ...newRange]));
+      setLastIntervalSelectionIds(newRange);
+    },
+    [intervalSelectionAnchor, itemsBetween, lastIntervalSelectionIds]
+  );
```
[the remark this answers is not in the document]
jzsfkzm: Shift-click being additive is the requirement, covered by tests.

ygrishajev (dismissed):

apps/deploy-web/src/hooks/useListSelection/useListSelection.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+      expectedItems: [2, 3, 4]
+    },
+    {
+      name: "can shift-select down and then up",
+      clicks: [
+        { id: 2, isShiftPressed: false },
+        { id: 4, isShiftPressed: true },
+        { id: 0, isShiftPressed: true }
+      ],
+      expectedItems: [0, 1, 2, 3, 4]
```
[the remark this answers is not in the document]
jzsfkzm: Additive behavior is the requirement.

apps/deploy-web/src/hooks/useListSelection/useListSelection.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+
+import type { UseListSelectionProps } from "./useListSelection";
+import { useListSelection } from "./useListSelection";
+
+import { act, renderHook, waitFor } from "@testing-library/react";
+
+const expectSelectedItems = (actualIds: number[], expectedIds: number[]) => {
+  expect(actualIds.length).toBe(expectedIds.length);
+  expect(expectedIds.every(id => actualIds.includes(id))).toBe(true);
+};
```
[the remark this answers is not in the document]
jzsfkzm: Prior length check is proving we have the same items in both arrays.

apps/deploy-web/src/hooks/useListSelection/useListSelection.spec.ts
```diff
@@ -0,0 +1,134 @@
+import { range } from "lodash";
+
+import type { UseListSelectionProps } from "./useListSelection";
+import { useListSelection } from "./useListSelection";
+
+import { act, renderHook, waitFor } from "@testing-library/react";
+
+const expectSelectedItems = (actualIds: number[], expectedIds: number[]) => {
+  expect(actualIds.length).toBe(expectedIds.length);
+  expect(expectedIds.every(id => actualIds.includes(id))).toBe(true);
```
stalniy: **nitpick:** you can use `expect(actualIds).toEqual(expect.arrayContaining(expectedIds))`

apps/deploy-web/src/hooks/useListSelection/useListSelection.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+        { id: 4, isShiftPressed: false },
+        { id: 6, isShiftPressed: true },
+        { id: 2, isShiftPressed: true },
+        { id: 0, isShiftPressed: true }
+      ],
+      expectedItems: [0, 1, 2, 3, 4, 5, 6]
+    }
+  ].forEach(({ name, clicks, expectedItems }) => {
+    it(name, async () => {
+      const { result } = renderHook(() => useListSelection({ ids: range(0, 10) }));
```
stalniy: **suggestion(non-blocking):** seems like it can be replaced with `setup({ ids: range(0, 10) })`

apps/deploy-web/src/hooks/useListSelection/useListSelection.spec.ts
ygrishajev: this assumes array may contain other ids, right?

apps/deploy-web/src/hooks/useListSelection/useListSelection.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  const onSelectItem = useCallback(
```
stalniy: **thought(non-blocking):** this should be `selectItem` based on this doc https://github.com/akash-network/console/discussions/1727 . @ygrishajev  am I right?

apps/deploy-web/src/hooks/useListSelection/useListSelection.spec.ts
stalniy: **non-blocking:** sorry guys, will add a bit of skepticism here:

To be honest I don't see much difference between:

```ts
{
      name: "can mark even more items up",
      clicks: [
        { id: 4, isShiftPressed: false },
        { id: 6, isShiftPressed: true },
        { id: 2, isShiftPressed: true },
        { id: 0, isShiftPressed: true }
      ],
      expectedItems: [0, 1, 2, 3, 4, 5, 6]
    }
```

and 

```ts
it("can mark even more items up", async () => {
   const hook = setup({ ids: range(0, 10) });

   [
        { id: 4, isShiftPressed: false },
        { id: 6, isShiftPressed: true },
        { id: 2, isShiftPressed: true },
        { id: 0, isShiftPressed: true }
    ].forEach(({ id, isShiftPressed }) => act(() => {
       result.current.onSelectItem({ id, isShiftPressed });
    });

    expectSelectedItems(hook.selectedItemIds, [0, 1, 2, 3, 4, 5, 6])
})
```

almost the same in terms of lines of code but configuration object adds additional level of abstraction on top and makes it harder to debug (cannot use `fit`, `xit` or `console.log` for specific test, stack trace is also not useful), adjusting single test behavior also harder. 

I usually prefer adding helper function instead of using configuration objects in tests for this reasons. 

In this particular case, `clicks` could be passed into `setup` function as an optional parameter.

stalniy (dismissed):

apps/deploy-web/src/hooks/useListSelection/useListSelection.spec.ts
stalniy: yes, you still need to check the length. just don't need to use `.every` and in case of failure jest will show better message

apps/deploy-web/src/hooks/useListSelection/useListSelection.spec.ts
ygrishajev: well, although i agree with the reasoning, the readability of the config stays unbeaten imo

stalniy: but I do not insist because sometimes I also see the benefit of configuration objects but in rare cases, probably when configuration is very small. Still debugging is possible with help of jest cli options and watch mode

ygrishajev: how do you like this @stalniy - mixes together these approaches.
```typescript
it("can mark even more items up", () => {
  testSelection({
    clicks: [
      { id: 4, isShiftPressed: false },
      { id: 6, isShiftPressed: true },
      { id: 2, isShiftPressed: true },
      { id: 0, isShiftPressed: true }
    ],
    expectedItems: [0, 1, 2, 3, 4, 5, 6]
  })
})
```

jzsfkzm: Oh this would be great. Asking Claude to turn it around.

PR #2031, opened 2025-10-08
test: add e2e test proving user can choose a template

### LINKED ISSUE #2004: E2E test that templates on new deployment page are clickable
## What 

ensure that on new deployment page all templates are clickable and corresponding SDL is used when template is clicked

Image

jzsfkzm:
closes #2004

## Summary by CodeRabbit

* **Accessibility**
  * Added a screen-reader label (aria-label="Template list") to the template list in the new deployment flow for improved navigation with assistive technologies.

* **Tests**
  * Updated the end-to-end UI test to find the labeled template list and iterate over every template (opening each in a new page) to confirm navigation to the deployment edit step for each template.

### DIFF
--- a/apps/deploy-web/src/components/new-deployment/TemplateList.tsx
+++ b/apps/deploy-web/src/components/new-deployment/TemplateList.tsx
@@ -162,7 +162,7 @@ export const TemplateList: React.FunctionComponent<Props> = ({ onChangeGitProvid
               </Link>
             </div>
 
-            <div className="grid grid-cols-1 gap-2 md:grid-cols-2 md:gap-4 lg:grid-cols-3">
+            <div className="grid grid-cols-1 gap-2 md:grid-cols-2 md:gap-4 lg:grid-cols-3" aria-label="Template list">
               {previewTemplates.map(template => (
                 <TemplateBox
                   key={template.id}
--- a/apps/deploy-web/src/components/new-deployment/TemplateList.tsx
+++ b/apps/deploy-web/src/components/new-deployment/TemplateList.tsx
@@ -162,7 +162,7 @@ export const TemplateList: React.FunctionComponent<Props> = ({ onChangeGitProvid
               </Link>
             </div>
 
-            <div className="grid grid-cols-1 gap-2 md:grid-cols-2 md:gap-4 lg:grid-cols-3" aria-label="Template list">
+            <section className="grid grid-cols-1 gap-2 md:grid-cols-2 md:gap-4 lg:grid-cols-3" aria-label="Template list">
               {previewTemplates.map(template => (
                 <TemplateBox
                   key={template.id}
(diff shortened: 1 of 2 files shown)


apps/deploy-web/src/components/new-deployment/TemplateList.tsx
```diff
@@ -162,7 +162,7 @@ export const TemplateList: React.FunctionComponent<Props> = ({ onChangeGitProvid
               </Link>
             </div>
 
-            <div className="grid grid-cols-1 gap-2 md:grid-cols-2 md:gap-4 lg:grid-cols-3">
+            <div className="grid grid-cols-1 gap-2 md:grid-cols-2 md:gap-4 lg:grid-cols-3" aria-label="Template list">
```
stalniy: replace the `<div>` with `<section>` otherwise semantic markup requires us to add `role=region`

apps/deploy-web/tests/ui/deploy-from-a-template.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+  const templateListPage = new DeployBasePage(context, page, "new-deployment");
+  await templateListPage.goto();
+
+  const templateList = page.locator('[aria-label="Template list"]');
+
+  await expect(templateList).toBeVisible();
+
+  const templateLinks = templateList.locator("> a");
+  expect(await templateLinks.count()).toBeGreaterThan(0);
+  await templateLinks.nth(random(0, (await templateLinks.count()) - 1)).click();
```
stalniy: **blocking**: adding randomness to the test is not good because then it's not reliable -> flaky. As far as I remember we had an issue that some of them where not clickable. So, we need to check that all can be clicked, anyway it should be pretty fast.

apps/deploy-web/tests/ui/deploy-from-a-template.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+
+test("user can choose a template from the templates page", async ({ page, context }) => {
+  const templateListPage = new DeployBasePage(context, page, "new-deployment");
+  await templateListPage.goto();
+
+  const templateList = page.locator('[aria-label="Template list"]');
+
+  await expect(templateList).toBeVisible();
+
+  const templateLinks = templateList.locator("> a");
```
stalniy: you can use `[role=link]` or `getAllByRole("link")`, also don't use structural selectors because adding a wrapper to markup breaks the test

stalniy (changes requested):

apps/deploy-web/tests/ui/deploy-from-a-template.spec.ts
jzsfkzm: Indeed, I had the feeling it would be slow to click all the template links, but I was wrong. Changed it, and it is not significantly slower now.

stalniy (approved):
🚀

PR #2047, opened 2025-10-13
Refactor/upgrade chain sdk

stalniy:
## Why

To adjust to the same signing alg between web and offline signers. Blocked by https://github.com/akash-network/chain-sdk/pull/110

Requires installation of chain-sdk version from that PR


stalniy:
closes in favor of https://github.com/akash-network/console/pull/2049

Issue #1919, opened 2025-09-16
Integrate ChainSDK into Deploy-Web


anilmurty:
- Search for everywhere we hardcoded "v1beta"
- Adapt with new deployment deposit messages

PR #2050, opened 2025-10-14
feat: chain sdk next web

baktun14:
https://github.com/akash-network/console/issues/1919


apps/deploy-web/src/utils/walletUtils.ts
```diff
@@ -1,4 +1,4 @@
-import type { NetworkId } from "@akashnetwork/akashjs/build/types/network";
```
stalniy, line 1: FYI: `NetworkId` is available in chain-sdk. So, maybe just easier to import it from there rather than remapping network id. It's hard to track right now where that mapping is needed and where not

baktun14: Ah I didn't know, I'm not sure what's the best solution. In the SDL it's expecting to be "mainnet" | "sandbox" | "testnet", it's the only place we do the reverse mapping, otherwise it uses the SupportedNetworks from the config file.

ygrishajev (dismissed):

Issue #1921, opened 2025-09-16
Integrate ChainSDK into Notifications


anilmurty:
- Search for everywhere we hardcoded "v1beta"
- Update with new chain-sdk package

PR #2051, opened 2025-10-14
feat(notifications): integrates sdk53 into the service

ygrishajev:


apps/notifications/package.json
```diff
@@ -39,6 +39,7 @@
   },
   "dependencies": {
     "@akashnetwork/akash-api": "1.4.0",
+    "@akashnetwork/chain-sdk": "^1.0.0-alpha.3",
```
stalniy: **suggestion**: upgrade to  alpha.8

apps/notifications/package.json
```diff
@@ -49,7 +50,7 @@
     "@cosmjs/encoding": "^0.32.4",
     "@cosmjs/proto-signing": "^0.33.0",
     "@cosmjs/stargate": "^0.33.0",
-    "@cosmjs/tendermint-rpc": "^0.33.1",
+    "@cosmjs/tendermint-rpc": "^0.36.1",
```
stalniy, line 53: **suggestion:** in this case it makes sense to upgrade all comjs libs, they live in monorepo and as far as I can tell all has single publishing version. So, right now you will have 2 sets of cosmjs libs in single application

stalniy: we anyway  need to update them to the latest but it's hard to do with all the moving parts

stalniy (dismissed):

stalniy:
Related https://github.com/akash-network/console/issues/1921

PR #2053, opened 2025-10-14
feat(provider): ensures compatibility with both versions of sdk

ygrishajev:
## Summary by CodeRabbit

* **New Features**
  * Certificate retrieval now supports both v1 and v1beta3 endpoints with automatic fallback.
* **Reliability**
  * Added a 5s timeout to certificate requests to prevent hanging.
  * Enhanced retry behavior for transient server errors (5xx, excluding 501).
* **Bug Fixes**
  * Clarified handling of 500 responses to return null on failure after retries, reducing ambiguous outcomes.


apps/provider-proxy/src/services/ProviderService/ProviderService.ts
```diff
@@ -37,6 +36,23 @@ export class ProviderService {
...
+    const response = await httpRetry(
+      () => this.fetch(`${baseUrl}/akash/cert/${this.certApiModuleVersion}/certificates/list?${queryParams}`, { signal: AbortSignal.timeout(5_000) }),
+      {
+        retryIf: response => response.status !== 501 && response.status >= 500,
+        logger: this.logger
+      }
+    );
+
+    if (this.certApiModuleVersion === "v1" && response.status === 501) {
+      this.certApiModuleVersion = "v1beta3";
```
stalniy: This works like a cache. Let’s just do it on every request. Certificates are cached anyway. Just try v1, doesn’t work v1beta3. On the next request again try v1

baktun14 (dismissed):

PR #2055, opened 2025-10-14
feat: indexer bseq

baktun14:


apps/indexer/drizzle/0001_add_bseq_to_bid_and_lease.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE "bid" ADD COLUMN "bseq" integer DEFAULT 0 NOT NULL;--> statement-breakpoint
```
stalniy: This may cause an issue during deployment. I assume we have many bids. For every bid db will need to set default value. Very likely it will lock the whole table. Don’t know whether only on write. 

@coderabbitai what are your suggestions on this for big amount of data?

stalniy (dismissed):
👍

apps/indexer/drizzle/0001_add_bseq_to_bid_and_lease.sql
baktun14: Yea this will probably take some time to execute

stalniy (approved):

PR #2075, opened 2025-10-17
feat: add otel namespace to logging package

### LINKED ISSUE #1801: add otel namespace to logging package
## Why

Currently we copy/paste configuration for otel for logger in provider-proxy and api. We should reuse it

## What

1. Create file `@akashnetwork/logging/otel` export in package.json
2. Export from it 
    ```ts
    export function createOtelLogger(options?: LoggerOptions): LoggerService {
      return new LoggerService({ ...options, mixin: collectOtel })
    }

    function collectOtel() { /* copy from apps*/ }
    ```
3. Adds `baggage` extraction, get the version of the value of `sentry-release`
4. Refactor provider-proxy to use `createOtelLogger` and remove `src/setupOpenTelemetry.ts` file
5. Refactor console-api to use `createOtelLogger` and remove `src/open-telemetry.ts` file

ritoban23:
## Add OTEL Namespace to Logging Package

Creates a reusable OpenTelemetry logger configuration to eliminate code duplication between provider-proxy and api apps.

### Changes
- Add `@akashnetwork/logging/otel` export with `createOtelLogger` function
- Add `@opentelemetry/api` as optional peer dependency  
- Refactor provider-proxy to use `createOtelLogger`
- Refactor api to set global OTEL mixin in logging.provider.ts
- Remove `apps/provider-proxy/src/setupOpenTelemetry.ts`
- Remove `apps/api/src/open-telemetry.ts`

### Benefits
- Eliminates code duplication
- Centralized OTEL configuration
- Consistent logging across apps
- Easier maintenance and updates

Closes #1801

## Summary by CodeRabbit

* **Refactor**
  * Made OpenTelemetry initialization opt-in by removing automatic startup instrumentation and centralizing OTEL hookup in the logging path.

* **New Features**
  * Added utilities to attach OTEL span context to logs and to create OTEL-enabled loggers for improved distributed-tracing observability.
  * Added an explicit side-effect hook to enable OTEL initialization when imported.

* **Chores**
  * Exported the new OTEL utilities and declared the OpenTelemetry API as an optional peer dependency.


stalniy:
Nice, thank you for contribution!

apps/api/src/core/providers/logging.provider.ts
```diff
@@ -1,7 +1,12 @@
 import { LoggerService as LoggerServiceOriginal } from "@akashnetwork/logging";
 import { HttpLoggerIntercepter } from "@akashnetwork/logging/hono";
+import { collectOtel } from "@akashnetwork/logging/otel";
 import { container, injectable } from "tsyringe";
 
+// Set up OpenTelemetry mixin for all LoggerService instances
+// This collects trace information and adds it to log entries
+LoggerServiceOriginal.mixin = collectOtel;
```
stalniy, line 8: issue(blocking): this needs to be moved back to open-telemetry.ts file in `api` app because this app imports and uses instances of `LoggerService` in other files. So, please restore open-telemetry.ts file in `api` for now (you can just do `LoggerService.mixin = collectOtel;`). SO, there may be the case when `LoggerService` is created before `logging.provider.ts` is required. In result, those logger instances won't use otel.

And in the follow up PR, you can replace `LoggerService` instances either with injected `LoggerService` from DI or with `createOtelLogger`

stalniy (changes requested):
Small remark but overall a nice job!

packages/logging/src/otel.ts
ygrishajev, line 1: let's move this file to a more appropriate location. There's `utils` dir for instance.

ygrishajev (dismissed):

packages/logging/src/otel.ts
stalniy: It's actually a different entry point, so I think it's better to keep in root

stalniy: probably it would make sense to move functions into utils and just re-export from this entry point

stalniy (approved):

PR #2077, opened 2025-10-17
fix: checkbox can optionally have a larger clickable wrapper

### LINKED ISSUE #1981: Lists checkboxes clickable area is too small
When trying it check a deployment it's easy to miss the checkbox causing navigation to a different page. It's caused by 2 potential issues:
- checkboxes are too small
- clickable area around checkboxes is too small

Fixing this would improve UX for this action.

jzsfkzm:
useful when the wrapping element is clickable too (eg. in lists)

closes #1981

## Summary by CodeRabbit

* **New Features**
  * Checkboxes now support an optional expanded touch target to improve accessibility and ease of use on mobile and touchscreens.
  * Deployment list rows now enable the expanded touch target for selectable items, making selection easier on touch devices.

### DIFF
--- a/apps/deploy-web/src/components/deployments/DeploymentListRow.tsx
+++ b/apps/deploy-web/src/components/deployments/DeploymentListRow.tsx
@@ -294,6 +294,7 @@ export const DeploymentListRow: React.FunctionComponent<Props> = ({ deployment, 
             {isSelectable && (
               <Checkbox
                 checked={checked}
+                expandedTouchTarget={true}
                 onClick={event => {
                   event.stopPropagation();
                   onSelectDeployment?.({ id: deployment.dseq, isShiftPressed: event.shiftKey });
--- a/packages/ui/components/checkbox.tsx
+++ b/packages/ui/components/checkbox.tsx
@@ -30,7 +30,7 @@ const Checkbox = React.forwardRef<
   return (
     <div
       ref={wrapperRef}
-      className={cn("inline-flex", expandedTouchTarget && "-m-3 cursor-pointer p-3")}
+      className={cn("inline-flex", expandedTouchTarget && "-m-3 cursor-pointer p-3 hover:bg-muted-foreground/10 rounded-full")}
       onClick={expandedTouchTarget ? triggerCheckboxClick : undefined}
     >
       <CheckboxPrimitive.Root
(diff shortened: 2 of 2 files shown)


apps/deploy-web/src/components/deployments/DeploymentListRow.tsx
```diff
@@ -294,6 +294,7 @@ export const DeploymentListRow: React.FunctionComponent<Props> = ({ deployment,
             {isSelectable && (
               <Checkbox
                 checked={checked}
+                expandedTouchTarget={true}
```
ygrishajev, line 297: non-blocking 

```suggestion
                expandedTouchTarget
```

packages/ui/components/checkbox.tsx
```diff
@@ -5,31 +5,59 @@ import { Check } from "lucide-react";
...
+const Checkbox = React.forwardRef<
+  React.ElementRef<typeof CheckboxPrimitive.Root>,
+  React.ComponentPropsWithoutRef<typeof CheckboxPrimitive.Root> & CheckboxTouchTargetProps
+>(({ className, onClick, expandedTouchTarget = false, ...props }, ref) => {
+  const checkboxRef = React.useRef<HTMLButtonElement>(null);
+  const wrapperRef = React.useRef<HTMLDivElement>(null);
+
+  React.useImperativeHandle(ref, () => checkboxRef.current!);
+
+  const handleWrapperClick = (e: React.MouseEvent<HTMLDivElement>) => {
```
ygrishajev: could you call is smth like
```suggestion
  const triggerCheckboxClick = (e: React.MouseEvent<HTMLDivElement>) => {
```

just to make the code a bit more readable

apps/deploy-web/src/components/deployments/DeploymentListRow.tsx
ygrishajev: do I get it right, that this prop exist to not break some other existing UI?

jzsfkzm: Yes, we have checkboxes at several places, and I saw it as a risky change just adding this margin. Should we check all the occurrences?

ygrishajev: I think this is fine

ygrishajev (approved):

Issue #1679, opened 2025-07-14
Update @cosmjs/* libs to latest version


stalniy:
## Why

In the latest version they migrated to safer cryptographic library. 

## What

This change is available from @cosmjs/*@0.34.0

stalniy:
https://github.com/cosmos/cosmjs/releases/tag/v0.34.0

stalniy:
this is partially blocked by https://github.com/hyperweb-io/cosmos-kit/issues/576

PR #2082, opened 2025-10-18
fix: upgrades cosmjs dependencies and removes unused ones

stalniy:
## Why

#1679


apps/indexer/src/providers/statusEndpointHandlers/grpc.ts
```diff
@@ -1,19 +1,14 @@
...
 import memoize from "lodash/memoize";
-import { promisify } from "util";
 
 import { parseDecimalKubernetesString, parseSizeStr } from "@src/shared/utils/files";
-import { FakeInsecureCredentials } from "./fake-insecure-credentials";
 import type { ProviderStatusInfo } from "./types";
 
 export async function fetchProviderStatusFromGRPC(provider: Provider, timeout: number): Promise<ProviderStatusInfo> {
   const data = await queryStatus(provider.hostUri, timeout);
+  console.log(data);
```
baktun14: :scissors:

apps/indexer/src/providers/statusEndpointHandlers/grpc.ts
```diff
@@ -96,34 +91,24 @@ export async function fetchProviderStatusFromGRPC(provider: Provider, timeout: n
...
   // TODO: fetch port from chain
-  const url = hostUri.replace(":8443", ":8444").replace("https://", "dns:///");
-  const credentials = FakeInsecureCredentials.createInsecure();
-  const channel = new Channel(url, credentials, {
-    "grpc.keepalive_time_ms": minutesToMilliseconds(5),
-    "grpc.keepalive_permit_without_calls": 1
+  const url = hostUri.replace(":8443", ":8444");
+  return createProviderSDK({
+    // TODO: pass transport options
+    // "grpc.keepalive_time_ms": minutesToMilliseconds(5),
```
baktun14, line 107: question(non-blocking): is it better to just leave them there for now?

baktun14 (dismissed):

apps/indexer/src/providers/statusEndpointHandlers/grpc.ts
stalniy: I'm not sure they are even needed because this is a pure unary http2 request (request/response). There is no point to enable keep alive. Well, maybe it's a bit faster but I don't think that few ms makes sense in context of fetching providers' data.

The issue is that it used akash-api to fetch provider data but akash-api is "old" in this terms, so let's just see. I don't think it will cause some issues

stalniy: There are 2 corresponding properties in Http2SessionManager which is used by new sdk. 

Analog to `grpc.keepalive_permit_without_calls`
Image

Analog to `grpc.keepalive_time_ms`
Image

Docs says that for stream request/response it sends ping requests automatically but for unary it doesn't. In this case we send unary requests to get status from provider. We fetch it every 10 secs, so having open connection I assume is useful to reduce latency

stalniy: I'll update it in the follow up PR

stalniy: follow up PR -> https://github.com/akash-network/console/pull/2103

stalniy: exposed options in this PR -> https://github.com/akash-network/chain-sdk/pull/138

PR #2124, opened 2025-10-28
feat(billing): uses ws transport for signer implementation

ygrishajev:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved explicit cleanup of signing client connections to prevent resource leaks and better handle sequence mismatches.

* **Chores**
  * Updated RPC endpoint used in functional test environment.

* **Refactor**
  * Switched signing client to eager initialization with explicit lifecycle controls (disconnect/dispose) and moved transport to a websocket-based implementation for more reliable connections.


apps/api/src/billing/lib/batch-signing-client/batch-signing-client.service.ts
```diff
@@ -83,16 +81,20 @@ export class BatchSigningClientService {
...
   }
 
-  private async initClient() {
-    return await backOff(
+  private async getClient(): Promise<SyncSigningStargateClient> {
+    if (this.clientAsPromised) {
+      return this.clientAsPromised;
+    }
+
+    const client = await backOff(
```
stalniy: **issue(blocking):** this is race condition vulnerable

apps/api/src/billing/lib/batch-signing-client/batch-signing-client.service.ts
```diff
@@ -82,94 +73,16 @@ export class BatchSigningClientService {
...
-    }
+  async dispose() {
+    await this.disconnect();
   }
 
-  private clearCachedAccountInfo(): void {
-    this.cachedAccountInfo = undefined;
-    this.accountInfoPromise = undefined;
+  private async init() {
+    this.chainId = await this.client.getChainId();
```
stalniy, line 85: **question(blocking):** can this be done lazily? I assume it sends a remote request somewhere and if we do this without `.catch` it will crash the app when the remote service is down

PR #2137, opened 2025-10-30
refactor: splits BackgroundTaskProvider into hook and service

stalniy:
## Why

Improves app rendering time and performance. According to https://github.com/akash-network/console/discussions/1978. Previously `<BackgroundTaskProvider>` caused unnecessary re-rendering of the whole application, after this change only 2 components will be re-rendered only if user navigates to them (i.e., LogsViewer, ShellFileDownloadModal)

## Summary by CodeRabbit

* **Refactor**
  * Switched deployments/logs and shell-download flows to a new provider API actions layer and unified the proxy URL for WebSocket access (removed the separate WS env var).
  * Removed the background task provider wrapper from app startup.

* **Bug Fixes**
  * Improved cancellation, error reporting, retry handling, and robustness for downloads and log/event streaming.


apps/deploy-web/src/services/provider-proxy/provider-proxy.service.ts
```diff
@@ -76,6 +87,202 @@ export class ProviderProxyService {
...
+    gseq: number;
+    oseq: number;
+    chainNetwork: string;
+    signal?: AbortSignal;
+    service: string;
+    filePath: string;
+  }): Promise<DownloadMessagesResult> {
+    const printCommand = "cat"; // print command on Windows is "type"
+    const command = `${printCommand} ${input.filePath}`;
+    const url = `${input.providerBaseUrl}/lease/${input.dseq}/${input.gseq}/${input.oseq}/shell?stdin=0&tty=0&podIndex=0${command
```
baktun14, line 215: comment(non-blocking): I'm sure if we have to now, but after the new upgrade we might have to add `bseq` to this, but I don't think it's implemented and we have to do anything there, just an observation.

baktun14 (approved):

PR #2138, opened 2025-10-30
refactor(billing): improves batch signing client

ygrishajev:
## Summary by CodeRabbit

* **New Features**
  * Batched sign-and-broadcast API with lifecycle controls (dispose, pending-state)
  * New client creation via HTTP endpoint
  * Async memoization utility exposed

* **Bug Fixes**
  * Improved retry/recovery for sequence-mismatch and transient errors
  * Better cleanup to prevent orphaned signing clients; added explicit dispose helper

* **Improvements**
  * Enhanced batching/state logging and shutdown cleanup (server now runs disposal hook)
  * Functional-test env: sampling adjusted and stdout log format set to pretty


apps/api/src/billing/lib/batch-signing-client/batch-signing-client.service.spec.ts
```diff
@@ -1,234 +1,194 @@
...
+    ];
+    const erroredTestData = createTransactionTestData(granter);
+    erroredTestData.hash = new Error("Test error");
+
+    const allTestData = [...successfulTestData, erroredTestData];
+    const { service, client } = setup(allTestData);
+
+    const results = await Promise.allSettled(allTestData.map(data => service.signAndBroadcast(data.messages)));
+
+    expect(results).toHaveLength(allTestData.length);
```
stalniy: **nitpick(non-blocking):** this is redundant because `allSettled` always returns an array of the same length as the one passed as the parameter.

apps/api/src/billing/lib/batch-signing-client/batch-signing-client.service.ts
```diff
@@ -6,373 +6,379 @@ import type { EncodeObject, Registry } from "@cosmjs/proto-signing";
...
+    const result = await retry(
+      handleWhenResult(res => {
+        if (res instanceof Err && "message" in res.val && res.val.message?.includes("account sequence mismatch")) {
+          this.logger.warn({ event: "ACCOUNT_SEQUENCE_MISMATCH", address });
+          return true;
         }
+        return false;
+      }),
+      { maxAttempts: 5, backoff: new ExponentialBackoff({ maxDelay: 5_000, initialDelay: 500 }) }
+    ).execute(context => {
```
stalniy: **nitpick:** retry + handleWhenResult can be moved outside of the function call, it's like strategy to use. then you can just use `this.executor.execute(contest => ... )`

apps/api/src/billing/lib/batch-signing-client/batch-signing-client.service.ts
```diff
@@ -6,373 +6,379 @@ import type { EncodeObject, Registry } from "@cosmjs/proto-signing";
...
-      this.logger.debug({
-        event: "SEQUENCE_INCREMENTED",
-        address: this.cachedAccountInfo.address,
-        oldSequence,
-        newSequence: this.cachedAccountInfo.sequence
+    if (!tx) {
+      const error = new Error("Failed to sign and broadcast transaction");
+      this.logger.error({
+        event: "SIGN_AND_BROADCAST_ERROR",
+        error: error.message
```
stalniy: **thought(non-blocking):** probably it should be just `error` because then we will see stack trace in logs but maybe we don't need this logging at all because eventually it will cause 500 error, right?

apps/api/src/billing/lib/batch-signing-client/batch-signing-client.service.ts
```diff
@@ -6,373 +6,379 @@ import type { EncodeObject, Registry } from "@cosmjs/proto-signing";
...
+   * Results are cached in an instance variable to avoid repeated wallet operations.
+   *
+   * @returns The first address from the wallet.
+   */
+  private async getAddress() {
+    if (!this.address) {
+      this.address = await this.wallet.getFirstAddress();
+    }
+    return this.address;
+  }
```
stalniy: **issue(non-blocking):** this is race condition unsafe code. if called multiple times, all callers will invoke `getFirstAddress`. Probably this doesn't cause any issues in this case but to make it race condition safe, you need to use sync function with promise or semaphore/async lock.

```ts
private async getAddress() {
   this.getAddressPromise ??= this.wallet.getFirstAddress();
   return this.getAddressPromise;
}
```

apps/api/src/billing/lib/batch-signing-client/batch-signing-client.service.ts
```diff
@@ -6,373 +6,379 @@ import type { EncodeObject, Registry } from "@cosmjs/proto-signing";
...
+   *
+   * Results are cached in an instance variable to avoid repeated client calls.
+   *
+   * @returns The chain ID.
+   */
+  private async getChainId() {
+    if (!this.chainId) {
+      this.chainId = await this.client.getChainId();
+    }
+    return this.chainId;
```
stalniy: **issue(non-blocking):** race condition unsafe code.

@coderabbitai could you please remember this pattern and let us know when the next time you see code like this?

apps/api/src/billing/lib/sync-signing-stargate-client/sync-signing-stargate-client.ts
```diff
@@ -1,23 +1,17 @@
...
-  ): Promise<SyncSigningStargateClient> {
-    const cometClient = await connectComet(endpoint);
-    return this.createWithSigner(cometClient, signer, options);
-  }
-
-  public static createWithSigner(cometClient: CometClient, signer: OfflineSigner, options: SigningStargateClientOptions = {}): SyncSigningStargateClient {
+  public static init(endpoint: string, signer: OfflineSigner, options: SigningStargateClientOptions = {}): SyncSigningStargateClient {
+    const client = new HttpClient(endpoint);
+    const cometClient = Comet38Client.create(client);
     return new SyncSigningStargateClient(cometClient, signer, options);
```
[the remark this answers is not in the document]
stalniy: Comet38Client.create returns a client not a Promise.

apps/api/src/billing/lib/sync-signing-stargate-client/sync-signing-stargate-client.ts
```diff
@@ -1,23 +1,17 @@
...
-    endpoint: string | HttpEndpoint,
-    signer: OfflineSigner,
-    options: SigningStargateClientOptions = {}
-  ): Promise<SyncSigningStargateClient> {
-    const cometClient = await connectComet(endpoint);
-    return this.createWithSigner(cometClient, signer, options);
-  }
-
-  public static createWithSigner(cometClient: CometClient, signer: OfflineSigner, options: SigningStargateClientOptions = {}): SyncSigningStargateClient {
+  public static init(endpoint: string, signer: OfflineSigner, options: SigningStargateClientOptions = {}): SyncSigningStargateClient {
```
stalniy: **suggestion(non-blocking):** probably it makes to call it `.create`, the same way it's done in `Coment38Client`

apps/api/src/billing/services/balances/balances.service.ts
```diff
@@ -84,14 +84,12 @@ export class BalancesService {
     const activeDeploymentsResponse = await this.deploymentHttpService.findAll({ owner: address, state: "active" });
     const activeDeployments = activeDeploymentsResponse.deployments;
 
-    const deploymentEscrowBalance = activeDeployments.reduce((total: number, deployment: DeploymentInfo) => {
+    return activeDeployments.reduce((total: number, deployment: DeploymentInfo) => {
       const escrowAccount = deployment.escrow_account;
       if (!escrowAccount) return total;
 
       return total + parseFloat(escrowAccount.state.funds.reduce((sum, { amount }) => sum + parseFloat(amount), 0).toFixed(18));
```
stalniy, line 91: **question(non-blocking):** what is the purpose of using `.toFixed(18)` if then this value is passed to `parseFloat`? `.toFixed` is just a formatting method and in this case I don't understand its purpose

stalniy (dismissed):

apps/api/src/billing/services/dedupe-signing-client/dedupe-signing-client.service.ts
```diff
@@ -27,15 +27,16 @@ export class DedupeSigningClientService {
...
 
     try {
-      return client.executeTx(messages, options);
+      return client.signAndBroadcast(messages, options);
     } finally {
       if (!client.hasPendingTransactions && this.clientsByAddress.has(key)) {
         this.logger.debug({ event: "DEDUPE_SIGNING_CLIENT_CLEAN_UP", key });
         this.clientsByAddress.delete(key);
+        client.dispose();
       }
```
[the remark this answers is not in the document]
stalniy: @ygrishajev coderabbit is right, it calls `finally` immediately, I tested this in isolated env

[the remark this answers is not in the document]
ygrishajev: Yea, I've got this fixed already locally. It doesn't fix the ws issue thought.

apps/api/src/billing/services/balances/balances.service.ts
ygrishajev: No idea. @baktun14 wdyt?

apps/api/src/billing/lib/sync-signing-stargate-client/sync-signing-stargate-client.ts
ygrishajev: it would conflict with the base method, so it should be named smth else.

apps/api/src/billing/services/balances/balances.service.ts
ygrishajev: i'm gonna resolve this anyway as this is irrelevant to this pr

stalniy (dismissed):

apps/api/src/billing/lib/sync-signing-stargate-client/sync-signing-stargate-client.ts
```diff
@@ -22,4 +39,17 @@ export class SyncSigningStargateClient extends SigningStargateClient {
...
+   * Ensures the underlying client connection is established.
+   *
+   * This method performs a test query to verify the connection is active.
+   * Useful for waiting for the connection to be ready before making queries.
+   *
+   * @returns A promise that resolves when the connection is confirmed.
+   * @throws If the connection cannot be established.
+   */
+  async connected() {
+    await this.forceGetCometClient().status();
```
stalniy, line 53: **question(non-blocking)**: is this race condition safe?

ygrishajev: what race conditions do you mean?

stalniy (dismissed):

apps/api/src/billing/lib/sync-signing-stargate-client/sync-signing-stargate-client.ts
stalniy: if this method is called by many callers, everybody will try to get status. So, instead of one there will be many calls to `this.forceGetCometClient().status()`

PR #2145, opened 2025-11-01
fix: omit userId from update

jzsfkzm:
refs #1934

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where user ID was being unintentionally overwritten during record updates, ensuring user identifiers remain stable and properly preserved across operations.


stalniy:
I don't think this is the reason, the issue was that prod db didn't have constraint connected to the index and that's why it failed with index_name constraint. I adjusted the prod db constraints to match what we have in migrations

PR #2147, opened 2025-11-03
fix(billing): refill trial wallets for non-anonymous

baktun14:
## Summary by CodeRabbit

* **Improvements**
  * Trial fees allowance increased to 1,000,000 units, extending support for trial users.
  * Wallet processing simplified to evaluate all wallets uniformly when assessing refills.

* **Tests**
  * Refill test streamlined to assert uniform fee allowance updates across all wallets, removing special-case distinctions.

* **Chores**
  * Reduced available network configurations and disabled testnet in initial network settings.


apps/api/src/billing/services/refill/refill.service.ts
```diff
@@ -19,13 +21,18 @@ export class RefillService {
...
-    private readonly analyticsService: AnalyticsService
+    private readonly analyticsService: AnalyticsService,
+    private readonly featureFlagsService: FeatureFlagsService
   ) {}
 
   async refillAllFees() {
-    const wallets = await this.userWalletRepository.findDrainingWallets({
-      fee: this.config.FEE_ALLOWANCE_REFILL_THRESHOLD
-    });
+    const isAnonymousTrialEnabled = this.featureFlagsService.isEnabled(FeatureFlags.ANONYMOUS_FREE_TRIAL);
```
stalniy: **issue(blocking):** make sure it actually works in console cli because I have a feeling that it doesn't work. Because FeatureFlagsService expects ExecutionContextService to be available but ExecutionContextService is not executed in console action:

```ts
program
  .command("refill-wallets")
  .description("Refill draining wallets")
  .action(async (options, command) => {
    await executeCliHandler(command.name(), async () => {
      await container.resolve(WalletController).refillWallets();
    });
  });
```

stalniy (changes requested):

apps/api/src/billing/services/refill/refill.service.ts
baktun14: Good point, just removed the check since we will most likely not revert to anonymous trials

apps/api/src/billing/repositories/user-wallet/user-wallet.repository.ts
```diff
@@ -83,12 +83,16 @@ export class UserWalletRepository extends BaseRepository<ApiPgTables["UserWallet
     return this.toOutput(item);
   }
 
-  async findDrainingWallets(thresholds = { fee: 0 }) {
-    const where = and(lte(this.table.feeAllowance, thresholds.fee.toString()), eq(this.table.isTrialing, false));
+  async findDrainingWallets(thresholds = { fee: 0 }, options?: { excludeTrialWallets?: boolean }) {
+    const conditions: SQL[] = [lte(this.table.feeAllowance, thresholds.fee.toString())];
+
+    if (options?.excludeTrialWallets) {
```
ygrishajev: non-blocking. as this option isn't used anymore, perhaps it makes sense to also remove it?

ygrishajev (dismissed):

stalniy (approved):

Issue #2141, opened 2025-10-31
SequelizeConnectionAcquireTimeoutError: Operation timeout


stalniy:
## What

Looks like indexer database sometimes overloaded and fails with timeout:

```
SequelizeConnectionAcquireTimeoutError: Operation timeout
    at ConnectionManager.getConnection (webpack://akashnetwork/node_modules/sequelize/lib/dialects/abstract/connection-manager.js:206:1)
    at async <anonymous> (webpack://akashnetwork/node_modules/sequelize/lib/sequelize.js:305:1)
    at async PostgresQueryInterface.select (webpack://akashnetwork/node_modules/sequelize/lib/dialects/abstract/query-interface.js:407:1)
    at async Function.findAll (webpack://akashnetwork/node_modules/sequelize/lib/model.js:1140:1)
    at async Function.findOne (webpack://akashnetwork/node_modules/sequelize/lib/model.js:1240:1)
    at async DeploymentRepository.findByIdWithGroups (webpack://akashnetwork/console-api/src/deployment/repositories/deployment/deployment.repository.ts:145:12)
    at async FallbackDeploymentReaderService.findByOwnerAndDseq (webpack://akashnetwork/console-api/src/deployment/services/fallback-deployment-reader/fallback-deployment-reader.service.ts:90:24)
    at async DeploymentReaderService.getDeployment (webpack://akashnetwork/console-api/src/deployment/services/deployment-reader/deployment-reader.service.ts:315:16)
    at async DeploymentReaderService.getDeploymentByOwnerAndDseq (webpack://akashnetwork/console-api/src/deployment/services/deployment-reader/deployment-reader.service.ts:208:24)
    at async DeploymentController.getByOwnerAndDseq (webpack://akashnetwork/console-api/src/deployment/controllers/deployment/deployment.controller.ts:94:24)
```

stalniy:
fails on GET http://console-api.akash.network/v1/transactions/6E254E526EDB70F22A9C38E8C52307BBC0AA4C8C5ADE09F3E45FF6D3953C2FB8

stalniy:
some of this things could be cached

stalniy:
GET http://console-api.akash.network/v1/addresses/akash1p4k0jer30g25mf62h5yc42agtjeumw4gxk8z37

stalniy:
userAgent is node, so it's server side code which fetches this data

stalniy:
[...]

PR #2149, opened 2025-11-04
refactor: rewrites SQL query to fetch transactions by address ref

stalniy:
## Why

Perf optimization. Relates to #2141

## Summary by CodeRabbit

* **Performance**
  * Faster transaction/address lookups via optimized query behavior and a new composite index.

* **Improvements**
  * Network list now only shows networks with at least one API or RPC endpoint.

* **Chores**
  * Added database migration, schema snapshot, and journal entry to apply/index the new index and related schema updates.

### DIFF
--- a/apps/indexer/drizzle/schema.ts
+++ b/apps/indexer/drizzle/schema.ts
@@ -587,6 +587,7 @@ export const addressReference = pgTable(
   table => [
     index("address_reference_address").using("btree", table.address.asc().nullsLast().op("text_ops")),
     index("address_reference_transaction_id").using("btree", table.transactionId.asc().nullsLast().op("uuid_ops")),
+    index("address_reference_address_transaction_id").using("btree", table.address.asc().nullsLast(), table.transactionId.asc().nullsLast()),
     foreignKey({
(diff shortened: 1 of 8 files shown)


ygrishajev (dismissed):

apps/indexer/drizzle/schema.ts
```diff
@@ -587,6 +587,7 @@ export const addressReference = pgTable(
   table => [
     index("address_reference_address").using("btree", table.address.asc().nullsLast().op("text_ops")),
     index("address_reference_transaction_id").using("btree", table.transactionId.asc().nullsLast().op("uuid_ops")),
+    index("address_reference_address_transaction_id").using("btree", table.address.asc().nullsLast(), table.transactionId.asc().nullsLast()),
```
baktun14, line 590: comment: I'm not sure, but since we have both the sequelize and the drizzle schema, we should probably keep both synced https://github.com/akash-network/console/blob/main/packages/database/dbSchemas/base/addressReference.ts

baktun14 (dismissed):

PR #2150, opened 2025-11-04
fix: implements factory produced di values dispose

ygrishajev:
## Summary by CodeRabbit

* **Infrastructure**
  * Updated RPC node endpoint for improved blockchain connectivity.
* **Chores / Stability**
  * Changed shutdown lifecycle to stop automatically disposing certain signing clients and related automatic teardown steps.
  * Added a centralized disposal registry to track and reliably clean up resources during shutdown.
* **Tests**
  * Added unit tests validating the disposal registry and its behavior when disposals fail.


apps/api/src/core/lib/disposable-registry/disposable-registry.ts
```diff
@@ -0,0 +1,73 @@
...
+   * that may return disposable instances.
+   *
+   * @param factory - The factory function to wrap.
+   * @returns A wrapped factory function that registers disposables automatically.
+   */
+  static registerFromFactory<T>(factory: FactoryProvider<T>["useFactory"]): FactoryProvider<T>["useFactory"] {
+    return container.resolve(DisposableRegistry).registerFromFactory(factory);
+  }
+
+  private readonly disposes: Array<() => Promise<void> | void> = [];
```
stalniy, line 29: **suggestion(non-blocking):** registry could collect instances of `Disposable`, so then there is no need to create wrapper functions around every `.dispose` method call

apps/api/src/rest-app.ts
```diff
@@ -201,7 +200,6 @@ export { appHono as app, connectUsingSequelize as initDb };
 export async function bootstrap(port: number) {
   await startServer(appHono, LoggerService.forContext("APP"), process, {
     port,
-    beforeStart: migratePG,
-    beforeEnd: () => disposeSigningClients()
+    beforeStart: migratePG
   });
```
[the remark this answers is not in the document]
ygrishajev: dispose is not actually needed there just yet

stalniy (approved):

stalniy:
https://github.com/microsoft/tsyringe/issues/274

apps/api/src/core/lib/disposable-registry/disposable-registry.ts
ygrishajev: i understand that and picked the approach i picked ;)

PR #2152, opened 2025-11-04
feat(billing): adds a unique rpc proxy header per instance of a batch signing client

ygrishajev:
## Summary by CodeRabbit

* **Refactor**
  * Reworked signing client wiring to use a modular factory approach for creating HTTP-backed signing clients with unique request identifiers, and removed an older synchronous client wrapper.
* **Tests**
  * Added and updated unit tests to cover the new signing-client factory and updated client usage.


baktun14 (approved):
Nice!

PR #2163, opened 2025-11-06
refactor: moves certs and transaction tests into unit scope

stalniy:
## Why

We should use functional tests only when it makes sense, if db level or unit level test can cover functionality then we should prefer it instead of writing functional or e2e test. 

e2e tests should be written only for business critical functionality

## Summary by CodeRabbit

* **Bug Fixes**
  * Certificates are now created using the authenticated user's identity.

* **New Features**
  * Certificate manager is provided via dependency injection and re-exported for API use.
  * Certificate creation endpoint now uses authenticated context automatically.

* **Tests**
  * Added unit tests for certificate creation success and failure paths.
  * Added repository tests for transaction retrieval and limits.
  * Removed several obsolete functional tests and simplified template-cache tests.


apps/api/src/certificate/services/certificate/certificate.service.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+import type { AuthService } from "@src/auth/services/auth.service";
+import type { UserWalletRepository } from "@src/billing/repositories";
+import type { RpcMessageService } from "@src/billing/services";
+import type { ManagedSignerService } from "@src/billing/services/managed-signer/managed-signer.service";
+import { CertificateService } from "./certificate.service";
+
+describe(CertificateService.name, () => {
+  describe("create", () => {
+    it("creates certificate successfully when wallet exists", async () => {
+      const userWallet = {
```
ygrishajev: https://github.com/akash-network/console/blob/main/apps/api/test/seeders/user-wallet.seeder.ts

ygrishajev (dismissed):

PR #2170, opened 2025-11-06
feat(onboarding): welcome step create deployment

### LINKED ISSUE #1185: New Onboarding "Start Page"
# Problem Statement

Today Console loses a majority of trial users through the onboarding funnel. The three most common causes for this are:

1. There are too many choices to choose from when they click "Deploy" (and that is assuming that they do click Deploy). 
2. Most users don't get a bid in response to their deployment request because all the trial providers are fully used up.
3. Takes too long to get through the process and see a deployed workload

# Solution

To address these issues we will implement the following changes to the onboaring experience in Console:

1. A streamlined onboarding experience where the user is just presented with 1 to 3 options to pick from for their first deployment. "Welcome to your first deployment on Akash Console" (basically a "hello world" app), "ComfyUI" (or some other 1 GPU workload) and Llama-8B (that uses 8 GPUs). If we find that people are picking the 3rd and not progressing further, we'll remove that option.
2. There will be no anonymous trial anymore. Users who decide to not use crypto will be required to sign up/ sign-in before they can start a trial or deploy with credits. This will increase the likelihood of a user getting a bid.
3. Once signed up (or signed-in) users will be nudged to enter their credit card information and will be granted some credits (amount TBD) for doing so. We will A/ B test two options here - one that grants users a small amount of credit (without requiring them to enter CC info) and another that grants a much larger (say 5-10x) if they put their CC info.
4. Users that provide CC will have access to a large set of trial providers which will also help improve the likelihood of getting bids.
5. All trial users' leases will have a timeout (TBD) associated with them

# Design

Today we do not have a custom experience for new (anonymous) users - they see the same screen that any returning user sees. We need a new "home" experience for first time users that is clean (without too many choices) and just has people do the following:
1. Start Trial
2. Deploy their first app (which will present the 3 templates and abstract away as much of the complexity or details of what it takes to deploy them as. possible). It should ideally just be an experience where they click the template and next they pick a bid and then see the lease screen.
3. For the Apps/ Tiles we present the user, I'm thinking:
    - CPU app: Hello Akash!
    - Single GPU (4090) app: ComfyUI (Stable Diffusion UI)
    - Multi-GPU (8 H100s) app: Llama-3.1-8b (or something like that)

@aktdenis - could you please work on the wireframes for this? (cc @baktun14 if there are additional things to add - please feel free to edit issue and add)

baktun14:
closes https://github.com/akash-network/console/issues/1185
image

## Summary by CodeRabbit

* **New Features**
  * Template-based onboarding deploys with per-template deploy flow and async completion
  * Trial balance hook and Trial Status bar show credits, expiry, progress
  * Template card component and a lightweight UI Skeleton for loading states
  * Back-to-console click now records an analytics event

* **Bug Fixes**
  * Credits Remaining now shows net available funds after escrow

* **Chores**
  * Wallet responses now include a creation timestamp


apps/deploy-web/src/components/onboarding/OnboardingContainer/OnboardingContainer.tsx
```diff
@@ -161,6 +178,105 @@ export const OnboardingContainer: React.FunctionComponent<OnboardingContainerPro
...
+            template: templateName,
+            dseq: dd.deploymentId.dseq
+          });
+
+          d.localStorage?.removeItem(ONBOARDING_STEP_KEY);
+          connectManagedWallet();
+          router.push(d.UrlService.newDeployment({ step: RouteStep.createLeases, dseq: dd.deploymentId.dseq }));
+        }
+      } catch (error) {
+        console.error("Error deploying template:", error);
```
stalniy: **nitpick:** no need to use console.error, since you rethrow the error and it will be visible in console and caught by sentry

apps/deploy-web/src/components/onboarding/OnboardingContainer/OnboardingContainer.tsx
```diff
@@ -37,7 +48,16 @@ const DEPENDENCIES = {
...
+  useTemplates,
+  useCertificate,
+  useSnackbar,
+  localStorage: typeof window !== "undefined" ? window.localStorage : null,
+  deploymentData,
+  validateDeploymentData,
+  appendAuditorRequirement,
+  helloWorldTemplate,
+  TransactionMessageData,
+  UrlService
```
stalniy, line 60: **suggestion(non-blocking):** this service is exposed via `useServices` hook.

I have a rule in my mind:
1. all services should be accessed via `useServices` hook
2. `dependencies` property of the component should contain only components, in rare cases other hooks

Also `deploymentData`, `validateDeploymentData`, `appendAuditorRequirement`, `TransactionMessageData`, `hellowWorldTemplate` and `localStorage` should be exposed via `useServices` because they belong to service layer

apps/deploy-web/src/components/onboarding/OnboardingContainer/OnboardingContainer.tsx
```diff
@@ -161,6 +178,105 @@ export const OnboardingContainer: React.FunctionComponent<OnboardingContainerPro
...
+          "llama-3.1-8b": {
+            id: "akash-network-awesome-akash-Llama-3.1-8B",
+            sdl: "",
+            name: "Llama 3.1 8B"
+          }
+        };
+
+        const templateConfig = templateMap[templateName];
+        if (!templateConfig) {
+          throw new Error(`Template ${templateName} not found`);
```
stalniy: **question(non-blocking):** wouldn't it better to show a message to user like this: 

"Template X" is no longer supported, please chose another one

You could additionally log the error to sentry

stalniy (dismissed):

apps/deploy-web/src/components/onboarding/OnboardingContainer/OnboardingContainer.tsx
baktun14: Yea makes sense, I'll create a an issue to refactor those into services

baktun14: https://github.com/akash-network/console/issues/2172

stalniy: You don't need to refactor them straight away, you could just pass them via `useServices` not via `DEPENDENCIES`. They serve different purposes

Issue #1779, opened 2025-08-05
Implement Auto Credit Reload UI


baktun14:
Should follow the same approach as AksahML

1. There should be the concept of a "default" credit card
2. All users should have at least one credit card set to "default". To start with this will be the CC they use to sign up
3. Auto Credit Reload should use the default credit card to add funds into their account, when some conditions are true
4. The conditions are:
     - The (auto credit reload) feature is enabled
     - There are two thresholds: One that is for when the reload is triggered and another for what amount is added. Both should have defaults but also should be configurable/ changeable by the user
     - The defaults should be $10 for the trigger and $50 for the amount to reload.
5. Implementing this requires that the user agree to letting us charge the credit card (authorization)

<img width="1331" height="1004" alt="Image" src="https://github.com/user-attachments/assets/7a70769e-efb3-4751-989e-a44af7b608d7" />

<img width="1026" height="918" alt="Image" src="https://github.com/user-attachments/assets/0fb1bdf6-c5d7-434a-aad9-414b74f76b24" />

baktun14:
@ygrishajev I think we can close this issue and milestone is completed

PR #2171, opened 2025-11-07
feat(notifications): upgrades pg-boss

ygrishajev:
refs #1779

## Summary by CodeRabbit

* **Chores**
  * Bumped pg-boss dependency to v11.0.4.

* **Bug Fixes**
  * Completed-job archival now applied per-queue and controlled by configuration for consistent cleanup.
  * Queue creation no longer includes an explicit name field in options.

* **Tests**
  * Updated tests to match per-queue archival behavior and revised queue options.

### DIFF
--- a/apps/notifications/src/infrastructure/broker/services/broker/broker.service.ts
+++ b/apps/notifications/src/infrastructure/broker/services/broker/broker.service.ts
@@ -36,7 +36,9 @@ export class BrokerService {
 
   async subscribe<ReqData>(eventName: string, options: { prefetchCount: number }, handler: SingleMsgWorkHandler<ReqData>) {
     const queueName = this.toQueueName(eventName);
-    await this.boss.createQueue(queueName);
+    await this.boss.createQueue(queueName, {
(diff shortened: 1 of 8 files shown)


baktun14 (dismissed):

apps/notifications/src/infrastructure/broker/services/broker/broker.service.ts
```diff
@@ -36,7 +36,9 @@ export class BrokerService {
 
   async subscribe<ReqData>(eventName: string, options: { prefetchCount: number }, handler: SingleMsgWorkHandler<ReqData>) {
     const queueName = this.toQueueName(eventName);
-    await this.boss.createQueue(queueName);
+    await this.boss.createQueue(queueName, {
+      deleteAfterSeconds: this.configService.getOrThrow("broker.EVENT_BROKER_ARCHIVE_COMPLETED_AFTER_SECONDS")
```
stalniy, line 40: **thought(non-blocking):** one important difference in v1 is that pg-boss no longer partition jobs table by default and if you need to enable this behavior, it should be explicit. https://timgit.github.io/pg-boss/#/./api/queues

Image

PR #2179, opened 2025-11-10
fix: updates jest in deploy-web to 30.x

stalniy:
## Why

To use newer version of jsdom which contains newer browser API (e.g., `AbortSignal.any`)

## Summary by CodeRabbit

* **Chores**
  * Updated Jest testing framework and jest-environment-jsdom from v29.7.0 to v30.2.0

* **Refactor**
  * Improved internal dependency injection patterns
  * Enhanced test infrastructure for better maintainability


apps/deploy-web/src/components/onboarding/OnboardingContainer/OnboardingContainer.tsx
```diff
@@ -67,7 +67,8 @@ export const OnboardingContainer: React.FunctionComponent<OnboardingContainerPro
   const router = d.useRouter();
   const { user } = d.useUser();
   const { data: paymentMethods = [] } = d.usePaymentMethodsQuery({ enabled: !!user?.stripeCustomerId });
-  const { analyticsService, urlService, authService, chainApiHttpClient, deploymentLocalStorage, appConfig, errorHandler } = d.useServices();
+  const { analyticsService, urlService, authService, chainApiHttpClient, deploymentLocalStorage, appConfig, errorHandler, windowLocation, windowHistory } =
```
ygrishajev, line 70: non-blocking: perhaps it makes more sense to just use `window` so that we wouldn't need to add every window prop in the future? 

```suggestion
  const { analyticsService, urlService, authService, chainApiHttpClient, deploymentLocalStorage, appConfig, errorHandler, windowLocation, window } =
```

ygrishajev (approved):

apps/deploy-web/src/components/onboarding/OnboardingContainer/OnboardingContainer.tsx
stalniy: from another hand, it will be harder to mock the whole `window` in tests. Though possible with `mockDeep<Window>()`. Honestly, I think this is a rare case, for example localStorage we need to encapsulate separately.

but the main reason behind this change is that I just needed to fix it for new jsdom version which doesn't allow to redefine `location` and `history` on `window`. Ideally we should use router for both cases, not sure whether react router exposes this information

PR #2181, opened 2025-11-10
fix: fix selecting services on Deployment Logs tab

### LINKED ISSUE #2105: Unselecting service in logs causes runtime exception
## What

1. Authenticate in console
2. Create any deployment
3. Go to logs
4. Unselect all services in log filter

## Expected result

Empty logs section, loading is stopped

## Actual result

Empty logs section, loading bar never stops

## Additional

This happens because of runtime error

https://github.com/user-attachments/assets/49179b4c-52f9-4513-a7e1-27dfab06f023

jzsfkzm:
closes #2105

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of deployment log streaming with enhanced connection state management.
  * Optimized service switching behavior to properly reset and reconnect log streams when changing selections.

### DIFF
--- a/apps/deploy-web/src/components/deployments/DeploymentLogs.tsx
+++ b/apps/deploy-web/src/components/deployments/DeploymentLogs.tsx
@@ -33,7 +33,8 @@ type Props = {
 
 export const DeploymentLogs: React.FunctionComponent<Props> = ({ leases, selectedLogsMode }) => {
   const { analyticsService } = useServices();
-  const [isLoadingLogs, setIsLoadingLogs] = useState(true);
+  const [isLoadingLogs, setIsLoadingLogs] = useState(false);
+  const [isConnecting, setIsConnecting] = useState(false);
   const [canSetConnection, setCanSetConnection] = useState(false);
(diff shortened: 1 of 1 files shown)


stalniy:
@jzsfkzm unfortunately we will need to close this one because I refactored this logic in https://github.com/akash-network/console/pull/2177 . There is no more a hook to work with websocket, this logic has been moved to service layer. 

Could you please double check whether the same is reproducible in the new version?

jzsfkzm:
> @jzsfkzm unfortunately we will need to close this one because I refactored this logic in #2177 . There is no more a hook to work with websocket, this logic has been moved to service layer.
> 
> Could you please double check whether the same is reproducible in the new version?

@stalniy I just checked it and now it works as expected:
- Unselecting one service loads logs for the other service only.
- Unselecting all/each and every service won't end up in an infinite progress indicator.

There's however one JS error in the logs when you open the select due to a prop deprecated in MUI5, I just created a PR for that:
https://github.com/akash-network/console/pull/2228

PR #2201, opened 2025-11-11
feat: can accept base64 payload while proxying to provider

jzsfkzm:
refs #2178

A follow-up change in a week or so should remove the new field and old version of decoding.

## Summary by CodeRabbit

* **Improvements**
  * Binary message payloads are now encoded as base64 when sent and are conditionally decoded on receipt to ensure accurate transmission and reconstruction.
  * Configuration/schema includes an optional flag to mark payloads as base64, enabling interoperable handling across services.

* **Tests**
  * Test expectations updated to assert concrete base64-encoded payloads to reflect the new encoding behavior.

### DIFF
--- a/apps/deploy-web/src/services/provider-proxy/provider-proxy.service.ts
+++ b/apps/deploy-web/src/services/provider-proxy/provider-proxy.service.ts
@@ -303,11 +303,12 @@ export class ProviderProxyService {
           url,
           auth: providerCredentialsToApiCredentials(input.providerCredentials),
           chainNetwork: this.netConfig.mapped(input.chainNetwork),
-          providerAddress: input.providerAddress
+          providerAddress: input.providerAddress,
+          isBase64: true
         };
 
         if (message.length > 0) {
-          remoteMessage.data = message.toString();
+          remoteMessage.data = btoa(String.fromCharCode(...message));
         }
 
         return JSON.stringify(remoteMessage);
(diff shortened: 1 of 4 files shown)


apps/deploy-web/src/services/provider-proxy/provider-proxy.service.ts
```diff
@@ -303,11 +303,17 @@ export class ProviderProxyService {
           url,
           auth: providerCredentialsToApiCredentials(input.providerCredentials),
           chainNetwork: this.netConfig.mapped(input.chainNetwork),
-          providerAddress: input.providerAddress
+          providerAddress: input.providerAddress,
+          isBase64: true
```
stalniy, line 308: **praise:** simple and elegant solution!

apps/deploy-web/src/services/provider-proxy/provider-proxy.service.ts
```diff
@@ -303,11 +303,17 @@ export class ProviderProxyService {
...
         };
 
         if (message.length > 0) {
-          remoteMessage.data = message.toString();
+          let binaryString = '';
+          for (let i = 0; i < message.length; i++) {
+            binaryString += String.fromCharCode(message[i]);
+          }
+
+          remoteMessage.data = btoa(binaryString);
```
stalniy: **suggestion(blocking):** please use helper function for base64 encoding from src/utils/encoding.ts

stalniy (changes requested):

apps/deploy-web/src/services/provider-proxy/provider-proxy.service.ts
stalniy: @copilot can you push the fix in this branch?

stalniy: @coderabbitai can you suggest committable suggestion for this comment?

stalniy (approved):

Issue #2157, opened 2025-11-04
Include count of bids received as prop in analytics event


anilmurty:
Add properties for sending count of bids received in the `bid_selected` event

[console/apps/deploy-web/src/services/analytics/analytics.service.ts](https://github.com/akash-network/console/blob/1febccc31a38ef85b18a3016970507a6d0144c0c/apps/deploy-web/src/services/analytics/analytics.service.ts#L117)

Line 117 in [1febccc](https://github.com/akash-network/console/commit/1febccc31a38ef85b18a3016970507a6d0144c0c)

 export type EventProperties = {

Issue #2156, opened 2025-11-04
Add field to collect organization


anilmurty:
Add a field to the stripe sign up page during new user creation flow that asks for "Organization" as an empty text field for the user to fill in. I believe this is also a standard field in the stripe API but if not we can make a custom one. 

<img width="882" height="359" alt="Image" src="https://github.com/user-attachments/assets/2c8b25db-5926-417c-b117-643b2453b352" />

PR #2203, opened 2025-11-11
feat(onboarding): improve onboarding analytics

### LINKED ISSUE #2155: Send user info as props to Analytics events
Add properties for sending user name, email, organization name that we will collect at sign up : https://github.com/akash-network/console/blob/1febccc31a38ef85b18a3016970507a6d0144c0c/apps/deploy-web/src/services/analytics/analytics.service.ts#L117

baktun14:
closes https://github.com/akash-network/console/issues/2155 https://github.com/akash-network/console/issues/2156 https://github.com/akash-network/console/issues/2157

## Summary by CodeRabbit

* **New Features**
  * Optional "Organization" field during payment setup; saved to the customer record and available via API.
  * Success animation shown after completing a payment method.
  * New lightweight Stripe input component for payment forms.

* **Bug Fixes**
  * Credits display now shows correct available amount.
  * Removing a payment method resets the setup flow; completing payment method no longer auto-advances prematurely.

* **Improvements**
  * Analytics: added "bids_received" event and richer registration tracking.
  * Minor UI text/layout updates in payment and onboarding screens.


apps/api/src/billing/controllers/stripe/stripe.controller.ts
```diff
@@ -196,4 +203,15 @@ export class StripeController {
       throw error;
     }
   }
+
+  @Protected([{ action: "create", subject: "StripePayment" }])
+  async updateCustomerOrganization(input: UpdateCustomerOrganizationRequest): Promise<UpdateCustomerOrganizationResponse> {
+    const { currentUser } = this.authService;
+
+    assert(currentUser.stripeCustomerId, 400, "Stripe customer ID not found");
```
stalniy: **suggestion(non-blocking):** I think we need to write something like "Payment method is not configured for this user". Since it's 400 (client error), telling him that there is no stripe customer id doesn't help to understand the client what is wrong

apps/api/src/billing/http-schemas/stripe.schema.ts
```diff
@@ -211,6 +211,14 @@ export const ErrorResponseSchema = z.object({
   type: z.string().optional()
 });
 
+export const UpdateCustomerOrganizationRequestSchema = z.object({
+  organization: z.string()
+});
+
+export const UpdateCustomerOrganizationResponseSchema = z.object({
+  success: z.boolean()
+});
```
stalniy: **suggestion(non-blocking):** if there is no plan to return `{ success: false }`, then better to just respond with 204 No content status

apps/api/src/billing/services/stripe/stripe.service.spec.ts
```diff
@@ -475,44 +482,87 @@ describe(StripeService.name, () => {
...
+            percent_off: null,
+            valid: true
+          }
         }
       });
+      const mockInvoice = createTestInvoice({ id: "in_123", status: "draft" });
+      const mockFinalizedInvoice = createTestInvoice({ id: "in_123", status: "paid" });
+
       jest.spyOn(service, "findPromotionCodeByCode").mockResolvedValue(stub(mockPromotionCode));
+      jest.spyOn(service.invoices, "create").mockResolvedValue(stub(mockInvoice));
```
stalniy: **issue(non-blocking):** `stub` is deprecated, it was a simple way to mimic what jest-mock-extended does. So, `stub` should not be used in new code

apps/api/src/billing/services/stripe/stripe.service.ts
```diff
@@ -495,6 +525,16 @@ export class StripeService extends Stripe {
     return reloaded.stripeCustomerId;
   }
 
+  async updateCustomerOrganization(customerId: string, organization: string): Promise<void> {
+    const customer = await this.customers.retrieve(customerId);
+
+    assert(!("deleted" in customer), 400, "Customer is deleted");
```
stalniy: **question(non-blocking):** does it make sense to return 404 instead? so the client doesn't know that customer is deleted

apps/deploy-web/src/components/onboarding/steps/PaymentMethodContainer/PaymentMethodContainer.tsx
```diff
@@ -156,7 +158,15 @@ export const PaymentMethodContainer: FC<PaymentMethodContainerProps> = ({ childr
     }
   }, [isConnectingWallet, managedWalletError]);
 
-  const handleSuccess = () => {
+  const handleSuccess = async (organization?: string) => {
+    if (organization) {
+      try {
+        await stripe.updateCustomerOrganization(organization);
+      } catch (error) {
+        sentry.captureException(error, { extra: { context: "Failed to update customer organization" } });
```
stalniy: **issue(blocking):** please use `errorHandler` service from `useServices`

stalniy (changes requested):

apps/deploy-web/src/components/onboarding/steps/PaymentMethodContainer/PaymentMethodContainer.tsx
stalniy: @coderabbitai can you provide committable suggestion for this issue?

apps/deploy-web/src/components/onboarding/steps/PaymentMethodContainer/PaymentMethodContainer.tsx
```diff
@@ -156,7 +158,15 @@ export const PaymentMethodContainer: FC<PaymentMethodContainerProps> = ({ childr
...
+      }
+    }
```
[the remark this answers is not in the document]
baktun14: This is only internal and we don't want to warn the user about something they can't fix, this is why I only put a sentry log so we can fix it internally if it's broken.

apps/api/src/billing/services/stripe/stripe.service.spec.ts
```diff
@@ -555,61 +616,65 @@ describe(StripeService.name, () => {
...
-        promotion_code: null
+      // Verify that the invoice was created and finalized before the wallet top-up failed
+      expect(service.invoices.create).toHaveBeenCalledWith({
+        customer: mockUser.stripeCustomerId,
+        auto_advance: false,
+        discounts: [{ promotion_code: mockPromotionCode.id }]
       });
+      expect(service.invoices.finalizeInvoice).toHaveBeenCalledWith(mockInvoice.id);
+      expect(refillService.topUpWallet).toHaveBeenCalledWith(1000, mockUser.id);
     });
```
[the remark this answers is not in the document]
baktun14: I think we're fine to deal with a new coupon in the case it failed to topup

stalniy (approved):

Issue #2189, opened 2025-11-11
Refactor provider-proxy to accept REST_API_NODE_URL


stalniy:
## Why

Right now, all backend services expect `RPC_NODE_ENDPOINT` (write) env variable and use this `REST_API_NODE_URL`(read) to talk to blockchain but provider-proxy relies on local `packages/net` package data (which is fetched on every commit from https://github.com/akash-network/net). And this makes it impossible to provide a custom proxy option. Right now, we want to switch provider-proxy to use our own rpc proxy which ensures that request is delivered to healthy blockchain node

## What

1. Add `REST_API_NODE_URL` env variable to provider-proxy
2. Add `.env.sample` put the value from https://github.com/akash-network/net/blob/main/sandbox-2/meta.json#L52C21-L52C26 and a link to https://github.com/akash-network/net in comment to look for other values 
3. Add `.env.mainnet` and `.env.sandbox` with corresponding values from https://github.com/akash-network/net
4. Use `REST_API_NODE_URL` and remove `chainNetwork` and `network` parameter from proxy requests https://github.com/akash-network/console/blob/main/apps/provider-proxy/src/utils/schema.ts#L26
5. Adjust deploy-web to create a new instance of `ProviderProxyService` when network changes (use `networkStore.selectedNetworkId`). Network change happens with page reload. Add corresponding env variables like it's done for api https://github.com/akash-network/console/blob/main/apps/deploy-web/env/.env#L16-L18

stalniy:
@jzsfkzm would you like to work on this?

stalniy:
this one cannot be merged without preparing https://github.com/akash-network/console/issues/2190 otherwise we won't be able to deploy provider-proxy for some time

jzsfkzm:
> [@jzsfkzm](https://github.com/jzsfkzm) would you like to work on this?

sure, assign it to me please

stalniy:
@jzsfkzm So yes, I cannot assign you to the issue until you leave a comment. So, it's a hard requirement

PR #2225, opened 2025-11-14
feat: provider-proxy accepts REST_API_NODE_URL

jzsfkzm:
refs #2189

## Summary by CodeRabbit

* **Configuration Updates**
  * Added REST_API_NODE_URL and made PORT configurable; server can be started with a full config object.

* **New Features**
  * Server start accepts an environment/config object for runtime configuration.

* **Bug Fixes**
  * Request timeout validated with bounds and defaults to 9000ms when unspecified.

* **Chores**
  * Added environment-loading support and simplified provider flow to use a single configured API URL (removed network-scoped routing).

* **Tests**
  * Improved test isolation and explicit wiring of the chain API URL per test.


stalniy:
@jzsfkzm could you please revert changes in deploy-web? I will adjust it in your PR which changes helm values and gha

apps/provider-proxy/env/.env.functional.test
```diff
@@ -0,0 +1 @@
+REST_API_NODE_URL=https://rest-akash.ecostake.com
```
stalniy: **suggestion(blocking):** we don't need this, please remove this file. Using real services in functional tests makes them e2e tests. 

there is a mock server for blockchain in functional tests, and we will need to configure app dynamically by overwriting this value

apps/provider-proxy/src/config/env.config.ts
```diff
@@ -0,0 +1,9 @@
+import { z } from "zod";
+
+export const envSchema = z.object({
+  REST_API_NODE_URL: z.string().url()
+});
+
+export const envConfig = envSchema.parse(process.env);
```
stalniy: **suggestion(blocking):**  please move this into DI, to get rid of side effect in the ES module

apps/provider-proxy/src/utils/schema.ts
```diff
@@ -21,12 +19,10 @@ export const providerRequestSchema = z.object({
...
   url: z.string().url(),
-  network: z.enum(netConfig.getSupportedNetworks() as [SupportedChainNetworks]).describe("Blockchain network"),
   providerAddress: z.string().refine(isValidBech32Address, "is not bech32 address").describe("Bech32 representation of provider wallet address"),
-  chainNetwork: z
-    .enum(netConfig.getSupportedNetworks() as [SupportedChainNetworks])
-    .describe('Deprecated blockchain network. Use "network" instead.')
-    .optional(),
+  // Deprecated fields for backward compatibility
+  network: z.string().describe("Deprecated blockchain network parameter - no longer used").optional(),
+  chainNetwork: z.string().describe("Deprecated blockchain network parameter - no longer used").optional(),
```
stalniy: **suggestion(blocking):** no need to keep them, since old clients will provide this value and we anyway ignore it. zod doesn't throw error if there are extra fields passed in object it validates

apps/provider-proxy/src/container.ts
```diff
@@ -14,14 +13,12 @@ export function createContainer() {
...
-      // for some reason when those libraries are used I receive MockSocket instead of TLSSocket
-      // @see https://github.com/mswjs/msw/discussions/2416
-      return process.env.TEST_CHAIN_NETWORK_URL || netConfig.getBaseAPIUrl(network);
-    },
+    // TEST_CHAIN_NETWORK_URL is hack for functional tests
+    // there is no good way to mock external server in nodejs
+    // both nock and msw do not work well when I need to use low level API like X509 certificate validation
+    // for some reason when those libraries are used I receive MockSocket instead of TLSSocket
+    // @see https://github.com/mswjs/msw/discussions/2416
+    process.env.TEST_CHAIN_NETWORK_URL || envConfig.REST_API_NODE_URL,
```
stalniy: **suggestion(blocking):** I guss you can remove `TEST_CHAIN_NETWORK_URL` after moving `envSchema.parse()` into `createContainer`. Then you can change `REST_API_NODE_URL` inside functional tests. This could even be passed as a parameter to `startServer` in `proxyServer.ts`.

Then you can also remove this big comment

apps/provider-proxy/test/functional/provider-proxy-http.spec.ts
```diff
@@ -14,10 +14,6 @@ describe("Provider HTTP proxy", () => {
   const network: SupportedChainNetworks = "sandbox-2";
   const ONE_HOUR = 60 * 60 * 1000;
 
-  beforeAll(async () => {
-    await startServer();
-  });
-
   afterAll(async () => {
```
stalniy: **question(blocking):** should this be changed to `afterEach`? Since you start a server in every test, you should also stop it after each test

apps/provider-proxy/test/setup-functional-env.ts
stalniy, line 1: Please remove this file, we don't need it

stalniy (changes requested):
there are few minor issues which I'd like to fix but I like how it looks 👍

apps/provider-proxy/env/.env.functional.test
jzsfkzm: Uh, my bad, removed this.

apps/provider-proxy/src/config/env.config.ts
jzsfkzm: Moved it, thanks.

apps/provider-proxy/env/.env.functional.test
jzsfkzm: Now tests are failing due to this value missing from env @stalniy, how should we proceed?

stalniy: I will take a look

stalniy (dismissed):

stalniy:
I configured env variables for this to work properly

PR #2228, opened 2025-11-16
fix: drop deprecated prop so no JS error is shown

### LINKED ISSUE #2105: Unselecting service in logs causes runtime exception

jzsfkzm:
closes #2105

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed menu anchoring for select-style checkboxes so dropdowns now rely on default anchoring. This improves dropdown positioning and visual alignment across different layouts and screen sizes, reducing occasional misplacement of menu content.

### DIFF
--- a/apps/deploy-web/src/components/shared/SelectCheckbox.tsx
+++ b/apps/deploy-web/src/components/shared/SelectCheckbox.tsx
@@ -17,7 +17,6 @@ const MenuProps = {
       width: 250
     }
   },
-  getContentAnchorEl: null,
   anchorOrigin: {
     vertical: "bottom" as const,
     horizontal: "center" as const


apps/deploy-web/src/components/shared/SelectCheckbox.tsx
```diff
@@ -17,7 +17,6 @@ const MenuProps = {
       width: 250
     }
   },
-  getContentAnchorEl: null,
```
stalniy, line 20: **question(blocking):** @jzsfkzm could you please explain a bit why we need to remove it? I looked into MenuProps and I don't see this property but there is `anchorEl`. Should we use it instead?

stalniy (changes requested):

apps/deploy-web/src/components/shared/SelectCheckbox.tsx
jzsfkzm: > **question(blocking):** @jzsfkzm could you please explain a bit why we need to remove it? I looked into MenuProps and I don't see this property but there is `anchorEl`. Should we use it instead?

Sure! With this prop being present, when you click the `Select` for selecting services, there's a JS warning thrown in Developer Console:
Screenshot 2025-11-17 at 11 08 23
This is not showing in prod for some reason though. Anyway, [docs for 5.x suggest just to remove the prop](https://v5.mui.com/material-ui/migration/v5-component-changes/#remove-getcontentanchorel-prop), and `anchorOrigin` and `transformOrigin` are well enough for positioning the menu.

stalniy (approved):

PR #2255, opened 2025-11-19
refactor: reduces usage of process.env

stalniy:
## Why

To reduce usage of process.env and amount of side effects. This will allow us to provide app config not from process.env but also from other sources

## Summary by CodeRabbit

* **Refactor**
  * Migrated configuration management system from direct environment variable access to dependency injection pattern for improved maintainability and testability.
  * Consolidated environment configuration handling across core, billing, deployment, authentication, and other modules.
  * Restructured configuration providers to use container-based injection for consistency.


apps/api/src/middlewares/privateMiddleware.ts
```diff
@@ -1,11 +1,14 @@
...
+import { container } from "tsyringe";
 
-import { env } from "@src/utils/env";
+import { CORE_CONFIG } from "@src/core/providers/config.provider";
 
 export async function privateMiddleware(c: Context, next: Next) {
-  if (!env.SECRET_TOKEN) {
+  const SECRET_TOKEN = container.resolve(CORE_CONFIG).SECRET_TOKEN;
+
+  if (!SECRET_TOKEN) {
```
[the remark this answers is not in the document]
stalniy: This is side effect which I'd like to avoid

apps/api/src/auth/services/user-auth-token/user-auth-token.service.ts
```diff
@@ -1,14 +1,20 @@
...
 import { kvStore } from "@src/middlewares/userMiddleware";
-import { env } from "@src/utils/env";
 import { getJwks, useKVStore, verify, VerifyRsaJwtEnv } from "@src/verify-rsa-jwt-cloudflare-worker-main";
+import { AuthConfigService } from "../auth-config/auth-config.service";
 
 @singleton()
 export class UserAuthTokenService {
+  readonly #authConfigService: AuthConfigService;
+
+  constructor(authConfigService: AuthConfigService) {
```
ygrishajev, line 11: why is this like so instead of regular
```suggestion
  constructor(private readonly authConfigService: AuthConfigService) {
```

is this to use # notation? I assume DI works this way too?

apps/api/src/billing/providers/config.provider.ts
```diff
@@ -1,11 +1,16 @@
...
-export const BILLING_CONFIG = "BILLING_CONFIG";
+export const BILLING_CONFIG: InjectionToken<BillingConfig> = Symbol("BILLING_CONFIG");
 
-container.register(BILLING_CONFIG, { useValue: config });
+container.register(BILLING_CONFIG, {
+  useFactory: instancePerContainerCachingFactory(c => envSchema.parse(c.resolve(RAW_APP_CONFIG)))
+});
 
-export type BillingConfig = typeof config;
+export type { BillingConfig };
```
ygrishajev, line 14: non-blocking: imo either this or class service approach should stay eventually.

ygrishajev (approved):

apps/api/src/auth/services/user-auth-token/user-auth-token.service.ts
stalniy: hashbang private fields do not work with short notation in constructor.

Issue #1846, opened 2025-08-26
support multiple handlers for single domain event


stalniy:
## Why

Currently domain event supports only single job handler but there may be cases when we will need to run multiple different job handlers. It makes sense because if job handler fails, it will be restarted and we want every handler to be idempotent, and restarted separately

## What

Implement support for multiple domain event handlers. This is possible with pgboss publish/subscribe

PR #2274, opened 2025-11-25
feat(billing): adds balance check job handler

ygrishajev:
refs #1779

## Summary by CodeRabbit

* **New Features**
  * Endpoints to mark/get a user’s default payment method; Stripe webhook route; Wallet reload scheduling service and immediate/singleton scheduling.

* **Improvements**
  * Local+remote default-payment-method sync, idempotent payment intent support, memoized fiat balance view, richer auto-reload validation and scheduling, handlers now receive job metadata, transactional helpers and type utilities.

* **Bug Fixes / DB**
  * Enforced not-null and new is_default column with unique partial index; removed obsolete wallet auto-reload fields.

* **Tests & Docs**
  * Extensive unit tests for reload flows and README for wallet reload logic.


apps/api/drizzle/0023_sad_adam_warlock.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "user_wallets" ALTER COLUMN "user_id" SET NOT NULL;
```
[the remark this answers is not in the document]
ygrishajev: no, this is not the case. user wallet is required to have a user id. in fact there is no wallets without user id in prod, so this is a fix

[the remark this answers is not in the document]
ygrishajev: user_id can be null for userSettings, not user_wallets. although we are dropping anonymous users support shortly anyway

apps/api/src/billing/services/stripe/stripe.service.ts
```diff
@@ -113,29 +119,97 @@ export class StripeService extends Stripe {
...
+  }
+
+  async hasPaymentMethod(paymentMethodId: string, user: UserOutput): Promise<boolean> {
+    const paymentMethod = await this.paymentMethods.retrieve(paymentMethodId);
+    const customerId = typeof paymentMethod.customer === "string" ? paymentMethod.customer : paymentMethod.customer?.id;
+
+    return customerId === user.stripeCustomerId;
+  }
+
+  @WithTransaction()
```
stalniy, line 161: **issue(non-blocking):** transaction around remote api calls. This may make our postgres instance slow to respond if it uses all available connections. We had the same issue with start trial. But I assume this is a temp solution, that's why it's non-blocking

apps/api/src/billing/repositories/payment-method/payment-method.repository.ts
```diff
@@ -66,6 +67,64 @@ export class PaymentMethodRepository extends BaseRepository<Table, PaymentMethod
...
+  async createAsDefault(input: Omit<PaymentMethodInput, "id" | "isDefault">) {
+    return this.ensureTransaction(async tx => {
+      const id = uuidv4();
+      const [output] = await Promise.all([
+        this.create({
+          ...input,
+          isDefault: true,
+          id
+        }),
+        this.#unmarkAsDefaultExcluding(id, tx)
```
stalniy, line 111: **issue(non-blocking):** this is concurrently unsafe code. You can end up with 2 `isDefault = true` records in db because newly created record is not visible to the rest of the transactions and if 2 execute in parallel and `unmarkAsDefaultExcluding` executes first in both and then `this.create`, we will have 2 `isDefault: true` payment methods.

Not sure whether it's a real case, that's why non-blocking

apps/api/src/billing/services/wallet-balance-reload-check/wallet-balance-reload-check.handler.ts
```diff
@@ -11,7 +43,192 @@ export class WalletBalanceReloadCheckHandler implements JobHandler<WalletBalance
...
+            }
+          ])
+        )
+      );
+    } catch (error) {
+      if (isHttpError(error)) {
+        return Err({
+          event: "ERROR_RETRIEVING_DEFAULT_PAYMENT_METHOD",
+          message: error.message,
+          source: error.source
```
stalniy: **question(non-blocking):** does it make sense to keep the whole error in result? Then it can be passed to logger and we will see all attached details to the error

apps/api/src/core/services/job-queue/job-queue.service.spec.ts
```diff
@@ -51,16 +52,28 @@ describe(JobQueueService.name, () => {
...
         retryBackoff: true,
-        retryDelayMax: 300,
-        retryLimit: 5
+        retryDelayMax: 5 * 60,
+        retryLimit: 5,
+        policy: undefined
       });
     });
+
+    it("throws error when multiple handlers register for the same queue", async () => {
```
stalniy, line 68: FYI: https://github.com/akash-network/console/issues/1846

apps/api/src/billing/services/wallet-balance-reload-check/wallet-balance-reload-check.handler.ts
```diff
@@ -11,7 +43,192 @@ export class WalletBalanceReloadCheckHandler implements JobHandler<WalletBalance
...
+
+    try {
+      if (resources.walletSetting.autoReloadThreshold >= resources.balance) {
+        await this.stripeService.createPaymentIntent({
+          customer: resources.user.stripeCustomerId,
+          payment_method: resources.paymentMethod.id,
+          amount: resources.walletSetting.autoReloadAmount,
+          currency: "usd",
+          confirm: true,
+          idempotencyKey: `${WalletBalanceReloadCheck.name}.${resources.job.id}`
```
stalniy: **question(non-blocking):** Do I understand correctly that request is idempotent per 1 job? 

If so, then in case of an error we may double charge, right? because the cycle is long: payment intent -> stripe webhook -> blockchain tx.

For example, when payment intent succeeds but scheduling of the next job fails. Then this job will be retried by pgboss and will create another payment intent

stalniy: So, [here](https://github.com/akash-network/console/pull/2274/files#diff-49160504661aa33039404a1f85f7be13f55a48cc4ea29e7c3fd22627f6113c3bR187) we need additional check for transaction state. Whether we received payment from stripe and it was added to the balance.

I assume it will be not easy since we do not have transactions in local db and need to add them first

stalniy (dismissed):

apps/api/src/core/services/job-queue/job-queue.service.spec.ts
ygrishajev: this was added by cursor. removing

apps/api/src/billing/services/wallet-balance-reload-check/wallet-balance-reload-check.handler.ts
ygrishajev: if you try to charge with the same `idempotencyKey` stripe wouldn't double charge

apps/api/src/billing/services/stripe/stripe.service.ts
ygrishajev: I'm not sure there's a good solution to this. Rolling back may as well fail. What do you think if we just add a timeout to the stripe call?

apps/api/src/billing/http-schemas/wallet.schema.ts
```diff
@@ -9,13 +9,6 @@ const WalletOutputSchema = z.object({
   createdAt: z.coerce.date().nullable().openapi({})
 });
 
-const ThreeDSecureAuthSchema = z.object({
```
baktun14, line 12: question: these we not used?

ygrishajev: yup

apps/api/src/billing/services/stripe-webhook/stripe-webhook.service.ts
```diff
@@ -140,17 +141,26 @@ export class StripeWebhookService {
...
 
-    await this.paymentMethodRepository.create({
-      userId: user.id,
-      fingerprint,
-      paymentMethodId: paymentMethod.id
-    });
+    const count = await this.paymentMethodRepository.countByUserId(user.id);
+    const isDefault = count === 0;
+
+    assertIsPayingUser(user);
```
baktun14, line 147: question: why are we asserting if the user is paying in the webhook? we also use payment methods for the free trial validation

apps/api/src/billing/services/stripe/stripe.service.ts
```diff
@@ -113,29 +119,100 @@ export class StripeService extends Stripe {
...
+        expand: ["invoice_settings.default_payment_method"]
+      }),
+      this.paymentMethodRepository.accessibleBy(ability, "read").findDefaultByUserId(user.id)
+    ]);
+
+    assert(!customer.deleted, 402, "Payment account has been deleted");
+
+    const remote = customer.invoice_settings.default_payment_method as Stripe.PaymentMethod;
+
+    if (remote && local) {
```
baktun14: comment: technically they should always both be there, maybe we should throw an error or a log so we can spot if something is wrong?

apps/api/src/billing/services/stripe-webhook/stripe-webhook.service.ts
ygrishajev: L125: user is looked up by `customerId`. So this is more of a type safety.

apps/api/src/billing/services/wallet-balance-reload-check/README.md
```diff
@@ -0,0 +1,183 @@
...
+**Why calculate costs for 7 days?**
+
+- Provides a meaningful projection window for deployment costs
+- Ensures reloads cover a full week of operations for deployments with auto top-up enabled
+- Balances UX (users don't get charged too frequently) with transaction costs
+- Reduces the number of payment transactions while maintaining adequate coverage
+
+**Why reload at 25% threshold?**
+
+- Reloads when balance can only cover less than ~1.75 days (25% of 7 days)
```
baktun14, line 37: Ah that's what I was saying every 1.75 days makes sense :)

baktun14 (dismissed):

apps/api/src/deployment/services/draining-deployment/draining-deployment.service.ts
```diff
@@ -84,4 +84,52 @@ export class DrainingDeploymentService {
...
+        if (predictedClosedHeight && predictedClosedHeight >= currentHeight && predictedClosedHeight <= targetHeight) {
+          const blocksNeeded = targetHeight - currentHeight;
+          const amountNeeded = Math.floor(blockRate * blocksNeeded);
+          totalAmount += amountNeeded;
+        }
+      }
+    }
+
+    return totalAmount;
+  }
```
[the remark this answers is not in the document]
ygrishajev: this method is called `calculateAllDeploymentCostUntilDate` which hints that it should calculate costs needed to run these deployments from now to target date. Also `WalletBalanceReloadCheckHandler.#tryToReload` method calculates the amount to charge based on the diff between the result of the method in question and user balance. So user balance is expected to contain funds sufficient to run these deployments. To me this looks correct. Can you spot any issues in this logic?

apps/api/src/billing/services/stripe/stripe.service.ts
```diff
@@ -113,29 +119,108 @@ export class StripeService extends Stripe {
...
+    const newLocal = await this.paymentMethodRepository.accessibleBy(ability, "create").createAsDefault({
+      userId: user.id,
+      fingerprint,
+      paymentMethodId
+    });
+
+    await this.markRemotePaymentMethodAsDefault(paymentMethodId, user);
+
+    return { ...remote, validated: newLocal.isValidated };
+  }
```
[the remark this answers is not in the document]
ygrishajev: @baktun14 I'm going to ignore this comment for now as it's not introduced in this PR. fingerprint is not null in the db. However the question is if this is expected?

PR #2276, opened 2025-11-26
fix: enforce stating security in routing

jzsfkzm:
refs #1701

## Release Notes

* **Refactor**
  * Centralized route creation through an internal wrapper to enforce explicit security declarations on all API routes
  * Updated public API endpoints to use standard OpenAPI path parameter format (e.g., `{id}` instead of `:id`)
  * Made numerous endpoints explicitly accessible without authentication requirements
  * Added structured parameter validation for certain request handlers


apps/api/src/core/lib/create-route/create-route.ts
```diff
@@ -0,0 +1,11 @@
...
+// eslint-disable-next-line no-restricted-imports
+import { createRoute as createOpenApiRoute } from "@hono/zod-openapi";
+
+export function createRoute<
+  R extends Omit<RouteConfig, "security"> & {
+    security: Required<RouteConfig>["security"];
+  }
+>(routeConfig: R) {
+  return createOpenApiRoute(routeConfig);
+}
```
stalniy, line 11: **suggestion:** probably would be good to move it into `lib` folder, we have it in core and this is not really a service, more a helper. However, I also put helper functions in `services` folder but this is probably wrong :)

apps/api/.eslintrc.js
```diff
@@ -1,5 +1,19 @@
...
   extends: [require.resolve("@akashnetwork/dev-config/.eslintrc.ts")],
+  rules: {
+    "no-restricted-imports": [
+      "error",
+      {
+        paths: [
+          {
+            name: "@hono/zod-openapi",
+            importNames: ["createRoute"],
+            message: "Import createRoute from '@src/core/services/create-route/create-route' instead to enforce security declaration."
```
stalniy: **praise:** nice!

apps/api/src/core/lib/create-route/create-route.ts
jzsfkzm: Funny thing is I put it there first but then I saw very few things there so I moved it to `services`. But correct, it's not a service. It's in `lib` now.

apps/api/.eslintrc.js
jzsfkzm: 🙇

stalniy (approved):

PR #2325, opened 2025-12-05
refactor: converts PricingProvider into a hook

stalniy:
## Why

To improve perf and reduce app re-renderings

## Summary by CodeRabbit

* **Refactor**
  * Pricing moved to a direct hook API and denomination resolution made network-aware for consistent price/conversion displays.
  * Services now use explicit denom configuration so wallet balances and active deployments show correct token balances.

* **Tests**
  * Added unit tests covering pricing calculations and conversion helpers.

* **Chores**
  * Updated build target and improved test wallet setup.


apps/deploy-web/src/utils/priceUtils.ts
```diff
@@ -1,13 +1,18 @@
...
-import { getUsdcDenom } from "@src/hooks/useDenom";
+import { READABLE_DENOMS, UAKT_DENOM, USDC_IBC_DENOMS } from "@src/config/denom.config";
+import networkStore from "@src/store/networkStore";
 import { averageDaysInMonth } from "./dateUtils";
 import { denomToUdenom } from "./mathHelpers";
 
 export const averageBlockTime = 6.098;
 
+export const getUsdcDenom = (networkId?: NetworkId): string => {
+  return USDC_IBC_DENOMS[networkId ?? (networkStore.selectedNetworkId as keyof typeof USDC_IBC_DENOMS)] as string;
```
baktun14, line 13: question: should we fallback to the mainnet usdc denom in case selectedNetworkId is not set? Shouldn't happen but in case

baktun14 (approved):

apps/deploy-web/src/utils/priceUtils.ts
stalniy: I just moved this function from another file here, to resolve circular dependency and added optional `networkId parameter to get rid of `networkStore` import in this file later. So, if we want to have a default it should be on `networkStore.selectedNetworkId` level.

I'd like to postpone this decision since I don't have a good answer on this

PR #2394, opened 2025-12-19
chore(dx): ensures DEPLOYMENT_ENV is loaded when it's an empty string

ygrishajev:
## Summary by CodeRabbit

* **Chores**
  * Treats an empty DEPLOYMENT_ENV value as unset so local environment is selected correctly.
  * Ensures local environment file from the parent directory is loaded before environment-specific files.
  * Continues loading environment-specific and default env files in the existing order to preserve compatibility.

### DIFF
--- a/packages/env-loader/src/index.js
+++ b/packages/env-loader/src/index.js
@@ -26,9 +26,14 @@ const config = path => {
   }
 };
 
+if (process.env.DEPLOYMENT_ENV === "") {
+  delete process.env.DEPLOYMENT_ENV;
+}
+
 if (!process.env.DEPLOYMENT_ENV) {
   config("../../.env.local");
 }
+
 config(`env/.env.${process.env.DEPLOYMENT_ENV || "local"}`);
 config(`env/.env.${process.env.NETWORK}`);
 config("env/.env");


packages/env-loader/src/index.js
```diff
@@ -26,9 +26,15 @@ const config = path => {
   }
 };
 
+if (process.env.DEPLOYMENT_ENV === "") {
+  delete process.env.DEPLOYMENT_ENV;
+}
```
stalniy, line 31: **question:** what is the purpose of deleting it if it's an empty string?

packages/env-loader/src/index.js
```diff
@@ -26,9 +26,15 @@ const config = path => {
...
+
 if (!process.env.DEPLOYMENT_ENV) {
   config("../../.env.local");
+  config(".env.local");
```
stalniy: as far as I know `.env.local` exists only in deploy-web/stats-web. Probably requires a comment because we have too much `.env.local`.

The initial idea was to use single `../../.env.local` which could define global variables for all apps (e.g., NETWORK, DEPLOYMENT_ENV, etc) but it's not picked up by docker because it's in ignore list and we don't mount root folder. maybe it makes sense to to mount that file manually as we do with `packages/docker/.env.sandbox.docker-compose-dev`?

ygrishajev: this would try one or another. I don't really see a reason to complicate rn

stalniy (dismissed):

packages/env-loader/src/index.js
stalniy: it's just that we have implemented unfinished ideas and then forget about them. And add other unfinished ideas :) 

but I agree, not a big deal

packages/env-loader/src/index.js
ygrishajev: docker defaults to an empty string which is then considered a value by dotenv and the loading flow is broken.

PR #2409, opened 2025-12-22
fix(billing): auto credit reload ui and setting update

### LINKED ISSUE #2408: ERROR (88921): Failed to schedule wallet balance reload check
## How to reproduce

- Enabled auto credit reload
- Disable it
- Re-enable it
- Throws the error

baktun14:
closes https://github.com/akash-network/console/issues/2408

## Summary by CodeRabbit

* **New Features**
  * Redesigned Account Overview: updated balance display, skeleton loading, enhanced auto-recharge info, and Add Funds action (disabled when no payment method).

* **Bug Fixes**
  * More robust wallet auto-reload scheduling and job handling with clearer failure/logging and safer cancel/complete behavior.

* **Style**
  * UI spacing and layout tweaks across billing views, sidebar, tabs, buttons, payment-methods card, slider visuals, and pie chart label colors.

* **Accessibility**
  * Date-range clear control moved to a separate, keyboard-accessible button.

### DIFF
--- a/apps/api/src/core/services/job-queue/job-queue.service.ts
+++ b/apps/api/src/core/services/job-queue/job-queue.service.ts
@@ -107,14 +107,16 @@ export class JobQueueService implements Disposable {
         name
       });
     } catch (error) {
-      // Job may already be in a terminal state (completed, cancelled, failed)
-      // This is expected when trying to cancel a job that has already finished
-      this.logger.warn({
-        event: "JOB_CANCEL_FAILED",
-        id,
-        name,
-        error
-      });
+      if (this.isTerminalStateError(error)) {
+        this.logger.warn({
+          event: "JOB_CANCEL_FAILED",
+          id,
+          name,
+          error
+        });
+      } else {
+        throw error;
+      }
     }
   }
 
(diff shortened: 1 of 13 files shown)


apps/api/src/core/services/job-queue/job-queue.service.ts
```diff
@@ -99,21 +99,43 @@ export class JobQueueService implements Disposable {
...
+    try {
+      await this.pgBoss.cancel(name, id);
+      this.logger.info({
+        event: "JOB_CANCELLED",
+        id,
+        name
+      });
+    } catch (error) {
+      // Job may already be in a terminal state (completed, cancelled, failed)
+      // This is expected when trying to cancel a job that has already finished
```
stalniy: **thought(non-blocking):** but it also may fail because of other reason. And this means the error won't be propagated and reported by our alerting stack. This means we need a separate alert for this kind of warning.

Probably better would be to detect whether it's a cancel error and then log a warning but if it's not a cancel error, then we need to re-throw it

apps/api/src/core/services/job-queue/job-queue.service.ts
```diff
@@ -99,21 +99,43 @@ export class JobQueueService implements Disposable {
...
+    try {
+      await this.pgBoss.complete(name, id);
+      this.logger.info({
+        event: "JOB_COMPLETED",
+        id,
+        name
+      });
+    } catch (error) {
+      // Job may already be in a terminal state (completed, cancelled, failed)
+      // This is expected when trying to complete a job that has already finished
```
stalniy: the same as here -> https://github.com/akash-network/console/pull/2409/changes#r2643137971

stalniy (dismissed):

PR #2422, opened 2025-12-26
refactor(billing): migrate wallet reload jobs to singleton policy with bulk cleanup

ygrishajev:
Change wallet balance reload check job management from tracking job IDs
in the database to using pg-boss singleton policy with bulk cleanup.

**Key changes:**
- Change handler policy from "short" to "singleton" to allow multiple
  scheduled jobs while maintaining single active execution
- Remove autoReloadJobId field from wallet-setting schema
- Remove job cancellation from wallet-settings service delete operation
- Add cancelCreatedBy method to JobQueueService for bulk cancellation
  of created jobs by singletonKey
- Update handler to use withCleanup flag when scheduling next check
- Simplify WalletReloadJobService by removing transaction wrapper and
  job ID tracking

This approach allows multiple scheduled jobs to exist while ensuring
only one active job runs at a time, with cleanup handled in bulk
rather than tracking individual job IDs.

Fixes race condition where REST handlers updating wallet settings could
mark an active job as cancelled even though it completes successfully,
causing state inconsistencies.

refs #1779

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved wallet auto-reload job management by simplifying and streamlining the scheduling and cleanup logic.
  * Enhanced consistency in wallet settings data handling with complete object returns.

* **Improvements**
  * Updated wallet reload job scheduling with more reliable singleton-based job handling and explicit cleanup controls.

### DIFF
--- a/apps/api/src/billing/services/wallet-balance-reload-check/wallet-balance-reload-check.handler.ts
+++ b/apps/api/src/billing/services/wallet-balance-reload-check/wallet-balance-reload-check.handler.ts
@@ -21,7 +21,7 @@ type ValidationError = {
 };
 
 type InitializedWallet = Require<Pick<UserWalletOutput, "address">, "address">;
-type ActionableWalletSetting = Pick<WalletSettingOutput, "id" | "userId" | "autoReloadJobId">;
+type ActionableWalletSetting = Pick<WalletSettingOutput, "id" | "userId">;
 
 type Resources = {
   walletSetting: ActionableWalletSetting;
@@ -38,7 +38,7 @@ export class WalletBalanceReloadCheckHandler implements JobHandler<WalletBalance
 
   public readonly concurrency = 10;
 
-  public readonly policy = "short";
+  public readonly policy = "singleton";
 
   #CHECK_INTERVAL_IN_MS = millisecondsInDay;
 
@@ -209,7 +209,7 @@ export class WalletBalanceReloadCheckHandler implements JobHandler<WalletBalance
     try {
       await this.walletReloadJobService.scheduleForWalletSetting(resources.walletSetting, {
         startAfter: this.#calculateNextCheckDate().toISOString(),
-        prevAction: "complete"
+        withCleanup: true
       });
     } catch (error) {
       this.instrumentationService.recordSchedulingError(resources.wallet.address, error);
--- a/apps/api/src/billing/services/wallet-settings/wallet-settings.service.ts
+++ b/apps/api/src/billing/services/wallet-settings/wallet-settings.service.ts
@@ -26,22 +26,14 @@ export class WalletSettingService {
     private readonly walletReloadJobService: WalletReloadJobService
   ) {}
 
-  async getWalletSetting(userId: string): Promise<Omit<WalletSettingOutput, "autoReloadJobId"> | undefined> {
+  async getWalletSetting(userId: string): Promise<WalletSettingOutput | undefined> {
     const { ability } = this.authService;
 
-    const maybeSetting = await this.walletSettingRepository.accessibleBy(ability, "read").findByUserId(userId);
-
-    if (!maybeSetting) {
-      return undefined;
-    }
-
-    const { autoReloadJobId, ...setting } = maybeSetting;
-
-    return setting;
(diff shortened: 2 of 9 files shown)


apps/api/src/billing/services/wallet-balance-reload-check/wallet-balance-reload-check.handler.ts
```diff
@@ -57,6 +57,7 @@ export class WalletBalanceReloadCheckHandler implements JobHandler<WalletBalance
     private readonly instrumentationService: WalletBalanceReloadCheckInstrumentationService
   ) {}
 
+  @WithTransaction()
```
stalniy: **issue:** this decorator is too risky, it wraps the whole function which sends external http calls (talks to stripe and blockchain). We already know from startTrial experience that this may cause issue because it will occupy 1 db connection while external calls are in flight and everything else will either wait or timeout. 1 job instance 1 connection, many job instances many connections will be occupied

stalniy: Probably we need to restrict its usage to repository lazyer only, where we know that possibility of external calls is minimal.

ygrishajev: i need a way to lock a wallet setting while job is in progress. db lock looks like the most reliable and logical approach in this case. startTrial was relying on blockchain transactions and that was the main bottleneck. this job on the contrary runs within 300-600ms.

ygrishajev: even though this is way less problematic imo your comment is valid and i was expecting it :)
we could also use smth like this https://www.npmjs.com/package/redis-semaphore
however it would a bit more complexity. I'd go with a db lock and if it causes issues would try to optimise. I anticipate the relevant feature to be used quite rarely so this case is really low risk

ygrishajev: > Probably we need to restrict its usage to repository lazyer only, where we know that possibility of external calls is minimal.

I can't agree with this. This should be used consciously for sure, but also this is a useful tool if used right

stalniy: we have separate containers for workers and for api, but I guess the postgres connection pool is shared between both. So, it will influence not only workers but also api.

we can discuss whether we can find solution without using lock at all

stalniy: > I can't agree with this. This should be used consciously for sure, but also this is a useful tool if used right

I consider this decorator dangerous because people will put it on top of some method without realizing consequences. Or you can put it consciously  but then when business requirements evolve, somebody will add a remote method call inside deep method without realizing that this method takes part in db transaction. 

There is no tool which could help us to detect this but then debugging and finding issues like this is hard

apps/api/src/billing/services/wallet-settings/wallet-settings.service.ts
```diff
@@ -123,11 +123,7 @@ export class WalletSettingService {
...
   async #arrangeSchedule(prev?: WalletSettingOutput, next?: WalletSettingOutput) {
     if (!prev?.autoReloadEnabled && next?.autoReloadEnabled) {
-      await this.walletReloadJobService.scheduleForWalletSetting(next, { prevAction: "cancel" });
-    }
-
-    if (!next?.autoReloadEnabled && next?.autoReloadJobId) {
-      await this.walletReloadJobService.cancel(next.userId, next.autoReloadJobId);
+      await this.walletReloadJobService.scheduleForWalletSetting(next, { prevAction: "none" });
     }
   }
```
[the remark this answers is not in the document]
ygrishajev: the point is that the job relies on the flag and would complete properly. so there is no need to cancel here and face concurrency issues

stalniy (dismissed):

baktun14 (dismissed):

ygrishajev:
@coderabbitai can you construct a diagram how the feature would work now

baktun14 (dismissed):

PR #2440, opened 2025-12-31
feat: adds react-query automatic proxy package

stalniy:
## Why

to have a way to automatically convert service into react-query observable hooks

## Summary by CodeRabbit

* **New Features**
  * Added a new react-query-proxy package providing a proxy-based integration to useQuery/useMutation for async SDKs, with nested support and customizable query-key mapping.

* **Documentation**
  * Added a README with installation, usage examples, query key/invalidation guidance, and license.

* **Tests**
  * Added comprehensive tests covering key generation, query/mutation flows, nested proxies, caching, and edge cases.

* **Chores**
  * Added package metadata and project configs (linting, testing, TypeScript) and simplified ESLint in the query-sdk package.

### DIFF
--- a/packages/react-query-proxy/README.md
+++ b/packages/react-query-proxy/README.md
@@ -1,0 +1,106 @@
+# React Query Proxy
+
+> Wrap any async service into fully-typed TanStack React Query hooks at runtime.
+> Works with REST, gRPC, XML-RPC, generated clients, or hand-written APIs — no codegen required.
+
+---
+
+## ✨ Why?
+
+Most teams already have an API client / SDK:
+- generated from OpenAPI / gRPC
+- written by hand
+- shared between frontend, backend, scripts, and tests
+
+This library lets you **reuse that SDK directly in React** by turning it into:
+- `useQuery` hooks
+- `useMutation` hooks
+- stable, predictable query keys
+
+All **at runtime**, without generating code or schemas.
+
+---
+
+## 📦 Installation
+
+```bash
+npm install @akashnetwork/react-query-proxy
+```
+
+> Peer dependencies:
+> - `react >= 18`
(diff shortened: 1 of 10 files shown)


packages/react-query-proxy/README.md
```diff
@@ -0,0 +1,106 @@
...
+  alerts: {
+    async list(input?: { page?: number }) {
+      return fetchAlerts(input)
+    },
+
+    async create(input: { name: string }) {
+      return createAlert(input)
+    },
+  },
+}
```
baktun14, line 51: question: how would this be defined? with the auto generation code from this pr? https://github.com/akash-network/console/pull/2437/changes

stalniy: these are 2 conflicting approaches. Here is the key difference:

react-query hooks inference (this PR):
- it doesn't care how you generate SDK, it can be protobuf based sdk, REST based sdk or even self-written service. It just wraps that service and expose react-query hooks for its methods.
- we have clear layer separation: SDK on service layer and this proxy acts as a glue to expose any SDK/service to react world (i.e., in View).

qraft PR:
- works only with OpenAPI defined APIs. Cannot be applied to chain-sdk or self-written services
- no clear separation between view and service layer, since it was written to support react and react SSR. This potentially may cause react & react-query libs to leak into server side bundle (in console-api/etc). IMO, this is code "smell"

stalniy: This is part of built console-api, you can see that it contains react-query code and very likely it also contains react. So, our API depends on react without a good reason to do so. Who knows which surprises this can bring in long term

Screenshot 2026-01-06 at 08 34 58

ygrishajev (approved):

PR #2476, opened 2026-01-12
fix(analytics): filter cookies in Amplitude proxy to prevent request rejections

ygrishajev:
Create custom API route handler for /api/collect that removes all cookies before proxying to Amplitude's API. This prevents Amplitude from rejecting requests due to excessive or irrelevant cookies while maintaining analytics functionality.

- Add /api/collect API route that proxies to Amplitude with cookie filtering
- Rewrite URL path from /api/collect to /2/httpapi to match Amplitude endpoint
- Initialize proxy server once and reuse for better performance
- Remove Next.js rewrite rule as API route takes precedence
image

## Summary by CodeRabbit

* **New Features**
  * Added a server-side API route that proxies external requests and exposes custom API runtime options.

* **Chores**
  * Removed client-side environment validation and the conditional client-side proxy rewrite configuration.
  * Simplified runtime configuration by eliminating the prior env-driven rewrite behavior.

### DIFF
--- a/apps/deploy-web/src/pages/api/collect/index.ts
+++ b/apps/deploy-web/src/pages/api/collect/index.ts
@@ -1,0 +1,57 @@
+import { defineApiHandler } from "@src/lib/nextjs/defineApiHandler/defineApiHandler";
+import type { AppServices } from "@src/services/app-di-container/server-di-container.service";
+
+let amplitudeProxy: ReturnType<AppServices["httpProxy"]["createProxyServer"]> | null = null;
+
+function getAmplitudeProxy(services: AppServices) {
+  if (!amplitudeProxy) {
+    amplitudeProxy = services.httpProxy.createProxyServer({
+      changeOrigin: true,
+      target: "https://api2.amplitude.com",
+      secure: true,
+      autoRewrite: false
+    });
+  }
+
+  return amplitudeProxy;
+}
+
+export default defineApiHandler({
+  route: "/api/collect",
+  async handler({ req, res, services }) {
+    delete req.headers.cookie;
+
+    req.url = "/2/httpapi";
+
+    const proxy = getAmplitudeProxy(services);
+
+    return new Promise<void>((resolve, reject) => {
+      const onProxyRes = (_proxyRes: unknown, _req: unknown, proxyRes: unknown) => {
+        if (proxyRes === res) {
+          proxy.off("error", onError);
+          resolve();
+        }
+      };
+      const onError = (error: Error, _req: unknown, errorRes: unknown) => {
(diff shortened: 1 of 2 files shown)


baktun14 (dismissed):

apps/deploy-web/src/pages/api/collect/index.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export default defineApiHandler({
+  route: "/api/collect",
+  async handler({ req, res, services }) {
+    delete req.headers.cookie;
+
+    req.url = "/2/httpapi";
+
+    req.headers["content-type"] = req.headers["content-type"] || "application/json";
+    req.headers["cf-connecting-ip"] = String(req.headers["cf-connecting-ip"] || req.socket.remoteAddress || "");
```
stalniy: **question(non-blocking):** does amplitude supports such forwarding?

stalniy (dismissed):

apps/deploy-web/src/pages/api/collect/index.ts
```diff
@@ -0,0 +1,53 @@
+import { defineApiHandler } from "@src/lib/nextjs/defineApiHandler/defineApiHandler";
+import type { AppServices } from "@src/services/app-di-container/server-di-container.service";
+
+let amplitudeProxy: ReturnType<AppServices["httpProxy"]["createProxyServer"]> | null = null;
+
+function getAmplitudeProxy(services: AppServices) {
+  if (!amplitudeProxy) {
+    amplitudeProxy = services.httpProxy.createProxyServer({
+      changeOrigin: true,
+      target: "https://api2.amplitude.com",
```
stalniy, line 10: **question(non-blocking):** shouldn't this be in env variable?

stalniy: you can use `new URL(services.privateConfig.AMPLITUDE_PROXY_URL)`, and then:

```ts
const url = new URL(services.privateConfig.AMPLITUDE_PROXY_URL);
url.hostname
url.pathname
```

apps/deploy-web/src/pages/api/collect/index.ts
ygrishajev: I don’t think so. Will remove

stalniy: you can use x-forwarded-for, probably this one should be supported

apps/deploy-web/src/pages/api/collect/index.ts
ygrishajev: This is hardly ever changing. So I decided not to bother

PR #2485, opened 2026-01-13
fix: ignore chrome extension errors sentry

baktun14:
## Summary by CodeRabbit

* **New Features**
  * Show a user-facing warning when a WalletConnect "proposal expired" error occurs, with rate-limiting to avoid repeated notifications.

* **Bug Fixes**
  * Improved error reporting by ignoring specific error patterns (e.g., "Proposal expired") and expanding filters to exclude errors from Chrome and Firefox extensions.

* **Chores**
  * Updated error-tracking configuration to use a revised filtering integration and override default integrations for quieter, more accurate telemetry.

### DIFF
--- a/apps/deploy-web/sentry.client.config.ts
+++ b/apps/deploy-web/sentry.client.config.ts
@@ -2,16 +2,7 @@
 // The config you add here will be used whenever a page is visited.
 // https://docs.sentry.io/platforms/javascript/guides/nextjs/
 
-import type { Event } from "@sentry/nextjs";
-import { init as initSentry, thirdPartyErrorFilterIntegration } from "@sentry/nextjs";
-
-function hasExtensionFrames(event: Event): boolean {
-  return (
-    event.exception?.values?.some(value =>
-      value.stacktrace?.frames?.some(frame => frame.filename?.startsWith("chrome-extension://") || frame.filename?.startsWith("moz-extension://"))
-    ) ?? false
-  );
-}
+import { inboundFiltersIntegration, init as initSentry, thirdPartyErrorFilterIntegration } from "@sentry/nextjs";
 
 initSentry({
   dsn: process.env.NEXT_PUBLIC_SENTRY_DSN,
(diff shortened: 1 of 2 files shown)


apps/deploy-web/sentry.client.config.ts
```diff
@@ -2,20 +2,37 @@
...
+    // Drop errors that originate from browser extensions
+    if (hasExtensionFrames(event)) {
+      return null;
+    }
+    return event;
+  },
   integrations: [
-    eventFiltersIntegration({
-      denyUrls: [/^chrome-extension:\/\//]
-    }),
```
stalniy, line 18: seems that this doesn't work because in defaultIntegrations there is inboundFiltersIntegration (not eventFiltersIntegration). They are basically the same but with different names. That's why eventFiltersIntegration doesn't overwrite options for inboundFiltersIntegration and inboundFiltersIntegration allow chrome-extension exception to be logged.

SO, I'd first try to replace eventFiltersIntegration with deprecated inboundFiltersIntegration and see how it works. 

The thing is that they work differently to what new `beforeSend` does. inboundFiltersIntegration will reject error only if it's top frame (the actual cause) is one of the denyUrls. While the beforeSend currently rejects error in case chrom-extension is specified anywhere in the stacktrace. This potentially may hide errors which we want to log, if they caused by our code but also involve extension code

stalniy: I tested this:

```
.captureEvent({
  level: "error",
  exception: {
    values: [
      {
        type: "TypeError",
        value: "Cannot redefine property: ethereum",
        stacktrace: {
          frames: [
            {
              filename: "chrome-extension://fiikommddbeccaoicoejoniammnalkfa/assets/inject.ts.js",
              function: "Object.defineProperty",
              lineno: 1,
              colno: 51391,
            },
            {
              filename: "https://console.akash.network/_next/static/chunks/pages/_app.js",
              function: "n",
              lineno: 54,
              colno: 5513,
            },
          ],
        },
      },
    ],
  },
})
```

on production and `eventsIntegration` filter picks the last url for comparison, in this test case `app:///_next/static/chunks/pages/_app.js` (it additionally rewrites it, so it doesn't contain domain). This last url is sentry error catching/processing code, so potentially we could have a filter that checks if there is only 1 frame from application and all the rest are third-party frames, then drop it.

Also I tried to throw exception from console and those exceptions are not caught by Sentry. This means that [this exception](https://cloudmos.sentry.io/issues/7185758844/?alert_rule_id=16085896&alert_type=issue&environment=production&notification_uuid=692461a7-00e8-40bb-8c24-940a2b096ae2&project=4504656428204032&referrer=slack) was initiated by our code and actually part of user flow, so would be nice to handle it/ignore it in app if possible

baktun14: Ok I reverted to `inboundFiltersIntegration` to see if it'll catch them or not. Thanks for the feedback

stalniy (dismissed):

PR #2502, opened 2026-01-14
fix: add proper logging for ssr

baktun14:
## Summary by CodeRabbit

* **Bug Fixes**
  * Pages now show clear "not found" messages for missing addresses, blocks, and transactions instead of failing.

* **Chores**
  * Added timeout fallback for feature flag initialization to avoid indefinite waits.
  * Improved logging and observability (including OpenTelemetry) to surface fetch errors and not-found events.
  * Server fetches configured to avoid caching to ensure fresh data.

### DIFF
--- a/apps/stats-web/package.json
+++ b/apps/stats-web/package.json
@@ -22,6 +22,7 @@
     "@date-fns/utc": "^2.1.0",
     "@json2csv/plainjs": "^7.0.4",
     "@nivo/line": "^0.87.0",
+    "@opentelemetry/api": "^1.9.0",
     "@radix-ui/react-icons": "^1.3.0",
     "@radix-ui/react-slot": "^1.0.2",
     "@tanstack/react-query": "^5.67.2",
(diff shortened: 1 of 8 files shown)


apps/stats-web/package.json
```diff
@@ -22,6 +22,7 @@
     "@date-fns/utc": "^2.1.0",
     "@json2csv/plainjs": "^7.0.4",
     "@nivo/line": "^0.87.0",
+    "@opentelemetry/api": "^1.9.0",
```
stalniy, line 25: **issue(non-blocking):** this probably will not work out of the box as for other services.

We probably want to use nextjs/sentry instead of opentelemetry here. Sentry wraps opentelemetry but I don't know how it will work together. Not sure it will work at all, so this requires testing.

Also autoinstrumentation is disabled for console-web and stats-web because of sentry. This requires a bit more investigation

baktun14: I had to add it for this https://github.com/akash-network/console/pull/2502/changes#diff-f237bf7da414f883386e11569032797941c44939b38b4434bbffcdbb681b0a8fR4

What do you suggest?

Issue #1048, opened 2025-03-17
Ensure dependencies are in sync across the workspaces


ygrishajev:
We're regularly having issues with dependencies going out of sync and resolution behaving in an unexpected ways. Most commonly this can be solved with certain tools to ensure dependencies are in sync.

One of them is https://jamiemason.github.io/syncpack/

**Acceptance criteria**:

- confirm the mentioned tool suites our needs or find another one
- set up linting on `postinstall`, `precommit`
- set up linting in CI

ygrishajev:
Example output against the current state of the project from the root:

```bash
$ syncpack lint
Formatting
✘ package.json
✘ packages/database/package.json
✘ packages/dev-config/package.json
✘ packages/docker/package.json
✘ packages/env-loader/package.json
✘ packages/http-sdk/package.json
✘ packages/logging/package.json
✘ packages/net/package.json
✘ packages/network-store/package.json
✘ packages/releaser/package.json
✘ packages/ui/package.json
✘ apps/api/package.json
✘ apps/deploy-web/package.json
✘ apps/indexer/package.json
✘ apps/notifications/package.json
✘ apps/provider-console/package.json
✘ apps/provider-proxy/package.json
✘ apps/stats-web/package.json
Versions
= Default Version Group ========================================================
✘ @akashnetwork/akash-api ^1.3.0 → ^1.4.0 apps/api/package.json > dependencies [HighestSemverMismatch]
✘ @akashnetwork/akash-api ^1.3.0 → ^1.4.0 apps/deploy-web/package.json > dependencies [HighestSemverMismatch]
✘ @akashnetwork/akash-api ^1.3.0 → ^1.4.0 apps/indexer/package.json > dependencies [HighestSemverMismatch]
✘ @akashnetwork/akashjs ^0.10.0 → ??? packages/network-store/package.json > devDependencies [UnsupportedMismatch]
  use syncpack prompt to fix this manually
✘ @akashnetwork/akashjs ^0.10.0 → ??? apps/api/package.json > dependencies [UnsupportedMismatch]
  use syncpack prompt to fix this manually
✘ @akashnetwork/akashjs ^0.10.0 → ??? apps/deploy-web/package.json > dependencies [UnsupportedMismatch]
  use syncpack prompt to fix this manually
✘ @akashnetwork/akashjs @akashnetwork/akashjs → ??? apps/stats-web/package.json > devDependencies [UnsupportedMismatch]
  use syncpack prompt to fix this manually
[...]

PR #2556, opened 2026-01-22
fix: improve Turborepo configuration

anthonyshew:
## Summary

- Add missing `build`, `lint`, and `test` tasks to `turbo.json` for proper caching and parallelization
- Implement transit node pattern so lint/test tasks run in parallel while still invalidating cache when dependency source changes
- Fix root `lint` scripts to delegate to `turbo run lint` instead of running `eslint .` directly
- Remove duplicate dependencies from root `package.json` that already exist in the apps that use them (`drizzle-orm`, `pg`, `nock`, `@interchain-ui/react`)

## Why

The previous configuration was missing task definitions for common scripts (`build`, `lint`, `test`) that exist across multiple packages. This meant:
1. Running `turbo run build/lint/test` wouldn't benefit from Turborepo's caching or parallelization
2. The root `lint` script bypassed Turborepo entirely, defeating the purpose of the monorepo setup
3. Root dependencies were duplicated from apps, adding unnecessary entries to the root package.json

## AI Usage
Chat transcript: https://opncd.ai/share/IwyhgsPI

## Summary by CodeRabbit

* **Chores**
  * Redesigned build system to leverage monorepo orchestration for consistent builds and improved performance across all projects
  * Refactored build, linting, and testing infrastructure with improved task management and caching strategies
  * Updated build and linting scripts for streamlined monorepo workflow
  * Removed unused package dependencies

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,8 +16,8 @@
     "dc:up:dev": "dc up",
     "format": "prettier --write ./*.{js,json} **/*.{ts,tsx,js,json}",
     "indexer:dev": "turbo dev --filter=\"./apps/indexer\"",
-    "lint": "eslint .",
-    "lint:fix": "npm run lint -- --fix",
+    "lint": "turbo run lint",
+    "lint:fix": "turbo run lint -- --fix",
     "prepare": "husky || true",
     "pretty": "prettier --write \"./**/*.{js,jsx,mjs,cjs,ts,tsx,json}\"",
     "safe-install": "script/safe-deps-install.sh",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -58,12 +58,6 @@
       "npm run validate:types -w apps/indexer"
     ]
   },
-  "dependencies": {
```
baktun14, line 61: @stalniy @ygrishajev  do you remember why we had dependencies at the root package?

ygrishajev: coz it was a pain to make sure that they appear in the root node_modules sometimes. If the package added in the root has different version it goes to the workspace's local node_modules. However it depends on some package in the root modules, so then it throws. The proper solution would syncing deps across workspaces so that all modules go to the root/deduped.

This change is potentially risky, need to test every app runs without issues. I guess tests should pick the issue if there's any

ygrishajev:
@anthonyshew could you pls rebase your repo

package.json
stalniy: yes, I spent a day yesterday fighting with package-lock issues. It appeared to be deeply nested package hoisting issue. To fix it, I added vitest and tsup into root devDependencies and it fixed the issue.

So, this hack just helps to solve the issue but we nee to get back to https://github.com/akash-network/console/issues/1048

PR #2559, opened 2026-01-22
fix(billing): add network error recovery for transaction retrieval

baktun14:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction signing/broadcasting resilience with a broader recovery path for transient network errors; adds recovery and success logging while preserving prior behavior for non-network errors.

* **New Features**
  * Public utility in the HTTP SDK to detect retriable network errors (now includes socket cases).

* **Tests**
  * Added unit tests covering transaction recovery for network/socket errors and non-recovery scenarios.

* **Documentation**
  * Expanded comments clarifying network-recovery behavior.

### DIFF
--- a/apps/api/src/billing/lib/batch-signing-client/batch-signing-client.service.ts
+++ b/apps/api/src/billing/lib/batch-signing-client/batch-signing-client.service.ts
@@ -350,15 +350,24 @@ export class BatchSigningClientService {
    *
    * Network errors can occur when the RPC connection is interrupted, but the
    * transaction may have already been successfully included in a block.
+   * Cosmjs wraps network errors in a "fetch failed" error with the actual
+   * error in the `.cause` property.
    *
    * @param error - The error to check.
    * @returns `true` if the error is a retriable network error.
    */
   private isRetriableNetworkError(error: unknown): boolean {
     if (!(error instanceof Error)) return false;
+
     if ("code" in error) {
       return isRetriableError(error as Error & { code: unknown });
     }
+
+    // Cosmjs wraps network errors in "fetch failed" with the actual error in .cause
+    if ("cause" in error && error.cause instanceof Error && "code" in error.cause) {
+      return isRetriableError(error.cause as Error & { code: unknown });
+    }
+
     return false;
   }
 
--- a/apps/api/src/billing/lib/batch-signing-client/batch-signing-client.service.ts
+++ b/apps/api/src/billing/lib/batch-signing-client/batch-signing-client.service.ts
@@ -1,4 +1,5 @@
 import { TxRaw } from "@akashnetwork/chain-sdk/private-types/cosmos.v1beta1";
+import { isRetriableError } from "@akashnetwork/http-sdk";
 import { createOtelLogger } from "@akashnetwork/logging/otel";
 import { sha256 } from "@cosmjs/crypto";
 import { toHex } from "@cosmjs/encoding";
@@ -7,7 +8,7 @@ import type { SigningStargateClient } from "@cosmjs/stargate";
 import { calculateFee, GasPrice } from "@cosmjs/stargate";
 import type { IndexedTx } from "@cosmjs/stargate/build/stargateclient";
 import { Sema } from "async-sema";
-import { ExponentialBackoff, handleWhen, handleWhenResult, retry } from "cockatiel";
+import { ExponentialBackoff, handleWhenResult, retry } from "cockatiel";
 import DataLoader from "dataloader";
 import type { Result } from "ts-results";
 import { Err, Ok } from "ts-results";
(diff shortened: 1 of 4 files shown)


apps/api/src/billing/lib/batch-signing-client/batch-signing-client.service.ts
```diff
@@ -337,6 +380,47 @@ export class BatchSigningClientService {
...
+  /**
+   * Checks if an error is a network-related error that might be recoverable.
+   *
+   * Network errors can occur when the RPC connection is interrupted, but the
+   * transaction may have already been successfully included in a block.
+   *
+   * @param error - The error to check.
+   * @returns `true` if the error appears to be network-related.
+   */
+  private isNetworkError(error: unknown): boolean {
```
stalniy: **suggestion:** please use this one https://github.com/akash-network/console/blob/main/packages/http-sdk/src/utils/createFetchAdapter/createFetchAdapter.ts#L78

Add additional check to `isRetriableError()` for code `UND_ERR_SOCKET`

apps/api/src/billing/lib/batch-signing-client/batch-signing-client.service.ts
```diff
@@ -190,18 +208,43 @@ export class BatchSigningClientService {
...
     }
 
-    const tx = await this.getTx(result.val);
+    const txHash = result.val;
+
+    try {
+      const tx = await this.getTx(txHash);
+
+      if (!tx) {
+        const error = new Error("Failed to sign and broadcast transaction");
```
stalniy: **issue:** this doesn't trigger retry

apps/api/src/billing/lib/batch-signing-client/batch-signing-client.service.ts
```diff
@@ -190,18 +208,43 @@ export class BatchSigningClientService {
       throw result.val;
     }
 
-    const tx = await this.getTx(result.val);
+    const txHash = result.val;
+
+    try {
+      const tx = await this.getTx(txHash);
```
stalniy: **suggestion(blocking):** instead of try/catch just use 

```
const recoveredTx = await this.tryRecoverTransaction(txHash);
```

in logs we will see multiple attemps to getTx, you can even add try /catch inside to additionally log failures

stalniy (changes requested):

apps/api/src/billing/lib/batch-signing-client/batch-signing-client.service.ts
stalniy: [logs](http://node1-elasticsearch-val.tail49ca1e.ts.net:32000/explore?schemaVersion=1&panes=%7B%22ckq%22:%7B%22datasource%22:%22beenf7rks2e4gd%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bnamespace%3D%5C%22prod%5C%22,%20service_name%3D%5C%22console-api-mainnet%5C%22%7D%20%7C%20json%20%7C%20log_traceId%3D%5C%221ef6763d39874fd477130ea05f265823%5C%22%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22beenf7rks2e4gd%22%7D,%22editorMode%22:%22code%22,%22direction%22:%22backward%22%7D%5D,%22range%22:%7B%22from%22:%22now-2d%22,%22to%22:%22now%22%7D%7D%7D&orgId=1)

Screenshot 2026-01-22 at 21 58 30

apps/api/src/billing/lib/batch-signing-client/batch-signing-client.service.ts
```diff
@@ -93,14 +94,21 @@ export class BatchSigningClientService {
...
+   * - Network errors occur during transaction retrieval
+   *
+   * This handles cases where the transaction may have been included in a block
+   * but is not yet indexed, or when the RPC connection fails.
    */
-  private readonly getTxExecutor = retry(
-    handleWhenResult(res => !res),
-    { maxAttempts: 5, backoff: new ExponentialBackoff({ maxDelay: 7_000, initialDelay: 500 }) }
+  private readonly txRecoveryExecutor = retry(
+    handleWhenResult(res => !res).orWhen(err => this.isRetriableNetworkError(err)),
```
stalniy, line 107: **issue(blocking):** in real app, it won't work because cosmjs fails with "fetch failed" error and this error has `.cause` property which explains the fetch failed error. And you need to check `err.cause` here as well

stalniy (changes requested):

apps/api/src/billing/lib/batch-signing-client/batch-signing-client.service.ts
stalniy: check this message for error example -> https://github.com/akash-network/console/pull/2559#discussion_r2718610542

baktun14: Ah good point!

stalniy (approved):

PR #2563, opened 2026-01-23
refactor: adds proper logger and move templates into module

stalniy:
## Why

To improve observability for templates 

## What

1. Replaced functional templates tests with unit ones
2. Added logger to template services
3. Cache templates forever after the first cache read
4. Speeds up search for template by its id

## Summary by CodeRabbit

* **Refactor**
  * Restructured template services with improved dependency injection, logging, and concurrency for more reliable and performant template fetching and serving.
  * Template processing now produces merged category data with an index for stable template lookup.

* **Tests**
  * Extensive new unit tests for template fetching, processing, and gallery behaviors.

* **Chores**
  * Cleaned up legacy test fixtures and sample cache files.


apps/api/src/template/services/template-gallery/template-gallery.service.ts
```diff
@@ -0,0 +1,185 @@
...
+    this.templateFetcher = new TemplateFetcherService(this.templateProcessor, this.#logger, fetch, getOctokit(options.githubPAT), {
+      categoryProcessingConcurrency: options.categoryProcessingConcurrency,
+      templateSourceProcessingConcurrency: options.templateSourceProcessingConcurrency
+    });
+    this.#options = options;
+    this.getTemplatesFromRepo = reusePendingPromise(this.getTemplatesFromRepo.bind(this), { getKey: options => options.repository });
+    this.getTemplateById = reusePendingPromise(this.getTemplateById.bind(this), { getKey: id => id });
+  }
+
+  @Memoize({ ttlInSeconds: Infinity })
```
baktun14, line 44: question: how does that work?

baktun14 (approved):

apps/api/src/template/services/template-gallery/template-gallery.service.ts
stalniy: the intention was that templates will be read from fs and then cached forever ( in `@Memoize` there is a check which checks whether cache expired, there is no bigger value then Infinity, so this cache will never expire).

But I see a potential issue with this. If somebody updates templates, and our api restarts it will try to fetch them in runtime. So, this change with ttl is not enough to prevent runtime rebuild

stalniy: I'll merge this as is and will adjust the logic in a separate PR, just so it will be smaller

PR #2565, opened 2026-01-24
feat: add cache control middleware and improve response handling

baktun14:
## Summary by CodeRabbit

* **New Features**
  * Global response compression, a cache-control middleware, and per-route cache policies.

* **Performance**
  * Short-lived in-memory caching for service lookups and optimized provider resolution to reduce latency.

* **Reliability**
  * 10s request timeouts added for deployment and lease API calls.

* **Chores / Data**
  * Added an addressReference height column and a batched backfill script.

* **Tests**
  * Unit tests covering route cache configuration lookup.

### DIFF
--- a/apps/api/src/core/lib/create-route/create-route.ts
+++ b/apps/api/src/core/lib/create-route/create-route.ts
@@ -16,21 +16,15 @@ export interface ExtendedRouteConfig<R extends RouteConfig> {
 const cacheConfigRegistry = new Map<string, CacheConfig>();
 
 export function getCacheConfig(path: string): CacheConfig | undefined {
-  // Try exact match first
+  // Try exact match first (for static paths)
   if (cacheConfigRegistry.has(path)) {
     return cacheConfigRegistry.get(path);
   }
 
-  // Try pattern matching (for paths with parameters like /v1/providers/{address})
-  for (const [pattern, config] of cacheConfigRegistry.entries()) {
-    const regexPattern = pattern.replace(/\{[^}]+\}/g, "[^/]+").replace(/\//g, "\\/");
-    const regex = new RegExp(`^${regexPattern}$`);
-    if (regex.test(path)) {
-      return config;
-    }
-  }
-
-  return undefined;
+  // Convert Hono's :param format to OpenAPI's {param} format for lookup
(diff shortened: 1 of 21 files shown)


apps/api/src/middlewares/cacheControlMiddleware.ts
```diff
@@ -0,0 +1,45 @@
+import type { Context, Next } from "hono";
+import { createMiddleware } from "hono/factory";
+
+interface CacheConfig {
+  maxAge: number;
+  staleWhileRevalidate?: number;
+}
+
+const CACHE_CONFIGS: Record<string, CacheConfig> = {
```
stalniy: suggestion: would be nice to include this into route definition. Right now this is separate and easily can be misconfigured

apps/api/src/core/lib/create-route/create-route.ts
```diff
@@ -2,10 +2,48 @@ import type { RouteConfig } from "@hono/zod-openapi";
...
+  }
+
+  // Try pattern matching (for paths with parameters like /v1/providers/{address})
+  for (const [pattern, config] of cacheConfigRegistry.entries()) {
+    const regexPattern = pattern.replace(/\{[^}]+\}/g, "[^/]+").replace(/\//g, "\\/");
+    const regex = new RegExp(`^${regexPattern}$`);
+    if (regex.test(path)) {
+      return config;
+    }
+  }
```
stalniy: **thought:** this seems redundant because `ctx.req.routePath` is a routePath, what was specified in `createRoute` object, not the actual `ctx.req.url`

I've tested this for ` path: "/v1/nodes/{network}",` and `ctx.req.routePath` contains `/v1/nodes/:network`

stalniy: so, we need to replace `{` with `:` and `}` with empty string in  `path`. Can be like this:

```ts
export function getCacheConfig(path: string): CacheConfig | undefined {
  // Try exact match first
  if (cacheConfigRegistry.has(path)) {
    return cacheConfigRegistry.get(path);
  }

 const adjustedPath = path.replaceAll('{', ':').replaceAll('}', '')
 return cacheConfigRegistry.get(adjustedPath)
}
```

stalniy (approved):

PR #2566, opened 2026-01-26
refactor: skips parsing JSON cache when serving templates

stalniy:
## Why

Currently combined templates cache size is 4.7Mb. `JSON.parse` is a sync operation and it may block event loop. Because of this liveness helthcheck fails and container is marked as unhealthy

## What

1. Removed obsolete /templates endpoint. It was replaced with /templates-list -> https://github.com/akash-network/console/issues/102
2. Restored templates-list shape. It supposed to return only basic information about templates but unexpectedly for me it serves the whole template data right now (including big readme and other information which we don't need on the list)
3. saves templates cache in kind of [nd-json](https://docs.mulesoft.com/dataweave/latest/dataweave-formats-ndjson) format. To have possibility to selectively parse only parts of the cache file, to reduce blocking time by `JSON.parse`
4. prebuilt serialized response for /templates-list. So, whenever somebody requests templates, API will just serve a string from RAM. Super fast response (50ms). At the moment, it takes 1-3 seconds on production to serve templates from cache. 

Screenshot 2026-01-26 at 08 22 34

## Summary by CodeRabbit

* **New Features**
  * Persistent on-disk template gallery cache and cache-builder to speed listings and detail retrieval.
  * Per-template in-memory caching for faster repeated loads.
* **Refactor**
  * Template list now returns streamlined summaries (id, name, logo, summary); full-templates endpoint removed.
  * Template detail responses standardized as JSON-wrapped data.
  * Internal merging now returns a flat array of categories instead of a wrapped result.
* **Tests**
  * Updated tests to align with array-based gallery and cache behavior.

### DIFF
--- a/apps/api/src/template/services/template-gallery/template-gallery.service.ts
+++ b/apps/api/src/template/services/template-gallery/template-gallery.service.ts
@@ -48,7 +48,7 @@ export class TemplateGalleryService {
   async getCachedTemplateGallery(): Promise<GalleriesCache> {
     if (this.#galleriesCache) return this.#galleriesCache;
 
-    const cacheFileExists = await this.#fs.access(this.#galleriesCachePath, fsConstants.W_OK | fsConstants.R_OK).then(
+    const cacheFileExists = await this.#fs.access(this.#galleriesCachePath, fsConstants.R_OK).then(
       () => true,
       () => false
     );
@@ -102,6 +102,7 @@ export class TemplateGalleryService {
     const serializeTemplatesContent = serializedTemplates.join("\n");
     const content = `${JSON.stringify(metadata)}\n${serializedCategoriesContent}\n${serializeTemplatesContent}`;
 
+    await this.#fs.mkdir(path.dirname(this.#galleriesCachePath), { recursive: true });
     await this.#fs.writeFile(this.#galleriesCachePath, content);
 
     return {
--- a/apps/api/src/template/services/template-gallery/template-gallery.service.ts
+++ b/apps/api/src/template/services/template-gallery/template-gallery.service.ts
@@ -1,12 +1,13 @@
 import { Octokit } from "@octokit/rest";
-import { constants as fsConstants, promises as fsp } from "node:fs";
+import type { promises as fsp } from "node:fs";
+import { constants as fsConstants } from "node:fs";
 import path from "node:path";
+import type z from "zod";
 
-import { Memoize, reusePendingPromise } from "@src/caching/helpers";
-import { LoggerService } from "@src/core";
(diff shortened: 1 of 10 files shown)


apps/api/src/template/services/template-gallery/template-gallery.service.ts
```diff
@@ -17,32 +18,108 @@ type Options = {
...
 
   constructor(logger: LoggerService, fs: FileSystemApi, options: Options) {
     this.#logger = logger;
     this.#fs = fs;
     this.templateProcessor = new TemplateProcessorService();
-    this.templateFetcher = new TemplateFetcherService(this.templateProcessor, this.#logger, fetch, getOctokit(options.githubPAT), {
-      categoryProcessingConcurrency: options.categoryProcessingConcurrency,
-      templateSourceProcessingConcurrency: options.templateSourceProcessingConcurrency
-    });
+    this.templateFetcher = options.githubPAT
```
baktun14, line 35: question: shouldn't we throw if we don't have the PAT?

apps/api/src/template/services/template-gallery/template-gallery.service.ts
```diff
@@ -183,3 +268,24 @@ export interface FileSystemApi {
...
+export interface GalleriesCache {
+  metadata: {
+    templatesRanges: Record<string, [start: number, end: number]>;
+    templatesOffset: number;
+  };
+  categories: string;
+  templates: string;
+}
+
+function buildContentRanges(content: string[]): Record<number, [start: number, end: number]> {
```
baktun14, line 281: question: what does that do?

baktun14 (dismissed):

apps/api/src/template/services/template-gallery/template-gallery.service.ts
stalniy: we don't need GITHUB_PAT in runtime anymore. So, not anymore. It's used only during build time

apps/api/src/template/services/template-gallery/template-gallery.service.ts
stalniy: it maps index of a string in `content` variable to start and end ranges inside concatenated text. So, then we quickly know where template is located and can slice it from `string` and parse only single template json, not everything

apps/api/src/template/services/template-gallery/template-gallery.service.ts
baktun14: Ah ok good

baktun14 (dismissed):

baktun14 (approved):

PR #2602, opened 2026-01-29
refactor: makes /pricing endpoint to be async

stalniy:
## Why

Currently pricing route is completely sync and cannot handle load

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a typo in the pricing endpoint response description.

* **Improvements**
  * Input validation: array pricing requests limited to 1–10 items.
  * Pricing computation now runs asynchronously and processes large batches in time‑bounded chunks.

* **New Features**
  * Middleware added to validate and enforce supported Content-Type headers.
  * New utility to iterate and process arrays in incremental, time‑bounded chunks.

* **Tests**
  * Added unit tests for chunked processing, content‑type middleware, and pricing service behavior.

* **Chores**
  * Removed legacy functional pricing tests.


apps/api/src/utils/array/array.ts
```diff
@@ -1,7 +1,43 @@
...
+
 type Matcher<T> = (a: T, b: T) => boolean;
 
 export function createFilterUnique<T>(matcher: Matcher<T> = (a, b) => a === b): (value: T, index: number, array: T[]) => boolean {
   return (value, index, array) => {
     return array.findIndex(other => matcher(value, other)) === index;
   };
 }
+
+export async function forEachInChunks<T>(
```
ygrishajev, line 12: https://superchargejs.com/docs/3.x/promise-pool we have this in the api. wouldn't it do the same thing?

stalniy: it's a bit different promise-pool is to process tasks concurrently. when you have data and want to process it concurrently. This function allows to process big arrays with concurrency 1, doing synchronous operations only.

Basically it's a version of `.forEach` which doesn't block event loop for more than specified in `maxTimeSpentPerChunk`

ygrishajev (dismissed):

PR #2607, opened 2026-01-29
refactor: replace per-file GitHub API calls with ZIP archive downloads

baktun14:
## Summary
- Replaces hundreds of individual Octokit `repos.getContent` API calls per refresh cycle with 3 ZIP archive downloads from GitHub's CDN (not rate-limited)
- Adds `GitHubArchiveService` that downloads and parses repo archives using `yauzl`, providing `readFile()` and `listDirectory()` via an `ArchiveReader` interface
- Chain registry metadata now fetched from `raw.githubusercontent.com` instead of the GitHub API
- Adds periodic template gallery refresh (default 15-minute TTL) via `TemplateRefreshService`, so new templates appear without redeploying the API
- Extracts `TemplateGalleryService` into a shared DI singleton so the controller and refresh service operate on the same in-memory cache
- Uses a `setTimeout` chain (not `setInterval`) to guarantee no overlapping refreshes and clean disposal on shutdown
- Refresh is a no-op if `GITHUB_PAT` is not configured

### How the refresh works
1. Every 15 minutes (configurable via `TEMPLATE_REFRESH_INTERVAL_SECONDS`), the refresh service calls `TemplateGalleryService.refreshCache()`
2. `refreshCache()` rebuilds the cache file (SHA-based — skips ZIP download if nothing changed), then resets in-memory caches (`#galleriesCache`, `#parsedTemplates`) and clears the `GitHubArchiveService` archive cache to free old entries
3. The next request reads the fresh cache file; requests during refresh continue serving the old in-memory data

### New files
| File | Purpose |
|------|---------|
| `providers/template-gallery.provider.ts` | Shared singleton `TemplateGalleryService` via DI |
| `providers/template-refresh.provider.ts` | Registers `TemplateRefreshService` as `APP_INITIALIZER` with `DisposableRegistry` |
| `services/template-refresh/template-refresh.service.ts` | `setTimeout`-chain periodic refresh with error logging and dispose cleanup |

## Test plan
- [x] TypeScript type checking passes (`npx tsc --noEmit`)
- [x] All 73 template unit tests pass
- [ ] Manual test: start API with `GITHUB_PAT` set, verify "TEMPLATE_REFRESH_REGISTERED" log appears, temporarily set `TEMPLATE_REFRESH_INTERVAL_SECONDS=30` and confirm refresh logs appear

## Summary by CodeRabbit

* **New Features**
  * Archive-based repository retrieval for template content (new archive service)
  * Automatic periodic template cache refresh (configurable via env)

* **Improvements**
  * Template fetching now uses archive reads for more reliable listings and file access
  * Service wiring updated to support cache/refresh lifecycle and simplified gallery sources (legacy linux-server removed)

* **Tests**
  * Tests migrated to archive-based mocks and expanded for archive error scenarios

### DIFF
--- a/apps/api/src/template/services/github-archive/github-archive.service.ts
+++ b/apps/api/src/template/services/github-archive/github-archive.service.ts
@@ -50,7 +50,7 @@ export class GitHubArchiveService {
 
   async #downloadAndParse(owner: string, repo: string, ref: string): Promise<ArchiveReader> {
     const url = `https://github.com/${owner}/${repo}/archive/${ref}.zip`;
-    const response = await fetch(url);
+    const response = await fetch(url, { signal: AbortSignal.timeout(60_000) });
 
     if (!response.ok) {
       throw new Error(`Failed to download archive from ${url}: ${response.status} ${response.statusText}`);
@@ -161,14 +161,18 @@ export class GitHubArchiveService {
     }
   }
 
+  static #normalizePath(value: string): string {
+    return value.replace(/^(?:\.\/|\/)+/, "").replace(/\/+$/, "");
+  }
+
   #createArchiveReader(parsed: ParsedArchive): ArchiveReader {
     return {
       async readFile(path: string): Promise<string | null> {
-        return parsed.files.get(path) ?? null;
+        return parsed.files.get(GitHubArchiveService.#normalizePath(path)) ?? null;
       },
 
       listDirectory(path: string): DirectoryEntry[] {
-        return parsed.directories.get(path) ?? [];
+        return parsed.directories.get(GitHubArchiveService.#normalizePath(path)) ?? [];
       }
     };
   }
--- a/apps/api/src/template/services/github-archive/github-archive.service.ts
+++ b/apps/api/src/template/services/github-archive/github-archive.service.ts
@@ -28,7 +28,8 @@ export class GitHubArchiveService {
   }
 
   async getArchive(owner: string, repo: string, ref: string, fileFilter?: (relativePath: string) => boolean): Promise<ArchiveReader> {
-    const cacheKey = `${owner}/${repo}/${ref}`;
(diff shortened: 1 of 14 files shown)


apps/api/src/template/services/template-fetcher/template-fetcher.service.ts
```diff
@@ -78,74 +79,51 @@ export class TemplateFetcherService {
...
-    }
-
-    return response.data as GithubDirectoryItem[];
+    }));
   }
 
   private async fetchChainRegistryData(chainPath: string): Promise<GithubChainRegistryChainResponse> {
-    const content = await this.fetchFileContent("cosmos", "chain-registry", `${chainPath}/chain.json`);
-    return JSON.parse(content) as GithubChainRegistryChainResponse;
+    const url = `https://raw.githubusercontent.com/cosmos/chain-registry/master/${chainPath}/chain.json`;
```
stalniy: **issue:** it should clone this repository as well. it can fail due to rate limiting

apps/api/src/template/services/github-archive/github-archive.service.ts
```diff
@@ -0,0 +1,101 @@
...
+  async #downloadAndParse(owner: string, repo: string, ref: string): Promise<ArchiveReader> {
+    const url = `https://github.com/${owner}/${repo}/archive/${ref}.zip`;
+    const response = await fetch(url);
+
+    if (!response.ok) {
+      throw new Error(`Failed to download archive from ${url}: ${response.status} ${response.statusText}`);
+    }
+
+    const buffer = await response.arrayBuffer();
+    const zip = await JSZip.loadAsync(buffer);
```
stalniy: **issue:** js-zip is not designed to work with big zip files. we are expecting zip files of >=50Mb

stalniy (dismissed):

PR #2625, opened 2026-01-30
fix: stats-web transaction page crashes and SSR memory leak

baktun14:
## Summary

- **Fix transaction page crashes**: The API returns snake_case field names (e.g. `from_address`, `delegator_address`) in transaction message data, but all frontend message components expect camelCase (`fromAddress`, `delegatorAddress`). Added a recursive `convertKeysToCamelCase` transform in `TxMessageRow` — the single entry point for all transaction message rendering — so every message component receives normalized keys. This is done in stats-web rather than the API to avoid affecting other consumers (deploy-web, provider-console, etc.).
- **Fix `MsgAccountDeposit` crash**: The component accessed `data?.account` but the API field is `id`, causing `JSON.parse(JSON.stringify(undefined))` to throw.
- **Guard against null address in `AddressLink`**: Added early return for falsy `address` prop to prevent `undefined.startsWith()` crashes when message fields are missing.
- **Fix SSR memory leak**: The module-level `QueryClient` singleton was shared across all SSR renders in the Next.js app router. Each render's `useQuery` hooks register observers on this shared instance, and without component lifecycle cleanup on the server, cached state accumulates indefinitely. Replaced with per-instance creation via `useState(() => new QueryClient())` in `CustomProviders`, following [React Query's SSR guidance](https://tanstack.com/query/latest/docs/framework/react/guides/ssr#initial-setup). The Jotai store remains a module-level singleton because `NetworkStore` hooks rely on `useAtom()` reading from the Provider context, requiring the same store instance.

## Test plan

- [ ] Verify stats-web homepage loads without errors at `/`
- [ ] Verify transaction detail pages render message fields correctly (e.g. `fromAddress`, `toAddress`, `delegatorAddress`)
- [ ] Verify `MsgAccountDeposit` transactions render without crashing
- [ ] Verify deploy-web transaction pages still work (API unchanged)
- [ ] Monitor stats-web container memory after deploy — should stabilize instead of growing indefinitely

## Summary by CodeRabbit

* **Bug Fixes**
  * Address links safely render nothing when an address is missing.
  * Transaction message data is normalized for more consistent display.
  * Deposit transaction account display corrected to show the appropriate data.

* **Refactor**
  * Query client now initialized locally to the app for improved initialization behavior.

### DIFF
--- a/apps/stats-web/src/queries/queryClient.ts
+++ b/apps/stats-web/src/queries/queryClient.ts
@@ -1,7 +1,0 @@
-"use client";
-
-import { QueryClient } from "@tanstack/react-query";
-
-const queryClient = new QueryClient();
-
-export { queryClient };
--- a/apps/stats-web/src/components/AddressLink.tsx
+++ b/apps/stats-web/src/components/AddressLink.tsx
@@ -12,6 +12,8 @@ type Props = {
 };
 
 export const AddressLink: React.FunctionComponent<Props> = ({ address }) => {
+  if (!address) return null;
+
   let href = null;
   if (address.startsWith("akashvaloper")) {
     href = UrlService.validator(address);
--- a/apps/stats-web/src/components/layout/CustomProviders.tsx
+++ b/apps/stats-web/src/components/layout/CustomProviders.tsx
@@ -1,8 +1,8 @@
 "use client";
-import React from "react";
+import React, { useState } from "react";
 import { TooltipProvider } from "@akashnetwork/ui/components";
 import { CustomSnackbarProvider } from "@akashnetwork/ui/context";
-import { QueryClientProvider } from "@tanstack/react-query";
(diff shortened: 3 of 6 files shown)


apps/stats-web/src/queries/queryClient.ts
```diff
@@ -1,7 +0,0 @@
-"use client";
```
stalniy, line 1: **thought:** here we have `"use client"` what means that this thing should not be used by SSR

stalniy: @coderabbitai can you explain why queryClient can result in OOM when it's marked as `use client`? nextjs 14.x

stalniy (approved):

apps/stats-web/src/queries/queryClient.ts
stalniy: is pages router app also affected in the same way?

PR #2635, opened 2026-01-30
fix: increase HTTP request timeout from 10s to 30s

baktun14:
## Summary by CodeRabbit

* **Improvements**
  * Increased HTTP request timeout thresholds from 10 to 30 seconds for deployment and lease operations, reducing timeout errors and improving reliability on slower network connections.

### DIFF
--- a/packages/http-sdk/src/deployment/deployment-http.service.ts
+++ b/packages/http-sdk/src/deployment/deployment-http.service.ts
@@ -146,7 +146,7 @@ export class DeploymentHttpService {
           "id.owner": owner,
           "id.dseq": dseq
         },
-        timeout: 10000,
+        timeout: 30000,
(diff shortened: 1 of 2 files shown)


packages/http-sdk/src/deployment/deployment-http.service.ts
```diff
@@ -196,7 +196,7 @@ export class DeploymentHttpService {
     return extractData(
       await this.httpClient.get<DeploymentListResponse>("/akash/deployment/v1beta4/deployments/list", {
         params,
-        timeout: 10000
+        timeout: 30000
```
stalniy, line 199: **question:** does it make sense to retry on timeout instead? waiting for 30s is a way too long

stalniy (approved):

PR #2639, opened 2026-02-01
fix: force user onboarding

baktun14:
## Summary by CodeRabbit

* **New Features**
  * Onboarding redirect/effect and guarded page gating (OnboardingRedirect, Guard, useIsOnboarded).
  * Close Deployment action while waiting for bids.
  * WalletConnectionButtons component for unified wallet controls.

* **Improvements**
  * Consolidated wallet connection UI across pages.
  * Smarter signup/login redirect to route to onboarding when appropriate.
  * Streamlined trial status UI and added logout button in onboarding.

* **Bug Fixes**
  * Analytics event renamed to onboarding_logout and logout clears trial sign‑in flag.


apps/deploy-web/src/pages/_app.tsx
```diff
@@ -60,7 +61,9 @@ const App: React.FunctionComponent<Props> = props => {
               <PaymentPollingProvider>
                 <CertificateProvider>
                   <NavigationGuardProvider>
-                    <Component {...pageProps} />
+                    <OnboardingGuard>
+                      <Component {...pageProps} />
+                    </OnboardingGuard>
```
stalniy: **issue:** this will add unnecessary  re-renderings for routes which don't require this. Causing additional API requests. Right now, templates page sends 2-3 requests for templates with short delay.

baktun14: True, changed it to use the existing hoc

stalniy (dismissed):

stalniy:
I think that we also need to remove `isSignedInWithTrial` from localStorage. This thing can probably be checked in a different way. Because it can result in false-positives

PR #2647, opened 2026-02-02
fix: switch container build target to production

stalniy:
## Why

Since we don't use nginx inside container in production (ports are mapped directly to our backend services), we should not build container which includes nginx. This will make them smaller, CI faster and resource utilization better

## Summary by CodeRabbit

* **Chores**
  * Updated Docker build configuration for api, provider-proxy, and deploy-web services to use production build targets
  * Removed analytics collection endpoint configuration from reverse proxy setup

### DIFF
--- a/apps/deploy-web/nginx.conf
+++ b/apps/deploy-web/nginx.conf
@@ -35,35 +35,5 @@ http {
             proxy_buffer_size 16k;
             proxy_cookie_path / "/; HTTPOnly; Secure";
         }
-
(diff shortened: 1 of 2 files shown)


stalniy:
removed proxy to amplitude because added it via k8s -> https://github.com/akash-network/helm-charts/pull/517

ygrishajev (approved):

baktun14 (approved):

PR #2663, opened 2026-02-03
fix: update template retrieval and response handling

baktun14:
## Summary by CodeRabbit

* **Bug Fixes**
  * Template page metadata now only includes preview images when a logo is present.
  * Legacy template redirects now preserve original query parameters when forwarding requests.

* **Chores**
  * API template responses standardized to a consistent { data: ... } envelope across services and client SDK.

### DIFF
--- a/apps/api/src/routers/legacyRouter.ts
+++ b/apps/api/src/routers/legacyRouter.ts
@@ -191,14 +191,10 @@ legacyRouter.get("/v1/templates-list.json", async c => {
   return c.redirect("/v1/templates-list", redirectStatusCode);
 });
 
-legacyRouter.get("/v1/templates/:id", async (c, next) => {
+legacyRouter.get("/v1/templates/:id.json", async c => {
   const id = c.req.param("id");
-  if (!id?.endsWith(".json")) {
-    return next();
-  }
-  const templateId = id.slice(0, -5); // Remove ".json" suffix
   const queryString = new URL(c.req.url).search;
-  return c.redirect(`/v1/templates/${templateId}${queryString}`, redirectStatusCode);
+  return c.redirect(`/v1/templates/${id}${queryString}`, redirectStatusCode);
 });
 
 legacyRouter.post("/user/subscribeToNewsletter", proxyToV1);
--- a/apps/api/src/routers/legacyRouter.ts
+++ b/apps/api/src/routers/legacyRouter.ts
@@ -191,10 +191,11 @@ legacyRouter.get("/v1/templates-list.json", async c => {
   return c.redirect("/v1/templates-list", redirectStatusCode);
 });
 
-legacyRouter.get("/v1/templates/:id.json", async c => {
+legacyRouter.get("/v1/templates/:id{.+\\.json}", async c => {
   const id = c.req.param("id");
+  const templateId = id.slice(0, -5); // Remove ".json" suffix
   const queryString = new URL(c.req.url).search;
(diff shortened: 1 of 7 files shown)


apps/api/src/routers/legacyRouter.ts
```diff
@@ -191,9 +191,14 @@ legacyRouter.get("/v1/templates-list.json", async c => {
   return c.redirect("/v1/templates-list", redirectStatusCode);
 });
 
-legacyRouter.get("/v1/templates/:id.json", async c => {
+legacyRouter.get("/v1/templates/:id", async (c, next) => {
```
stalniy: **question(blocking):** this basically registers 2 routes with the same url handling (one using openapi createRoute and another using legacyRouter). does hono call both? which one is called first? (I guess order matters in this case)

packages/http-sdk/src/template/template-http.service.ts
```diff
@@ -39,7 +39,8 @@ export class TemplateHttpService {
   }
 
   async findById(id: string): Promise<TemplateOutput> {
-    return extractData(await this.#httpClient.get<TemplateOutput>(`/v1/templates/${id}.json`));
+    const response = extractData(await this.#httpClient.get<{ data: TemplateOutput }>(`/v1/templates/${id}.json`));
+    return response.data;
```
stalniy, line 43: @ygrishajev this is what I was saying that if we add `.data` we should return `.data` to the end user and not do this inside sdk functions because then we loose the point of adding this `.data`. Obviously we won't break contract on network level but by adding some metadata we will break contract on sdk level.

Eventually some sdk methods will return `{ data: resource, metadata }` and others will return just `resource` because of such normalization. in `ApiHttpService` we even had special method to do this extraction - `extractApiData`.

stalniy, line 43: let's discuss this first

stalniy (changes requested):

packages/http-sdk/src/template/template-http.service.ts
stalniy: let's keep this .data as is for now.

baktun14: Should I cancel this pr?

apps/api/src/routers/legacyRouter.ts
baktun14: Hono captures .json as part of the :id parameter in /v1/templates/:id.json, so the redirect goes back to itself. The previous version creates a redirect loop.

baktun14: Good catch — you're right that the previous approach registered two routes with the same /:id pattern, which was problematic. In Hono, the first matching route runs first, so 
  the legacy router (mounted earlier in rest-app.ts) would intercept every /v1/templates/:id request before the OpenAPI handler, relying on next() to fall through for non-.json 
  requests. That's fragile and order-dependent.                                                                                                                                

  I've updated the route to use Hono's regex parameter syntax instead:

  legacyRouter.get("/v1/templates/:id{.+\\.json}", async c => {

  This way the legacy route only matches URLs ending in .json (e.g. /v1/templates/foo.json), and non-.json requests like /v1/templates/foo go directly to the OpenAPI handler
  without ever touching the legacy router. No duplicate patterns, no next() fallthrough, no dependency on registration order.

stalniy (approved):

PR #2672, opened 2026-02-05
fix(user): shows friendly auth0 message and skips username update

ygrishajev:
## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented unintended username changes when updating existing user records during registration/upsert flows.
  * Signup error messaging now surfaces the backend's friendliest message first, with previous fallbacks used only when needed.

* **Tests**
  * Updated/added tests to verify username preservation and that signup returns friendly backend messages on error.

### DIFF
--- a/apps/api/src/user/repositories/user/user.repository.ts
+++ b/apps/api/src/user/repositories/user/user.repository.ts
@@ -66,13 +66,14 @@ export class UserRepository extends BaseRepository<ApiPgTables["Users"], UserInp
       );
   }
 
-  async upsertByUserId(data: UserInput): Promise<UserOutput> {
+  async upsertOnExternalIdConflict(data: UserInput): Promise<UserOutput> {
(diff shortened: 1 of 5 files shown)


apps/api/src/user/repositories/user/user.repository.ts
```diff
@@ -66,13 +66,14 @@ export class UserRepository extends BaseRepository<ApiPgTables["Users"], UserInp
       );
   }
 
-  async upsertByUserId(data: UserInput): Promise<UserOutput> {
```
stalniy, line 69: **suggestion(blocking):** upsert may be based on different things. We have several unique indexes on this table, so better to keep in the name

stalniy (dismissed):

stalniy (dismissed):

stalniy (dismissed):

baktun14 (dismissed):

PR #2688, opened 2026-02-07
feat: improve template list

### LINKED ISSUE #108: Templates Section Improvements
The current "[Templates](https://deploy.cloudmos.io/templates)" section could benefit from improvements. 

I've prepared a draft of a possible improvement here:
Screenshot 2024-02-21 at 14 01 23

At the top of the page we could recommend templates. This would allow us to feature certain templates that we would like to highlight at a specific point of time. For example if a new LLM comes around, or not long ago we had Zephyr mining on Akash - which could be highlighted at that time.

Next to that we could provide another option where users could view "most popular templates". For all the users who like to explore templates based on popularity. 

The filtering option could be added to the section of "All templates" as well.

baktun14:
closes #108 

NEW
image

## Summary by CodeRabbit

* **New Features**
  * Templates now show "Recommended" and "Popular" badges and support tag-driven highlighting and ordering.
  * Template gallery: improved sorting (featured → recommended → popular), category labeling, counts, and search placement.
  * Template summaries may include tags in API responses.

* **Bug Fixes**
  * Better display for payment methods without card details; expiry shows "N/A" when missing.
  * Improved error messaging for unsupported payment types.


apps/deploy-web/src/components/templates/CuratedTemplatesSection.tsx
```diff
@@ -0,0 +1,66 @@
...
+  "akash-network-awesome-akash-DeepSeek-V3.1",
+  "akash-network-awesome-akash-DeepSeek-R1"
+];
+
+const mostPopularTemplateIds = [
+  "akash-network-awesome-akash-openclaw",
+  "akash-network-awesome-akash-ssh-ubuntu",
+  "akash-network-awesome-akash-comfyui",
+  "akash-network-awesome-akash-DeepSeek-V3.1"
+];
```
stalniy: **suggestion(non-blocking):** this could be part of /v1/templates-list response. Can be implemented with tags on templates (tags=[popular, recomended]) or as separate arrays as it's done here. If you move it to api, we will be able to change this values without application redeployment: change in code and then trigger workflow to rebuild templates.

apps/deploy-web/src/components/templates/TemplateGallery.tsx
```diff
@@ -14,12 +14,32 @@ import { domainName, UrlService } from "@src/utils/urlUtils";
...
+  Business: 7,
+  Chat: 8,
+  "Data Analytics": 9,
+  Gaming: 10,
+  Hosting: 11,
+  Media: 12,
+  Social: 13,
+  Storage: 14,
+  Tools: 15
+};
```
stalniy: **suggestion(non-blocking):** also better to do at templates build time, so they are generated in the way we need them and it makes updates easier

stalniy (dismissed):

apps/deploy-web/src/components/templates/CuratedTemplatesSection.tsx
baktun14: have them as env vars? or feature flags values?

stalniy: no, let's discuss today, easier to  show

apps/api/src/template/services/template-refresh/template-refresh.service.ts
stalniy, line 1: **suggestion:** let's remove this service. We rely on cloudflare, it cloudflare is down -> all our services will be down.

baktun14: Alright will do

apps/api/src/template/providers/template-gallery.provider.ts
```diff
@@ -3,20 +3,65 @@ import type { InjectionToken } from "tsyringe";
...
+  type TemplateTagsConfig
+} from "../services/template-gallery/template-gallery.service";
 import { TEMPLATE_CONFIG } from "./config.provider";
 
 export const TEMPLATE_GALLERY_SERVICE: InjectionToken<TemplateGalleryService> = Symbol("TEMPLATE_GALLERY_SERVICE");
 
+function getTagsConfigFromFlags(featureFlags: FeatureFlagsService, logger: LoggerService): TemplateTagsConfig {
+  const recommendedVariant = featureFlags.getVariant(FeatureFlags.TEMPLATE_RECOMMENDED_IDS);
+  const popularVariant = featureFlags.getVariant(FeatureFlags.TEMPLATE_POPULAR_IDS);
+  const priorityVariant = featureFlags.getVariant(FeatureFlags.TEMPLATE_CATEGORY_PRIORITY);
```
stalniy: **thought(non-blocking):**  this is not what I meant in another comment. you could just hardcode these values on API side, the thing is that changes to templates can be done without app re-deployment. So, you could make changes in api, merge PR and then rebuild templates manually. 

Probably the best option would be to keep this in templates itself but anyway it's done.

baktun14: I misunderstood what you meant.

> Probably the best option would be to keep this in templates itself but anyway it's done.

What do you mean by that, keep this in templates?

baktun14: For now the default values are still hardcoded, with the possiblity to override with a feature flag variant

stalniy (approved):

PR #2689, opened 2026-02-08
feat: add unit tests for BalancesService and enhance wallet limit refresh logic

baktun14:
## Summary by CodeRabbit

* **Tests**
  * Added comprehensive tests for wallet limit refresh covering trial-ending, unchanged limits, limit changes, and non-trial wallets.

* **Refactor**
  * Updated the update flow so trial-end status is applied before deciding to skip further updates, ensuring trial transitions are handled correctly.


apps/api/src/billing/services/balances/balances.service.ts
```diff
@@ -23,23 +24,38 @@ export class BalancesService {
...
+  private async cancelTrialJobsByWalletId(walletId: number): Promise<void> {
+    await Promise.all([
+      this.jobQueueService.cancelCreatedByPattern({
+        name: "CloseTrialDeployment",
+        singletonKeyPattern: `closeTrialDeployment.%.${walletId}`
+      }),
+      this.jobQueueService.cancelCreatedByPattern({
+        name: "NotificationJob",
+        singletonKeyPattern: `notification.beforeCloseTrialDeployment.%.${walletId}`
+      })
```
stalniy: **thought(blocking):** the idea of relying on internal table structure of pgboss to cancel jobs by conditions puts a bit of maintenance on us. This job handlers have checks that converts them into no op if user is not in trial mode anymore:

- [CloseTrialDeployment](https://github.com/akash-network/console/blob/main/apps/api/src/app/services/close-trial-deployment/close-trial-deployment.handler.ts#L49) skips doing stuff if wallet is not in trial anymore
- [beforeCloseTrialDeployment](https://github.com/akash-network/console/blob/main/apps/api/src/app/services/trial-deployment-lease-created/trial-deployment-lease-created.handler.ts#L65) notification also is triggered only if user is in trial

So, this cancellation seems to be redundant. Could you please elaborate why you decided to add them?

stalniy (changes requested):

apps/api/src/billing/services/balances/balances.service.ts
baktun14: You're right this is not ideal, removed it. It was to remove unnecessary jobs that won't have any side effects since the trial is not valid anymore, but we do check in the job so we can keep it like this.

stalniy (dismissed):

apps/api/src/billing/services/balances/balances.service.ts
```diff
@@ -29,14 +29,14 @@ export class BalancesService {
...
 
-    if (!Object.keys(update).length) {
-      return;
-    }
-
     if (options?.endTrial && userWallet.isTrialing) {
       update.isTrialing = false;
     }
 
+    if (!Object.keys(update).length) {
```
baktun14, line 38: @ygrishajev can you validate this if the check for update was intended please? There could be issues where the trial flag isn't updated when it should.

ygrishajev: `getFreshLimitsUpdate` method can resolve with an empty object. In this case no update is needed. If there are issues I doubt it's due to this check: if the object is empty there is nothing to udpate

ygrishajev: so it kinda makes sense to move the check below to make sure that trial can be ended even if the balance is changed. However the only place it is used is in `BalanceService.topUpWallet` meaning that the balance should have been changed. Basically if balance stays unchanged the issue is most prob elsewhere where the balance is not set into the db properly (e.g. rpc node sync or some unhandled tx failures on our side)

baktun14: it's also called in `executeDecodedTxByUserWallet`, but in any case it doesn't hurt to fix the order

ygrishajev (approved):

PR #2720, opened 2026-02-12
fix: remove PgSemaphore to fix the payment deadlock

ygrishajev:
PgSemaphore is no longer needed in the console API. All operations using it are either logically safe without locking or already idempotent. BatchSigningClientService, the only remaining consumer, is only used for testing or in single instance mode, so a simple in-process Sema(1) is sufficient.

## Summary by CodeRabbit

* **Refactor**
  * Removed distributed PostgreSQL semaphore-based synchronization and replaced with in-process semaphore for relevant billing flows.
  * Simplified concurrency control in payment confirmation, coupon application, and wallet creation flows.
* **Tests**
  * Removed semaphore-related unit and integration tests and a race-condition test for wallet creation.
* **Chores**
  * Removed semaphore provider initialization and related integration setup.


apps/api/src/billing/controllers/stripe/stripe.controller.ts
```diff
@@ -78,7 +76,6 @@ export class StripeController {
     return { data: [] };
   }
 
-  @Semaphore()
```
ygrishajev, line 81: this is logically redundant. We have it properly implemented on the frontend and if implemented via api - it would not do a double call. in the worst case we could do a refund

apps/api/src/billing/controllers/stripe/stripe.controller.ts
```diff
@@ -124,7 +121,6 @@ export class StripeController {
     }
   }
 
-  @Semaphore()
   @Protected([{ action: "create", subject: "StripePayment" }])
   async applyCoupon(
```
ygrishajev, line 125: apply coupon is already idempotent op in stripe.

apps/api/src/billing/services/refill/refill.service.ts
```diff
@@ -111,7 +110,6 @@ export class RefillService {
     this.logger.info({ event: "WALLET_BALANCE_REDUCED", userId, amountUsd, previousLimit: currentLimit, nextLimit });
   }
 
-  @Semaphore()
```
ygrishajev, line 114: user wallet has a unique index on userId so sema is redundant here

stalniy (approved):

PR #2728, opened 2026-02-13
feat: add ManagedDeploymentLeaseCreated event and handler

baktun14:
## Summary by CodeRabbit

* **New Features**
  * Auto-enable deployment alerts for managed leases, using deployment escrow to set a sensible threshold (30%).

* **Improvements**
  * Notification service now ensures a default channel, fetches deployment details, and skips gracefully when prerequisites (email, channel, or deployment) are missing.
  * Managed-lease events are published during the signing flow and the corresponding handler is registered at startup to trigger alerting.

* **Tests**
  * Added unit tests covering the alert flow and managed-lease event handler.


stalniy (dismissed):

ygrishajev:
please make sure to follow pull request template and describe the changes

apps/api/src/notifications/services/notification/notification.service.ts
```diff
@@ -13,7 +18,12 @@ const DEFAULT_BACKOFF_OPTIONS: BackoffOptions = {
 
 @singleton()
 export class NotificationService {
-  constructor(@inject(NOTIFICATIONS_API_CLIENT) private readonly notificationsApi: NotificationsApiClient) {}
+  constructor(
+    @inject(NOTIFICATIONS_API_CLIENT) private readonly notificationsApi: NotificationsApiClient,
+    private readonly userRepository: UserRepository,
+    private readonly deploymentHttpService: DeploymentHttpService,
```
ygrishajev: this class is responsible for notifications logic, I assume there may be use cases when `deploymentHttpService` can be used in it however I also believe this is not it.

The new method added now does some deployment specific logic which breaks boundaries of concerns. IMO it would be better if that method accepts deployment params as a part of the input. So it can be done inside the handler and most prob the handler can utilise one of the existing services created for this purpose.

apps/api/src/notifications/services/notification/notification.service.ts
```diff
@@ -65,6 +75,77 @@ export class NotificationService {
       throw new Error("Failed to create default notification channel", { cause: result.error });
     }, DEFAULT_BACKOFF_OPTIONS);
   }
+
+  async autoEnableDeploymentAlert(input: { userId: string; walletAddress: string; dseq: string }): Promise<void> {
```
ygrishajev: This method is objectively too long and takes time to understand. I'd suggest to break it down and extract smaller methods with verbose names. E.g. `getNotificationChannel` etc. Then it would much readable, reduce cognitive load on the reviewer.

apps/api/src/billing/events/managed-deployment-lease-created.ts
```diff
@@ -0,0 +1,17 @@
+import type { DomainEvent } from "@src/core/services/domain-events/domain-events.service";
+import { DOMAIN_EVENT_NAME } from "@src/core/services/domain-events/domain-events.service";
+
+export class ManagedDeploymentLeaseCreated implements DomainEvent {
```
ygrishajev: Semantically this should be a command, extend a `Job` and renamed to something like
```suggestion
export class EnableDeploymentAlertCommand implements Job {
```

Coz it seems like it'll never do anything else and in this case it should say what it does, not what it reacts to. The handler class should follow the same approach.

baktun14:
image
Tested locally and both are enabled by default when creating a deployment

ygrishajev (approved):

ygrishajev (approved):

PR #2731, opened 2026-02-13
feat: enhance deployment reader service

baktun14:
https://github.com/akash-network/support/issues/415

## Summary by CodeRabbit

* **Documentation**
  * Clarified test guidance: projects use Vitest or are migrating from Jest; preserved concrete test-run commands under the Unit Tests subsection.

* **Bug Fixes / API Changes**
  * Deployment list responses simplified: top-level deployment IDs reduced and lease entries no longer include a default null status.
  * Deployment list item shape updated to a dedicated, cleaner list schema.

### DIFF
--- a/.claude/instructions/general-projects-description.manual.md
+++ b/.claude/instructions/general-projects-description.manual.md
@@ -87,7 +87,7 @@ npm run build
 
 ## Running Tests
 
-All applications and packages use **vitest** for unit testing or are moving from jest.
+All applications and packages use **Vitest** for unit testing or are migrating from **Jest**.
 
 ### Unit Tests
 
--- a/.claude/instructions/general-projects-description.manual.md
+++ b/.claude/instructions/general-projects-description.manual.md
@@ -87,9 +87,27 @@ npm run build
 
 ## Running Tests
 
-### Unit Tests
+### Test Frameworks
 
-All applications and packages use **Jest** for unit testing.
+The monorepo is migrating from Jest to Vitest. Check the project's `package.json` test scripts to confirm which framework it uses.
+
+**Vitest projects:**
+- `apps/api`
+- `packages/http-sdk`
+- `packages/logging`
+- `packages/net`
+- `packages/ui`
+
+**Jest projects:**
+- `apps/deploy-web`
+- `apps/indexer`
+- `apps/log-collector`
+- `apps/notifications`
+- `apps/provider-console`
+- `apps/provider-proxy`
+- `apps/tx-signer`
+
+### Unit Tests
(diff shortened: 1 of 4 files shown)


apps/api/src/deployment/services/deployment-reader/deployment-reader.service.ts
```diff
@@ -139,6 +142,43 @@ export class DeploymentReaderService {
...
+    const providerAuthMap = new Map<string, Awaited<ReturnType<ProviderService["toProviderAuth"]>>>();
+    await PromisePool.withConcurrency(10)
+      .for(uniqueProviders)
+      .process(async provider => {
+        const auth = await this.providerService.toProviderAuth({ walletId: wallet.id, provider }, ["status"]);
+        providerAuthMap.set(provider, auth);
+      });
+
+    const statusMap = new Map<string, Awaited<ReturnType<ProviderService["getLeaseStatus"]>> | null>();
+    await PromisePool.withConcurrency(100)
```
stalniy: **question:** did you load test this? 

I have doubts whether it will work reliably. 10 requests will generate 1000 connections, this may end up with OOM or other system level limits (like max open files)

baktun14: I didn't but I was considering to remove the status from the list endpoint instead of implementing this, and only have the status being fetched from the detail endpoint. Makes sense?

.claude/instructions/general-projects-description.manual.md
```diff
@@ -87,9 +87,9 @@ npm run build
 
 ## Running Tests
 
-### Unit Tests
+All applications and packages use **Vitest** for unit testing or are migrating from **Jest**.
 
-All applications and packages use **Jest** for unit testing.
+### Unit Tests
```
ygrishajev, line 90: non-blocking: afaiu this would be dropped based on the current workflows and discussions setup

apps/api/src/deployment/http-schemas/deployment.schema.ts
```diff
@@ -73,6 +73,10 @@ export const DeploymentResponseSchema = z.object({
   })
 });
 
+const DeploymentListItemSchema = DeploymentResponseSchema.extend({
```
ygrishajev:
```suggestion
const DeploymentLeaseListItemSchema = DeploymentResponseSchema.extend({
```

or perhaps the one above is wrong?

ygrishajev (changes requested):

ygrishajev (approved):

PR #2745, opened 2026-02-16
chore(notifications): migrates from jest to vitest

ygrishajev:
## Summary by CodeRabbit

* **Chores**
  * Replaced Jest with Vitest for the notifications project, removed old Jest config, added a new Vitest configuration, and updated test scripts and dev tooling.

* **Tests**
  * Migrated all test suites and mocks to Vitest (unit and functional), updated timing/mocking APIs and test setups; no runtime behavior or public APIs changed.


apps/notifications/vitest.config.ts
```diff
@@ -0,0 +1,66 @@
...
+  ],
+  resolve: {
+    alias: {
+      "@src": path.resolve(__dirname, "./src"),
+      "@test": path.resolve(__dirname, "./test")
+    }
+  },
+  test: {
+    environment: "node",
+    globals: true,
```
stalniy: **suggestion:** ask it to also add explicit imports for `describe`, `it` and co. Explicit imports are better because the we don't need to deal with `types` option in `tsconfig` which will default to empty array from ts@6.0

PR #2762, opened 2026-02-17
fix: api memory improvement

baktun14:
## Why

The memory-cache doesn't have any limit

## What

Replace with LRUCache

## Summary by CodeRabbit

* **Improvements**
  * Switched to bounded LRU caching across the API for better memory use and predictable eviction.
  * Adjusted health-check/readiness logic for clearer success tracking and stricter initial-failure behavior.

* **Chores**
  * Removed legacy in-memory cache dependency and migrated caching helpers and usages to the new implementation.

* **Tests**
  * Added comprehensive cache unit tests and updated test cleanup to use the new caching helpers.


apps/api/src/caching/memoryCacheEngine.ts
```diff
@@ -1,13 +1,16 @@
-import mcache from "memory-cache";
+import { LRUCache } from "lru-cache";
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const cache = new LRUCache<string, any>({ max: 500 });
```
ygrishajev:
```suggestion
export type CacheValue = NonNullable<unknown>;
const cache = new LRUCache<string, CacheValue>({ max: 500 });
....
storeInCache<T extends CacheValue>(key: string, data: T, duration?: number) {
```

apps/api/src/caching/memoryCacheEngine.ts
```diff
@@ -1,13 +1,16 @@
...
    * @param {*} key
    */
   getFromCache<T>(key: string): T | false {
-    const cachedBody = mcache.get(key);
-    if (cachedBody) {
+    const cachedBody = cache.get(key);
+    if (cachedBody !== undefined) {
       return cachedBody as T;
     }
     return false;
```
ygrishajev: question: `false` is a valid value to store. Any idea why we return it as a fallback?

ygrishajev:
although this is not introducing radical changes I think the cache engine should be reworked as an interface utilising various storages. E.g. in local dev LRUCache is totally find, in prod we should use a distributed storage like redis.

@stalniy wdyt?

ygrishajev (changes requested):

stalniy:
> although this is not introducing radical changes I think the cache engine should be reworked as an interface utilising various storages. E.g. in local dev LRUCache is totally find, in prod we should use a distributed storage like redis.
> 
> @stalniy wdyt?

I actually like this one because it's minus 1 dependency. And this is the thing which I wanted to do by myself for a long time :) 

I think that we should make a discussion for this and consider other options like https://bentocache.dev/docs/introduction

stalniy (approved):

ygrishajev (approved):

PR #2763, opened 2026-02-17
fix(deployment): implement countByOwner method and corresponding tests

baktun14:
## Why                                                                                                                                                                                                              
                                                                                          
  The deployments table at stats.akash.network/addresses/{address}/deployments shows "Page 1 of 1" even when an address has many deployments. The Cosmos SDK node returns an incorrect pagination.total value equal
   to the page size (e.g. "10") instead of the true total count, resulting in pageCount = ceil(10/10) = 1 and making pagination navigation impossible.
   https://stats.akash.network/addresses/akash10lq8uyfl52d6t467qmd58e7jzl9ecmvet2m57p/deployments
                                                                                                                                                                                                                   
 ## What                                                                                                                                                                                                             
                                                                                                                                                                                                                   
  - Added countByOwner(owner, status?) method to DeploymentRepository that performs a fast COUNT query against the database, optionally filtering by active/closed status
  - Modified DeploymentReaderService.listWithResources to use the DB count instead of the unreliable Cosmos SDK pagination.total
  - Parallelized the lease fetch, provider list fetch, and DB count query with Promise.all for improved performance
  - Set countTotal: false in the Cosmos SDK request since we no longer rely on its total
  - Added unit tests for DeploymentRepository.countByOwner (3 tests) and DeploymentReaderService.listWithResources (7 tests)

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Deployment listings now return counts sourced from the backend and include owner/dseq mapping in results.

* **Tests**
  * Added comprehensive tests for deployment counting and listing behavior across status filters and pagination.

* **Chores**
  * Improved listing performance by parallelizing backend requests and optimizing count retrieval for faster, more reliable responses.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/apps/api/src/deployment/repositories/deployment/deployment.repository.spec.ts
+++ b/apps/api/src/deployment/repositories/deployment/deployment.repository.spec.ts
@@ -1,0 +1,52 @@
+import { Deployment } from "@akashnetwork/database/dbSchemas/akash";
+import { Op } from "sequelize";
+
+import { DeploymentRepository } from "./deployment.repository";
+
+describe(DeploymentRepository.name, () => {
+  describe("countByOwner", () => {
+    it("counts all deployments for an owner when no status filter", async () => {
+      const owner = "akash1abc123";
+      const expectedCount = 42;
+      const { repository, countSpy } = setup({ count: expectedCount });
+
+      const result = await repository.countByOwner(owner);
+
+      expect(countSpy).toHaveBeenCalledWith({ where: { owner } });
+      expect(result).toBe(expectedCount);
+    });
+
+    it("counts only active deployments when status is 'active'", async () => {
+      const owner = "akash1abc123";
+      const expectedCount = 10;
+      const { repository, countSpy } = setup({ count: expectedCount });
+
+      const result = await repository.countByOwner(owner, "active");
+
+      expect(countSpy).toHaveBeenCalledWith({
+        where: { owner, closedHeight: null }
+      });
+      expect(result).toBe(expectedCount);
+    });
+
+    it("counts only closed deployments when status is 'closed'", async () => {
+      const owner = "akash1abc123";
+      const expectedCount = 32;
+      const { repository, countSpy } = setup({ count: expectedCount });
+
+      const result = await repository.countByOwner(owner, "closed");
+
+      expect(countSpy).toHaveBeenCalledWith({
+        where: { owner, closedHeight: { [Op.ne]: null } }
+      });
+      expect(result).toBe(expectedCount);
+    });
+  });
+
+  function setup(input: { count: number }) {
+    const countSpy = jest.spyOn(Deployment, "count").mockResolvedValue(input.count);
+    const repository = new DeploymentRepository();
+
+    return { repository, countSpy };
+  }
+});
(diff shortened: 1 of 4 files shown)


ygrishajev:
I believe that requesting data from 1 source and pagination from another would cause another set of issues eventually. Switching entirely to the indexer is obviously not an option too due to performance issues.

Why can't this be fixed on the rpc node? Have this been requested to @akash-network/core ?

baktun14:
The hybrid approach (RPC for deployments, DB for count) is intentional because `pagination.count_total` is broken at the Akash RPC node level. It always returns the page size instead of the actual total:

```
# limit=5, count_total=true → total is 5 (wrong, should be 146)
curl "https://rpc.akt.dev/rest/akash/deployment/v1beta4/deployments/list?filters.owner=akash1p2e73vphy9umsx02y6xqr49yeu0dn9s3pytkvk&filters.state=active&pagination.limit=5&pagination.count_total=true"
→ { "total": "5", "count": 5 }

# limit=1000, all 146 active fit in one page → total matches but only because page_size == actual_count
curl "https://rpc.akt.dev/rest/akash/deployment/v1beta4/deployments/list?filters.owner=akash1p2e73vphy9umsx02y6xqr49yeu0dn9s3pytkvk&filters.state=active&pagination.limit=1000"
→ { "total": "146", "count": 146 }

# limit=1000, no state filter, >1000 deployments → total is 1000 (wrong, capped at page size)
curl "https://rpc.akt.dev/rest/akash/deployment/v1beta4/deployments/list?filters.owner=akash1p2e73vphy9umsx02y6xqr49yeu0dn9s3pytkvk&pagination.limit=1000&pagination.count_total=true"
→ { "total": "1000", "next_key": "...", "count": 1000 }
```

This happens with both offset-based and key-based pagination, regardless of whether `count_total=true` is set. The node simply ignores it.

Since we can't get an accurate total count from the RPC, the DB count is the only viable option for pagination. This should probably also be reported to @akash-network/node as a node-level bug.

ygrishajev:
@baktun14 let's first escalate this to to @akash-network/core, see how fast we can get this fixed. Coz it forces us to apply this hack which is not ideal also

cloud-j-luna:
Hey! In regard to the `node`, that behaviour is as designed due to the performance hit the node gets when going through all the deployments. That means the total is set to the total amount of deployments returned on the query. 

One possible solution could be to offer an "infinite scroll" UX where scrolling down queries the next batch of *n* deployments. This way we don't need to hack and it might be a good compromise given the cost/benefit of pagination in this specific view. Thoughts?

baktun14:
> Hey! In regard to the `node`, that behaviour is as designed due to the performance hit the node gets when going through all the deployments. That means the total is set to the total amount of deployments returned on the query.
> 
> One possible solution could be to offer an "infinite scroll" UX where scrolling down queries the next batch of _n_ deployments. This way we don't need to hack and it might be a good compromise given the cost/benefit of pagination in this specific view. Thoughts?

I don't think this is a good solution for pagination ui's when displaying with a table, which is what we use everywhere.

What is the issue and why can't it be fixed?

PR #2764, opened 2026-02-18
chore: refactor release flow to create release on every push

stalniy:
## Why

We have release PR right now which collects changes and then we have 1 big release and deployment. This causes issues because:
- context is lost — by the time we release, we’ve already forgotten why something was done or what trade-offs were made or what deployment strategy we should apply
- risk accumulates — many unrelated changes ship together, making rollback and root-cause analysis harder.
- feedback loop is too long — bugs and architectural mistakes are discovered late, when fixing them is more expensive.

## Summary by CodeRabbit

* **Chores**
  * Added workflow-level concurrency and a per-run "cancel-if-stale" option, broadened release triggers, and updated deploy inputs/flags to avoid stale deployments.
  * Migrated release tooling to ESM, simplified releaser packaging, removed legacy per-app release scripts, and added a new bump-and-changelog script to compute versions and generate changelogs.

* **Documentation**
  * Rewrote release workflow docs to reflect automated, GitHub Actions–driven releases and standardized tagging.

* **Revert**
  * Removed the pre-release PR creation workflow.


.github/workflows/all-release.yml
```diff
@@ -3,10 +3,9 @@ on:
   push:
     branches: ["main"]
     paths:
-      - "apps/*/package.json"
-
-concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+      - "apps/**/*"
+      - "packages/**/*"
+      - package-lock.json
```
stalniy: releases can be executed in parallel

.github/workflows/console-api-release.yml
```diff
@@ -8,6 +8,9 @@ on:
         required: false
         type: string
 
+concurrency:
+  group: ${{ github.workflow }}-${{ inputs.image_tag || github.ref_name }}
```
stalniy: deployments should be executed sequentially, though github doesn't support this. So, the best we can have is one in progress + 1 one latest in pending state.

.github/workflows/reusable-create-github-release.yml
```diff
@@ -13,7 +13,7 @@ on:
         value: ${{ jobs.release.outputs.git_tag }}
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}-${{ inputs.app }}
+  group: ${{ github.workflow }}-${{ github.sha }}-${{ inputs.app }}
```
stalniy: one release per sha, since all releases are executed on main branch

.github/workflows/reusable-create-github-release.yml
```diff
@@ -28,81 +28,62 @@ jobs:
...
+          {
+            echo "changelog<<CHANGELOG_EOF"
+            echo "$changelog"
+            echo "CHANGELOG_EOF"
+          } >> "$GITHUB_OUTPUT"
 
       - name: Create Release
         if: ${{ steps.bumps.outputs.git_tag != '' }}
-        id: create_release
-        uses: actions/create-release@v1
```
stalniy, line 102: this action is archived, so I replaced it with gh cli

ygrishajev (approved):

packages/releaser/recommended-bump.js
```diff
@@ -0,0 +1,162 @@
...
+  bumper.tag({ prefix: values["tag-prefix"] });
+}
+
+if (values.path?.length) {
+  const additionalPaths = findLocalPackageDependencies(values.path);
+  bumper.commits({ path: [values.path, ...additionalPaths] });
+}
+
+const lastTag = await bumper.getLastSemverTag();
+const result = await bumper.bump(whatBump);
```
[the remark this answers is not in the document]
stalniy: I tested and it works

.github/workflows/reusable-deploy-setup.yml
```diff
@@ -53,3 +60,55 @@ jobs:
...
+                const [otherPrefix, otherRawVersion] = otherImageTag.split('/', 2);
+                if (otherPrefix !== app) continue;
+
+                const otherVersion = otherRawVersion.replace(/^v/, '').replace(/-beta$/, '');
+                if (otherVersion === version) continue;
+
+                if (isNewer(otherVersion, version)) {
+                  throw new Error(`Skipping ${app}@${version}: found newer ${otherVersion} (status: ${status})`);
+                }
+              }
```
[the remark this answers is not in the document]
stalniy: those inputs are taken for workflow_dispatch workflow. This workflow fetches caller workflow id and then fetches workflowruns for that workflow id

packages/releaser/recommended-bump.js
```diff
@@ -0,0 +1,163 @@
...
+    nextVersion = `${major}.${minor + 1}.0`;
+    break;
+  default:
+    nextVersion = `${major}.${minor}.${patch + 1}`;
+    break;
+}
+
+const nextTag = `${prefix}${nextVersion}`;
+const repoUrl = values["repo-url"] || "";
+const changelog = generateChangelog([result.commits[0]], { repoUrl, lastTag, nextTag, nextVersion });
```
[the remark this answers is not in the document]
stalniy: this is intentional, because release creation is done in parallel and previous tags may not be pushed yet, so commit analyzer will receive more than 1 commit but changelog should generate only for a single commit because other instances of release workflow will create tags and releases for other commits

.github/workflows/reusable-create-github-release.yml
```diff
@@ -28,81 +28,62 @@ jobs:
...
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        with:
-          tag_name: ${{ steps.bumps.outputs.git_tag }}
-          release_name: ${{ steps.bumps.outputs.git_tag }}
-          body: ${{ steps.bumps.outputs.changelog }}
+          RELEASE_CHANGELOG: ${{ steps.bumps.outputs.changelog }}
+          RELEASE_VERSION: ${{ steps.bumps.outputs.git_tag }}
+        run: |
+          gh release create "$RELEASE_VERSION" \
```
stalniy, line 84: **issue(blocking):** handle race condition when release with such tag exists. Need to retry and recompute the version or maybe just add a suffix '${version}-${timestamp}'

.github/workflows/reusable-create-github-release.yml
```diff
@@ -13,7 +13,7 @@ on:
         value: ${{ jobs.release.outputs.git_tag }}
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}-${{ inputs.app }}
+  group: ${{ github.workflow }}-${{ github.sha }}-${{ inputs.app }}
   cancel-in-progress: true
```
stalniy: **suggestion(blocking):** remove concurrency group because it doesn't make sense to have group per sha, because sha are unique

.github/workflows/all-release.yml
```diff
@@ -3,10 +3,9 @@ on:
   push:
     branches: ["main"]
     paths:
-      - "apps/*/package.json"
-
-concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+      - "apps/**/*"
+      - "packages/**/*"
```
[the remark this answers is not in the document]
stalniy: they are actually important but I need to remap them into apps. Find all dependent apps

.github/workflows/reusable-deploy-setup.yml
```diff
@@ -53,3 +60,57 @@ jobs:
...
+              return false;
+            };
+
+            for (const status of ['queued', 'in_progress', 'completed']) {
+              const { data } = await github.rest.actions.listWorkflowRuns({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                workflow_id: currentRun.workflow_id,
+                status,
+                per_page: 10
```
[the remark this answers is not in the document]
stalniy: unlikely to have more than 10 concurrent runs per app

PR #2792, opened 2026-02-19
fix(stats): prevent SSR crash loop caused by Sentry consoleIntegration

baktun14:
## Why

The `stats-web` pod has **70 container restarts** due to a two-stage crash:

1. **SSR timeout:** `serverFetch()` has no timeout, so when the API is slow, Next.js's internal SSR timeout fires
2. **Fatal secondary crash:** Sentry's `consoleIntegration` patches `console.error`. When the SSR timeout calls `console.error`, Node's `util.inspect` crashes at `formatProperty` trying to access `.value` on an undefined property descriptor of Next.js's internal error object. This uncaught exception kills the process.

Without Sentry's patch, the SSR timeout would be handled gracefully by Next.js's error boundary.

## What

- **Disable Sentry `Console` integration on the server** — prevents Sentry from patching `console.error`, eliminating the fatal secondary crash. Sentry still captures errors via `captureException`.
- **Add 10s `AbortController` timeout to `serverFetch`** — defense in depth to abort slow API calls before Next.js's own SSR timeout fires.
- **Add 404 handling to deployment detail page** — matches the pattern used by blocks/address/transaction pages. Returns a "not found" message instead of throwing.
- **Add error logging to address page** — adds `logger.error()` before throwing on non-404 errors, matching other pages.
- **Add vitest setup + `serverFetch` unit tests** — 5 tests covering timeout behavior, signal forwarding, and header merging.

## Summary by CodeRabbit

* **Bug Fixes**
  * Show a “Deployment not found” message for missing deployments.
  * Improved logging for failed address and deployment fetches.
  * Added a 10s timeout to API requests.
  * Disabled a problematic console integration in error tracking.

* **Tests**
  * Added unit tests covering server-side fetch behavior, timeouts, header forwarding, and teardown.
  * Added Vitest configuration and a new test:unit script.

* **Chores**
  * Added Vitest as a dev dependency and updated scripts.

### DIFF
--- a/apps/stats-web/package.json
+++ b/apps/stats-web/package.json
@@ -10,7 +10,8 @@
     "dev:inspect": "cross-env NODE_OPTIONS=\"--inspect\" ../../node_modules/.bin/next start -p 3001",
     "format": "prettier --write ./*.{ts,js,json} **/*.{ts,tsx,js,json}",
     "lint": "eslint .",
-    "start": "next start"
+    "start": "next start",
+    "test:unit": "NODE_ENV=test vitest run"
   },
   "dependencies": {
     "@akashnetwork/chain-sdk": "1.0.0-alpha.24",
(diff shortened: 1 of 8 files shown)


stalniy (approved):
@baktun14 I'd like to migrate stats to pure react app, only frontend without SSR and do pre-rendering using headless browser or cloudflare pre-render. This SSR issues take a lot of time, like proxying IPs, observability, memory leaks, slow response, etc. This should not be such a pain for so simple app

PR #2797, opened 2026-02-20
refactor: uses local instrumentation package for provider-proxy

stalniy:
## Why

The internal package aims to load only otel libs we need for instrumentation and improve application bootstrap time

Default otel auto-instrumentation depends on many packages which we don't need. Even though, we configured it in [helm](https://github.com/akash-network/helm-charts/blob/main/charts/nodejs-app/templates/_app-env.tpl#L7-L10) to apply only  instrumentations that we need, it still loads all libraries and Nodejs spends time to parse and evaluate their code.

## Summary by CodeRabbit

* **Chores**
  * Disabled telemetry collection across production and staging deployments.
  * Status endpoint now reports application version from an environment variable.

* **Refactor**
  * Instrumentation centralized into a new internal package and initialized at startup; legacy tracing dependencies replaced.

* **Tests**
  * Added a functional test validating the /status endpoint and its response shape.


ygrishajev (approved):

stalniy:
closed in favor of https://github.com/akash-network/console/pull/2800, don't understand why CI fails

PR #2815, opened 2026-02-24
fix(deployment): optimizes provider deployment query

ygrishajev:
## Why

The `GET /v1/providers/{provider}/deployments/{skip}/{limit}` endpoint is the top query by total execution time in `pg_stat_statements` — 817 calls with **11.2 seconds mean execution time** (9.1M ms total). This is caused by Sequelize generating a correlated subquery when combining `include` (JOIN) with `LIMIT/OFFSET` pagination. For every row in the `deployment` table, Postgres runs a subquery against `lease`, resulting in a full table scan.

## What

- Replace the Sequelize `findAll` with `include` + pagination in `ProviderDeploymentsService.getProviderDeployments` with a raw SQL query using `JOIN + GROUP BY` for the initial paginated dseq lookup. This eliminates the correlated subquery and allows Postgres to use the existing `(providerAddress, closedHeight, createdHeight)` index on the `lease` table.
- The second query (fetching full deployment/lease details by dseq list) remains unchanged as it operates on a small bounded set.
- Add integration tests covering: full response shape, active/closed status filtering, pagination ordering, provider isolation, and resource aggregation across multiple leases.

### Before (Sequelize-generated SQL)
```sql
SELECT "deployment".* FROM (
  SELECT ... FROM "deployment"
  WHERE (
    SELECT "deploymentId" FROM "lease"
    WHERE "leases"."providerAddress" = $1
      AND "leases"."deploymentId" = "deployment"."id"
    LIMIT 1
  ) IS NOT NULL
  ORDER BY ... LIMIT $2 OFFSET $3
) AS "deployment" INNER JOIN "lease" ...
```

### After (raw SQL)
```sql
SELECT d."dseq"
FROM "deployment" d
INNER JOIN "lease" l ON l."deploymentId" = d."id"
  AND l."providerAddress" = $1
GROUP BY d."id"
ORDER BY d."createdHeight" DESC, d."dseq" DESC
LIMIT $2 OFFSET $3
```

Some more detailed explanation: the correlated subquery has no independent filter on deployment. The only         
  condition is:                                                                                 
          
```sql                                                                                      
WHERE (                                                                                       
  SELECT "deploymentId" FROM "lease"                      
  WHERE "leases"."providerAddress" = $1                                                       
    AND "leases"."deploymentId" = "deployment"."id"  -- references outer row                  
  LIMIT 1
) IS NOT NULL
```

Postgres has to evaluate this for every row in deployment because it can't know which
deployments have matching leases without checking each one. The subquery is correlated — it
depends on "deployment"."id" from the outer query — so Postgres can't optimize it into a join.

With the raw query fix, we flip the approach: start from lease (which has the index on
providerAddress), find the matching rows, then join to deployment. Postgres scans only the
relevant lease rows for that provider, then looks up the corresponding deployments.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive integration tests for provider deployment listing, filtering, pagination, and multi-lease resource aggregation to improve reliability.

* **Performance**
  * Improved provider deployment retrieval for faster, more consistent listing and pagination, including more accurate status filtering and resource summaries.

### DIFF
--- a/apps/api/src/provider/services/provider-deployments/provider-deployments.service.ts
+++ b/apps/api/src/provider/services/provider-deployments/provider-deployments.service.ts
@@ -1,6 +1,7 @@
 import { Block } from "@akashnetwork/database/dbSchemas";
 import { Deployment, Lease } from "@akashnetwork/database/dbSchemas/akash";
 import type { WhereOptions } from "sequelize";
+import { QueryTypes } from "sequelize";
 import { Op } from "sequelize";
 import { singleton } from "tsyringe";
 
@@ -22,22 +23,27 @@ export class ProviderDeploymentsService {
   }
 
   async getProviderDeployments(provider: string, skip: number, limit: number, status?: ProviderDeploymentsQuery["status"]) {
-    const leaseFilter: WhereOptions<Lease> = { providerAddress: provider };
-
-    if (status) {
-      leaseFilter.closedHeight = { [status === "active" ? Op.is : Op.ne]: null };
+    let statusCondition = "";
+    if (status === "active") {
+      statusCondition = `AND l."closedHeight" IS NULL`;
+    } else if (status === "closed") {
+      statusCondition = `AND l."closedHeight" IS NOT NULL`;
     }
 
-    const deploymentDseqs = await Deployment.findAll({
-      attributes: ["dseq", "createdHeight"],
-      include: [{ model: Lease, attributes: [], required: true, where: leaseFilter }],
-      order: [
-        ["createdHeight", "DESC"],
-        ["dseq", "DESC"]
-      ],
-      offset: skip,
-      limit: limit
-    });
+    const deploymentDseqs = await Deployment.sequelize!.query<{ dseq: string }>(
+      `/* provider-deployments:paginated-dseqs */ SELECT d."dseq"
+       FROM "deployment" d
+       INNER JOIN "lease" l ON l."deploymentId" = d."id"
+         AND l."providerAddress" = $1
+         ${statusCondition}
+       GROUP BY d."id"
+       ORDER BY d."createdHeight" DESC, d."dseq" DESC
+       LIMIT $2 OFFSET $3`,
+      {
+        bind: [provider, limit, skip],
+        type: QueryTypes.SELECT
+      }
+    );
 
     const deployments = await Deployment.findAll({
       where: {
@@ -56,7 +62,10 @@ export class ProviderDeploymentsService {
         { model: Block, required: true, as: "createdBlock" },
         { model: Block, required: false, as: "closedBlock" }
       ],
-      order: [["createdHeight", "DESC"]]
+      order: [
+        ["createdHeight", "DESC"],
+        ["dseq", "DESC"]
+      ]
     });
 
     return deployments.map(d => ({
--- a/apps/api/src/provider/services/provider-deployments/provider-deployments.service.integration.ts
+++ b/apps/api/src/provider/services/provider-deployments/provider-deployments.service.integration.ts
@@ -1,0 +1,352 @@
+import type { Lease } from "@akashnetwork/database/dbSchemas/akash";
+import type { Block } from "@akashnetwork/database/dbSchemas/base";
+import { faker } from "@faker-js/faker";
+import type { CreationAttributes } from "sequelize";
+import { container } from "tsyringe";
+
+import { CHAIN_DB } from "@src/chain";
+import { ProviderDeploymentsService } from "./provider-deployments.service";
+
+import { createAkashAddress, createAkashBlock, createDeployment, createDeploymentGroup, createLease, createProvider } from "@test/seeders";
+
+function uniqueHeight() {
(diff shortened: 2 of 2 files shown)


apps/api/src/provider/services/provider-deployments/provider-deployments.service.ts
```diff
@@ -22,22 +23,27 @@ export class ProviderDeploymentsService {
...
-      offset: skip,
-      limit: limit
-    });
+    const deploymentDseqs = await Deployment.sequelize!.query<{ dseq: string }>(
+      `/* provider-deployments:paginated-dseqs */ SELECT d."dseq"
+       FROM "deployment" d
+       INNER JOIN "lease" l ON l."deploymentId" = d."id"
+         AND l."providerAddress" = $1
+         ${statusCondition}
+       GROUP BY d."id"
```
stalniy, line 39: **issue:** one deployment may have multiple leases, one lease per deployment group. there will be one dseq and multiple gsec. Each gsec may have a separate lease

ygrishajev: The query uses `GROUP BY d."id"` which deduplicates at the deployment level — a deployment with multiple leases (one per gseq) produces a single row. This is verified by two integration tests:
- ["aggregates resources from multiple leases"](https://github.com/akash-network/console/pull/2815/changes#diff-7100ec48c375063a392fd33f5d0dfe8ee98d0ec3718fb0c9bc09e69a71287668R140) creates a deployment with 2 leases (gseq=1, gseq=2) and confirms both are returned with resources correctly summed;
- ["does not count multiple leases per deployment as separate pagination entries"](https://github.com/akash-network/console/pull/2815/changes#diff-7100ec48c375063a392fd33f5d0dfe8ee98d0ec3718fb0c9bc09e69a71287668R197) confirms a multi-lease deployment occupies a single pagination slot.

ygrishajev:
## Query plan comparison

### Before (correlated subquery)

Scans **3,897,358 deployment rows**, running a subquery per row. Estimated cost: **33,539,235**.

```
Nested Loop  (cost=78.97..173.30 rows=1 width=29)
  Join Filter: (deployment.id = leases."deploymentId")
  ->  Limit  (cost=78.55..164.60 rows=10 width=29)
        ->  Incremental Sort  (cost=78.55..33539235.20 rows=3897358 width=29)
              Sort Key: deployment."createdHeight" DESC, deployment.dseq DESC
              Presorted Key: deployment."createdHeight"
              ->  Index Scan Backward using deployment_created_height_closed_height on deployment  (cost=0.43..33366880.35 rows=3897358 width=29)
                    Filter: ((SubPlan 1) IS NOT NULL)
                    SubPlan 1
                      ->  Limit  (cost=0.42..8.44 rows=1 width=16)
                            ->  Index Scan using lease_deployment_id on lease leases_1  (cost=0.42..8.44 rows=1 width=16)
                                  Index Cond: ("deploymentId" = deployment.id)
                                  Filter: (("closedHeight" IS NULL) AND (("providerAddress")::text = '...'::text))
  ->  Materialize  (cost=0.42..8.45 rows=1 width=16)
        ->  Index Scan using lease_provider_address_closed_height_created_height on lease leases  (cost=0.42..8.44 rows=1 width=16)
              Index Cond: ((("providerAddress")::text = '...'::text) AND ("closedHeight" IS NULL))
```

### After (JOIN + GROUP BY)

Starts from the `lease` index on `(providerAddress, closedHeight)`, finds matching leases, then looks up deployments by primary key. Estimated cost: **16.92**.

```
Limit  (cost=16.92..16.92 rows=1 width=29)
  ->  Sort  (cost=16.92..16.92 rows=1 width=29)
        Sort Key: d."createdHeight" DESC, d.dseq DESC
        ->  Group  (cost=16.90..16.91 rows=1 width=29)
              Group Key: d.id
              ->  Sort  (cost=16.90..16.91 rows=1 width=29)
                    Sort Key: d.id
                    ->  Nested Loop  (cost=0.85..16.89 rows=1 width=29)
                          ->  Index Scan using lease_provider_address_closed_height_created_height on lease l  (cost=0.42..8.44 rows=1 width=16)
                                Index Cond: ((("providerAddress")::text = '...'::text) AND ("closedHeight" IS NULL))
                          ->  Index Scan using deployment_pkey on deployment d  (cost=0.43..8.45 rows=1 width=29)
                                Index Cond: (id = l."deploymentId")
```

### Summary

| | Before | After |
|---|---|---|
| Strategy | Scan all deployments, subquery per row | Index scan on lease, PK lookup on deployment |
| Rows scanned | 3,897,358 | Only matching leases |
| Estimated cost | 33,539,235 | 16.92 |
| Cost reduction | | ~2,000,000x |

apps/api/src/provider/services/provider-deployments/provider-deployments.service.ts
stalniy: thanks for clarification, then should be good

stalniy (approved):

PR #2825, opened 2026-02-25
feat: adds support for uact for custodial wallet flows

stalniy:
## Why

ref https://github.com/akash-network/console/issues/2859

## Summary by CodeRabbit

* **New Features**
  * ACT (Akash Compute Token) support added across pricing, deposits, balances and conversions; UI shows ACT where supported.

* **Configuration**
  * Akash testnet bumped to testnet-8 with staking enabled and updated network endpoints.

* **UI**
  * Wallet popups updated to display ACT vs USDC labels/values when supported.

* **Tests**
  * Expanded unit tests covering wallet popups, pricing, deposits and ACT-related flows.


apps/deploy-web/src/chains/akash-testnet.ts
```diff
@@ -4,13 +4,31 @@ import { akash, akashAssetList } from "./akash";
...
   chain_name: "akash-testnet",
   pretty_name: "Akash-Testnet",
   apis: {
     rpc: [{ address: "https://testnetrpc.akashnet.net", provider: "ovrclk" }],
     rest: [{ address: "https://testnetapi.akashnet.net", provider: "ovrclk" }]
+  },
+  fees: {
+    fee_tokens: [
+      {
+        denom: "uact",
```
baktun14: I think the fees and staking stay in `uakt`

stalniy: yes, you are right, I checked this with Scott after submitting this PR

apps/deploy-web/src/hooks/useSupportsACT/useSupportsACT.ts
```diff
@@ -0,0 +1,7 @@
+import { useSettings } from "@src/context/SettingsProvider";
+
+export function useSupportsACT(): boolean {
+  const { settings } = useSettings();
+  const nodeInfo = settings.selectedNode?.nodeInfo?.node_info;
+  return nodeInfo?.network === "testnet-8";
```
baktun14, line 6: question: should we use an var or ff for this?

baktun14 (approved):

packages/database/chainDefinitions.ts
```diff
@@ -83,17 +83,17 @@ export const chainDefinitions: { [key: string]: ChainDef } = {
...
-    genesisFileUrl: "https://raw.githubusercontent.com/akash-network/net/master/testnet-7/genesis.json",
+    genesisFileUrl: "https://raw.githubusercontent.com/akash-network/net/master/testnet-8/genesis.json",
     coinGeckoId: "akash-network",
     logoUrlSVG: "https://raw.githubusercontent.com/cosmos/chain-registry/master/akash/images/akt.svg",
     logoUrlPNG: "https://console.akash.network/images/chains/akash.png",
     customIndexers: ["AkashStatsIndexer"],
     bech32Prefix: "akash",
-    denom: "akt",
-    udenom: "uakt",
+    denom: "act",
```
baktun14, line 95: I'm not sure about this, I think this is the same as fees?

apps/deploy-web/src/components/wallet/CustodialWalletPopup/CustodialWalletPopup.spec.tsx
```diff
@@ -0,0 +1,127 @@
...
+    setup({ isSignedInWithTrial: false });
+
+    expect(screen.getByTestId("connect-managed-wallet")).toBeInTheDocument();
+  });
+
+  it("renders ConnectManagedWalletButton when signed in with trial but user exists", () => {
+    setup({ isSignedInWithTrial: true, user: { id: "user-1" } });
+
+    expect(screen.getByTestId("connect-managed-wallet")).toBeInTheDocument();
+  });
```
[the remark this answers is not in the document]
stalniy: using `getBy*` in combo with `toBeInTheDocument` makes it actually better because in case of failure it prints the DOM structure

apps/deploy-web/src/hooks/useSupportsACT/useSupportsACT.ts
stalniy: No both var and ff are worse in this case. This is blockchain characteristic and it depends on what rpc node our app talks to.

Having 2 variables like `use ACT` and `RPC_NODE_URL` opens a door for mistakes. For example `useACT = true` and `RPC_NODE_URL` points to node that doesn't support ACT. The same with feature flag.

The right way is to autodiscover capabilities of blockchain node based on its status or version. Then we also don't need to re-deploy anything. User just need to refresh the page after network upgrade and everything will just silently work

stalniy: Here I used the most primitive way to start migration. But we need to talk to core and find out how can we discover this.

packages/database/chainDefinitions.ts
stalniy: this is used in `AddressBalanceMonitor` and in `DeploymentBalanceMonitor`, that's why I guess it should be ACT but I'm not sure as well

apps/deploy-web/src/hooks/useSupportsACT/useSupportsACT.ts
stalniy: I will make it depend on the application version in the follow up PR. If `application_version.version` is higher than 2.1.0-a22 then ACT is enabled

packages/database/chainDefinitions.ts
baktun14: It should be fine I think, these are not actively used anymore.

apps/deploy-web/src/chains/akash-testnet.ts
```diff
@@ -4,13 +4,20 @@ import { akash, akashAssetList } from "./akash";
...
   chain_name: "akash-testnet",
   pretty_name: "Akash-Testnet",
   apis: {
     rpc: [{ address: "https://testnetrpc.akashnet.net", provider: "ovrclk" }],
     rest: [{ address: "https://testnetapi.akashnet.net", provider: "ovrclk" }]
+  },
+  staking: {
+    staking_tokens: [
+      {
+        denom: "uact"
```
baktun14: this should be uakt, not that it matters, we won't stake on the testnet.

baktun14 (approved):

stalniy:
socketio doesn't run PR check for some reason, ignore it with a label

PR #2845, opened 2026-03-02
fix(alert): disable deployment alerts before intentional close

baktun14:
## Why

When a user intentionally closes a deployment via managed wallet, the `deployment-closed` chain event fires and triggers an unwanted alert notification. This is annoying for users since they deliberately closed the deployment.

## What

Disable deployment alerts **before** the close transaction is signed, so the chain event finds no enabled alerts to trigger.

- Added `disableDeploymentAlerts` method to `NotificationService` that fetches current alerts and upserts them with `enabled: false` (best-effort, errors are caught and logged)
- Added `getDeploymentAlerts` private helper to `NotificationService`
- Hooked into `ManagedSignerService.executeDecodedTxByUserWallet()` to detect `MsgCloseDeployment` and disable alerts before signing — this covers both the direct close endpoint (`DELETE /v1/deployments/{dseq}`) and the generic transaction path (`POST /v1/tx`)
- Alerts must be disabled **before** close because the notifications service validates deployment existence on-chain; after closing, the upsert would fail with "Deployment not found"
- Added 6 unit tests for `disableDeploymentAlerts` covering: both alerts enabled, already disabled, suppressed by system, no alerts, error handling, partial disable

## Summary by CodeRabbit

* **New Features**
  * Automatically disable deployment alerts when a deployment is closed to prevent unnecessary notifications.

* **Updates**
  * Notification service now exposes a public operation to disable deployment alerts for a given user/wallet/deployment; managed signing flow invokes it when closes are detected.

* **Tests**
  * Added comprehensive tests covering multiple close events, skipped cases, partial disables, error handling, and retry/backoff.

### DIFF
--- a/apps/api/src/billing/services/managed-signer/managed-signer.service.ts
+++ b/apps/api/src/billing/services/managed-signer/managed-signer.service.ts
@@ -1,5 +1,5 @@
 import { MsgAccountDeposit } from "@akashnetwork/chain-sdk/private-types/akash.v1";
-import { MsgCreateDeployment } from "@akashnetwork/chain-sdk/private-types/akash.v1beta4";
+import { MsgCloseDeployment, MsgCreateDeployment } from "@akashnetwork/chain-sdk/private-types/akash.v1beta4";
 import { MsgCreateLease } from "@akashnetwork/chain-sdk/private-types/akash.v1beta5";
 import { LeaseHttpService } from "@akashnetwork/http-sdk";
 import { EncodeObject, Registry } from "@cosmjs/proto-signing";
@@ -19,6 +19,7 @@ import { TxManagerService } from "@src/billing/services/tx-manager/tx-manager.se
 import { WalletReloadJobService } from "@src/billing/services/wallet-reload-job/wallet-reload-job.service";
 import { DomainEventsService } from "@src/core/services/domain-events/domain-events.service";
 import { Trace, withSpan } from "@src/core/services/tracing/tracing.service";
+import { NotificationService } from "@src/notifications/services/notification/notification.service";
 import { UserRepository } from "@src/user/repositories";
 import { BalancesService } from "../balances/balances.service";
 import { BillingConfigService } from "../billing-config/billing-config.service";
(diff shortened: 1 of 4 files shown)


apps/api/src/billing/services/managed-signer/managed-signer.service.ts
```diff
@@ -112,6 +114,24 @@ export class ManagedSignerService {
...
+      const uniqueDseqs = [
+        ...new Set(closeDeploymentMessages.filter(message => message.value?.id?.dseq != null).map(message => message.value.id!.dseq.toString()))
+      ];
+
+      for (const dseq of uniqueDseqs) {
+        await this.notificationService.disableDeploymentAlerts({
+          userId: userWallet.userId,
+          walletAddress: userWallet.address!,
+          dseq
+        });
```
stalniy, line 131: **question:** does it make sense to keep this logic on notification service? for example, when it receives a block or event (not sure how it does it) then it can check whether there is a MsgCloseDeployment message for that dseq and if so, it means user intentionally closed it and it can skip notification.

stalniy:
I think from awareness standpoint, it's good to receive this notification even if it was initiated by user. Because he receives confirmation that he won't be billed for that deployment anymore. Also from audit trail standpoint it may be useful, since blockchain is abstracted from managed wallet user.

So, probably it makes sense to make it configurable via "notify on explicit close" checkbox, which can be disabled by default

apps/api/src/billing/services/managed-signer/managed-signer.service.ts
ygrishajev: This may be a cleaner approach indeed. Some logic to identify the event source would be needed [here](https://github.com/akash-network/console/blob/16caa25c2a1e8bdc0318b0e25f7bf4e7af8dafc5/apps/notifications/src/modules/chain/services/tx-events-service/tx-events.service.ts). Here's what the trigger event looks like:

```json
[
  {
    "type": "akash.deployment.v1.EventDeploymentClosed",
    "attributes": [
      {
        "key": "id",
        "value": "{\"owner\":\"akash1sjds2vaquyax6zf88jhl9kt4enp3tqdn4r9erv\",\"dseq\":\"2330094\"}"
      },
      {
        "key": "msg_index",
        "value": "0"
      }
    ]
  }
]
```

ygrishajev (commented):
Although I understand the purpose of this PR I'd like to throw in some concern:
We currently do not support SASS like product organization. Meaning if any user wants to use akash as an organization they are forced to share credentials and use the same account. This said I as a user may want to receive notifications if other users close deployments intentionally.

PR #2852, opened 2026-03-03
fix(billing): add chain error middleware to tx-signer

ygrishajev:
## Why

The tx-signer service is too noisy in alerting because all errors are returned as 500 (Internal Server Error), even when they are caused by invalid client input (e.g., insufficient funds, deployment closed, invalid GPU attributes). This makes it difficult to distinguish actual server issues from expected client errors.

## What

Added a ChainErrorMiddleware Hono middleware to tx-signer that classifies known chain/blockchain error patterns into proper HTTP status codes (400/402) instead of letting them fall through as 500s. The middleware:

- Matches 13 known chain error patterns (mirroring the API's ChainErrorService)
- Preserves the original error message so the API layer can still parse it for its own classification
- Only applies to /v1/tx/* routes
- Non-matching errors pass through unchanged to the existing HonoErrorHandlerService as 500s

## Release Notes

* **New Features**
  * Enhanced error handling for transaction operations, providing clearer and more appropriate error messages for common blockchain-related issues.

### DIFF
--- a/apps/tx-signer/src/server.ts
+++ b/apps/tx-signer/src/server.ts
@@ -9,6 +9,7 @@ import { container } from "tsyringe";
 
 import { healthzRouter, txRouter } from "@src/routes";
 import { AppConfigService } from "@src/services/app-config/app-config.service";
+import { ChainErrorMiddleware } from "@src/services/chain-error/chain-error.middleware";
 import { HonoErrorHandlerService } from "@src/services/hono-error-handler/hono-error-handler.service";
 import { startServer } from "@src/services/start-server/start-server";
 import type { AppEnv } from "@src/types/app-context";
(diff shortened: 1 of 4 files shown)


apps/tx-signer/src/server.ts
```diff
@@ -9,13 +9,15 @@ import { container } from "tsyringe";
...
 import { AppConfigService } from "@src/services/app-config/app-config.service";
+import { ChainErrorMiddleware } from "@src/services/chain-error/chain-error.middleware";
 import { HonoErrorHandlerService } from "@src/services/hono-error-handler/hono-error-handler.service";
 import { startServer } from "@src/services/start-server/start-server";
 import type { AppEnv } from "@src/types/app-context";
 
 export const app = new Hono<AppEnv>();
 app.use("*", otel());
 app.use(container.resolve(HttpLoggerInterceptor).intercept());
+app.use("/v1/tx/*", container.resolve(ChainErrorMiddleware).intercept());
```
stalniy, line 20: **suggestion(non-blocking):** you can attach middleware in `createRoute` call. It accepts `middleware` option or you can just wrap corresponding controller logic.

This keeps related things together

stalniy (approved):

PR #2860, opened 2026-03-05
refactor: removes POST /v1/certificates endpoint since it is not needed anymore

stalniy:
## Why

Some users pass certificate properties incorrectly, this causes issues and requires additional validation of client provided certificate and documentation improvements. But all providers support JWT authentication, so we don't need to support mTLS anymore and we should not confuse our users by supporting 2 types of authentication for deployment api

## Summary by CodeRabbit

* **Breaking Changes**
  * Certificate creation endpoint and related APIs removed.
  * Certificate fields removed from deployment and lease requests.
  * Provider authentication now uses JWT tokens exclusively (mTLS removed).
  * Bid responses no longer include certificate requirement information.

* **Tests**
  * Test suites and end-to-end flows updated to remove certificate handling and mTLS paths.


ygrishajev (approved):

baktun14:
we also need to remove it from the documentation website https://akash.network/docs/api-documentation/console-api/api-reference/#post-v1certificates-optional https://github.com/akash-network/website/blob/6521bffb17f35d14467782725436b2866bd506ad/src/content/Docs/api-documentation/console-api/api-reference/index.md?plain=1#L554

PR #2863, opened 2026-03-05
feat(log-collector): polling-based pod discovery with AbortSignal support

ygrishajev:
## Why

When a target pod restarts (gets a new name), the log-collector stops receiving logs because it only discovers pods once at startup. This adds polling-based pod discovery to detect pod additions and removals at runtime.

> **Note:** This PR is stacked on top of https://github.com/akash-network/console/pull/2862 and should only be reviewed after that PR.

## What

- Add `POD_POLL_INTERVAL_MS` config (default: 5000ms)
- Add `watchPods(callback)` to `PodDiscoveryService` with polling loop that detects pod additions/removals and notifies via `PodCallback = (podInfo, signal) => void`
- Add `isPodReady` filter to only track pods with `Ready=True` condition
- Add `AbortSignal` support to `PodLogsCollectorService` for graceful teardown when a pod disappears (early return, k8s stream abort, logStream destroy)
- Update `PodLogsCollectorFactory` to pass `AbortSignal` through
- Rename `K8sLogCollectorService` → `K8sCollectorService` with fire-and-forget per-pod collection using the `watchPods` callback pattern
- Extract `splitChunkIntoLines()` and `createDeduplicator()` from `write()` for readability
- The `PodCallback`/`AbortSignal` interface is designed so PR 3 (K8s Watch API) only replaces `startPodWatch()` internals — no changes needed in consumers

## Release Notes

* **New Features**
  * Added configurable pod discovery polling interval via environment variable
  * Implemented continuous pod monitoring with real-time tracking
  * Added support for canceling active log collection operations
  * Improved log output with automatic deduplication of duplicate lines

* **Bug Fixes**
  * Enhanced pod filtering to exclude non-ready pods from collection
  * Improved same-deployment pod exclusion logic


apps/log-collector/src/services/pod-logs-collector/pod-logs-collector.service.ts
```diff
@@ -166,41 +170,20 @@ export class PodLogsCollectorService {
...
-                continue;
-              }
+          logStream.on("close", () => {
+            this.loggerService.info({
+              event: "LOG_STREAM_CLOSED",
+              podName: this.podInfo.podName,
+              namespace: this.podInfo.namespace
+            });
+            resolve();
+          });
```
ygrishajev, line 180: @coderabbitai The remainingBuffer holds bytes that arrived without a trailing \n — i.e. an incomplete, truncated line. The close event fires when either:

  1. The pod's AbortSignal fires (pod untracked) — collection resumes from the last written timestamp when the pod is re-discovered
  2. The K8s log stream ends (container terminates) — the partial line is genuinely incomplete/truncated

  In both cases, writing a truncated line would produce malformed output. The data will either be re-collected on resume, or was never complete to begin with.

  ---
  Does that make sense to you?

apps/log-collector/src/services/pod-logs-collector/pod-logs-collector.service.ts
```diff
@@ -218,13 +201,76 @@ export class PodLogsCollectorService {
...
+            namespace: this.podInfo.namespace,
+            duplicateLine: line.substring(0, 100) + "..."
+          });
+          continue;
+        } else {
+          if (!firstChunkTimeout) {
+            firstChunkTimeout = setTimeout(() => {
+              isFirstChunk = false;
+            }, 300);
+          }
```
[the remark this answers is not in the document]
ygrishajev: @coderabbitai This deduplication logic is pre-existing (predates this PR). The current behavior is intentional: the timer starts after the first non-duplicate line so that the deduplication window stays open until we've actually moved past the replayed lines from Kubernetes sinceTime resumption. Starting the timer immediately could cause it to expire before duplicates arrive on slow connections, letting them through to the file.

apps/log-collector/src/services/pod-discovery/pod-discovery.service.spec.ts
```diff
@@ -355,6 +400,134 @@ describe(PodDiscoveryService.name, () => {
     });
   });
 
+  describe("watchPods", () => {
+    it("should call callback for each initially discovered pod", async () => {
+      const namespace = faker.internet.domainWord();
+      const { podDiscoveryService, k8sClient } = setup({
+        KUBERNETES_NAMESPACE_OVERRIDE: namespace
+      });
```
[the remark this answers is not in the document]
ygrishajev: @coderabbitai This test covers the initial discovery phase of watchPods — that the callback is invoked for each initially discovered pod with correct PodInfo and an AbortSignal. The other two tests cover the polling loop (new pod detection, removed pod signal abort). The never-resolving second listNamespacedPod call just keeps the polling loop suspended so the test can assert on initial discovery without dealing with poll timing. No short poll interval needed since it never reaches the poll.

apps/log-collector/src/services/k8s-collector/k8s-collector.service.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  async start(): Promise<void> {
+    try {
+      await this.podDiscoveryService.watchPods((podInfo, signal) => {
+        this.startCollectionForPod(podInfo, signal);
+      });
+    } catch (error) {
+      if (this.errorHandlerService.isForbidden(error)) {
+        this.loggerService.warn({
```
stalniy, line 30: **question:** isn't it a critical error?

apps/log-collector/src/services/pod-logs-collector/pod-logs-collector.service.ts
```diff
@@ -218,13 +201,76 @@ export class PodLogsCollectorService {
...
+   * Splits a buffer chunk into complete lines, carrying over incomplete trailing lines.
+   *
+   * @param remainingBuffer - Leftover bytes from the previous chunk
+   * @param chunk - The new data chunk to process
+   * @returns Object with complete lines and any remaining incomplete line bytes
+   */
+  private splitChunkIntoLines(remainingBuffer: Buffer<ArrayBufferLike>, chunk: Buffer): { lines: string[]; remaining: Buffer<ArrayBufferLike> } {
+    const combinedBuffer = Buffer.concat([remainingBuffer, chunk]);
+    const parts = combinedBuffer.toString().split("\n");
+    const isLastLineComplete = combinedBuffer[combinedBuffer.length - 1] === 10;
```
stalniy: **question:** what is the meaning of `10`?

stalniy (approved):

apps/log-collector/src/services/k8s-collector/k8s-collector.service.ts
ygrishajev: it is and there is nothing we can do about it. this would stop the process and pod would restart. the error message states what's wrong

apps/log-collector/src/services/pod-logs-collector/pod-logs-collector.service.ts
ygrishajev: 10 is the ASCII character code for the newline character (\n)

extracted into a var

PR #2865, opened 2026-03-06
fix: allow user to navigate freely while waiting for the deployment to close

### LINKED ISSUE #2776: UX - Unable to access app while closing deployment - pop up
While closing any deployment - It doesn't gives good user experience if someone wants to start something and don't want to wait until this deployment gets closed - he needs to wait for the deployment to close , because the app is disabled with close deployment pop up

instead we can push that closing deployment to notifications some where in headers section so, that can help user to move around the app and with notifying user from notification once the deployment closed in the header section 

Image

eb1609:
## Why
When closing a deployment, a blocking popup prevents the user from 
interacting with the rest of the app until the transaction completes.

## What
When closing a deployment, a modal blocks the user from interacting with the app until the transaction completes.

This fix navigates the user back to the deployments list immediately after they confirm closing, allowing them to freely use the app while the transaction processes in the background. The existing snackbar notification system handles the issue by showing progress, success, or error feedback.

Closes #2776

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved responsiveness of the close deployment flow by displaying the deployment list immediately when the action is initiated, rather than waiting for the transaction to complete.

### DIFF
--- a/apps/deploy-web/src/components/deployments/DeploymentDetailTopBar.tsx
+++ b/apps/deploy-web/src/components/deployments/DeploymentDetailTopBar.tsx
@@ -77,12 +77,15 @@ export const DeploymentDetailTopBar: React.FunctionComponent<Props> = ({
       return;
     }
 
+    //push to deployment list immediately to avoid showing a deployment that is in the process of closing,
+    // the deployment detail page will handle showing the correct state while the transaction is being processed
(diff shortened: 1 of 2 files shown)


baktun14:
Thanks for the feedback, but this is not the best implementation. We'll take note of the ux improvement and implement it.

PR #2869, opened 2026-03-06
fix: resolve Swagger UI validation errors for dseq and market-data endpoints (Issue #2778)

### LINKED ISSUE #2778: swagger - console api - few api not working to test in swagger UI
While testing the console api in swagger few of the endpoints getting the below errors

{
"error": "BadRequestError",
"message": "Validation error",
"code": "validation_error",
"type": "validation_error",
"data": [
{
"received": "{coin",
"code": "invalid_enum_value",
"options": [
"akash-network",
"usd-coin"
],
"path": [
"coin"
],
"message": "Invalid enum value. Expected 'akash-network' | 'usd-coin', received '{coin'"
}
]
}
akt price issue because payload is going as coin instead of akash-network Image

Image

Image
Image
Image

eb1609:
## Why

Several Swagger UI endpoints were unusable for testing due to validation errors:
- `GET /v1/market-data/{coin}` was sending the literal string `{coin}` instead of the selected enum value

- `GET /v1/deployments/{dseq}`, `GET /v1/bids`, `GET /v1/bids/{dseq}` were rejecting valid numeric input with `Value must follow pattern ^d+$`
## What

- Removed `?` from `{coin?}` path param in the market-data router — `?` is not valid OpenAPI syntax and caused Swagger to not resolve the parameter
- Changed `DseqSchema` from `z.bigint({ coerce: true })` to `z.string().regex(/^\d+$/)` — `bigint` is not a standard OpenAPI type, causing Swagger to generate a broken regex pattern

Fixes #2778

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced deployment closing experience with improved navigation timing and reduced UI delays.

* **Updates**
  * Market-data API endpoint now requires explicit coin parameter in requests.
  * Refined validation for numeric identifiers with stricter string pattern enforcement.

### DIFF
--- a/apps/deploy-web/src/components/deployments/DeploymentDetailTopBar.tsx
+++ b/apps/deploy-web/src/components/deployments/DeploymentDetailTopBar.tsx
@@ -77,12 +77,15 @@ export const DeploymentDetailTopBar: React.FunctionComponent<Props> = ({
       return;
     }
 
+    //push to deployment list immediately to avoid showing a deployment that is in the process of closing,
+    // the deployment detail page will handle showing the correct state while the transaction is being processed
+    router.push(UrlService.deploymentList());
+
     const message = TransactionMessageData.getCloseDeploymentMsg(address, deployment.dseq);
     const response = await signAndBroadcastTx([message]);
     if (response) {
       onDeploymentClose();
       removeLeases();
-      loadDeploymentDetail();
 
       analyticsService.track("close_deployment", {
         category: "deployments",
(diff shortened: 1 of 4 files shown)


apps/deploy-web/src/components/deployments/DeploymentDetailTopBar.tsx
```diff
@@ -77,12 +77,15 @@ export const DeploymentDetailTopBar: React.FunctionComponent<Props> = ({
       return;
     }
 
+    //push to deployment list immediately to avoid showing a deployment that is in the process of closing,
+    // the deployment detail page will handle showing the correct state while the transaction is being processed
+    router.push(UrlService.deploymentList());
```
ygrishajev, line 82: Hey, I don't think this is a desired behaviour. Transaction may fail for various reasons and deployment state would not be obvious to a user.

PR #2879, opened 2026-03-10
fix(notifications): fix hung shutdown on crashed poller

ygrishajev:
## Why

Incident: the chain events poller silently crashed and went undetected for 5 days.

**Root cause:** `finishPolling()` in `onModuleDestroy` did not check whether the abort signal was already aborted. When the poller crashed, the `.catch` handler aborted the signal, but `finishPolling` still called `once(signal, "complete")` waiting for an event that would never fire (since `processBlocksLooping` had already rejected and never reached the line that dispatches `"complete"`). This caused `app.close()` to hang indefinitely, preventing the full shutdown lifecycle — so the container stayed alive with a dead poller and passing health checks.

When a release eventually restarted the container, the poller recovered and processed all ~72,000 missed blocks without delay, triggering spurious notifications to users.

## What

- Fix `finishPolling` to skip waiting when the abort signal is already aborted
- Implement `HealthzService` on `ChainEventsPollerService` so liveness/readiness probes fail when the poller is down (defense in depth)
- Add a dedicated `HealthzController` for the `chain-events` interface that includes the poller in health checks

## Release Notes

* **New Features**
  * Added health check endpoints (readiness and liveness probes) for the notifications service to enable monitoring of service health and chain events poller status.

* **Tests**
  * Added comprehensive unit tests for health check functionality and chain events poller status reporting.

### DIFF
--- a/apps/notifications/src/modules/chain/services/chain-events-poller/chain-events-poller.service.ts
+++ b/apps/notifications/src/modules/chain/services/chain-events-poller/chain-events-poller.service.ts
@@ -9,6 +9,7 @@ import { setTimeout as delay } from "timers/promises";
 import { eventKeyRegistry } from "@src/common/config/event-key-registry.config";
 import { LoggerService } from "@src/common/services/logger/logger.service";
 import { ShutdownService } from "@src/common/services/shutdown/shutdown.service";
+import type { HealthzService, ProbeResult } from "@src/common/types/healthz.type";
 import { BrokerService } from "@src/infrastructure/broker";
 import type { ChainEventsConfig } from "@src/modules/chain/config";
 import { BlockCursorRepository } from "@src/modules/chain/repositories/block-cursor/block-cursor.repository";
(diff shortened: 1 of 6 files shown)


apps/notifications/src/modules/chain/services/chain-events-poller/chain-events-poller.service.ts
```diff
@@ -134,12 +137,24 @@ export class ChainEventsPollerService implements OnModuleInit, OnModuleDestroy {
...
+    return this.getReadinessStatus();
+  }
+
   async onModuleDestroy() {
     await this.finishPolling();
   }
 
   private async finishPolling() {
-    if (this.abortController) {
+    if (this.abortController && !this.abortController.signal.aborted) {
```
ygrishajev, line 157: this is the root cause of the issue

stalniy (approved):

PR #2919, opened 2026-03-13
refactor: moves blockchain nodes configuration to UI application responsibility

stalniy:
## Why

The blockchain node configuration (RPC/API endpoints) currently lives in a shared config fetched at runtime from the console API, which in turn pulls it from the GitHub API's main branch. This makes it impossible to reconfigure endpoints locally without pushing and merging changes — a poor developer experience and an unnecessary coupling between apps and a centralized config source.

## What

Moved the blockchain config resolution into each UI app (deploy-web, stats-web) as a local API route (/api/blockchain-config). Each app now resolves node endpoints directly using @akashnetwork/net defaults, with the ability to override mainnet RPC/API endpoints via server environment variables (DEFAULT_REST_API_NODE_URL_MAINNET, DEFAULT_RPC_NODE_URL_MAINNET). Sandbox and testnet endpoints are not overridable since they are autogenerated from the akash-network/net repository which works for us in 99.9% of cases

## Additional context

Don't remove console-api /v1/nodes/:network endpoint and corresponding configs from repo for some time because some users will use outdated version of stats-web and deploy-web, don't want to disrupt their flows

## Summary by CodeRabbit

* **New Features**
  * Added a public /api/blockchain-config endpoint to expose network-specific blockchain API/RPC info for mainnet, sandbox, and testnet.

* **Chores**
  * Added environment variables for mainnet API and RPC node URLs.
  * Switched runtime API base to use the centralized blockchain-config endpoint (client stores now use "/api").
  * Network configuration updated to source node info from the new centralized endpoint and removed legacy version URL usage.


apps/deploy-web/src/pages/api/blockchain-config.ts
```diff
@@ -0,0 +1,41 @@
...
+      mainnet: [
+        {
+          id: "mainnet-1",
+          api: services.privateConfig.DEFAULT_REST_API_NODE_URL_MAINNET ?? netConfig.getBaseAPIUrl("mainnet"),
+          rpc: services.privateConfig.DEFAULT_RPC_NODE_URL_MAINNET ?? netConfig.getBaseRpcUrl("mainnet")
+        }
+      ],
+      sandbox: [
+        {
+          id: "sandbox-1",
```
baktun14, line 25: question: what are these ids? are they supposed to match the network ids?

baktun14 (approved):

packages/network-store/src/network.config.ts
```diff
@@ -8,10 +8,9 @@ export const getInitialNetworksConfig = ({ apiBaseUrl }: { apiBaseUrl: string })
     id: MAINNET_ID,
     title: "Mainnet",
     description: "Akash Network mainnet network.",
-    nodesUrl: `${apiBaseUrl}/v1/nodes/mainnet`,
+    nodesUrl: `${apiBaseUrl}/blockchain-config?network=mainnet`,
```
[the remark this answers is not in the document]
stalniy: provider-console doen't use this functionality. they construct it separately

apps/deploy-web/src/pages/api/blockchain-config.ts
stalniy: it's just any unique value for chain node. In the prev code it was hostname of the node. But since we have only 1 and going to have only 1 because we now have rpc proxy, it doesn't matter I believe. I used `1` because this it the 1st element in the array, that's why it's `sandbox-1`

PR #2944, opened 2026-03-17
fix(deployment): use correct denom for managed wallet deposit modal

baktun14:
## Why

Fixes CON-93

Managed wallet users see "Balance: 0 USD" and incorrect deposit amounts in the deployment creation confirmation modal. Two root causes:

1. `yaml.template()` interprets `${VAR}` env variable syntax in SDLs as template variables, throwing a `ReferenceError` and silently falling back to `uakt` as the denom
2. The deposit modal always receives `sdlDenom` (which defaults to `uakt`) instead of the managed wallet's actual denom (USDC)

## What

- Replace `yaml.template()` with `js-yaml`'s `load()` for SDL denom extraction — it treats `${VAR}` as literal strings instead of template variables
- Pass `managedDenom` to `DeploymentDepositModal` when wallet is managed, so `useDenomData` reads the correct USDC balance instead of the (empty) AKT balance

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed deployment fee denomination selection to correctly use the managed wallet denomination when applicable.
  * Improved manifest editor YAML parsing and error handling to reduce parsing failures during manifest edits.

### DIFF
--- a/apps/deploy-web/src/components/new-deployment/ManifestEdit/ManifestEdit.tsx
+++ b/apps/deploy-web/src/components/new-deployment/ManifestEdit/ManifestEdit.tsx
@@ -2,7 +2,6 @@
 import type { ComponentProps, Dispatch, SetStateAction } from "react";
 import { useEffect, useMemo, useRef, useState } from "react";
 import type { SDLInput } from "@akashnetwork/chain-sdk/web";
-import { yaml } from "@akashnetwork/chain-sdk/web";
 import { Alert, Button, CustomTooltip, FileButton, Input, Snackbar, Spinner } from "@akashnetwork/ui/components";
 import { cn } from "@akashnetwork/ui/utils";
 import type { EncodeObject } from "@cosmjs/proto-signing";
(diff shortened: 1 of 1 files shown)


apps/deploy-web/src/components/new-deployment/ManifestEdit/ManifestEdit.tsx
```diff
@@ -451,7 +451,7 @@ export const ManifestEdit: React.FunctionComponent<Props> = ({
         <d.DeploymentDepositModal
           handleCancel={() => setIsDepositingDeployment(false)}
           onDeploymentDeposit={onDeploymentDeposit}
-          denom={sdlDenom}
+          denom={wallet.isManaged ? managedDenom : sdlDenom}
```
stalniy, line 454: **thought:** we have this logic in a different place. I'd replace `yaml.template` with just pure yaml parsing. Spreading this logic over multiple places will make maintenance even more complicated

https://github.com/akash-network/console/blob/main/apps/deploy-web/src/components/new-deployment/ManifestEdit/ManifestEdit.tsx#L140

stalniy: I created this https://github.com/akash-network/console/pull/2948

baktun14: I should close this pr?

PR #2962, opened 2026-03-19
feat(wallet): add BalanceWatchProvider for generic balance change polling

ygrishajev:
## Why

Part of CON-94

Multiple features (credit card payments, ACT minting) need to detect when a user's balance has increased after an async operation. `PaymentPollingProvider` handles this for payments but is tightly coupled to payment-specific logic (trial status, analytics, snackbar messages). We need a generic, reusable provider that only tracks balance changes — callers decide what to do with the result via callbacks.

## What

- **New `BalanceWatchProvider`** — generic provider mounted in `_app.tsx` that polls `refetchBalance` and signals when a specified balance field increases. Supports custom balance keys (e.g. `balanceUACT` for mint/burn) with `totalUsd` as default.
- **Callback support** — `start()` accepts optional `onSuccess`/`onTimeOut` callbacks stored in a ref, so they survive page navigations.
- **Integrated into `MintBurnPage`** — after a successful mint/burn tx, starts balance watch with callbacks that manage persistent pending/success/timeout snackbars.
- **Follow-up work**: `PaymentPollingProvider` can be simplified to use `BalanceWatchProvider` + callbacks. Trial-to-paid transition detection can be handled by checking `isTrialing` in the payment flow's `onSuccess` callback — no separate provider needed.

## Summary by CodeRabbit

* **New Features**
  * Automatic balance monitoring after mint/burn with real-time polling and callbacks
  * Submit button shows "Updating balance..." and is disabled while syncing
  * Enhanced notifications flow: persistent info, success ("Conversion complete!"), and timeout warnings

* **Bug Fixes**
  * Non-supported/non-custodial wallets render a 404-style page instead of blank

* **Tests**
  * New comprehensive tests for balance-watching behavior and provider integration

### DIFF
--- a/apps/deploy-web/src/context/BalanceWatchProvider/BalanceWatchProvider.tsx
+++ b/apps/deploy-web/src/context/BalanceWatchProvider/BalanceWatchProvider.tsx
@@ -1,0 +1,135 @@
+"use client";
+import React, { createContext, useCallback, useContext, useEffect, useRef } from "react";
+
+import type { WalletBalance } from "@src/hooks/useWalletBalance";
+import { useWalletBalance } from "@src/hooks/useWalletBalance";
+
+const POLLING_INTERVAL_MS = 2000;
+const MAX_POLLING_DURATION_MS = 30000;
+const MAX_ATTEMPTS = MAX_POLLING_DURATION_MS / POLLING_INTERVAL_MS;
+
+export const DEPENDENCIES = {
+  useWalletBalance
+};
+
+export interface BalanceWatchContextType {
+  /** Start watching for a balance increase. Only one watch can be active at a time; calls while already active are no-ops. */
+  start: (snapshotBalance: number, balanceKey?: keyof WalletBalance) => void;
+  stop: () => void;
+  isActive: boolean;
+  /** Becomes true once the polled balance exceeds the snapshot. Resets on next start. */
+  isSuccess: boolean;
+  /** True if polling timed out without detecting a change. Resets on next start. */
+  isTimeOut: boolean;
+}
+
+interface ActivePoll {
+  snapshotBalance: number;
+  balanceKey: keyof WalletBalance;
+  attemptCount: number;
+  interval: ReturnType<typeof setInterval>;
+}
+
+const BalanceWatchContext = createContext<BalanceWatchContextType | null>(null);
+
+export interface BalanceWatchProviderProps {
+  children: React.ReactNode;
+  dependencies?: typeof DEPENDENCIES;
+}
+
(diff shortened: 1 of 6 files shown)


apps/deploy-web/src/context/BalanceWatchProvider/BalanceWatchProvider.tsx
stalniy, line 1: **question(blocking):** what is the reason for this functionality to be implemented as a provider?

ygrishajev: Notification stays live if you navigate away from the page

baktun14:
## Alternative: Query ledger record status via BME REST (LCD) endpoint

Per discussion with Artur — instead of (or in addition to) polling balance changes, the frontend can query the ledger record status directly using the BME LCD endpoint:

```
GET {LCD_BASE_URL}/akash/bme/v1/ledger?filters.source={userAddress}&filters.status=ledger_record_status_pending
```

e.g. `https://testnetapi.akashnet.net/akash/bme/v1/ledger?filters.source=akash1abc...`

Available filters ([proto definition](https://github.com/akash-network/chain-sdk/blob/bf19b1a7fe2a1e2f14987267efed547a1843476b/proto/node/akash/bme/v1/query.proto#L34)):

| Filter | Description | Example |
|--------|-------------|---------|
| `filters.source` | Account address that initiated the burn/mint | `akash1abc...` |
| `filters.denom` | Burn denomination | `uakt` or `uact` |
| `filters.to_denom` | Mint denomination | `uact` or `uakt` |
| `filters.status` | Record status | `ledger_record_status_pending`, `ledger_record_status_executed` |

This would let us check the exact status of a user's mint/burn operation (pending → executed / canceled) rather than inferring completion from a balance increase. Could be a more reliable signal and would also let us surface *why* a conversion didn't happen (canceled vs still pending).

Other useful BME LCD endpoints:
- `GET /akash/bme/v1/vault` — current vault state (balances, total burned/minted, remint credits)
- `GET /akash/bme/v1/status` — circuit breaker status (mints_allowed, refunds_allowed, collateral_ratio)
- `GET /akash/bme/v1/params` — module parameters

apps/deploy-web/src/context/BalanceWatchProvider/BalanceWatchProvider.tsx
stalniy: can this be implemented with jotai store?

ygrishajev: I don't think this would work well with jotai. Coz besides state itself the provider does quite some logic with side effects which should be done without coupling to components

PR #2975, opened 2026-03-22
fix(managed-wallet): switch sandbox denom from uakt to uact

baktun14:
## Why

Sandbox chain has migrated to uact as its native denom with the BME upgrade. Managed wallets and the indexer need to use the correct denom.

## What

- Update `DEPLOYMENT_GRANT_DENOM` from `uakt` to `uact` in sandbox env
- Update sandbox chain definition `denom`/`udenom` from `akt`/`uakt` to `act`/`uact`

## Release Notes

* **Chores**
  * Updated deployment grant denomination configuration for sandbox and mainnet environments.


stalniy (changes requested):
The commit message is wrong and will not trigger release pipeline.

stalniy (approved):

PR #2976, opened 2026-03-23
fix(deployment): add backward compat for legacy deposit params endpoint

ygrishajev:
## Why

PR #2971 removes the legacy cosmos params endpoint entirely, but nodes running pre-2.0.0 don't serve the module-native `/akash/deployment/v1beta4/params` endpoint. This breaks backward compatibility for users connected to older nodes.

Part of CON-116

## What

- Add `legacyDepositParams` URL builder to `ApiUrlService` preserving the old `/cosmos/params/v1beta1/params` endpoint
- Add `getLegacyDepositParams` fetcher that parses the legacy response format
- `useDepositParams` now calls `useSupportsACT` to pick the correct endpoint and parser
- Injectable `DEPOSIT_PARAMS_DEPS` for testability following existing codebase patterns
- Tests cover both ACT-supported (module endpoint) and legacy (cosmos params) paths


stalniy (approved):

baktun14:
it does return currently the values since sdk53, we just never noticed https://rpc.akt.dev/rest/akash/deployment/v1beta4/params

ygrishajev:
> it does return currently the values since sdk53, we just never noticed https://rpc.akt.dev/rest/akash/deployment/v1beta4/params

@baktun14 yes, I figured later

PR #2981, opened 2026-03-23
fix(stats): hide percentage chip when baseline value is zero

baktun14:
## Why

The BME dashboard shows absurd percentage increases (e.g. 2,157,201,574,800%) because it just launched and there's no comparison data from 24h ago. When the baseline value is 0, `percIncrease` returned the raw value as a percentage instead of hiding the chip.

## What

Fixed `percIncrease` in all three copies (`stats-web`, `deploy-web`, `provider-console`):
- When baseline (`a`) is 0: return 0 (hides the chip since `StatsCard` checks `!!diffPercent`)
- When current (`b`) is 0 and baseline is non-zero: return -1 (-100%) instead of the raw negative value
- Added unit tests for `percIncrease` in `stats-web`

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined percentage-increase calculation to simplify logic and unify edge-case outputs (notably when baseline or current values are zero).

* **Tests**
  * Added comprehensive test suites validating positive/negative changes, equal values, zero baselines, and drops-to-zero behavior across apps.

* **Documentation**
  * Added repo guidance on writing tests and a new "Writing Tests" section.

### DIFF
--- a/apps/stats-web/src/lib/mathHelpers.spec.ts
+++ b/apps/stats-web/src/lib/mathHelpers.spec.ts
@@ -2,7 +2,7 @@ import { describe, expect, it } from "vitest";
 
 import { percIncrease } from "./mathHelpers";
 
-describe("percIncrease", () => {
+describe(percIncrease.name, () => {
   it("returns standard percentage change for non-zero values", () => {
     expect(percIncrease(100, 150)).toBe(0.5);
   });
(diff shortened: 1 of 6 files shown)


stalniy (approved):

apps/stats-web/src/lib/mathHelpers.spec.ts
```diff
@@ -0,0 +1,33 @@
+import { describe, expect, it } from "vitest";
+
+import { percIncrease } from "./mathHelpers";
+
+describe("percIncrease", () => {
```
stalniy: **suggesstion:** should be `percIncrease.name`

baktun14: Fixed in 1eb1dd437

PR #2994, opened 2026-03-25
refactor: remove CoinGecko after oracle migration

stalniy:
## Why

Once all services use the oracle, CoinGecko code is dead. USDC no longer exists (replaced by ACT under BME), so there's no reason to keep CoinGecko at all. Ref CON-109

## What

- Deleted `packages/http-sdk/src/coin-gecko/` (service, types, barrel export)
- Removed `CoinGeckoHttpService` container registration from `apps/api`
- Simplified `DenomExchangeService` to always use the on-chain oracle path, removing `CoinGeckoHttpService` and `BlockchainCapabilitiesService` dependencies
- Updated `denom-exchange.service.spec.ts` — removed CoinGecko mock tests, kept oracle-only tests

## Summary by CodeRabbit

* **Refactor**
  * Market data now comes exclusively from the on-chain oracle: price and 24h change reflect oracle values; volume, market cap, and market cap rank are reported as 0.

* **Chore**
  * Removed external CoinGecko integration and its exports.
  * Supported market-data coins narrowed to "akash-network" and "akt".
  * Made chain/testnet configuration access safer to avoid runtime errors.
  * Deployment-grant denom handling tightened; unsupported denoms now error.

* **Tests**
  * Unit and functional tests updated for oracle-only market data and the tightened denom behavior.


apps/api/src/chain/services/denom-exchange/denom-exchange.service.ts
```diff
@@ -1,43 +1,20 @@
...
-          marketCap: response.market_data.market_cap.usd,
-          marketCapRank: response.market_cap_rank,
-          priceChange24h: response.market_data.price_change_24h,
-          priceChangePercentage24: response.market_data.price_change_percentage_24h
-        };
-      }
-
       const legacyToNewMapping: Record<string, string> = {
         "akash-network": "akt",
         "usd-coin": "usdc"
```
baktun14: we don't need usdc anymore

apps/deploy-web/src/chains/akash-testnet.ts
```diff
@@ -5,14 +5,22 @@ import { akash, akashAssetList } from "./akash";
 
 export const akashTestnet: Chain = {
   ...akash,
-  chain_id: netConfig.mapped("testnet"),
+  chain_id: safe(() => netConfig.mapped("testnet")) ?? "testnet",
```
baktun14, line 8: perhaps we could handle this safe paring in the mapped function?

baktun14 (approved):

apps/deploy-web/src/chains/akash-testnet.ts
stalniy: I'd like to postpone this decision because it affects only testnet which appears for short period of time and then disappears. And this is the only place we currently have the issue.

apps/api/src/chain/services/denom-exchange/denom-exchange.service.ts
stalniy: I left it here because I'm not sure whether some website or app fetches usdc price and just wanted to keep protocol backward compatible.


I also cannot find `/v1/market-data/{coin}` calls in our apps, we could simplify interface, should I remove it coin parameter as well?

apps/deploy-web/src/chains/akash-testnet.ts
baktun14: Sure

apps/api/src/chain/services/denom-exchange/denom-exchange.service.ts
baktun14: yea I think it makes sense to just have akt price

stalniy: removal of `{coin}` increases amount of changes, so I decided not to touch it in this PR

PR #3001, opened 2026-03-27
fix(stats): filter out incomplete last snapshot from graph views

baktun14:
## Why

The last snapshot in the time series is always in-progress and incomplete, causing misleading dips at the end of all stats graphs (network, BME, and provider views).

## What

- Introduces `useCompletedSnapshots` hook that slices off the final (incomplete) snapshot entry
- Applied to all three graph containers: `graph`, `bme-graph`, and `provider-graph`

## Summary by CodeRabbit

* **Bug Fixes**
  * Graphs and CSV exports now consistently exclude the latest incomplete snapshot, ensuring visualizations and downloads reflect only completed data.
  * Graphs and export controls wait until the completed-snapshot subset is available before rendering or enabling CSV download, preventing partial or inconsistent outputs.


stalniy (approved):
don't really understand what it fixes. Would be nice to have some screenshots. Also just a thought, maybe it makes sense to exclude the latest snapshot on API side?

baktun14:
> don't really understand what it fixes. Would be nice to have some screenshots. Also just a thought, maybe it makes sense to exclude the latest snapshot on API side?

On any of the graph charts, we used to show only the previous day to not have daily stats reset to 0 on the chart, it got changed while fixing a visual bug in a previous pr. https://stats.akash.network/graph/daily-usd-spent

Issue #1550, opened 2025-06-24
Utilize Nodejs permission model to restrict access to host machine


stalniy:
## Why

The [Node.js permission model](<https://nodejs.org/api/permissions.html>) allows improving security of our applications by restricting access to the host machine — limiting what each service can read, write, or execute beyond its intended scope.

## What

Enable the Node.js permission model for every console app to enforce least-privilege access at runtime. Each app should only be granted the permissions it legitimately needs.

- [X] provider-proxy
- [X] tx-signer
- [ ] console-api
- [X] notifications
- [X] indexer

PR #3024, opened 2026-03-30
feat: enable node.js permission model for provider-proxy

stalniy:
## Why

Ref https://github.com/akash-network/console/issues/1550

Enable the Node.js Permission Model (`--permission`) on provider-proxy to restrict filesystem access at runtime. This is a defense-in-depth security measure that prevents the process from reading or writing files outside explicitly allowed paths, and
blocks spawning child processes, worker threads, and loading native addons.

## What

- Updated the `prod` script in `apps/provider-proxy/package.json` to run with `--permission` and minimal `--allow-fs-read` paths
(`./dist/`, `./env/`, `./node_modules/`, `../../node_modules/`, `../../packages/env-loader/`)
- No write, worker, child-process, or addon permissions are granted — provider-proxy only needs read access

## Summary by CodeRabbit

* **Chores**
  * Simplified the development watch command to rely on the build tool's default entry handling while preserving watch behavior.
  * Production start now runs in a stricter Node sandbox with explicit read permissions and source maps enabled for better diagnostics.
  * Watch-mode builds now invoke the updated production start command automatically after successful builds.

### DIFF
--- a/apps/provider-proxy/package.json
+++ b/apps/provider-proxy/package.json
@@ -8,11 +8,11 @@
   "scripts": {
     "build": "NODE_ENV=production tsup && tsc -p tsconfig.build.json",
     "build:container": "dc build provider-proxy",
-    "dev": "tsup src/server.ts --watch",
+    "dev": "tsup --watch",
     "dev:inspect": "NODE_OPTIONS=\"--inspect\" npm run dev",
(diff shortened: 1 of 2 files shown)


ygrishajev (approved):
💪

PR #3038, opened 2026-04-02
fix(deployment): require state filter for offset-based pagination

ygrishajev:
## Why

The Akash blockchain endpoint `/akash/deployment/v1beta4/deployments/list` rejects requests that use `pagination.offset` without `filters.state`, returning a 400 error:

> invalid request parameters. if offset is set, filter.state must be provided

The `listWithResources` route (`/v1/addresses/:address/deployments/:skip/:limit`) always provides `skip` (offset) as a path param but `status` was an optional query param, so requests without `status` were failing.

## What

**BREAKING CHANGE**: The `status` query parameter is now **required** on `GET /v1/addresses/{address}/deployments/{skip}/{limit}`. Requests without `?status=active` or `?status=closed` will return 400.

- Made `status` required in the `ListWithResourcesQuerySchema` (OpenAPI)
- Made `status` required in `DeploymentReaderService.listWithResources` signature, removing the silent `"active"` default
- Added `FindAllParams` union type in `packages/http-sdk` that enforces `state` is required when `pagination.offset` is set — catches violations at compile time
- Updated `list` method to branch on `skip` for type narrowing
- Updated stats-web `AddressDeployments` to default the status filter to `"active"` instead of `"*"` so it always sends a valid status
- Updated empty state message to say "no active deployments"
- Added unit tests for `DeploymentHttpService.findAll` and `DeploymentReaderService.listWithResources`
- Added functional test for missing status returning 400

## Release Notes

* **Improvements**
  * Enhanced deployment filtering to require explicit status selection in API calls
  * Updated default deployment view to display active deployments
  * Refined trial deployment limit calculation to count only active deployments
  * Improved pagination handling for better consistency across deployment list queries

* **Bug Fixes**
  * Fixed validation to properly enforce status parameter requirements in deployment endpoints

### DIFF
--- a/apps/api/src/deployment/http-schemas/deployment.schema.ts
+++ b/apps/api/src/deployment/http-schemas/deployment.schema.ts
@@ -166,7 +166,7 @@ export const ListWithResourcesParamsSchema = z.object({
 });
 
 export const ListWithResourcesQuerySchema = z.object({
-  status: z.enum(["active", "closed"]).optional().openapi({
+  status: z.enum(["active", "closed"]).openapi({
     description: "Filter by status",
     example: "closed"
   }),
--- a/apps/api/src/billing/services/trial-validation/trial-validation.service.spec.ts
+++ b/apps/api/src/billing/services/trial-validation/trial-validation.service.spec.ts
@@ -1,0 +1,84 @@
+import { MsgCreateDeployment } from "@akashnetwork/chain-sdk/private-types/akash.v1beta4";
+import { MsgCreateLease } from "@akashnetwork/chain-sdk/private-types/akash.v1beta5";
+import type { EncodeObject } from "@cosmjs/proto-signing";
+import { mock } from "vitest-mock-extended";
+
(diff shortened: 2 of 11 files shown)


apps/api/src/deployment/http-schemas/deployment.schema.ts
```diff
@@ -166,7 +166,7 @@ export const ListWithResourcesParamsSchema = z.object({
 });
 
 export const ListWithResourcesQuerySchema = z.object({
-  status: z.enum(["active", "closed"]).optional().openapi({
+  status: z.enum(["active", "closed"]).openapi({
```
baktun14, line 169: question: I think the status is still optional

baktun14 (approved):

apps/api/src/deployment/http-schemas/deployment.schema.ts
ygrishajev: It's required now — the diff drops `.optional()` on this line. The functional test in `apps/api/test/functional/addresses.spec.ts` pins the 400 when `status` is missing.

PR #3050, opened 2026-04-09
fix(deployment): reduce noisy websocket errors in Sentry

baktun14:
## Why

Sentry issue DEPLOY-WEB-27J is escalating — 595 occurrences across 79 users in ~2 days. The `Error: websocket error` fires from the deployment logs WebSocket stream (`followLogs`) on `/deployments/[dseq]`.

Most of these errors are caused by users navigating away from the LOGS tab, which aborts the WebSocket connection and triggers the error handler unnecessarily. The remaining genuine errors have a generic message that makes Sentry triage difficult.

Fixes CON-204

## What

- **Skip Sentry reporting when abort signal has fired** — when the `useEffect` cleanup runs (user navigates away), the abort controller fires, closing the WS. Previously this still reported to Sentry; now we check `abortController.signal.aborted` before reporting.
- **Downgrade severity to `warning`** — genuine WebSocket connection failures (provider down, network issue) are transient and expected in a streaming context; `warning` is more appropriate than `error`.
- **Include close code and reason in error message** — instead of the generic `"websocket error"`, the message now reads `"websocket error: 1006 unknown reason"`, making Sentry grouping and triage much easier.

## Summary by CodeRabbit

* **Bug Fixes**
  * Deployment log streaming: avoid updating loading/connection indicators and prevent duplicate/error-level reports when a log request is intentionally cancelled — such cases are now reported as warnings.
  * Websocket errors: improved error messages include close-event reasons when available and use a more consistent, standardized message for non-close errors.


apps/deploy-web/src/components/deployments/DeploymentLogs.tsx
```diff
@@ -135,8 +135,11 @@ export const DeploymentLogs: React.FunctionComponent<Props> = ({ leases, selecte
...
       onLogReceived
     ).catch(error => {
+      if (abortController.signal.aborted) return;
+
       errorHandler.reportError({
         error,
+        severity: "warning",
         tags: { category: "deployments", label: "followLogs" }
       });
     });
```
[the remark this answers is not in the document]
baktun14: Fixed in upcoming commit — added `setIsLoadingLogs(false)` and `setIsConnectionEstablished(false)` in the non-abort catch path.

apps/deploy-web/src/components/deployments/DeploymentLogs.tsx
```diff
@@ -308,3 +306,21 @@ export const DeploymentLogs: React.FunctionComponent<Props> = ({ leases, selecte
     </div>
   );
 };
+
+export function handleLogsStreamError(
+  error: unknown,
+  signal: AbortSignal,
+  errorHandler: ErrorHandlerService,
+  setIsLoadingLogs: (v: boolean) => void,
+  setIsConnectionEstablished: (v: boolean) => void
```
ygrishajev: passing around local state setters may not be a good idea and eventually cause issues. I'd just put this fn inside the component and used them directly.

apps/deploy-web/src/components/deployments/DeploymentLogs.tsx
```diff
@@ -308,3 +306,21 @@ export const DeploymentLogs: React.FunctionComponent<Props> = ({ leases, selecte
...
+): void {
+  if (signal.aborted) return;
```
ygrishajev: why don't you want to report errors if signal is aborted?

apps/deploy-web/src/lib/websocket/createWebsocket.ts
```diff
@@ -137,7 +137,8 @@ export function waitForEvent<T>(target: WsEvents, type: string): Promise<T | und
 
 function createWsError(event: Event): Error {
   if (event instanceof CloseEvent) {
-    return new Error("websocket error", {
+    const reason = event.reason || "unknown reason";
+    return new Error(`websocket error: ${event.code} ${reason}`, {
```
ygrishajev: this may result in reporting various websocket errors and those would not be grouped together. perhaps a better idea would be sending cause field as extra in the reporter.

apps/deploy-web/src/components/deployments/DeploymentLogs.tsx
baktun14: I think this is what is causing the spam of this error no? Why do we want to report an aborted error?

apps/deploy-web/src/components/deployments/DeploymentLogs.tsx
baktun14: Sure

apps/deploy-web/src/components/deployments/DeploymentLogs.tsx
ygrishajev: honestly, idk. i'm just asking. my concern is we could miss some important ones

apps/deploy-web/src/components/deployments/DeploymentLogs.tsx
```diff
@@ -135,9 +135,16 @@ export const DeploymentLogs: React.FunctionComponent<Props> = ({ leases, selecte
...
+        setIsLoadingLogs(false);
+        setIsConnectionEstablished(false);
+      }
+
       errorHandler.reportError({
         error,
-        tags: { category: "deployments", label: "followLogs" }
+        severity: abortController.signal.aborted ? "warning" : "error",
+        tags: { category: "deployments", label: "followLogs" },
+        cause: error instanceof Error ? (error as Error & { cause?: unknown }).cause : undefined
```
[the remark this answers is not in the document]
baktun14: We intentionally kept reporting for aborted streams — the reviewer raised a concern about silently missing important errors. The current approach differentiates by severity: aborted streams get `"warning"`, real failures get `"error"`. This way nothing is silenced, but aborted noise is clearly separated in Sentry. What's your take on this tradeoff vs silencing aborts entirely?

ygrishajev (approved):

apps/deploy-web/src/components/deployments/DeploymentLogs.tsx
stalniy: no need to log cause explicitly, sentry walks over `.cause`s, additionally our logger logs error as well and that info becomes a part of sentry issue breadcrumbs

apps/deploy-web/src/components/deployments/DeploymentLogs.tsx
```diff
@@ -135,9 +135,16 @@ export const DeploymentLogs: React.FunctionComponent<Props> = ({ leases, selecte
       }),
       onLogReceived
     ).catch(error => {
+      if (!abortController.signal.aborted) {
+        setIsLoadingLogs(false);
+        setIsConnectionEstablished(false);
+      }
+
       errorHandler.reportError({
```
stalniy, line 143: **thought:** I doubt that this PR actually reduces websocket errors because all `AbortError`s are ignored at errorHandler level -> https://github.com/akash-network/console/blob/87fe42b32f04b49b1981d222e28ce56b4bd93aa1/apps/deploy-web/src/services/error-handler/error-handler.service.ts#L24

stalniy: browser websocket on(error) event doesn't expose details about the actual issue. That's why we probably need to send additional http request to websocket, in case of error to get the actual error details (for example, 404, provider network issues, provider is down, etc)

stalniy: I tested it in console-beta and I was able to reproduce the error case for websocket. it happens websocket is closed before the connection has been established.

I think that we just need to ignore this situation in websocket abstraction

baktun14: Good to know

stalniy (approved):

PR #3067, opened 2026-04-15
feat: expose providers host uris by attributes and signatures

stalniy:
## Why

ref CON-220


baktun14 (approved):

stalniy:
this one is on hold until we agree on the implementation path

PR #3097, opened 2026-04-22
feat(deployment): add shell-exec endpoint for synchronous command execution

### LINKED ISSUE #3079: Add synchronous shell-exec endpoint for API-token access
## Why

API consumers (CI/CD pipelines, scripts, SDK users) cannot execute commands inside deployment containers via the Console API. Shell access currently requires a direct WebSocket connection through `provider-proxy` with provider-level credentials (mTLS/JWT). This forces automation workflows to manage WebSocket sessions and provider certificates manually.

Adding an HTTP exec endpoint to `apps/api` would let API-token holders run one-shot commands and receive structured output.

**User story:** As a DevOps engineer with a Console API key, I want to execute a command in my deployment container via a REST call, so that I can automate health checks and maintenance without managing WebSocket connections or provider credentials.

### Key use case: post-deploy secret injection

A primary motivation is enabling **secrets to be injected after deployment** rather than baked into the SDL. Today, secrets (DB passwords, API keys, TLS certificates, service mesh tokens) must be embedded in the SDL at deploy time, which means they are written to the Akash on-chain record.

With shell-exec, containers can start in a "waiting for init" state and receive secrets only after the deployment is live:

```
Deploy container (no secrets in SDL)
    ↓
Container starts → waits for secrets (polling /run/secrets/ or similar)
    ↓
shell-exec pushes secrets into the container
    ↓
Container detects secrets → proceeds with full initialization
```

Example container entrypoint pattern:

```bash
#!/bin/sh
until [ -f /run/secrets/db_password ] && [ -f /run/secrets/api_key ]; do
  echo "Waiting for secrets..."
  sleep 2
done
exec myapp
```

This pattern also applies to infrastructure-level secrets (e.g., service mesh gossip keys, ACL bootstrap tokens, TLS certificates) that would otherwise require a full close + redeploy cycle to rotate. The wait loop should include a hard timeout (e.g., 120s) to surface failures cleanly.

### Proposed endpoint

```
POST /v1/deployments/{dseq}/leases/{gseq}/{oseq}/shell-exec
```

**Request body:**
```json
{
  "command": "string",
  "service": "string",
  "timeout": 60
}
```

**Response body:**
```json
{
  "stdout": "string",
  "stderr": "string",
  "exitCode": 0,
  "truncated": false
}
```

Auth: `SECURITY_BEARER_OR_API_KEY` (existing pattern).

### Secret injection coverage

| Scenario | Without shell-exec | With shell-exec |
|---|---|---|
| DB credentials / API keys | ❌ Baked into SDL (on-chain) | ✅ Injected post-deploy via file write |
| TLS certificate rotation | ❌ Requires redeploy | ✅ Write cert + trigger reload signal |
| Service mesh tokens (e.g., Consul ACL) | ❌ Requires redeploy | ✅ Init pattern: container waits, shell-exec pushes |
| Image pull credentials | ❌ Requires redeploy | ❌ Still requires redeploy |
| Initial env vars (process-level) | ❌ Deploy-time only | ❌ Cannot update live env vars |

### Acceptance criteria

- [ ] Given a valid API key and owned deployment, when the endpoint is called with a command and service name, then stdout, stderr, and exitCode are returned in the response body
- [ ] Given an unauthenticated request, the endpoint returns 401
- [ ] Given a deployment the user does not own, the endpoint returns 403
- [ ] Given a command that exceeds the timeout (default 60s), the connection is closed and a timeout error is returned
- [ ] Given output exceeding 1MB, the response is truncated and includes a `truncated: true` flag
- [ ] The endpoint is documented in the OpenAPI spec with request/response schemas

### Implementation approach

1. **Blocking open question:** Verify provider shell endpoint works with non-interactive mode (`stdin=0`, `tty=0`) and determine required JWT scope
2. Add `ShellExecService` that opens a WebSocket to provider-proxy, sends the command, collects output frames, and returns structured result
3. Extend `ProviderJwtTokenService` to generate tokens with shell-exec scope (if providers enforce scopes)
4. Create POST route with Zod request/response schemas and OpenAPI documentation
5. Add ownership validation via existing CASL ability checks on the deployment
6. Write unit tests for `ShellExecService` and functional tests for the route

### Size estimate

**Medium** — API auth, provider-proxy communication, and route patterns all exist. The novel part is the WebSocket-to-HTTP bridge in `ShellExecService`, but `downloadFileFromShell` in the frontend already demonstrates the exact pattern.

### Not doing (in this issue)

- Interactive shell (WebSocket upgrade on API)
- Streaming output (SSE)
- File upload/download through this endpoint
- Concurrent command execution
- Shell session persistence
- Provider-side changes
- Updating live process environment variables (requires redeploy)

jobordu:
See PR description at https://github.com/jobordu/console/pull/1

Closes #3079

## Release Notes

* **New Features**
  * Added a new `POST /v1/deployments/{dseq}/leases/{gseq}/{oseq}/shell-exec` endpoint to run commands on active deployment leases.
  * Added request/response validation and OpenAPI documentation for the shell-exec contract (including stdin, stdout/stderr, exit codes, and truncation).
* **Tests**
  * Added controller and service test suites covering error mapping, lease/provider selection, stdin streaming, timeouts, auth-expiry handling, and output truncation.
* **Chores**
  * Updated websocket support by adding `ws` and its TypeScript typings to the API package.

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -106,6 +106,7 @@
     "tsyringe": "^4.10.0",
     "unleash-client": "^6.6.0",
     "uuid": "^9.0.1",
+    "ws": "^7.5.10",
     "zod": "3.*"
   },
   "devDependencies": {
@@ -127,6 +128,7 @@
     "@types/semver": "^7.5.2",
     "@types/tar": "^6.1.13",
     "@types/uuid": "^8.3.1",
+    "@types/ws": "^8.5.4",
     "@typescript-eslint/eslint-plugin": "^7.12.0",
     "@vitest/coverage-v8": "^4.0.18",
     "drizzle-kit": "^0.22.7",
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -106,6 +106,7 @@
     "ts-results": "^3.3.0",
     "tsyringe": "^4.10.0",
     "unleash-client": "^6.6.0",
+    "ws": "^8.18.2",
     "zod": "3.*"
   },
   "devDependencies": {
@@ -126,6 +127,7 @@
     "@types/pg": "^8.11.6",
     "@types/semver": "^7.5.2",
     "@types/tar": "^6.1.13",
+    "@types/ws": "^8.5.4",
     "@typescript-eslint/eslint-plugin": "^7.12.0",
     "@vitest/coverage-v8": "^4.1.5",
     "drizzle-kit": "^0.22.7",
--- a/apps/api/src/deployment/controllers/shell-exec/shell-exec.controller.spec.ts
+++ b/apps/api/src/deployment/controllers/shell-exec/shell-exec.controller.spec.ts
@@ -1,0 +1,213 @@
+import { Err, Ok } from "ts-results";
+import { container } from "tsyringe";
+import { describe, expect, it } from "vitest";
+import { mock } from "vitest-mock-extended";
+
+import { AuthService } from "@src/auth/services/auth.service";
+import type { WalletReaderService } from "@src/billing/services/wallet-reader/wallet-reader.service";
+import type { DeploymentReaderService } from "@src/deployment/services/deployment-reader/deployment-reader.service";
+import type { ShellExecService } from "@src/deployment/services/shell-exec/shell-exec.service";
+import type { ProviderService } from "@src/provider/services/provider/provider.service";
+import { ShellExecController } from "./shell-exec.controller";
+
+import { createUser } from "@test/seeders/user.seeder";
+
+describe(ShellExecController.name, () => {
+  it("throws 404 when deployment not found", async () => {
+    const { controller, deploymentReaderService } = setup();
+    deploymentReaderService.findByUserIdAndDseq.mockResolvedValue(undefined as never);
+
+    const error = await captureError(() => controller.exec({ dseq: "1234", gseq: 1, oseq: 1, command: "ls", service: "web", timeout: 60 }));
+
+    expect(error.status).toBe(404);
+    expect(error.message).toBe("Deployment not found");
+  });
+
+  it("throws 404 when no lease matches the provided gseq and oseq", async () => {
+    const { controller } = setup();
+
+    const error = await captureError(() => controller.exec({ dseq: "1234", gseq: 99, oseq: 99, command: "ls", service: "web", timeout: 60 }));
+
+    expect(error.status).toBe(404);
+    expect(error.message).toBe("Lease not found");
+  });
+
+  it("throws 500 when lease provider address is an empty string", async () => {
+    const { controller } = setup({ provider: "" });
+
+    const error = await captureError(() => controller.exec({ dseq: "1234", gseq: 1, oseq: 1, command: "ls", service: "web", timeout: 60 }));
+
+    expect(error.status).toBe(500);
+    expect(error.message).toBe("Lease provider address not found");
+  });
+
+  it("throws 400 when lease state is not active", async () => {
+    const { controller } = setup({ state: "closed" });
+
+    const error = await captureError(() => controller.exec({ dseq: "1234", gseq: 1, oseq: 1, command: "ls", service: "web", timeout: 60 }));
+
+    expect(error.status).toBe(400);
+    expect(error.message).toBe("Lease is not active");
+  });
+
+  it("throws 502 when shell exec service returns an error result", async () => {
+    const { controller, shellExecService } = setup();
+    shellExecService.execute.mockResolvedValue(Err("Command timed out"));
+
+    const error = await captureError(() => controller.exec({ dseq: "1234", gseq: 1, oseq: 1, command: "ls", service: "web", timeout: 60 }));
+
+    expect(error.status).toBe(502);
+    expect(error.message).toBe("Command execution timed out");
+  });
+
+  it("returns the shell exec result on successful execution", async () => {
+    const { controller, shellExecService } = setup();
+
+    const result = await controller.exec({ dseq: "1234", gseq: 1, oseq: 1, command: "ls", service: "web", timeout: 60 });
+
+    expect(result).toEqual({ stdout: "output", stderr: "", exitCode: 0, truncated: false });
+    expect(shellExecService.execute).toHaveBeenCalledWith({
+      providerBaseUrl: "https://provider.example.com",
+      providerAddress: "akash1provider",
+      dseq: "1234",
+      gseq: 1,
+      oseq: 1,
+      service: "web",
+      command: "ls",
+      timeout: 60,
+      jwtToken: "test-token"
+    });
+  });
+
+  it("throws 404 when provider info lookup returns null", async () => {
+    const { controller, providerService } = setup();
+    providerService.getProvider.mockResolvedValue(null as never);
+
+    const error = await captureError(() => controller.exec({ dseq: "1234", gseq: 1, oseq: 1, command: "ls", service: "web", timeout: 60 }));
+
+    expect(error.status).toBe(404);
+    expect(error.message).toBe("Provider not found");
+  });
+
+  it("throws 404 when deployment has an empty leases array", async () => {
+    const { controller, deploymentReaderService, deployment } = setup();
+    deploymentReaderService.findByUserIdAndDseq.mockResolvedValue({ ...deployment, leases: [] } as never);
+
+    const error = await captureError(() => controller.exec({ dseq: "1234", gseq: 1, oseq: 1, command: "ls", service: "web", timeout: 60 }));
+
+    expect(error.status).toBe(404);
+    expect(error.message).toBe("Lease not found");
+  });
+
+  it("finds the correct lease among multiple leases by gseq and oseq", async () => {
+    const { controller, deploymentReaderService, deployment, shellExecService } = setup();
+
+    const multiLeaseDeployment = {
+      ...deployment,
+      leases: [
+        { ...deployment.leases[0], id: { ...deployment.leases[0].id, gseq: 1, oseq: 1 }, state: "active" },
+        { ...deployment.leases[0], id: { ...deployment.leases[0].id, gseq: 2, oseq: 1, provider: "akash1provider2" }, state: "active" },
+        { ...deployment.leases[0], id: { ...deployment.leases[0].id, gseq: 1, oseq: 2, provider: "akash1provider3" }, state: "active" }
+      ]
+    };
+    deploymentReaderService.findByUserIdAndDseq.mockResolvedValue(multiLeaseDeployment as never);
+
+    const result = await controller.exec({ dseq: "1234", gseq: 2, oseq: 1, command: "ls", service: "web", timeout: 60 });
+
+    expect(result).toEqual({ stdout: "output", stderr: "", exitCode: 0, truncated: false });
+    expect(shellExecService.execute).toHaveBeenCalledWith(expect.objectContaining({ providerAddress: "akash1provider2" }));
+  });
+
+  it("throws 400 when deployment state is closed", async () => {
+    const { controller, deploymentReaderService, deployment } = setup();
+    deploymentReaderService.findByUserIdAndDseq.mockResolvedValue({ ...deployment, deployment: { ...deployment.deployment, state: "closed" } } as never);
+
+    const error = await captureError(() => controller.exec({ dseq: "1234", gseq: 1, oseq: 1, command: "ls", service: "web", timeout: 60 }));
+
+    expect(error.status).toBe(400);
+    expect(error.message).toBe("Deployment is not active");
+  });
+
+  it("throws 502 with stable message when WS connection fails", async () => {
+    const { controller, shellExecService } = setup();
+    shellExecService.execute.mockResolvedValue(Err("WebSocket connection failed: ECONNREFUSED"));
+
+    const error = await captureError(() => controller.exec({ dseq: "1234", gseq: 1, oseq: 1, command: "ls", service: "web", timeout: 60 }));
+
+    expect(error.status).toBe(502);
+    expect(error.message).toBe("Failed to connect to provider");
+  });
+
+  it("throws 502 with stable message when provider returns error", async () => {
+    const { controller, shellExecService } = setup();
+    shellExecService.execute.mockResolvedValue(Err("Provider error: internal error"));
+
+    const error = await captureError(() => controller.exec({ dseq: "1234", gseq: 1, oseq: 1, command: "ls", service: "web", timeout: 60 }));
+
+    expect(error.status).toBe(502);
+    expect(error.message).toBe("Provider returned an error");
+  });
+
+  async function captureError(fn: () => Promise<any>): Promise<any> {
+    try {
+      await fn();
+      throw new Error("Expected function to throw");
+    } catch (error) {
+      return error;
+    }
+  }
+
+  function setup(overrides?: { provider?: string; state?: string }) {
+    const user = createUser();
+    const deploymentReaderService = mock<DeploymentReaderService>();
+    const providerService = mock<ProviderService>();
+    const shellExecService = mock<ShellExecService>();
+    const authService = mock<AuthService>({ currentUser: user });
+    const walletReaderService = mock<WalletReaderService>();
+
+    container.register(AuthService, { useValue: authService });
+
+    const controller = new ShellExecController(deploymentReaderService, providerService, shellExecService, authService, walletReaderService);
+
+    const provider = overrides?.provider ?? "akash1provider";
+    const state = overrides?.state ?? "active";
+
+    const deployment = {
+      deployment: {
+        id: { owner: "akash1owner", dseq: "1234" },
+        state: "active",
+        hash: "abc123",
+        created_at: "12345"
+      },
+      leases: [
+        {
+          id: { owner: "akash1owner", dseq: "1234", gseq: 1, oseq: 1, provider, bseq: 0 },
+          state,
+          price: { denom: "uakt", amount: "100" },
+          created_at: "12345",
+          closed_on: "0",
+          status: null
+        }
+      ],
+      escrow_account: {
+        id: { scope: "deployment", xid: "1234" },
+        state: {
+          owner: "akash1owner",
+          state: "open",
+          transferred: [],
+          settled_at: "12345",
+          funds: [{ denom: "uakt", amount: "1000" }],
+          deposits: []
+        }
+      }
+    };
+
+    deploymentReaderService.findByUserIdAndDseq.mockResolvedValue(deployment as never);
+    walletReaderService.getWalletByUserId.mockResolvedValue({ id: 1, address: "akash1wallet" } as never);
+    providerService.toProviderAuth.mockResolvedValue({ type: "jwt" as const, token: "test-token" });
+    providerService.getProvider.mockResolvedValue({ hostUri: "https://provider.example.com" } as never);
+    shellExecService.execute.mockResolvedValue(new Ok({ stdout: "output", stderr: "", exitCode: 0, truncated: false }));
+
+    return { controller, deploymentReaderService, providerService, shellExecService, authService, walletReaderService, user, deployment };
+  }
+});
--- a/apps/api/src/deployment/controllers/shell-exec/shell-exec.controller.spec.ts
+++ b/apps/api/src/deployment/controllers/shell-exec/shell-exec.controller.spec.ts
@@ -8,7 +8,7 @@ import type { WalletReaderService } from "@src/billing/services/wallet-reader/wa
 import type { DeploymentReaderService } from "@src/deployment/services/deployment-reader/deployment-reader.service";
 import type { ShellExecService } from "@src/deployment/services/shell-exec/shell-exec.service";
 import type { ProviderService } from "@src/provider/services/provider/provider.service";
-import { ShellExecController } from "./shell-exec.controller";
+import { mapExecError, ShellExecController } from "./shell-exec.controller";
 
 import { createUser } from "@test/seeders/user.seeder";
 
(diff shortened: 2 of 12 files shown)


jobordu:
@coderabbitai full review

jobordu:
@baktun14 Ready ! 

btw, tu es Québécois ?

baktun14:
Have you looked at the api docs? https://akash.network/docs/api-documentation/console-api/api-reference/#wss-lease-shell

jobordu:
The intent of this PR isn’t to introduce new shell functionality—the provider WSS lease shell API already exists. The goal is to improve the developer experience by exposing a simple synchronous HTTP endpoint for API consumers, so they don’t need to implement provider JWT acquisition, WebSocket lifecycle management, stream demultiplexing, timeout handling, and output aggregation themselves.

That said, I’d appreciate a closer review of the security model. Since this endpoint effectively exposes shell execution through the Console API, I want to make sure we’re not unintentionally broadening access or introducing privilege-escalation risks compared to the existing WSS flow.

apps/api/package.json
```diff
@@ -106,6 +106,7 @@
     "ts-results": "^3.3.0",
     "tsyringe": "^4.10.0",
     "unleash-client": "^6.6.0",
+    "ws": "^8.18.2",
```
stalniy, line 109: **suggestion:** nodejs 24 supports builtin `WebSocket` client and it should also support custom `headers`

apps/api/src/deployment/routes/shell-exec/shell-exec.router.ts
```diff
@@ -0,0 +1,59 @@
...
+    502: {
+      description: "Provider proxy error"
+    }
+  }
+});
+
+shellExecRouter.openapi(shellExecRoute, async function routeShellExec(c) {
+  const params = c.req.valid("param");
+  const body = c.req.valid("json");
+  const result = await container.resolve(ShellExecController).exec({ ...params, ...body });
```
stalniy, line 74: **question:** what do you think about returning SSE response (Server Sent Events), so then you can receive the whole output in a stream?

apps/api/src/deployment/services/shell-exec/shell-exec.service.ts
```diff
@@ -0,0 +1,284 @@
...
+      ws.on("open", () => {
+        const message: WebSocketOutgoingMessage = {
+          type: "websocket",
+          url,
+          auth,
+          providerAddress: input.providerAddress,
+          isBase64: true
+        };
+        ws.send(JSON.stringify(message));
+      });
```
stalniy: **issue(blocking):** this mixes provider websocket api and provider-proxy websocket api. For now it's better to forward traffic via provider-proxy because it will additionally properly validate providers that use self-signed SSL certificate.

as far as I understood you try to establish direct provider connection which will fail for self signed certs

apps/api/src/deployment/services/shell-exec/shell-exec.service.ts
```diff
@@ -0,0 +1,284 @@
...
+ * `sh -c "echo SECRET=v > /run/secrets/.env"`), which must survive as the three
+ * tokens `["sh", "-c", "echo SECRET=v > /run/secrets/.env"]`.
+ *
+ * Rules: unquoted whitespace delimits tokens (space/tab/newline — matching the
+ * POSIX default `IFS`, so a carriage return is preserved, not a delimiter); single quotes
+ * preserve their contents literally; double quotes preserve contents but allow
+ * `\` to escape `"` and `\`; a backslash outside quotes escapes the next char.
+ * Quote characters are removed from the emitted tokens (shell semantics).
+ */
+export function tokenizeCommand(command: string): string[] {
```
stalniy: **question:** I assume that those commands which will be sent to shell, will be manually written by people. If this assumption is correct, could you please share the use case where using raw command string improves DX or simplifies request?

If API will require user to pass argv style command then we won't need this tokenization function

jobordu:
Thanks for the thorough review — pushed a revision (`b960b54a1`) that addresses all of it.

**What changed**
- **Shell traffic now routes through provider-proxy** (`PROVIDER_PROXY_URL` + the standard `{type:"websocket",…}` envelope) instead of a direct provider connection, so self-signed provider certs are validated by the proxy as before. This also corrected the receive parser to the real proxy frame protocol (`message.message.data`, `LeaseShellCode` 100/101/102/103, error frames handled **before** base64-decode, dual JSON / little-endian-int32 exit-code, `4001/4003` auth-expiry, and read-to-exit after the 1 MB truncation).
- **`command` is now argv (`string[]`)** — the custom tokenizer is gone.
- **Kept the synchronous JSON response** for this iteration (SSE reasoning in-thread).
- **Kept `ws`** (already `^8.18.2`, CVE-free; reasoning in-thread).
- **Added stdin-secret injection**: an optional `stdin` field is streamed as a `104` frame (+ EOF) so secrets stay **out of the provider URL** (which the proxy logs). E.g. `command: ["sh","-c","cat > /run/secrets/.env && chmod 600 /run/secrets/.env"]`, `stdin: "SECRET=…"`. Previously a secret embedded in the command would have been logged in the provider URL.
- CodeRabbit residuals: `dseq` numeric regex, OpenAPI `500` response, provider lookup before JWT mint, typed test mocks.

66 unit tests pass; typecheck + lint clean. Per-point rationale (with pros/cons) is in the threads below. I review this repo asynchronously and can be slow to reply — please merge or push back as you see fit; none of the notes should block your review.

apps/api/src/deployment/services/shell-exec/shell-exec.service.ts
jobordu: You're right — it was mixing the two APIs (direct provider socket + proxy envelope) and would fail on self-signed provider certs. Fixed in `b960b54a1`.

- **Why route through the proxy:** provider-proxy validates the provider cert against its on-chain identity (`CertificateValidator`) and handles self-signed certs — the exact validation a direct connection skips. It's also the path `deploy-web`'s `connectToShell` already uses, and `PROVIDER_PROXY_URL` was already wired into `apps/api`. The only upside of a direct connection was one fewer hop, which isn't worth losing cert validation.
- Now connects to `PROVIDER_PROXY_URL` and sends the `{type:"websocket",…}` envelope; the receive side was corrected to the real proxy frames (markers 100/101/102/103, error-handling before base64-decode, dual JSON/LE-int32 exit-code). Resolved.

apps/api/src/deployment/services/shell-exec/shell-exec.service.ts
jobordu: Agreed — with argv there's no need to tokenize. The API now takes `command: string[]` and `tokenizeCommand` is removed; `buildShellUrl` maps the array straight to `cmd0..cmdN` (matching `connectToShell`).

- **Pros:** no shell-parsing to reimplement/maintain, no quote/escape attack surface, unambiguous, and consistent with the existing client contract.
- **Con:** callers must structure the command as an array — but that's the shape the provider consumes anyway, and it composes cleanly with the new `stdin` field for secret injection (e.g. `["sh","-c","cat > /run/secrets/.env"]` + `stdin`).

apps/api/src/deployment/routes/shell-exec/shell-exec.router.ts
jobordu: Good call for streaming. For this iteration I kept the synchronous JSON response and would propose SSE as a follow-up:

- **Sync-JSON pros:** simplest client contract, trivial error handling, and the endpoint is bounded (≤120 s, 1 MB cap) so buffering the whole result is safe.
- **SSE pros (later):** incremental stdout/stderr for long/verbose commands.
- **SSE cons now:** a different client contract, needs proxy response-buffering disabled (nginx buffers by default), reconnect/partial-error semantics, and a larger test surface — and it's additive/non-breaking to add later.

Happy to prioritize SSE if you'd prefer, but I'd suggest not blocking this on it. (One caveat I've documented: the max timeout should stay under the ingress/LB idle timeout.)

apps/api/package.json
jobordu: You're right that Node's built-in `WebSocket` can do this now (undici's `WebSocketInit.headers` + a custom `dispatcher` for TLS). I kept `ws` for this PR:

- **Keep-`ws` pros:** already CVE-patched (`^8.18.2`), already a dependency, and consistent with `provider-proxy` + `deploy-web`, which both use `ws`.
- **Built-in pros:** drops a dependency.
- **But:** after the proxy-routing change the connection needs **neither** custom headers (auth rides in the envelope) **nor** TLS bypass — so the built-in's advantage doesn't apply here — and its header/dispatcher support is documented only in undici's docs, not on nodejs.org. Net: keeping `ws` is lower-risk with no functional downside. Happy to drop the dep in a separate change if you'd like.

apps/api/src/deployment/controllers/shell-exec/shell-exec.controller.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    const providerAddress = lease.id.provider;
+
+    const wallet = await this.walletReaderService.getWalletByUserId(userId);
+
+    const auth = await this.providerService.toProviderAuth({ walletId: wallet.id, provider: providerAddress }, ["shell"]);
+
+    const providerInfo = await this.providerService.getProvider(providerAddress);
+
+    assert(providerInfo, 404, "Provider not found");
```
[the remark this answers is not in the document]
jobordu: Addressed in `b960b54a1`. `deployment.state === "active"` is now checked before the lease lookup, and `getProvider` + `assert(providerInfo, 404)` run **before** `toProviderAuth`, so no JWT is minted for a non-existent provider. Resolving.

apps/api/src/deployment/controllers/shell-exec/shell-exec.controller.ts
```diff
@@ -0,0 +1,63 @@
...
+      oseq: input.oseq,
+      service: input.service,
+      command: input.command,
+      timeout: input.timeout,
+      jwtToken: auth.token
+    });
+
+    if (!result.ok) {
+      assert(false, 502, result.val);
+    }
```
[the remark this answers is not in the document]
jobordu: Addressed in `b960b54a1`. Service errors are mapped to stable, user-facing messages (timed-out / connect-failed / auth-expired / provider-error / generic) before the 502 — the raw internal `result.val` is never returned to the client. Resolving.

apps/api/src/deployment/http-schemas/shell-exec.schema.ts
```diff
@@ -0,0 +1,24 @@
...
+  dseq: z.string(),
+  gseq: z.number(),
+  oseq: z.number()
+});
+
+export const ShellExecRequestSchema = z.object({
+  command: z.string().min(1),
+  service: z.string().min(1),
+  timeout: z.number().min(1).max(120).default(60)
+});
```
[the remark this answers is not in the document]
jobordu: Addressed in `b960b54a1`. `dseq` is `z.string().regex(/^\d+$/)`, `gseq`/`oseq` are `z.coerce.number().int().nonnegative()`, `command` is now an argv `z.array(z.string().min(1)).min(1).max(64)`, and `timeout` is `.int().min(1).max(120)`. Resolving.

apps/api/src/deployment/services/shell-exec/shell-exec.service.ts
```diff
@@ -0,0 +1,284 @@
...
+
+  if (hasToken) tokens.push(current);
+  return tokens;
+}
+
+export function buildShellUrl(input: Pick<ShellExecInput, "providerBaseUrl" | "dseq" | "gseq" | "oseq" | "service" | "command">): string {
+  const tokens = tokenizeCommand(input.command);
+  const cmdParts = tokens.length > 0 ? tokens.map((token, i) => `&cmd${i}=${encodeURIComponent(token)}`) : [`&cmd0=`];
+  const baseUrl = input.providerBaseUrl.replace(/\/$/, "");
+  return `${baseUrl}/lease/${encodeURIComponent(input.dseq)}/${input.gseq}/${input.oseq}/shell?stdin=0&tty=0&podIndex=0&service=${encodeURIComponent(input.service)}${cmdParts.join("")}`;
```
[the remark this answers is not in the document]
jobordu: No longer applicable as of `b960b54a1`: `tokenizeCommand` was removed and `command` is now `z.array(z.string().min(1)).min(1)`, so blank/empty commands are rejected at the schema boundary. Resolving.

apps/api/src/deployment/controllers/shell-exec/shell-exec.controller.spec.ts
```diff
@@ -0,0 +1,264 @@
...
+        funds: [{ denom: "uakt", amount: "1000" }],
+        deposits: []
+      }
+    }
+  } as unknown as DeploymentResponse;
+}
+
+function createProviderInfo(overrides: Partial<{ hostUri: string }> = {}): ProviderInfo {
+  return { hostUri: overrides.hostUri ?? "https://provider.example.com" } as unknown as ProviderInfo;
+}
```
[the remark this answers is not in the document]
jobordu: Done in `09be47f8e`. The factory helpers (`createLease`/`createDeployment`/`createProviderInfo`) and the wallet resolved value now use `vitest-mock-extended`'s `mock<T>()`; zero `as unknown as` / `as never` casts remain in the file. Resolving.

apps/api/src/deployment/controllers/shell-exec/shell-exec.controller.spec.ts
```diff
@@ -0,0 +1,264 @@
...
+      dseq: "1234",
+      gseq: 1,
+      oseq: 1,
+      command: ["sh", "-c", "cat > /run/secrets/.env"],
+      service: "web",
+      timeout: 60,
+      stdin: "SECRET=value"
+    });
+
+    expect(shellExecService.execute).toHaveBeenCalledWith(expect.objectContaining({ stdin: "SECRET=value" }));
```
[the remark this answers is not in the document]
jobordu: Done in `09be47f8e`. The stdin test secret is now generated dynamically, and the same treatment was applied to the stdin tests in the service spec. Resolving.

apps/api/src/deployment/http-schemas/shell-exec.schema.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export const ShellExecRequestSchema = z.object({
+  command: z.array(z.string().min(1)).min(1).max(64),
+  service: z.string().min(1).max(253),
+  timeout: z.number().int().min(1).max(120).default(60),
+  stdin: z.string().max(1_048_576).optional().openapi({
+    description:
+      'Optional raw UTF-8 data streamed to the command\'s standard input (max 1 MiB). Put secrets HERE (env files, tokens, passwords) — never in `command`, whose tokens are placed in the provider-proxy request URL, which is logged. Example command: ["sh","-c","cat > /run/secrets/.env"].',
+    example: "SECRET=value"
+  })
```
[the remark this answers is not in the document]
jobordu: Done in `09be47f8e`. `stdin` now enforces the 1 MiB cap via `Buffer.byteLength(s, "utf8") <= 1_048_576`, and the service's output-truncation accounting tracks accumulated UTF-8 bytes (`Buffer.byteLength(chunk, "utf8")`) instead of UTF-16 string length. Resolving.

package-lock.json
```diff
@@ -104,6 +104,7 @@
         "ts-results": "^3.3.0",
```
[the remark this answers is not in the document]
jobordu: Reviewed — the current `package-lock.json` diff vs `main` is just two additions, `"ws": "^8.18.2"` and `"@types/ws": "^8.5.4"` (the runtime dep this feature uses + its types), matching `apps/api/package.json`. No downgrades or `node_modules` churn on the branch (the larger diff shown here reflected an earlier transient state). Resolving.

apps/api/src/deployment/http-schemas/shell-exec.schema.ts
```diff
@@ -0,0 +1,33 @@
...
+  oseq: z.coerce.number().int().nonnegative()
+});
+
+export const ShellExecRequestSchema = z.object({
+  command: z.array(z.string().min(1)).min(1).max(64),
+  service: z.string().min(1).max(253),
+  timeout: z.number().int().min(1).max(120).default(60),
+  stdin: z
+    .string()
+    .refine(s => Buffer.byteLength(s, "utf8") <= 1_048_576, { message: "stdin must not exceed 1 MiB (UTF-8 bytes)" })
```
stalniy: **issue(blocking)**: Each in-flight exec can hold up to ~1 MiB of `stdout`/`stderr` (`MAX_OUTPUT_SIZE` in `shell-exec.service.ts`) plus this stdin buffer, and nothing bounds concurrent execs — ~1000 concurrent requests ≈ 1.5 GB resident with no malicious payload.

None of the intended secret-injection payloads (env files, API keys, DB passwords, a TLS cert+key chain, service-mesh tokens) are anywhere near 1 MiB — they are kilobytes. Suggest capping stdin at **16 KiB** and pulling anything larger from inside the container. Also set an explicit `bodyLimit` on the route to match, so the request body limit and this schema limit cannot drift.

apps/api/src/deployment/services/shell-exec/shell-exec.service.spec.ts
```diff
@@ -0,0 +1,631 @@
...
+      }),
+      send: vi.fn(),
+      close: vi.fn(),
+      _trigger(event: string, ...args: unknown[]) {
+        (handlers[event] || []).forEach(h => h(...args));
+      }
+    };
+  }
+
+  function createService() {
```
stalniy: **suggestion:** Per [`setup-function-instead-of-beforeeach`](https://github.com/akash-network/console/blob/main/.claude/instructions/setup-function-instead-of-beforeeach-in-unit-service-level-tests.md), the object-under-test factory should be named `setup()`.  Every other deployment service spec follows this convention.

apps/api/src/deployment/controllers/shell-exec/shell-exec.controller.ts
```diff
@@ -0,0 +1,76 @@
...
+        : result.val.startsWith("WebSocket connection failed")
+          ? "Failed to connect to provider"
+          : result.val.startsWith("Auth expired")
+            ? "Provider authentication expired"
+            : result.val.startsWith("Invalid provider host")
+              ? "Invalid provider host"
+              : result.val.startsWith("Provider error")
+                ? "Provider returned an error"
+                : "Shell execution failed";
+      assert(false, 502, message);
```
stalniy: **suggestion:** This 6-level nested ternary mapping error prefixes to messages/status is hard to read and extend. Prefer a small named helper (e.g. `mapExecError(result.val)` returning `{ status, message }`) backed by a lookup table. It also reads as if all branches are 502, but the timeout case in particular should arguably be a 408/504 rather than 502 (see separate note). And auth expired should be 403. Invalid provider host = 400

apps/api/src/deployment/services/shell-exec/shell-exec.service.ts
```diff
@@ -0,0 +1,363 @@
+import { isIP } from "node:net";
+import { Err, Ok, Result } from "ts-results";
+import { inject, singleton } from "tsyringe";
+import { WebSocket } from "ws";
+
+import { LoggerService } from "@src/core";
+import { DEPLOYMENT_CONFIG, type DeploymentConfig } from "@src/deployment/config/config.provider";
+
+const MAX_OUTPUT_SIZE = 1024 * 1024;
```
stalniy: **suggestion(blocking):** set to 64Kb. In would be enough with a streaming solution. I'd like to avoid having big buffers in memory which can make DoS attack simple

stalniy (changes requested):

apps/api/src/deployment/http-schemas/shell-exec.schema.ts
jobordu: Done in `0157d43fc`. `stdin` is now capped at **16 KiB** (`MAX_STDIN_BYTES`) and the route carries an explicit `bodyLimit` (`SHELL_EXEC_BODY_LIMIT_BYTES`) that is **derived arithmetically** from the schema field caps (`MAX_COMMAND_ARGS·MAX_COMMAND_ARG_BYTES + MAX_STDIN_BYTES + SERVICE_NAME_MAX + JSON_ENVELOPE_OVERHEAD_BYTES`) — so the request-body limit and the schema caps import from one source of truth and can't drift.

`command` args are now bounded per-arg by **UTF-8 byte length** (64 × 1 KiB) rather than `z.string().max()` (which counts UTF-16 code units), so worst-case body size is finite. Combined with the output cap dropping to 64 KiB (see the other thread), steady-state per-exec resident memory is now ~kilobytes rather than ~1 MiB+.

apps/api/src/deployment/services/shell-exec/shell-exec.service.ts
jobordu: Done in `0157d43fc`. `MAX_OUTPUT_SIZE` is now **64 KiB** and exported so the tests assert against the constant instead of a hardcoded literal.

I kept the synchronous buffered response for this iteration since the buffer is now small enough that concurrent execs can't be used as a cheap DoS. I'm happy to follow up with an SSE/streaming variant as a separate, non-breaking change if you'd like incremental stdout/stderr for long-running commands — flagged it as a follow-up rather than folding a client-contract change into this PR.

apps/api/src/deployment/controllers/shell-exec/shell-exec.controller.ts
jobordu: Done in `0157d43fc`. The 6-level ternary is replaced with a pure `mapExecError()` helper backed by an ordered prefix→`{ status, message }` lookup table (unit-tested per prefix). Statuses:

- timeout → **504** (we're a gateway to the provider; 408 would say the *client* was slow to send its request, which isn't the case)
- auth expired → **403**
- provider / ws-connection / closed-without-exit-code → **502**, each with a distinct message so e.g. "Provider connection closed unexpectedly" is triageable separately from a generic provider error

On **invalid provider host** I went with **502** rather than 400: `providerBaseUrl`/`hostUri` is server-derived from on-chain provider data (`providerService.getProvider(providerAddress).hostUri`), not client input — so a malformed host is an upstream/config fault, and a 400 would misattribute blame to a caller who sent a perfectly valid request. It's a one-line change in the table if you'd still prefer 400 here.

The router OpenAPI now documents 403/504/413 as well.

apps/api/src/deployment/services/shell-exec/shell-exec.service.spec.ts
jobordu: Done in `0157d43fc`. Renamed the object-under-test factory `createService()` → `setup()` per the convention.

jobordu:
@stalniy pushed a revision (`0157d43fc`) addressing all four threads. Per-point replies are in-line; summary:

**Memory / DoS (both blocking):**
- `MAX_OUTPUT_SIZE` 1 MiB → **64 KiB** (exported).
- `stdin` → **16 KiB**; `command` args bounded per-arg by **UTF-8 bytes** (64 × 1 KiB).
- Explicit route `bodyLimit` = `SHELL_EXEC_BODY_LIMIT_BYTES`, **derived arithmetically** from the schema field caps so the body limit and schema can't drift.
- Net effect: per-exec resident memory drops from ~1 MiB+ to ~kilobytes; ~1000 concurrent execs is now ~tens of MB, not ~1.5 GB.

**Error mapping:**
- 6-level ternary → pure `mapExecError()` lookup table (unit-tested per prefix). timeout **504**, auth **403**, provider/connection **502** (distinct messages). Invalid-host kept at **502** (host is server-derived on-chain data, not client input — rationale in-thread; trivially switchable to 400 if you prefer). Router OpenAPI documents 403/504/413.

**Tests:** factory `createService()` → `setup()`; truncation tests use the exported constant; new schema spec covers the stdin/command byte caps (incl. a multi-byte/emoji case) and the derived body-limit; `mapExecError` contract test. 85 shell-exec unit tests pass; typecheck + lint clean.

Kept the synchronous buffered response this round (buffer is now small); SSE remains an easy non-breaking follow-up if you'd like incremental output.

PR #3099, opened 2026-04-22
feat(managed-wallet): add managed wallet deployment E2E and refactor test infrastructure

ygrishajev:
## Why

Part of CON-153, CON-154

The managed wallet deployment lifecycle (create, lease, close) had no E2E coverage. The existing test infrastructure used `data-testid` selectors instead of accessible attributes, and page objects had unnecessary inheritance hierarchies.

## What

**New E2E tests:**
- `managed-wallet-deployment.spec.ts` — full deployment lifecycle: login → deploy hello-world → top up ACT if needed via billing → create lease → verify active lease → verify auto-deposit → close → verify closed
- `deploy-self-custody-hello-world.spec.ts` — replaces `deploy-hello-world.spec.ts` with steps and wallet type support
- `custom-container-form.spec.ts` — replaces `deploy-custom-template.spec.ts` with descriptive name

**Test infrastructure refactors:**
- `DeployPage` replaces `DeployBasePage` + 3 subclasses with `walletType: "api" | "extension"` and `withTxAccepted` pattern
- `Sidebar` page object for shared layout navigation (deploy button)
- `BillingPage.submitPayment()` for reusable payment flow
- `DeployPage.openDepositDialog()` extracted for reuse
- `deploy-from-a-template` fixed with `context.newPage()` and named steps
- Provider whitelist switched from hostUri to owner addresses for stability

**Accessibility improvements (replacing data-testid):**
- `DeployOptionBox`: `role="button" aria-label={title}`
- `TemplateList`: hello-world link `aria-label="Hello World"`
- `DeploymentDetailTopBar`: dropdown `aria-label="Deployment actions"`, close item `aria-label="Close deployment"`
- `LeaseRow`: state span `aria-label="Lease N state"`
- `BidRow`: radio `aria-description={provider.owner}` for stable provider selection
- `Editor`: wrapper `aria-label="SDL editor"`
- `SimpleServiceFormControl`: OS image select `aria-label="OS image"`
- `AccountMenu`: trigger `aria-label="Account menu"` (from #3081)

**Other fixes:**
- Permanent diagnostic hints in mailsac-code strategy error messages
- Increased wallet connection timeout for flaky extension init
- Wait for wallet indicator after managed wallet login
- Editor unit test updated for aria-label change

## Summary by CodeRabbit

* **Accessibility**
  * Replaced test-only selectors with ARIA labels/roles, added keyboard activation for interactive cards, and added descriptive metadata to provider selections.

* **Infrastructure**
[...]

PR #3102, opened 2026-04-23
feat(alert): add E2E deployment alerts test and fix threshold rounding

ygrishajev:
## Why

Part of CON-155

Deployment alerts configuration and the global alerts page had no E2E coverage. Additionally, the threshold display had a floating point rounding bug, and the unsaved changes detection had a race condition.

## What

**E2E test (`managed-wallet-alerts.spec.ts`):**
- Deploy hello-world → open Alerts tab → verify both escrow balance and deployment close alerts enabled by default
- Update escrow threshold → save → verify value persisted
- Toggle escrow alert off → save → re-enable → save
- Navigate to global alerts page → verify alerts listed → toggle from list
- Navigate back to deployment via DSEQ link → close deployment

**Page objects:**
- `AlertsPage` — global alerts list with tab navigation and alert toggle
- `DeploymentAlertsForm` — deployment detail alerts form (scoped via `aria-label` on Fieldset sections)
- `DeployPage.openTab(name)` — generic tab opener for deployment details
- `Sidebar.openAlerts()` — sidebar navigation

**Shared action:**
- `createManagedDeployment()` extracted to `actions/deploy.ts` — handles template selection, deposit dialog, top-up if needed, lease creation. Used by both deployment and alerts tests.

**App fixes:**
- `DeploymentAlertsContainer`: replace `ceilDecimal` with `roundDecimal(value * price, 6)` for threshold display — `ceilDecimal` added `Number.EPSILON` which caused `0.075` to display as `0.076`
- `DeploymentAlertsView`: reset `hasChanges` and notify parent after successful save to prevent stale navigation guard

**Accessibility:**
- `Fieldset` component: added `aria-label` prop passthrough
- `DeploymentBalanceAlert`: `aria-label="Escrow Balance alert"`
- `DeploymentCloseAlert`: `aria-label="Deployment Close alert"`

Depends on #3099

## Summary by CodeRabbit

* **Bug Fixes**
  * Deployment alert threshold now uses quantized rounding for more accurate balance display.

* **Accessibility**
  * Fieldset containers in deployment alert UIs accept and expose aria-labelledby references; specific alert titles wired for improved screen‑reader clarity.

* **Tests**
  * Added end-to-end coverage for managed‑wallet deployments and alerts.
  * Introduced reusable test helper flows and page‑object helpers (Alerts, DeploymentAlertsForm, Sidebar, DeployPage) to drive and validate alert workflows.

### DIFF
--- a/apps/deploy-web/src/components/shared/Fieldset.tsx
+++ b/apps/deploy-web/src/components/shared/Fieldset.tsx
@@ -7,12 +7,12 @@ type Props = {
   subLabel?: string;
   className?: string;
   children: React.ReactNode;
-  "aria-label"?: string;
+  "aria-labelledby"?: string;
 };
 
-export const Fieldset: React.FunctionComponent<Props> = ({ label, subLabel, className = "", children, "aria-label": ariaLabel }) => {
+export const Fieldset: React.FunctionComponent<Props> = ({ label, subLabel, className = "", children, "aria-labelledby": ariaLabelledBy }) => {
   return (
-    <Card className={className} aria-label={ariaLabel}>
+    <Card className={className} aria-labelledby={ariaLabelledBy}>
       <CardHeader>
         {label}
         {subLabel && <p className="text-gray-400">{subLabel}</p>}
--- a/apps/deploy-web/src/components/shared/Fieldset.tsx
+++ b/apps/deploy-web/src/components/shared/Fieldset.tsx
@@ -7,11 +7,12 @@ type Props = {
   subLabel?: string;
   className?: string;
   children: React.ReactNode;
+  "aria-label"?: string;
 };
 
-export const Fieldset: React.FunctionComponent<Props> = ({ label, subLabel, className = "", children }) => {
+export const Fieldset: React.FunctionComponent<Props> = ({ label, subLabel, className = "", children, "aria-label": ariaLabel }) => {
   return (
-    <Card className={className}>
+    <Card className={className} aria-label={ariaLabel}>
       <CardHeader>
         {label}
         {subLabel && <p className="text-gray-400">{subLabel}</p>}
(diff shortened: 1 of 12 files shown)


apps/deploy-web/src/components/shared/Fieldset.tsx
```diff
@@ -7,11 +7,12 @@ type Props = {
...
   className?: string;
   children: React.ReactNode;
+  "aria-label"?: string;
 };
 
-export const Fieldset: React.FunctionComponent<Props> = ({ label, subLabel, className = "", children }) => {
+export const Fieldset: React.FunctionComponent<Props> = ({ label, subLabel, className = "", children, "aria-label": ariaLabel }) => {
   return (
-    <Card className={className}>
+    <Card className={className} aria-label={ariaLabel}>
```
stalniy: **question:** why not a `label`?

stalniy (approved):

apps/deploy-web/src/components/shared/Fieldset.tsx
stalniy: probably also possible to use `aria-labeledby` https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-labelledby

ygrishajev: `<label>` is semantically for form controls (`<input>`, `<select>`), not for section containers — so it doesn't apply here.

That leaves `aria-label` vs `aria-labelledby`. `aria-labelledby` would require adding an `id` to the heading and referencing it — coupling two elements via a string id that can silently break if one side is renamed. `aria-label` is self-contained on one element and simpler for this use case. wdyt?

PR #3130, opened 2026-05-03
fix(onboarding): prevent signup flash on refresh for authenticated users

baktun14:
## Why

Authenticated users (logged in, with a managed wallet) saw the signup / free-trial onboarding screen briefly flash on every page refresh before being redirected back to where they were. This made the app feel broken and could mislead returning users into thinking they'd been signed out.

The flash is caused by a race in `WalletProvider`: `isLoading` reads `false` in the window between Auth0 hydrating and the managed-wallet query resolving. This happens because `selectedWalletType` defaults to `"custodial"` for the very first render before `atomWithStorage` rehydrates from `localStorage`, and the existing `isLoading` derivation only counted `isManagedWalletLoading` when `selectedWalletType === "managed"`. During that window `OnboardingRedirectEffect` saw `user.userId && !hasManagedWallet && !isWalletConnected` and called `router.replace(\"/signup\")` — until the managed-wallet query settled and `OnboardingContainer` navigated back.

Closes CON-279

## What

Keep `isLoading` true while an authenticated user's managed-wallet query is still loading. Both `OnboardingRedirectEffect` and `OnboardingContainer` already gate on `isWalletLoading`, so fixing the source closes the race for every consumer.

The derivation is extracted into a small pure helper (`deriveWalletIsLoading`) so it can be unit-tested without the full `WalletProvider` DI surface. `isManagedWalletLoading` is React Query's `isInitialLoading`, which becomes `false` once the query has succeeded, errored, or been disabled — so this won't hang.

## Test plan
- [ ] Sign in with a managed-wallet account; navigate to `/deployments`, hard refresh — no flash of the signup / free-trial screen
- [ ] Repeat on `/templates` and `/settings` — no flash
- [ ] Sign in with custodial wallet connected — no regression on refresh
- [ ] Brand-new authenticated user with no wallet — still correctly redirected to `/signup`
- [ ] Signed-out user — `/signup` still reachable directly
- [ ] `deriveWalletIsLoading.spec.ts` covers the race-condition case (5 tests, all pass)
- [ ] `OnboardingRedirectEffect.spec.tsx` — all 11 tests still pass

## Summary by CodeRabbit

* **Refactor**
  * Reorganized wallet loading state calculation logic for improved maintainability.

* **Tests**
  * Added comprehensive test coverage for wallet loading state calculation, including various authentication and wallet type scenarios.

### DIFF
--- a/apps/deploy-web/src/context/WalletProvider/useIsWalletLoading.ts
+++ b/apps/deploy-web/src/context/WalletProvider/useIsWalletLoading.ts
@@ -1,4 +1,4 @@
-type SelectedWalletType = "managed" | "custodial";
+import type { SelectedWalletType } from "@src/store/walletStore";
 
 export type UseIsWalletLoadingInput = {
   hasAuthenticatedUserId: boolean;
@@ -8,9 +8,6 @@ export type UseIsWalletLoadingInput = {
 };
 
 export const useIsWalletLoading = (input: UseIsWalletLoadingInput): boolean => {
-  return (
-    (input.hasAuthenticatedUserId && input.isManagedWalletLoading) ||
-    (input.selectedWalletType === "managed" && input.isManagedWalletLoading) ||
-    (input.selectedWalletType === "custodial" && input.isCustodialConnecting)
-  );
+  const isManagedQueryRelevant = input.hasAuthenticatedUserId || input.selectedWalletType === "managed";
+  return (isManagedQueryRelevant && input.isManagedWalletLoading) || (input.selectedWalletType === "custodial" && input.isCustodialConnecting);
 };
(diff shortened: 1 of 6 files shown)


apps/deploy-web/src/context/WalletProvider/useIsWalletLoading.ts
```diff
@@ -0,0 +1,13 @@
+import type { SelectedWalletType } from "@src/store/walletStore";
+
+export type UseIsWalletLoadingInput = {
+  hasAuthenticatedUserId: boolean;
+  selectedWalletType: SelectedWalletType;
+  isManagedWalletLoading: boolean;
+  isCustodialConnecting: boolean;
+};
+
+export const useIsWalletLoading = (input: UseIsWalletLoadingInput): boolean => {
```
ygrishajev: nit: this is named as if it's a react hook, however it's a simple pure function. I see a benefit in it being independent (testing), however it could be named to avoid confusion without a `use` prefix.

On the contrary it could rely on `useMemo` and become a hook

ygrishajev (approved):

apps/deploy-web/src/context/WalletProvider/useIsWalletLoading.ts
baktun14: Renamed to `deriveWalletIsLoading` in a4fc64435.

PR #3131, opened 2026-05-03
feat(wallet): support links and persistent dismiss in generic banner

baktun14:
## Why

Part of CON-266. Before pulling self-custody features out of Console, the team needs a way to surface an in-app announcement that:

- links out to the Console Air repo and the blog post
- stays dismissed across page reloads
- can be turned on/off and reworded by ops via Unleash

Rather than introduce a one-off banner component, this enhances the existing `generic_banner` flag so any future announcement can do the same.

## What

- Extends `GenericBannerDetails` payload with optional `links: { label, href }[]` and `dismissId: string`.
- `GenericBanner` renders payload links as `target="_blank" rel="noopener noreferrer"` anchors after the message.
- Extracts `useGenericBannerVisibility` from `useTopBanner`. When `dismissId` is set, dismiss is persisted in `localStorage` under `generic_banner_dismissed:<dismissId>`; if `dismissId` is omitted, the prior session-only dismiss behavior is preserved. Changing `dismissId` re-shows the banner (new announcement = fresh dismiss state).
- Adds `useTopBanner.spec.tsx` covering: flag on/off, prior dismiss, dismiss persistence, no-op without `dismissId`, reset on `dismissId` change.

### How ops will use it for Console Air (CON-266)

Configure the `generic_banner` Unleash flag with a payload like:

```json
{
  "message": "Self-custody is moving to Console Air.",
  "links": [
    { "label": "Read the announcement", "href": "https://akash.network/blog/introducing-console-air-self-host-self-custody" },
    { "label": "Open Console Air", "href": "https://github.com/akash-network/console-air" }
  ],
  "dismissId": "console-air-2026-05"
}
```

### Not in this PR

- The "self-custody only" audience gate from the CON-266 AC is not implemented — the banner shows to all users when the flag is on. Easiest follow-up if needed: add an `audience: "self-custody"` field to the payload and gate on `isCustodial`.
- The blog post itself is content work, tracked by CON-266.

## Test plan

- [ ] `npm run test:unit` passes in `apps/deploy-web` (6 new tests in `useTopBanner.spec.tsx`).
- [ ] `npx tsc --noEmit` is clean for changed files.
- [ ] `npm run lint -- --quiet` is clean for changed files.
- [ ] Manually verify in dev with the Unleash flag enabled:
  - [ ] Banner renders with the configured message and links.
  - [ ] Clicking a link opens it in a new tab.
  - [ ] Dismissing the banner hides it for the rest of the session.
  - [ ] Reloading the page keeps the banner dismissed (with `dismissId` set).
  - [ ] Without `dismissId`, dismiss is session-only (prior behavior).
  - [ ] Changing the `dismissId` payload value re-shows the banner.

## Summary by CodeRabbit

* **New Features**
  * Top banners can include clickable external links that open in a new tab and are validated for safety.
  * Banner dismissal is now persisted per announcement so dismissed banners stay closed across sessions.
  * Banner layout spacing improved for clearer presentation.

* **Tests**
  * Added comprehensive tests covering banner visibility, dismissal persistence, and link safety validation.

### DIFF
--- a/apps/deploy-web/src/components/layout/TopBanner.tsx
+++ b/apps/deploy-web/src/components/layout/TopBanner.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useMemo, useState } from "react";
+import { Fragment, useEffect, useMemo, useState } from "react";
 import { useIntl } from "react-intl";
 import { Button } from "@akashnetwork/ui/components";
 import { Xmark } from "iconoir-react";
@@ -74,11 +74,21 @@ function MaintenanceBanner({ onClose }: { onClose: () => void }) {
 }
 
 function GenericBanner({ onClose }: { onClose: () => void }) {
-  const { message } = useGenericBannerDetails();
+  const { message, links } = useGenericBannerDetails();
 
   return (
-    <div className="fixed top-0 z-10 flex h-[40px] w-full items-center justify-center bg-primary px-3 py-2 text-primary-foreground md:space-x-4">
-      <span className="text-xs font-semibold md:text-sm">{message}</span>
+    <div className="fixed top-0 z-10 flex h-[40px] w-full items-center justify-center gap-2 bg-primary px-3 py-2 text-primary-foreground md:gap-4">
+      <span className="text-xs font-semibold md:text-sm">
+        {message}
+        {links?.map(link => (
+          <Fragment key={link.href}>
+            {" "}
+            <a href={link.href} target="_blank" rel="noopener noreferrer" className="underline hover:text-primary-foreground/80">
+              {link.label}
+            </a>
+          </Fragment>
+        ))}
+      </span>
       <Button variant="text" className="rounded-full hover:text-primary-foreground/80" size="icon" onClick={onClose}>
         <Xmark />
       </Button>
--- a/apps/deploy-web/src/components/layout/TopBanner.tsx
+++ b/apps/deploy-web/src/components/layout/TopBanner.tsx
@@ -1,4 +1,4 @@
-import { Fragment, useEffect, useMemo, useState } from "react";
+import { useEffect, useMemo, useState } from "react";
 import { useIntl } from "react-intl";
 import { Button } from "@akashnetwork/ui/components";
 import { Xmark } from "iconoir-react";
@@ -81,12 +81,9 @@ function GenericBanner({ onClose }: { onClose: () => void }) {
       <span className="text-xs font-semibold md:text-sm">
         {message}
         {links?.map(link => (
-          <Fragment key={link.href}>
-            {" "}
-            <a href={link.href} target="_blank" rel="noopener noreferrer" className="underline hover:text-primary-foreground/80">
-              {link.label}
-            </a>
-          </Fragment>
+          <a key={link.href} href={link.href} target="_blank" rel="noopener noreferrer" className="ml-1 underline hover:text-primary-foreground/80">
+            {link.label}
+          </a>
         ))}
       </span>
       <Button variant="text" className="rounded-full hover:text-primary-foreground/80" size="icon" onClick={onClose}>
--- a/apps/deploy-web/src/hooks/useTopBanner.tsx
+++ b/apps/deploy-web/src/hooks/useTopBanner.tsx
@@ -37,9 +37,7 @@ export function useGenericBannerVisibility(input: { isFlagEnabled: boolean; dism
   const [isOpen, setIsOpenAtom] = useAtom(IS_GENERIC_BANNER_ATOM);
 
   useEffect(() => {
-    if (!isFlagEnabled) return;
-    if (isGenericBannerDismissed(dismissId)) return;
-    setIsOpenAtom(true);
+    setIsOpenAtom(isFlagEnabled && !isGenericBannerDismissed(dismissId));
   }, [isFlagEnabled, dismissId, setIsOpenAtom]);
 
   const setIsOpen = useCallback(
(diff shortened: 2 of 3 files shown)


apps/deploy-web/src/hooks/useTopBanner.tsx
```diff
@@ -20,16 +32,41 @@ interface ITopBannerContext {
...
+      setIsOpenAtom(open);
+      if (!open) {
+        persistGenericBannerDismiss(dismissId);
+      }
+    },
+    [setIsOpenAtom, dismissId]
+  );
+
+  return [isOpen, setIsOpen] as const;
+}
```
ygrishajev, line 46: **Suggestion: replace the manual `localStorage` persistence with `atomWithStorage` keyed per `dismissId`.**

The `isGenericBannerDismissed` / `persistGenericBannerDismiss` helpers + the syncing `useEffect` are essentially a hand-rolled version of `atomWithStorage`. Since the storage key is dynamic (`generic_banner_dismissed:<dismissId>`), we can lift it into a per-`dismissId` atom cache and let Jotai own read/write/SSR-guarding:

```tsx
import { atom, useAtom } from "jotai";
import { atomWithStorage } from "jotai/utils";

const GENERIC_BANNER_DISMISSED_PREFIX = "generic_banner_dismissed:";

// Map cache instead of the deprecated `atomFamily` (removed in Jotai v3).
const dismissedAtomCache = new Map<string, ReturnType<typeof atomWithStorage<boolean>>>();
const getGenericBannerDismissedAtom = (dismissId: string) => {
  let dismissedAtom = dismissedAtomCache.get(dismissId);
  if (!dismissedAtom) {
    dismissedAtom = atomWithStorage<boolean>(GENERIC_BANNER_DISMISSED_PREFIX + dismissId, false);
    dismissedAtomCache.set(dismissId, dismissedAtom);
  }
  return dismissedAtom;
};

// Session-only fallback when the payload omits `dismissId` (preserves prior behavior).
const SESSION_GENERIC_BANNER_DISMISSED_ATOM = atom(false);

export function useGenericBannerVisibility(input: { isFlagEnabled: boolean; dismissId: string | undefined }) {
  const { isFlagEnabled, dismissId } = input;
  const [dismissed, setDismissed] = useAtom(
    dismissId ? getGenericBannerDismissedAtom(dismissId) : SESSION_GENERIC_BANNER_DISMISSED_ATOM
  );
  const isOpen = isFlagEnabled && !dismissed;
  const setIsOpen = useCallback((open: boolean) => setDismissed(!open), [setDismissed]);
  return [isOpen, setIsOpen] as const;
}
```

Alternatively, if we'd rather not hand-roll the cache, the official [`jotai-family`](https://github.com/jotaijs/jotai-family) package provides a v3-supported `atomFamily` replacement (it's the `jotaijs` migration target for the deprecation). Same shape as before:

```tsx
import { atomFamily } from "jotai-family";
import { atomWithStorage } from "jotai/utils";

const genericBannerDismissedAtom = atomFamily((dismissId: string) =>
  atomWithStorage<boolean>(GENERIC_BANNER_DISMISSED_PREFIX + dismissId, false)
);
```

…and use `genericBannerDismissedAtom(dismissId)` in `useAtom`. Trade-off: adds a small dependency, but skips the manual Map plumbing and gives us `.remove(key)` if we ever need to evict stale announcement atoms.

What either approach gets us:
- Drops both `window.localStorage` helpers and the `typeof window` SSR guards (`atomWithStorage` does that internally).
- Drops the `useEffect` that mirrors `localStorage` → atom — the atom *is* the persistence.
- Cross-tab sync via the browser `storage` event comes for free.
- "Changing `dismissId` re-shows the banner" falls out of switching to a different cached atom (whose stored value defaults to `false`); no explicit reset needed.

I tried the Map-cache version locally against this branch's `useTopBanner.spec.tsx` and all 6 tests pass without modification — `JSON.stringify(true) === "true"`, so the `expect(getItem(...)).toBe("true")` assertion in *"persists dismiss to localStorage"* still holds. The `jotai-family` version would behave identically for the test suite.

Caveats (apply to both):
- `atomWithStorage` reads `localStorage` lazily on the client, so SSR's first paint shows the default (`false` → banner open). That matches the current behavior, where the `useEffect` recomputes on mount.
- One boolean atom is retained per past `dismissId` for the lifetime of the page (in-memory only — not persisted growth). For announcement banners that's negligible.
- The session-only branch (`dismissId === undefined`) needs a stable shared atom (the `SESSION_*` constant above), otherwise dismiss state would be lost on each render.

Not a blocker — current code works — but it shrinks the hook by ~25 lines and removes the imperative sync, which felt like the heaviest part of this PR.

ygrishajev (approved):

apps/deploy-web/src/hooks/useTopBanner.tsx
baktun14: Good point

apps/deploy-web/src/components/layout/TopBanner.tsx
```diff
@@ -74,11 +74,18 @@ function MaintenanceBanner({ onClose }: { onClose: () => void }) {
...
-    <div className="fixed top-0 z-10 flex h-[40px] w-full items-center justify-center bg-primary px-3 py-2 text-primary-foreground md:space-x-4">
-      <span className="text-xs font-semibold md:text-sm">{message}</span>
+    <div className="fixed top-0 z-10 flex h-[40px] w-full items-center justify-center gap-2 bg-primary px-3 py-2 text-primary-foreground md:gap-4">
+      <span className="text-xs font-semibold md:text-sm">
+        {message}
+        {links?.map(link => (
+          <a key={link.href} href={link.href} target="_blank" rel="noopener noreferrer" className="ml-1 underline hover:text-primary-foreground/80">
+            {link.label}
+          </a>
+        ))}
```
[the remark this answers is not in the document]
baktun14: Validating in `useGenericBannerDetails` via `isSafeLink` (8e96fc126): rejects entries whose label/href are not strings, or whose href doesn't parse to http/https. Filtered list flows through to `GenericBanner`, so unsafe schemes never reach the anchor.

apps/deploy-web/src/hooks/useTopBanner.tsx
baktun14: Went with the Map-cache + `atomWithStorage` variant in 8e96fc126. Drops the manual localStorage helpers, the SSR guards, and the sync `useEffect`; existing 6 tests still pass without modification.

PR #3132, opened 2026-05-04
fix(deployment): prefer bash with sh fallback in lease shell

baktun14:
## Why

The in-browser deployment shell defaulted to `/bin/sh`, which has no readline support — pressing **Tab** inserts a literal tab character instead of triggering autocomplete, and **Up arrow** does not recall history. The existing FAQ entry and in-app warning acknowledged this as a workaround ("type `bash` to switch") rather than fixing it.

This PR makes the shell prefer `bash` when the container has it, falling back to `sh` transparently — so users get autocomplete + history out of the box without manual intervention, and alpine/busybox/distroless images keep working as before.

## What

- `connectToShell` now defaults to `sh -c 'command -v bash >/dev/null 2>&1 && exec bash || exec sh'` instead of `/bin/sh`. Tested locally against a deployment that has bash (autocomplete works) and the array path itself was confirmed by intermediate iterations.
- Refactored `connectToShell`'s `command` parameter from `string` (space-split) to `string[]`. The old encoding couldn't represent argv elements containing spaces — so it dropped the buggy `c.replace(" ", "+")` workaround. As a side effect, this closes a latent shell-injection footgun in `downloadFileFromShell` where a `filePath` containing spaces would split into multiple `kubectl exec` args.
- Removed the now-misleading "UP arrow and TAB autocompletion" warning and its arrow-key sniffing from `DeploymentLeaseShell`, plus the dead `Link` / `UrlService` / `OpenInWindow` imports.
- Removed the corresponding FAQ entry (`shell-arrows-and-completion`).
- Updated `provider-proxy.service.spec.ts` to reflect the new default URL shape and the array-form `command` input.

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed the UP-arrow/TAB autocompletion warning from the deployment shell interface.

* **Documentation**
  * Removed the related FAQ entry about shell autocompletion issues.

* **Improvements**
  * Improved remote shell command handling and file download reliability, resulting in more robust and predictable remote command execution.

### DIFF
--- a/apps/deploy-web/src/components/deployments/DeploymentLeaseShell.tsx
+++ b/apps/deploy-web/src/components/deployments/DeploymentLeaseShell.tsx
@@ -2,8 +2,7 @@
 import React, { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { Alert, AlertDescription, AlertTitle, Button, Spinner } from "@akashnetwork/ui/components";
 import { cn } from "@akashnetwork/ui/utils";
-import { OpenInWindow, WarningCircle } from "iconoir-react";
-import Link from "next/link";
+import { WarningCircle } from "iconoir-react";
 
 import { ViewPanel } from "@src/components/shared/ViewPanel";
 import { useServices } from "@src/context/ServicesProvider";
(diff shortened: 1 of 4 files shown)


baktun14:
## Fixes Applied Successfully

Fixed 2 file(s) based on 2 CodeRabbit feedback item(s).

**Files modified:**
- `apps/deploy-web/src/services/provider-proxy/provider-proxy.service.ts`
- `apps/deploy-web/src/services/provider-proxy/provider-proxy.service.spec.ts`

**Commit:** `30f0513af`

The double-ampersand in the shell websocket URL is removed: each `cmdN=...` is now joined with `&` instead of being prefixed with `&` and joined empty.

stalniy (approved):

PR #3137, opened 2026-05-05
fix: uses RegExp.escape to escape USDC denoms

stalniy:
## Why

Sanitizing untrusted input is a common technique for preventing injection attacks such as SQL injection or cross-site scripting. Usually, this is done by escaping meta-characters such as quotes in a domain-specific way so that they are treated as normal characters.

However, directly using the string replace method to perform escaping is notoriously error-prone. Common mistakes include only replacing the first occurrence of a meta-character, or backslash-escaping various meta-characters but not the backslash itself.

In the former case, later meta-characters are left undisturbed and can be used to subvert the sanitization. In the latter case, preceding a meta-character with a backslash leads to the backslash being escaped, but the meta-character appearing un-escaped, which again makes the sanitization ineffective.

Even if the escaped string is not used in a security-critical context, incomplete escaping may still have undesirable effects, such as badly rendered or confusing output.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved parsing of insufficient-funds error messages so denomination values (including an additional denom) are reliably extracted and shown.
  * Made regex escaping type-safe to ensure denomination matching is robust and avoids mis-parsing.
  * Enhanced extraction of available and required amounts from error text for clearer user-facing messages.

### DIFF
--- a/apps/api/src/billing/services/chain-error/chain-error.service.ts
+++ b/apps/api/src/billing/services/chain-error/chain-error.service.ts
@@ -141,10 +141,10 @@ export class ChainErrorService {
     denom: string;
   } | null {
     const usdcDenoms = Object.values(this.billingConfigService.get("USDC_IBC_DENOMS"))
-      .map(denom => denom.replace(/\//g, "\\/"))
+      .map(denom => RegExp.escape(denom))
       .join("|");
 
-    const pattern = new RegExp(`(\\d+)(uakt|${usdcDenoms}) is smaller than (\\d+)\\2`);
+    const pattern = new RegExp(`(\\d+)(uakt|uact|${usdcDenoms}) is smaller than (\\d+)\\2`);
 
     const match = message.match(pattern);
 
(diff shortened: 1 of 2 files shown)


apps/api/src/billing/services/chain-error/chain-error.service.ts
```diff
@@ -141,7 +141,7 @@ export class ChainErrorService {
     denom: string;
   } | null {
     const usdcDenoms = Object.values(this.billingConfigService.get("USDC_IBC_DENOMS"))
-      .map(denom => denom.replace(/\//g, "\\/"))
+      .map(denom => RegExp.escape(denom))
       .join("|");
 
     const pattern = new RegExp(`(\\d+)(uakt|${usdcDenoms}) is smaller than (\\d+)\\2`);
```
baktun14: perhaps we should adapt this error to `uact`?

baktun14 (approved):

PR #3158, opened 2026-05-08
fix(deployment): cache cosmos REST fallback routes

baktun14:
## Why

`/akash/deployment/{version}/deployments/info` and `/list` mirror the Cosmos REST API but are served from console's chain DB by `FallbackDeploymentReaderService`. They exist so external Cosmos tooling can keep working against deployments whose state has been pruned from chain RPC nodes.

Logs show ~50K requests in 6 hours from a single Azure IP polling one specific draining deployment (`dseq=26735741`) with `userAgent: node`. Every call goes through Sequelize against the chain DB — the same DB that's saturated with 235+ active connections (see #3157).

This PR caches at two layers, since neither alone is sufficient:

- `Cache-Control` headers help browsers and any cooperating CDN.
- **In-process memoization** is what actually defeats server-side bots that ignore HTTP cache headers — identical queries within a block window are served from an in-memory LRU and never reach Sequelize.

## What

1. **Route-level `cache: { maxAge: 10, staleWhileRevalidate: 30 }`** on both fallback routes — emits `Cache-Control: public, max-age=10, stale-while-revalidate=30`.
2. **Service-level memoization** in `FallbackDeploymentReaderService`:
   - `findByOwnerAndDseq`: `@Memoize({ ttlInSeconds: averageBlockTime })` (matches the existing pattern in `AkashBlockService`).
   - `findAll`: wrapped with `cacheResponse(averageBlockTime, ...)` keyed on the params (the `@Memoize` decorator only handles primitive args).

The `cacheResponse` helper has stale-while-revalidate semantics in-process and dedupes concurrent identical requests via `pendingRequests`, so a request burst still results in a single DB hit.

## Summary by CodeRabbit

* **Performance Improvements**
  * Added HTTP caching to deployment list and info endpoints to reduce load and speed up responses.
  * Introduced result caching/memoization for deployment lookups to minimize repeated database queries and improve endpoint responsiveness.

[Review Change Stack](https://app.coderabbit.ai/change-stack/akash-network/console/pull/3158)


apps/api/src/deployment/services/fallback-deployment-reader/fallback-deployment-reader.service.ts
```diff
@@ -94,6 +100,7 @@ export class FallbackDeploymentReaderService {
     };
   }
 
+  @Memoize({ ttlInSeconds: averageBlockTime })
```
stalniy, line 103: **question:** not sure about the details of `@Memoize` but it seems that cache key is highly variable, wouldn't it open another point of failure like OOM?

stalniy: and probably in case of blockchain being available, we could just redirect request to blockchain api

stalniy: also we need to find out the reason why pool is exhausted:
- is this because of slow db query?
- is this because amount of requests too high?

stalniy (approved):

apps/api/src/deployment/services/fallback-deployment-reader/fallback-deployment-reader.service.ts
baktun14: This is for the fallback when the chain in unavailable, I think from what I've seen is bursts of queries. But yea this is not a permanent fix for the problem.

PR #3204, opened 2026-05-20
feat(onboarding): add /onboarding template picker with phased deploy

ygrishajev:
## Why

Closes CON-313, CON-314, CON-315.

Introduces a non-discoverable `/onboarding` entry point that takes a newly registered user from picking a template to a running deployment in a single in-place flow, so the first deploy is one click instead of a multi-step build-your-own-SDL detour.

## What

- New `/onboarding` page (`apps/deploy-web/src/pages/onboarding/index.tsx`) gated behind a feature flag, with a template picker (Hello world, LLM chatbot, Stable Diffusion).
- Phased deploy flow: `createDeployment` → wait for bid match → `createLease` → redirect to deployment details, driven by `usePhasedDeploymentFlow` + `useGradualProgress`.
- `PhasedDeploymentProgress` / `PhasedDeploymentContainer` components render the progress card over an extended `Globe` with focus tween + camera biases.
- New `useProvidersQuery` query for bid-matching, plus template-card images under `public/images/onboarding/`.
- Playwright e2e (`onboarding-quick-deploy.spec.ts`) picks Hello world and verifies the deployment is created and closed.

https://github.com/user-attachments/assets/9aa5a99a-a13e-4aae-b2b3-05285dd03988

## Summary by CodeRabbit

* **New Features**
  * Onboarding picker page with interactive deployment template cards and deploy flow.
  * Deployment template picker card component with recommended/disabled states.
  * Phased deployment container and progress overlay with phase-based progress bar and provider selection.
  * Globe enhancements: focus animations, biasing, and marker sizing.
  * Provider reachability check hook and phased deployment flow hook.
  * Added onboarding route.

* **Tests**
  * Extensive unit/integration tests for new components and hooks.
  * End-to-end UI test for onboarding quick deploy.

* **Chores**
  * Added new feature flag value.

[Review Change Stack](https://app.coderabbit.ai/change-stack/akash-network/console/pull/3204?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


apps/deploy-web/public/images/onboarding/stable-diffusion.png
stalniy, line 1: **suggestion(blocking)**: use https://imagecompressor.com/ to minify images

ygrishajev: these are placeholders and would eventually change as i got it. but did this anyway

apps/deploy-web/src/hooks/usePhasedDeploymentFlow/usePhasedDeploymentFlow.ts
```diff
@@ -0,0 +1,222 @@
...
+              provider: activeBid.bid.id.provider
+            }
+          ]
+        },
+        { onSuccess: onCreateLeaseSuccess, onError: onCreateLeaseError }
+      );
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [phase, activeBid, reachableProvider, dseq, manifest]
+  );
```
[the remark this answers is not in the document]
ygrishajev: Skipping — the scenario isn't reachable through the current UI:

- `retry()` is only invoked by the "Start over" button in `PhasedDeploymentProgress`, which renders solely when `state.kind === "error"`.
- The only transition from `"preparing"` → `"error"` is `onCreateLeaseError`, which fires *after* `createLease` has settled (i.e. when the mutation is no longer in flight).
- `failOnBidTimeout` is the only other path to error, but its cleanup runs the moment we leave `"matching"`, so it can't fire during `"preparing"`.

A closure-based `cancelled` flag also wouldn't work here because `setPhase("preparing")` inside the effect triggers an immediate cleanup, so the flag would be flipped before the real callbacks fire. If we ever need this guard (e.g. if "Start over" gets surfaced from non-error states), a ref-based request token like the one you suggested is the right shape.

apps/deploy-web/src/pages/onboarding/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+      ctx.services.template.findById("akash-network-awesome-akash-Llama-3.1-8B")
+    ]);
+
+    return {
+      props: {
+        templates: {
+          helloWorld: helloWorldTemplate.content,
+          imageGen: imageGenTemplate.deploy,
+          llmChatbot: llmChatbotTemplate.deploy
+        }
```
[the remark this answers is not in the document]
ygrishajev: Skipping — `findById` returns `Promise<TemplateOutput>` (never null) and `defineServerSideProps` already handles the failure modes:

- The wrapper catches 404/400 HTTP errors and converts them to `notFound: true` (`defineServerSideProps.ts:79-88`).
- Other errors propagate up through `wrapGetServerSidePropsWithSentry`, so they're auto-reported to Sentry and surface as the standard error page.

Adding a redirect-to-`/` swallow here would actually obscure the auto-report and override the wrapper's `notFound` behavior, so we'd rather rely on the existing infrastructure.

apps/deploy-web/src/hooks/usePhasedDeploymentFlow/usePhasedDeploymentFlow.ts
```diff
@@ -0,0 +1,222 @@
...
+  onSuccess: (dseq: string) => void;
+};
+
+type DeploymentPhase = {
+  id: DeploymentPhaseId;
+  label: string;
+  status: DeploymentPhaseStatus;
+};
+
+type State = { kind: "creating" | "matching" | "preparing" | "success" } | { kind: "error"; message?: string };
```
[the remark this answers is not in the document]
ygrishajev: Skipping — the error overlay (`PhasedDeploymentProgress.tsx:37-58`) renders an entirely separate view: a heading + message + "Start over"/"Contact support" buttons. The progress bar and phase checklist are not rendered in `state.kind === "error"`, so `phases` / `progressPercent` / `phaseIndex` aren't consumed there. The phase-0 flattening has no visual effect today.

If we ever surface a phase-aware error overlay (e.g. "Failed at: Preparing deployment"), the `failedPhase` pattern you suggested is what we'd want.

apps/deploy-web/src/queries/useProvidersQuery.ts
```diff
@@ -57,6 +57,35 @@ export function useProviderStatus(
   });
 }
 
+/** Per-provider `/status` probe timeout. Keeps the loop from stalling on a single unresponsive provider so we move on to the next candidate quickly. */
+const PROVIDER_STATUS_PROBE_TIMEOUT_MS = 5000;
+
+export function useFirstReachableProvider(
+  providers: ApiProviderList[] | undefined | null,
```
baktun14, line 64: question: is this based on the bids we receive?

baktun14 (approved):

apps/deploy-web/src/queries/useProvidersQuery.ts
ygrishajev: this fn doesn't care if that's based on bids. however it is used against providers from bids by the client code

Issue #2809, opened 2026-02-22
Concurrent deployment requests fail due to identical dseq values


stalniy:
### Summary
Concurrent deployment requests for a single user result in failed transactions because both receive the same `dseq` value (derived from `getLatestBlock`). This manifests in two ways:

**Chain-side failure:**
```
failed to execute message; message index: 0: Deployment exists
```

**DB-side failure (from #1757):**
```
PostgresError: duplicate key value violates unique constraint "dseq_user_id_idx" (code: 23505)
```

### Steps to Reproduce
1. Trigger multiple deployment requests in parallel for the same user.
2. Observe that both deployments receive the same dseq value.
3. Attempt to submit both transactions.

### Actual Result
- Only one transaction succeeds; the other fails with `Deployment exists` (chain) or `duplicate key value violates unique constraint "dseq_user_id_idx"` (DB).

### Expected Result
- Each concurrent request should use a unique dseq to avoid transaction conflicts.

### Context
```
stdout | test/e2e/managed-api-deployment-flow.spec.ts > Managed Wallet API Deployment Flow > executes a full deployment cycle with provider JWT auth { data: { dseq: '2141770', ... } }
stdout | ... signTx: { code: 3, rawLog: 'failed to execute message; message index: 0: Deployment exists' }
```

DB stack trace (from #1757):
```
PostgresError: duplicate key value violates unique constraint "dseq_user_id_idx" (code: 23505)
    at ErrorResponse (postgres/cjs/src/connection.js:788:1)
    at handle (postgres/cjs/src/connection.js:474:1)
    ...
```

### Suggestion
- Implement a way to ensure unique dseq values for concurrent deployments (for example, using a monotonic counter or unique ID per request).

### Environment
- Repo: akash-network/console
- Impact: Affects automated flows, CI/CD, and any users issuing multiple deployments in parallel.

---
**Consolidates:** #1757

baktun14:
Moved to Linear: CON-85

PR #3215, opened 2026-05-24
fix(deployment): bump dseq on collision and surface non-zero tx code

baktun14:
## Why

Closes CON-85

Mainnet incident on 2026-05-23 — two failed `MsgCreateDeployment` transactions landed on-chain (heights 26955138 and 26955799), both for the same user's derived wallet. The deployment creation flow had two latent bugs that made these failures harmful instead of recoverable:

1. **DSeq generated from current block height with no idempotency.** [DeploymentWriterService.create:39](https://github.com/akash-network/console/blob/main/apps/api/src/deployment/services/deployment-writer/deployment-writer.service.ts#L39) did `dseq = await this.blockHttpService.getCurrentHeight()`. Two POST `/v1/deployments` requests in the same Akash block (~6 s) generated the *same* dseq; the second landed on-chain with `deployment exists`.
2. **Non-zero tx codes propagated silently.** In [ManagedSignerService.executeDecodedTxByUserWallet](https://github.com/akash-network/console/blob/main/apps/api/src/billing/services/managed-signer/managed-signer.service.ts#L120-L144), the code path after `executeDerivedTx` didn't check `tx.code`. A transaction that landed with `code !== 0` still triggered side effects: `TrialDeploymentLeaseCreated`, `EnableDeploymentAlertCommand`, `balancesService.refreshUserWalletLimits`, `walletReloadJobService.scheduleImmediate`. Alerts and trial events were created for deployments that don't exist on-chain.

## What

- **DSeq from millisecond timestamp in `DeploymentWriterService.create`**: replace block-height-derived dseq with `Date.now()`. dseq is uint64 on-chain and stored as a string, so ms timestamps (~1.7e12) fit safely. This eliminates the same-block collision class entirely. The vanishingly rare same-millisecond case (two POSTs from the same user inside one ms) is cleanly handled by defense (2) — returns a 4xx, no orphan side effects.
- **Defense-in-depth `code !== 0` assertion in `ManagedSignerService.executeDecodedTxByUserWallet`**: immediately after `executeDerivedTx`, check `tx.code === 0` and throw through `ChainErrorService.toAppError(new Error(tx.rawLog), messages)` if not. This guarantees side effects (domain events, balance refresh, reload schedule) only fire on transactions that succeeded on-chain. The existing `ChainErrorService` pattern matcher already maps `deployment exists` / `deployment closed` / etc. to the right 4xx — no new mappings needed.

Note: the self-custodial frontend deploy path in [apps/deploy-web/src/utils/deploymentData/v1beta3.ts](https://github.com/akash-network/console/blob/main/apps/deploy-web/src/utils/deploymentData/v1beta3.ts) still uses block height. A single user firing two browser deploys in the same ~6 s block is far rarer than the managed-wallet server case; leaving it out keeps this PR tight. Can follow up separately if needed.

Companion to #PR_TX_SIGNER (tx-signer-side throw on `code !== 0`). This PR's defense-in-depth makes the api side robust independently — works whether the tx-signer PR has shipped or not.

## Test plan

- [ ] `cd apps/api && npm run test:unit src/billing/services/managed-signer src/deployment/services/deployment-writer` — 28 tests pass
- [ ] `npm run lint -- --quiet` — clean
- [ ] `npx tsc --noEmit` — no new errors (7 deployment-writer spec errors remain, all pre-existing patterns; count went down from 9 since the two collision tests were removed)
- [ ] Manual staging smoke: hand-craft a duplicate dseq POST `/v1/deployments` → API returns 400 `"Deployment with provided dseq and owner already exists"`, no `TrialDeploymentLeaseCreated` in logs
- [ ] Trigger two rapid deployments → both succeed with distinct ms-timestamp dseqs (~13 digits) matching what landed on-chain
- [ ] After deploy, monitor for orphaned alerts/events; should drop to zero

Related to CON-296.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling for on-chain transactions that land with errors: clear app errors are thrown and downstream side effects are skipped.

* **Performance Improvements**
  * Deployment creation now uses client-side timestamps to avoid extra network calls.

* **Refactor**
  * Removed dependence on the chain API client from deployment flows and UI components.

* **Tests**
  * Updated tests and mocks to match timestamp behavior and added mock cleanup.

* **Documentation**
  * Clarified dseq description to allow millisecond timestamps.

[Review Change Stack](https://app.coderabbit.ai/change-stack/akash-network/console/pull/3215?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/apps/api/src/billing/services/managed-signer/managed-signer.service.spec.ts
+++ b/apps/api/src/billing/services/managed-signer/managed-signer.service.spec.ts
@@ -231,6 +231,36 @@ describe(ManagedSignerService.name, () => {
       });
     });
 
+    it("throws and skips side-effects when the broadcast tx lands on-chain with a non-zero code", async () => {
+      const wallet = createUserWallet({ userId: "user-123", feeAllowance: 100, deploymentAllowance: 100, isTrialing: true });
+      const user = createUser({ userId: "user-123" });
+      const deploymentMessage: EncodeObject = {
+        typeUrl: MsgCreateLease.$type,
+        value: MsgCreateLease.fromPartial({
+          bidId: { dseq: 123, owner: "akash1test", gseq: 1, oseq: 1, provider: "akash1test" }
+        })
+      };
+
+      const { service, domainEvents, balancesService, walletReloadJobService, chainErrorService } = setup({
+        findOneByUserId: jest.fn().mockResolvedValue(wallet),
+        findById: jest.fn().mockResolvedValue(user),
+        signAndBroadcastWithDerivedWallet: jest.fn().mockResolvedValue({
+          code: 17,
+          hash: "tx-hash",
+          rawLog: "deployment exists"
+        }),
+        refreshUserWalletLimits: jest.fn().mockResolvedValue(undefined),
+        transformChainError: jest.fn().mockResolvedValue(new Error("Deployment with provided dseq and owner already exists"))
+      });
+
+      await expect(service.executeDerivedDecodedTxByUserId("user-123", [deploymentMessage])).rejects.toThrow("Deployment with provided dseq");
+
+      expect(chainErrorService.toAppError).toHaveBeenCalledWith(expect.objectContaining({ message: "deployment exists" }), [deploymentMessage]);
+      expect(domainEvents.publish).not.toHaveBeenCalled();
+      expect(balancesService.refreshUserWalletLimits).not.toHaveBeenCalled();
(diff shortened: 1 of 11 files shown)


stalniy:
https://github.com/akash-network/console/issues/2809 

I think we should not use current block for dseq

baktun14:
> #2809
> 
> I think we should not use current block for dseq

Ah right, linked the issue. We can use a timestamp or something more unique.

stalniy (approved):

PR #3242, opened 2026-05-29
fix(template): reject path traversal in public template-by-id endpoint

baktun14:
## Why

`GET /v1/templates/{id}` (public, **unauthenticated**) built a file path by interpolating the caller-supplied `id` straight into the template cache path. Hono routes with `decodeURI` (so a URL-encoded `%2F` survives as a single segment), then decodes the param with `decodeURIComponent`, so a payload like `..%2F..%2Fsecret` arrived at the service as `../../secret` and escaped the cache directory — disclosing arbitrary `.json` files the API process can read (other tenants' cached templates, service-account key JSON, config JSON).

Found via an internal repo-wide security audit.

Fixes CON-428

## What

> Reworked per review (was a `resolvePathWithinDir()` util guarding the read inside the service).

- Reject template ids containing a path separator at the route's param schema (`GetTemplateByIdParamsSchema`): `id: z.string().regex(/^[^/\\]+$/)`. A rejected id never reaches the service and returns a **400** validation error (no info leak). The fix sits at the HTTP boundary; `getTemplateById()` and the `#templateCachePath()` read/write encapsulation are left untouched.
- A separator denylist (not a charset allowlist) is both safe and complete here: template ids are generated as `{owner}-{repo}-{path}` with `path.replace(/[\\/]+/g, "-")`, so a legit id never contains `/` or `\`; and traversal fundamentally requires a separator in the string passed to `path.resolve`. Blocking separators closes the vector while still allowing dots/spaces (e.g. `…-DeepSeek-V3.1`).
- Added a `400` response to the route's OpenAPI definition so the documented contract matches.

### Tests
- Functional test (`test/functional/templates.spec.ts`): at the real endpoint, single-encoded payloads (`..%2F..%2F..%2Fetc%2Fpasswd`, `%2Fetc%2Fpasswd`, `..%2F..%2Fsecret`) return **400**; the double-encoded payload (`..%252F..%252Fsecret`) decodes to a literal `%2F` (no separator), so it resolves to a missing file → **404**. Each case also asserts the response body leaks no file contents.
- Service unit tests keep the standard `getTemplateById` coverage (incl. a legit dot-id resolving to the correct path).

No breaking changes; no DB migrations.

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened template ID validation to reject invalid characters and block path-traversal attempts.

* **Documentation**
  * Clarified API docs with a 400 response for invalid template IDs and an explanatory message.

* **Tests**
  * Added unit and functional tests verifying template ID handling, encoded traversal cases, and that file contents are not exposed.


apps/api/src/template/services/template-gallery/template-gallery.service.ts
```diff
@@ -244,10 +245,16 @@ export class TemplateGalleryService {
   }
 
   async getTemplateById(id: Required<Template>["id"]): Promise<Template | null> {
+    const cachePath = resolvePathWithinDir(`${this.#galleriesCachePath}/v1/templates`, `${id}.json`);
+    if (!cachePath) {
+      this.#logger.warn({ event: "TEMPLATE_PATH_TRAVERSAL_BLOCKED", templateId: id });
+      return null;
+    }
```
stalniy: **issue(blocking):** seems like it's the wrong place to fix it. It also breaks encapsulation of `#templateCachePath` function. 

To prevent path traversal it's enough to remove all slashes or block them on http schema (zod) level

stalniy (changes requested):
overcomplicated implementation but nice catch!

apps/api/src/template/services/template-gallery/template-gallery.service.ts
baktun14: Moved it to the HTTP boundary: `GetTemplateByIdParamsSchema.id` now rejects ids containing `/` or `\` via `/^[^/\\]+$/`, so a traversal id fails validation with a 400 before it reaches the service. `getTemplateById` is back to `this.#fs.readFile(this.#templateCachePath(id), "utf8")` and the `resolvePathWithinDir` util is removed, so `#templateCachePath` encapsulation is restored. This is safe because ids are generated as `{owner}-{repo}-{path}` with `path.replace(/[\\/]+/g, "-")`, so a valid id never contains a separator. 7982952eb

apps/api/test/functional/templates.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+  describe("GET /v1/templates/{id} path traversal", () => {
+    // %2F survives Hono routing as a single segment and is only decoded to "/" when the param is
+    // read, so the id schema sees the real separator and rejects it with a 400 before any file is
+    // read. The double-encoded payload decodes to a literal "%2F" (no separator), so it passes
+    // validation and resolves to a missing file -> 404. Either way no file contents leak (CON-428).
+    [
+      { encodedId: "..%2F..%2F..%2Fetc%2Fpasswd", expectedStatus: 400 }, // ../../../etc/passwd
+      { encodedId: "%2Fetc%2Fpasswd", expectedStatus: 400 }, // /etc/passwd (absolute escape)
+      { encodedId: "..%2F..%2Fsecret", expectedStatus: 400 },
+      { encodedId: "..%252F..%252Fsecret", expectedStatus: 404 } // double-encoded: stays a literal "%2F"
```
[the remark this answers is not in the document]
baktun14: Added backslash-encoded cases to `templates.spec.ts`: `..%5C..%5Csecret` → 400 (the `\` decodes to a real separator and the id schema rejects it) and `..%255C..%255Csecret` → 404 (double-encoded, stays a literal `%5C`, so it resolves to a missing file). Both verified against the real endpoint. ced346de4

stalniy (approved):

PR #3263, opened 2026-06-03
refactor(sdl): placement-first form values schema

ygrishajev:
## Why

Closes CON-413. Foundational data-layer change for the redesigned configure-deployment screen (CON-411): every subsequent pane (CON-414..427) reads or writes one shared deployment model, and that model now treats placements as first-class entities. With the legacy embedded \`service.placement\` shape, empty placements have no host, renames cascade through every referencing service, and region is a global \`generateSdl\` arg with no per-placement granularity — none of which work for a UI that lets users create multiple placements and pick a provider per placement.

The new screen also needs a single source of truth that SDL generation and validation can reuse without re-authoring rules. Evolving the existing schema in place (rather than branching a parallel model) keeps the legacy wizard and the new redesign on one shape, and the legacy form continues to render identically against it.

## What

- **\`SdlBuilderFormValuesSchema\` reshape** (\`apps/deploy-web/src/types/sdlBuilder/sdlBuilder.ts\`): \`placements\` becomes a top-level array (each with a stable \`id\` and an optional \`region\`); services drop their embedded \`placement\` and gain \`placementId\` + a top-level \`pricing\` field. A top-level \`superRefine\` guards against orphan \`placementId\` references.
- **\`generateSdl\`** now takes the full \`SdlBuilderFormValuesType\`, resolves each service's placement via id, and writes \`location-region\` as a per-placement attribute. The \`(services, region?)\` signature is gone.
- **\`importSimpleSdl\`** now returns \`SdlBuilderFormValuesType\`, deduplicates placements by SDL name, lifts \`attributes["location-region"]\` onto \`placement.region\`, and lifts per-service pricing onto \`service.pricing\`. SDL YAML wire format is unchanged.
- **\`defaultServiceWithPlacement\` helper** (\`utils/sdl/data.ts\`) collapses the typical "init the form" pattern that consumers (\`SdlBuilder\`, \`RemoteDeployUpdate\`, \`SimpleSdlBuilderForm\`) all repeated.
- **Legacy form controls** (\`PlacementFormModal\`, \`SignedByFormControl\`, \`AttributesFormControl\`, \`SimpleServiceFormControl\`, \`LogCollectorControl\`, \`ImportSdlModal\`, \`SimpleSdlBuilderForm\`) rebind from \`services.\${i}.placement.X\` paths to \`placements.\${p}.X\` (shared) or \`services.\${i}.pricing.X\` (per-service). A new \`usePlacementIndexForService\` hook in \`PlacementFormModal\` resolves service-index → placement-index.
- **\`SdlBuilder.tsx\`** gains try/catch around \`generateSdl\` in the watch subscription (surfaces errors via \`setError\` instead of silently killing the RHF subscription), and \`getSdl\` handles non-\`TransformError\` throws.
- **Roundtrip test**: imports the canonical \`two-services-sdl.yml\` fixture, regenerates, parses both via \`yaml.load\`, asserts deep equality. Guards against silent drift between generator and importer.

### Bonus pre-existing bug fixes surfaced by the refactor

- \`signedBy.allOf\`-only placements were silently dropped from SDL output — the legacy guard tested \`anyOf?.length > 0 || anyOf?.length > 0\` (both halves identical typo) instead of \`anyOf || allOf\`. Now both branches are emitted independently.
- A second IP-using service overwrote the first's \`endpoints\` block due to non-idempotent endpoints initialization. Now initialized lazily (\`||= {}\`).

### Out of scope

- IP endpoints stay as \`expose[].ipName: string\`. CON-416 decides if they get a first-class array.
- Multi-group services. The "only one placement available" assumption from the legacy importer is preserved.
- The new redesign panes under \`components/deployments/ConfigureDeployment/\` — they don't read this schema yet (CON-415+).
- \`apps/api\`, \`apps/indexer\`, etc. — SDL wire format is unchanged.

## Test plan

- [ ] \`npm run lint -- --quiet\`, \`npx tsc --noEmit\`, \`npm test\` pass under \`apps/deploy-web\`
- [ ] Legacy SDL builder at \`/new-deployment\` renders identically — template selection, manifest edit step, create-lease step
- [ ] Legacy \`/sdl-builder\` page renders and Save Template / Preview / Reset all work
- [ ] Remote-deploy update flow (\`RemoteDeployUpdate\`) renders existing deployments and edits stay coherent
- [ ] Importing a representative existing SDL (single placement, multiple services, GPU, signedBy with allOf-only, location-region attribute) roundtrips cleanly
- [ ] Log-collector sync still ties the log-collector service to the target service's placement and pricing

## Summary by CodeRabbit

* **Refactor**
  * Reorganized deployment form structure to separate placement configuration from service pricing, improving form clarity and validation.

* **Bug Fixes**
  * Enhanced error handling during SDL import and generation with clearer user-facing error messages when operations fail.

* **Tests**
  * Updated test suite to reflect internal form structure changes and added validation tests for placement references.

### DIFF
--- a/apps/deploy-web/src/types/sdlBuilder/sdlBuilder.ts
+++ b/apps/deploy-web/src/types/sdlBuilder/sdlBuilder.ts
@@ -198,23 +198,21 @@ export const ExposeSchema = z.object({
 });
 
 export const PlacementSchema = z.object({
+  id: z.string().min(1, { message: "Placement id is required." }),
   name: z
     .string()
     .min(1, { message: "Placement name is required." })
     .regex(/^[a-z0-9-]+$/, { message: "Invalid placement name. It must only be lower case letters, numbers and dashes." })
     .regex(/^[a-z]/, { message: "Invalid starting character. It can only start with a lowercase letter." })
     .regex(/[^-]$/, { message: "Invalid ending character. It can only end with a lowercase letter or number" }),
+  region: z.string().optional(),
   attributes: z.array(PlacementAttributeSchema).optional(),
   signedBy: z
     .object({
       allOf: z.array(SignedBySchema),
       anyOf: z.array(SignedBySchema)
     })
-    .optional(),
-  pricing: z.object({
-    amount: z.number().min(1, { message: "Pricing amount is required." }),
-    denom: z.string().min(1, { message: "Pricing denom is required." })
-  })
+    .optional()
 });
 
 const validateCpuAmount = (value: number, serviceCount: number, context: z.RefinementCtx) => {
@@ -348,7 +346,11 @@ export const ServiceSchema = z
     expose: z.array(ExposeSchema),
     command: CommandSchema.optional(),
     env: z.array(EnvironmentVariableSchema).optional(),
-    placement: PlacementSchema,
+    placementId: z.string().min(1, { message: "Placement reference is required." }),
+    pricing: z.object({
+      amount: z.number().min(1, { message: "Pricing amount is required." }),
+      denom: z.string().min(1, { message: "Pricing denom is required." })
+    }),
     count: z.number().min(1, { message: "Service count is required." }),
     sshPubKey: z.string().optional(),
     params: z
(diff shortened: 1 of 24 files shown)


apps/deploy-web/src/types/sdlBuilder/sdlBuilder.ts
```diff
@@ -198,23 +198,21 @@ export const ExposeSchema = z.object({
...
     .regex(/[^-]$/, { message: "Invalid ending character. It can only end with a lowercase letter or number" }),
+  region: z.string().optional(),
   attributes: z.array(PlacementAttributeSchema).optional(),
   signedBy: z
     .object({
       allOf: z.array(SignedBySchema),
       anyOf: z.array(SignedBySchema)
     })
-    .optional(),
-  pricing: z.object({
```
baktun14, line 214: pricing is not on the placement?

baktun14 (approved):

apps/deploy-web/src/types/sdlBuilder/sdlBuilder.ts
ygrishajev: yes, basically it's easier to keep it so as pricing is per service (one-to-one) and placement is one-to-many

PR #3268, opened 2026-06-03
feat(billing): migrate denom exchange oracle to V2 with V1 fallback

baktun14:
## Why

Oracle V2 activates on Akash mainnet with the **v2.1.0 node upgrade (June 11)**. At that point the V1 oracle query stops working, and `DenomExchangeService` — the only code querying the oracle module — would silently degrade to its **daily CoinGecko** fallback, hurting real-time pricing for **live ACT minting** (`master-wallet-mint`).

Part of CON-195

## What

Migrates `DenomExchangeService.getExchangeRateToUSD` from Oracle **V1 (block-height)** to **V2 (time-based)** with a **version-aware fallback**: try V2 → fall back to V1 when the V2 query service is unavailable (pre-upgrade nodes don't register it, so the call throws) → existing CoinGecko net. This decouples the api deploy from the June 11 node release and self-heals across the upgrade.

- V2 `getPrices` uses a `startTime`/`endTime` window instead of `height`; `getAggregatedPrice` is unchanged between versions.
- The 24h price-history query (edge-of-retention under V2's 24h pruning, and feeds only the currently-unused `priceChange` fields) is decoupled so a failure there cannot degrade the billing-critical current price.
- Return shape is unchanged, so both consumers (`master-wallet-mint`, `stats`) are transparent.
- **No indexer change**: the `Oracle.v1` event listener must stay V1 for re-index correctness (re-indexing replays frozen historical V1 events) — see canceled CON-409.

**Not breaking**: the `getExchangeRateToUSD` return contract is unchanged.

**Follow-up** (not in this PR): remove the V1 fallback path once Oracle V2 is stable on mainnet (post-v2.1.0).

### Verification

- `npm run test:unit -- denom-exchange` — 9 passed (covers V2 path, V1 fallback, 24h-history degrade, unhealthy → CoinGecko, both-versions-fail → CoinGecko)
- `npx tsc --noEmit` — no new errors (changed files + both consumers type-clean)
- `npm run lint` — clean

## Summary by CodeRabbit

* **New Features**
  * App now prefers the newer oracle price feed and avoids repeatedly probing unsupported endpoints, improving stability.

* **Bug Fixes**
  * More reliable USD exchange rates via version-aware feed selection and resilient fallbacks when primary feeds or history queries fail.
  * More accurate 24‑hour change calculations using a precise ~23h time window and graceful degradation to avoid misleading change values.
  * Emits clearer warnings when historical pricing is unavailable while preserving expected behavior during partial outages.

* **Tests**
  * Expanded coverage for multi-version feed behavior, fallback scenarios, and improved test helpers for independent feed mocking.

### DIFF
--- a/apps/api/src/chain/services/denom-exchange/denom-exchange.service.ts
+++ b/apps/api/src/chain/services/denom-exchange/denom-exchange.service.ts
@@ -1,4 +1,4 @@
-import { minutesToMilliseconds } from "date-fns";
+import { minutesToMilliseconds, subHours } from "date-fns";
 import { inject, singleton } from "tsyringe";
 
 import { memoizeAsync } from "@src/caching/helpers";
@@ -27,17 +27,7 @@ export class DenomExchangeService {
       const mappedDenom = legacyToNewMapping[denom] ?? denom;
 
       try {
-        const latestBlock = await this.#chainSdk.cosmos.base.tendermint.v1beta1.getLatestBlock();
-        const currentHeight = latestBlock.block?.header?.height?.toBigInt() ?? 0n;
-        const blocksPerDay = (24n * 60n * 60n) / BigInt(Math.ceil(averageBlockTime));
-        const blockHeight24hAgo = currentHeight > blocksPerDay ? currentHeight - blocksPerDay : 1n;
-        const [oracleRate, rate24hAgo] = await Promise.all([
-          this.#chainSdk.akash.oracle.v1.getAggregatedPrice({ denom: mappedDenom }),
-          this.#chainSdk.akash.oracle.v1.getPrices({
-            filters: { assetDenom: mappedDenom, baseDenom: "usd", height: blockHeight24hAgo },
-            pagination: { limit: 1 }
-          })
-        ]);
+        const { oracleRate, rate24hAgo } = await this.#fetchOracleRate(mappedDenom);
 
         if (!oracleRate.priceHealth?.isHealthy) {
           this.#logger.warn({ event: "ORACLE_PRICE_UNHEALTHY", denom: mappedDenom });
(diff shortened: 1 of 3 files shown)


apps/api/src/chain/services/denom-exchange/denom-exchange.service.ts
```diff
@@ -63,6 +53,55 @@ export class DenomExchangeService {
...
+   * Version-aware oracle fetch: prefer Oracle V2, falling back to V1 when the V2 query service is
+   * unavailable (e.g. on nodes that have not yet run the v2.1.0 upgrade, which do not register the
+   * V2 query service so the call throws). The V1 fallback is removed once V2 is stable on mainnet.
+   */
+  async #fetchOracleRate(mappedDenom: string) {
+    try {
+      return await this.#fetchOracleRateV2(mappedDenom);
+    } catch (error) {
+      this.#logger.warn({ event: "ORACLE_V2_UNAVAILABLE_FALLBACK_V1", denom: mappedDenom, error });
+      return await this.#fetchOracleRateV1(mappedDenom);
```
stalniy, line 79: **suggestion:** we had similar solution on provider-proxy in the past. YOu need to check the error response status. if it's 501 (not implemented), then you can fallback. Also it's better to cache this decision until it breaks.

apps/api/src/chain/services/denom-exchange/denom-exchange.service.ts
```diff
@@ -63,6 +53,55 @@ export class DenomExchangeService {
...
+    }
+  }
+
+  async #fetchOracleRateV2(mappedDenom: string) {
+    const endTime = new Date();
+    const startTime = subHours(endTime, 24);
```
stalniy: **issue:** oracle v2 stores only 24h of prices feed, so at exact 24h back there may not be any price. So, I'd use something a bit smaller than 24 (e.g., 23 or 23.5). But this things was implemented as a fallback to prev coingecko api. As far as I know we don't use rate24Ago price anywhere in the code

stalniy (approved):

apps/api/src/chain/services/denom-exchange/denom-exchange.service.ts
baktun14: Fixed — using a 23h window now so we stay inside the ~24h retention.

apps/api/src/chain/services/denom-exchange/denom-exchange.service.ts
baktun14: Done in cb2d0ae — only falls back to V1 on Unimplemented (gRPC code 12) now, and caches that until V1 breaks.

PR #3276, opened 2026-06-05
feat(deployment): stop auto-top-up for reclaimed leases

baktun14:
## Why

Part of CON-376.

With AEP-82 resource reclamation (v2.1.0, mainnet June 11), a provider can reclaim a lease. The lease enters the `reclaiming` state and reclamation **always proceeds to close** — a tenant cannot cancel it. Today auto-top-up keeps funding such deployments right up until the chain closes them, wasting tenant funds on a deployment that is going away.

> **Scope note:** CON-376 originally also called for a deploy-web "paused group" status + a **"Restart group"** action signing `MsgStartGroup`. That premise turned out to be wrong — reclamation is terminal and there is no tenant restart path (recovery is close & redeploy). The frontend portion has been dropped from this PR; CON-376 needs re-scoping for the read-only paused-status + redeploy guidance. This PR ships only the backend behavior, which is correct regardless of how the UI surfaces it.

## What

- Add an optional `isTerminal` flag to `DrainingDeploymentOutput`. It is `true` only when **every** lease of a deployment is terminal (closed **or** `reclaiming`). Populated by the RPC source only; the DB source lacks reclaim state and leaves it `undefined`.
- `DrainingDeploymentRpcService` now treats a `reclaiming` lease as terminal (named constant `RECLAIMING_LEASE_STATE`), aggregated so a deployment with a mix of reclaiming + still-active leases keeps being topped up.
- `DrainingDeploymentService` excludes terminal deployments from top-up and marks their setting `closed: true`, exactly as it already does for closed leases.

The exact REST state string (`"reclaiming"`) was confirmed against the chain-sdk serializer (`lease_StateToJSON`).

### Tests
- `draining-deployment-rpc.service.spec.ts`: reclaiming → terminal; partial reclaiming → not terminal; all-terminal (reclaiming + closed) → terminal; active → not terminal.
- `draining-deployment.service.spec.ts`: a terminal (reclaiming) deployment is marked closed and excluded from top-up batches.

All affected api specs pass (36), lint clean, build type-check clean.

## Summary by CodeRabbit

* **New Features**
  * Enhanced lease state tracking to properly identify and handle terminal deployments with reclaiming leases.

* **Tests**
  * Added comprehensive test coverage for terminal lease state determination based on lease conditions.

### DIFF
--- a/apps/api/src/deployment/repositories/lease/lease.repository.ts
+++ b/apps/api/src/deployment/repositories/lease/lease.repository.ts
@@ -11,6 +11,13 @@ export interface DrainingDeploymentOutput {
   blockRate: number;
   predictedClosedHeight: number;
   closedHeight?: number;
+  /**
+   * True when every lease of the deployment has reached a terminal state (closed or reclaiming).
+   * A reclaiming lease is terminal because a provider-initiated reclamation cannot be cancelled,
+   * so reclamation always proceeds to close. Treated like a closed deployment for auto-top-up.
+   * Only populated by the RPC source; the DB source lacks reclaim state and leaves it undefined.
+   */
+  isTerminal?: boolean;
 }
 
 export interface DatabaseLeaseListParams {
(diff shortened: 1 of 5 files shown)


baktun14:
Closing — this backend change is unnecessary and slightly harmful. A reclaiming lease has `closed_on === "0"`, so `main` already treats it as active and keeps auto-top-up running until the lease actually closes at the reclamation deadline (at which point top-up stops on its own). Stopping top-up early at the `reclaiming` state would close the deployment *before* the grace-period deadline, cutting the tenant's grace window short — and unused escrow is refunded on close anyway, so there's no real waste to prevent. Correct behavior is to keep topping up reclaiming deployments until the reclaim completes, which is the existing behavior. CON-376 remains as frontend-only work.

PR #3278, opened 2026-06-05
feat(notifications): email tenant when lease enters reclaiming state

baktun14:
## Why

When a provider starts reclaiming a lease, the chain emits `EventLeaseReclaimStarted` with a deadline after which the lease may be closed. The in-app reclamation card isn't enough — a tenant who doesn't open the dashboard misses the window. This adds an email push to the affected tenant.

Fixes CON-377

## What

Notifications-only. The reclaim email rides on the existing **deployment-closed alert** as the `(owner, dseq) → channel` record — no new configurable alert, no codegen, no migration.

- **Ingestion**: recognize `akash.market.v1.EventLeaseReclaimStarted` in `TxEventsService`; the poller fetches market events alongside deployment events and publishes `akash.v1.market.lease-reclaim-started`.
- **Matching**: a direct `(owner, dseq)` lookup (owner lives in the closed alert's `conditions.value[]`, dseq in `params`). Runs in a system context with **no CASL ability scope**, so it matches both to avoid borrowing another tenant's channel on a dseq collision.
- **Exactly-once / replay-safe**: a `params.reclaimNotifiedAt` claim-then-send. The pipeline is at-least-once (pg-boss redelivery + block reprocessing), so the atomic claim guards replay.
- **Opt-out honored**: if the deployment-closed alert is `enabled = false`, skip. No alert on file → skip, no crash.
- **Email content**: stated reason (complete `LeaseClosedReason` map + fallback), deadline as absolute UTC + relative, provider identity, and a deep link to `/deployments/{dseq}`.

`reason` and `deadline` tolerate the ambiguous on-chain string-or-number JSON encoding and are normalized in the service (the chain-sdk does not export a `fromJSON` helper in its runtime bundle).

Adds `apps/notifications/CONTEXT.md` documenting the alert / notification / reclaim domain language.

### Tests
Unit (parsing, poller wiring, reason/deadline helpers, reclaim service skip/dedupe branches, handler routing) and a functional spec against a real DB covering single-send, dedupe on replay, owner-scoping, opt-out, and no-alert. `npm test` → 241 passing.

## Summary by CodeRabbit

* **New Features**
  * Added lease-reclaim alerting: notifications when a lease-reclaim-started event occurs, with human-readable reason text, formatted UTC deadline + relative timing, and a console link to the deployment.

* **Behavior**
  * Deduplication prevents duplicate notifications on event replays.
  * Alerts respect enabled/disabled state and atomically track/claim reclaim notifications to avoid re-sends.

* **Tests**
  * Added unit and functional tests covering event ingestion, message publishing, deadline formatting, reason mapping, and claim/dedup flows.

### DIFF
--- a/apps/notifications/src/modules/alert/services/reclaim-alert/reclaim-alert.service.ts
+++ b/apps/notifications/src/modules/alert/services/reclaim-alert/reclaim-alert.service.ts
@@ -1,0 +1,74 @@
+import { Injectable } from "@nestjs/common";
+import { ConfigService } from "@nestjs/config";
+
+import { LoggerService } from "@src/common/services/logger/logger.service";
+import type { AlertConfig } from "@src/modules/alert/config";
+import { AlertRepository } from "@src/modules/alert/repositories/alert/alert.repository";
+import type { AlertMessagePayload } from "@src/modules/alert/services/alert-message/alert-message.service";
+import { getLeaseClosedReasonText, toLeaseClosedReason } from "@src/modules/alert/services/reclaim-alert/lease-closed-reason";
+import { formatReclaimDeadline } from "@src/modules/alert/services/reclaim-alert/reclaim-deadline";
+import type { MessageCallback } from "@src/modules/alert/types/message-callback.type";
+
+export interface ReclaimAlertEvent {
+  owner: string;
+  dseq: string;
+  provider: string;
+  reason: string | number;
+  deadline: string | number;
+}
+
+@Injectable()
+export class ReclaimAlertService {
+  constructor(
+    private readonly alertRepository: AlertRepository,
+    private readonly configService: ConfigService<AlertConfig>,
+    private readonly loggerService: LoggerService
+  ) {
+    this.loggerService.setContext(ReclaimAlertService.name);
+  }
+
+  async alertFor(event: ReclaimAlertEvent, onMessage: MessageCallback): Promise<void> {
+    const alert = await this.alertRepository.findDeploymentClosedAlertByOwnerAndDseq(event.owner, event.dseq);
+
+    if (!alert) {
+      this.loggerService.debug({ event: "RECLAIM_ALERT_SKIPPED", reason: "NO_DEPLOYMENT_CLOSED_ALERT", owner: event.owner, dseq: event.dseq });
+      return;
+    }
+
+    if (!alert.enabled) {
+      this.loggerService.debug({ event: "RECLAIM_ALERT_SKIPPED", reason: "OPTED_OUT", alertId: alert.id });
+      return;
+    }
+
+    const claimedAlert = await this.alertRepository.claimReclaimNotification(alert.id);
+
+    if (!claimedAlert) {
+      this.loggerService.debug({ event: "RECLAIM_ALERT_SKIPPED", reason: "ALREADY_NOTIFIED", alertId: alert.id });
+      return;
+    }
+
+    await onMessage({
+      notificationChannelId: claimedAlert.notificationChannelId,
+      payload: this.buildMessage(event)
+    });
+  }
+
+  private buildMessage(event: ReclaimAlertEvent): AlertMessagePayload {
+    const reasonText = getLeaseClosedReasonText(toLeaseClosedReason(event.reason));
+    const deadline = formatReclaimDeadline(Number(event.deadline), Date.now());
+    const link = this.getConsoleLink(event.dseq);
+
+    return {
+      summary: `Deployment ${event.dseq} is being reclaimed by the provider`,
+      description:
+        `The provider ${event.provider} has started reclaiming deployment ${event.dseq} because ${reasonText}. ` +
+        `You have until ${deadline} to take action before the lease is closed. ` +
+        `Please visit ${link} to manage your deployment.`
+    };
+  }
+
+  private getConsoleLink(dseq: string): string {
+    const baseUrl = this.configService.getOrThrow("alert.CONSOLE_WEB_URL");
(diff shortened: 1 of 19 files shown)


apps/notifications/src/modules/alert/services/reclaim-alert/reclaim-alert.service.ts
```diff
@@ -0,0 +1,74 @@
...
+    const claimedAlert = await this.alertRepository.claimReclaimNotification(alert.id);
+
+    if (!claimedAlert) {
+      this.loggerService.debug({ event: "RECLAIM_ALERT_SKIPPED", reason: "ALREADY_NOTIFIED", alertId: alert.id });
+      return;
+    }
+
+    await onMessage({
+      notificationChannelId: alert.notificationChannelId,
+      payload: this.buildMessage(event)
```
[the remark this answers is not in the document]
baktun14: Fixed in 04509943e — routing now uses `claimedAlert.notificationChannelId` from the `UPDATE … RETURNING` row.

apps/notifications/src/modules/alert/services/reclaim-alert/reclaim-alert.service.ts
```diff
@@ -0,0 +1,74 @@
...
+    });
```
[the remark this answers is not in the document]
baktun14: The literal reorder (publish-before-claim) would break the exactly-once AC: on pg-boss redelivery, a publish that already succeeded would re-send, producing duplicate emails — which the dedupe exists to prevent.

Claim-then-send is deliberate and consistent with the existing deployment-closed flow (`ChainAlertService.alertFor` marks `status=TRIGGERED` before `onMessage`, and `forEachAlert` only re-queries `status=OK`, so it has the same publish-error window). Here `onMessage` just enqueues a pg-boss `createNotification` job (a single insert), so the loss window requires the DB to fail between the claim commit and that insert — at which point the whole handler throws anyway.

A genuinely loss-free *and* duplicate-free fix needs the claim `UPDATE` and the job enqueue in one transaction (outbox, like `BrokerService.publishAll`). Leaving this open for a maintainer to decide whether that hardening belongs in this PR or as a follow-up across the reclaim + deployment-closed flows together.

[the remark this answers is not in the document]
baktun14: No need — we track follow-ups in our own tracker, not GitHub issues. Resolving this as an accepted, documented tradeoff.

baktun14:
## Fixes Applied Successfully

Fixed 1 file (+ 1 test) based on 1 CodeRabbit feedback item.

**Issue addressed:** _Prevent duplicate emissions when filters overlap_ — `getBlockEvents` flat-mapped results across all filters, so overlapping filters could emit the same chain event more than once and cascade into duplicate tenant emails. The merged events are now deduplicated (preserving first-seen order) via a new `dedupeEvents` helper, with a regression test covering an overlapping-filter case.

**Files modified:**
- `apps/notifications/src/modules/chain/services/tx-events-service/tx-events.service.ts`
- `apps/notifications/src/modules/chain/services/tx-events-service/tx-events.service.spec.ts`

**Commit:** `7667f1beb`

Validation: unit tests pass (3/3), lint clean, no new type errors.

The latest autofix changes are on the `feat/notifications-lease-reclaim-email` branch.

apps/notifications/src/modules/chain/services/chain-events-poller/chain-events-poller.service.ts
```diff
@@ -145,14 +145,12 @@ export class ChainEventsPollerService implements OnApplicationBootstrap, OnModul
 
       const block = await this.blockMessageService.getMessages(nextBlockHeight, [MsgCloseDeployment["$type"], MsgCreateDeployment["$type"]], this.signal);
 
-      const txEvents = await this.txEventsService.getBlockEvents(
+      const chainEvents = await this.txEventsService.getBlockEvents(
```
ygrishajev: nit: why? these are events extracted from a block.

apps/notifications/src/modules/chain/services/tx-events-service/tx-events.service.ts
```diff
@@ -87,6 +95,24 @@ export class TxEventsService {
...
 
+  /**
+   * Removes duplicate events that arise when overlapping filters match the same
+   * chain event, preserving first-seen order. Without this, a single chain event
+   * could be published — and the tenant emailed — multiple times.
+   *
+   * @param events - Events merged from all filters
+   * @returns Events with duplicates removed
+   */
+  private dedupeEvents(events: ProcessedEvent[]): ProcessedEvent[] {
```
ygrishajev: q: why have you added this? any real use case?

ygrishajev (approved):

apps/notifications/src/modules/chain/services/chain-events-poller/chain-events-poller.service.ts
baktun14: Good call — renamed to `blockEvents` since these are extracted from the block. Done in d405b07d6.

apps/notifications/src/modules/chain/services/tx-events-service/tx-events.service.ts
baktun14: No real use case — removed it in d405b07d6. It was added speculatively for a CodeRabbit "overlapping filters" concern, but the only caller passes mutually exclusive filters (`deployment`/`deployment-closed` vs `market`/`lease-reclaim-started`), so no event can ever match both and the dedupe was dead code. Kept a multi-filter test covering the real deployment + market case.

PR #3284, opened 2026-06-08
feat(provider): align attributes schema with audit requirements

Zblocker64:
## Summary
- Sync provider attributes with audit requirements: GPU models from provider-configs, SHM/CUDA/persistent storage keys, and `discord-username`
- Remove Chia attributes from schema, API, and UI
- Regenerate OpenAPI, `console-api-types`, and API docs snapshot via `sdk:gen`

## Test plan
- [x] `npm run sdk:gen`
- [x] `provider.service.spec.ts` unit tests
- [x] `docs.spec.ts` functional snapshot updated (with local Postgres)
- [ ] CI green on PR

## Summary by CodeRabbit

* **New Features**
  * Added provider attributes: Shared Memory (SHM), GPU capabilities, persistent storage class/capability, CUDA support, datacenter, and Discord username.

* **Bug Fixes**
  * Improved location-region detection with fallback to legacy region for backward compatibility.

* **Improvements**
  * UI and API now surface the new hardware capability fields; removed obsolete Chia/workload and disk-related attributes.

* **Tests / Seeders**
  * Updated test seeders and specs to cover new attribute keys and mappings.


apps/api/src/provider/services/provider-attributes-schema/provider-attributes-schema.service.ts
```diff
@@ -9,7 +11,12 @@ export class ProviderAttributesSchemaService {
   constructor(private readonly gitHubHttpService: GitHubHttpService) {}
 
   @Memoize({ ttlInSeconds: minutesToSeconds(5) })
-  async getProviderAttributesSchema() {
+  async getProviderAttributesSchema(): Promise<ProviderAttributesSchema> {
+    if (process.env.DEPLOYMENT_ENV === "test") {
+      const schemaPath = path.join(__dirname, "../../../../../../config/provider-attributes.json");
+      return JSON.parse(await fs.readFile(schemaPath, "utf8")) as ProviderAttributesSchema;
+    }
```
baktun14: This puts a test-only branch in the production path — every prod call now reads `process.env.DEPLOYMENT_ENV` directly (we go through injected config everywhere else, e.g. `configService.get("DEPLOYMENT_ENV")`) and loads a fixture off disk via `../../../../../../config/...`.

It also shadows an existing double: `DEPLOYMENT_ENV=test` in both `.env.unit.test` and `.env.functional.test`, so this branch bypasses the `GitHubHttpService` fake that `provider-regions.service.integration.ts` injects — that fake is dead code now.

Is there actually a test that needs this? Every consumer already stubs the schema — unit via `mock<ProviderAttributesSchemaService>()`, the integration test via its own injected fake — so I think it can just be dropped.

If something does need a container-resolved schema, it belongs in the test bootstrap, not the service — same override we already do for `RAW_APP_CONFIG` in `setup-functional-tests.ts`:

```ts
container.register(GitHubHttpService, {
  useValue: mock<GitHubHttpService>({ getProviderAttributesSchema: async () => loadLocalProviderAttributesSchema() })
});
```

Service stays a passthrough, anything resolving it through the container gets the local schema, and the integration test can drop its `as unknown as GitHubHttpService` cast too.

apps/api/src/utils/map/provider.spec.ts
```diff
@@ -0,0 +1,131 @@
+import type { Provider } from "@akashnetwork/database/dbSchemas/akash";
+import type { ProviderAttributesSchema } from "@akashnetwork/http-sdk";
+import { describe, expect, it } from "vitest";
+
+import { createProviderSeed } from "../../../test/seeders/provider.seeder";
+import { mapProviderToList } from "./provider";
+
+const schemaDetail = { key: "test", type: "string" as const, required: false, description: "test", values: null };
+
+const providerAttributeSchemaStub: ProviderAttributesSchema = {
```
baktun14: This `providerAttributeSchemaStub` is a second copy of the one in `provider.service.spec.ts`, and they've already drifted — that one uses bare `schemaDetail` for every key, this one sets per-key `key`/`type`. Worth pulling into a `createProviderAttributesSchema()` seeder under `test/seeders/` and using it in both — kills ~30 lines in each and stops them diverging further.

While you're in there, give the `multiple-option` entries real `values`. With `values: null`, `getProviderAttributeValue` short-circuits to `null` for `hardware-shm`/`hardware-gpu-capability`, so `hardwareShm`/`hardwareGpuCapabilities` can't actually be asserted — and right now they aren't tested at all. A seeder with representative `values` lets you add a case for each.

apps/api/src/provider/services/provider/provider.service.ts
```diff
@@ -254,8 +254,12 @@ export class ProviderService {
       this.providerAttributesSchemaService.getProviderAttributesSchema()
     ]);
 
+    const mapped = mapProviderToList(provider, providerAttributeSchema, auditors, lastSuccessfulSnapshot ?? undefined);
+
     return {
-      ...mapProviderToList(provider, providerAttributeSchema, auditors, lastSuccessfulSnapshot ?? undefined),
+      ...mapped,
+      hardwareGpuCapabilities: mapped.hardwareGpuCapabilities ?? [],
```
baktun14, line 261: `ProviderDetail` singles out `hardwareGpuCapabilities`/`hardwareShm` as non-null `string[]` — via `Omit<ProviderList, …>` in `types/provider.ts` plus this `?? []` — but `hardwareGpuModels`, same shape and same mapper path, stays nullable. And the list side contradicts itself: `ProviderListResponseSchema` keeps these `.nullable()` while deploy-web's `ApiProviderList` types them non-null `string[]`, so the list contract lies and throws the moment anything `.map`s them.

Cleaner to normalize once in the mapper — have `getStringArrayAttribute` return `[]` instead of `null` — and drop the `Omit` + the `?? []` here.

apps/provider-console/src/components/become-provider/ProviderAttributes.tsx
```diff
@@ -163,15 +163,15 @@ export const ProviderAttributes: React.FunctionComponent<ProviderAttributesProps
       const updatedProviderProcess = {
         ...providerProcess,
         attributes: data.attributes.map(attr => ({
-          key: attr.key === "unknown-attributes" ? attr.customKey || "" : attr.key || "",
+          key: attr.customKey?.trim() ? attr.customKey : attr.key || "",
```
baktun14: This now routes *any* attribute with a non-empty `customKey` onto the chain key, not just the old `"unknown-attributes"` sentinel — needed since the GPU caps moved onto `hardware-gpu-capability`, but it's a broader change. Worth confirming nothing else in the form carries a `customKey` that would now silently override its real `key`.

Same `.map` is also duplicated in both branches — pull it into one helper.

apps/provider-console/src/components/become-provider/ProviderAttributes.tsx
```diff
@@ -81,24 +81,24 @@ const createGpuAttributes = (gpuConfigs: GpuConfig[] | undefined) => {
...
         value: "true",
-        customKey: `capabilities/gpu/vendor/${vendor}/model/${model}/ram/${memory}/interface/${iface}`
+        customKey: `capabilities/gpu/vendor/${vendor}/model/${model}/interface/${iface}`
       },
       {
-        key: "unknown-attributes",
+        key: "hardware-gpu-capability",
         value: "true",
-        customKey: `capabilities/gpu/vendor/${vendor}/model/${model}/interface/${iface}`
+        customKey: `capabilities/gpu/vendor/${vendor}/model/${model}/interface/${iface}/ram/${memory}`
```
baktun14: This flips the fully-specified GPU capability key from `…/model/<m>/ram/<r>/interface/<i>` to `…/model/<m>/interface/<i>/ram/<r>` and drops the canonical ram-first combined key (old #3 became model+interface). `customKey` is what we publish on-chain (mapped at :166/:174), and the chain bid gate string-matches it order-sensitively — so this stops providers from bidding on GPU deployments that pin both ram and interface.

End to end:
- The deployment's GPU requirement key is built **ram-before-interface** by the SDK we ship — `@akashnetwork/chain-sdk` `transformGpuAttributes` (`sdl/manifest/manifestUtils`): `key = vendor/<v>/model/<m>` → `if (model.ram) key += /ram/<r>` → `if (model.interface) key += /interface/<i>`.
- Matching is a whole-string glob, not field-by-field: `groupspec.go MatchResourcesRequirements` → `GetCapabilitiesMap("gpu")` (rejoins segments verbatim) → `Attribute.SubsetOf` = `filepath.Match(deploymentKey, providerKey)`. (Our own `gpu-attribute-parser` is order-tolerant, but that's the post-lease hardware matcher, not the chain bid gate.)
- So `filepath.Match("…/ram/<r>/interface/<i>", "…/interface/<i>/ram/<r>")` → false → `shouldBid` declines. Model-only / model+ram / model+interface still match (those keys stay), so only the fully-specified case breaks — which is the common datacenter case (a100 `ram/80Gi` + `interface/sxm`).

Live mainnet backs this up: across providers, ram-before-interface beats interface-before-ram ~3.4× (208 vs 61 occurrences). The interface-first minority are the providers already silently failing to match fully-specified requests today; this moves console-onboarded providers into that bucket.

Heads up: the akash.network provider-attributes doc (`akash-network/website`, `…/operations/provider-attributes/index.md`) shows the combined key as `interface/ram`, which probably drove this — but the doc is wrong/non-matchable for that key; the SDK + chain matcher are ground truth.

Fix: keep ram-before-interface — revert this combined key to `…/ram/<r>/interface/<i>`. The new model+interface (no-ram) key is fine to keep as extra coverage, and the `hardware-gpu-capability` rename + `customKey?.trim()` logic are fine. It's only the order flip (and dropping the ram-first combined key) that breaks matching.

apps/provider-console/src/components/become-provider/ProviderAttributes.tsx
baktun14, line 1: @jigar-arc10 can you confirm this doesn't affect provider console in a bad way?

apps/api/test/setup-integration-tests.ts
```diff
@@ -1,13 +1,17 @@
...
 import { container } from "tsyringe";
 
 import { cacheEngine } from "@src/caching/helpers";
+import { registerLocalProviderAttributesSchemaGitHubHttpService } from "./helpers/register-local-provider-attributes-schema";
 import { TestDatabaseService } from "./services/test-database.service";
 
 const testPath = expect.getState().testPath;
 const dbService = new TestDatabaseService(testPath!);
 
+registerLocalProviderAttributesSchemaGitHubHttpService();
```
baktun14: this is global setup, is it needed for all the tests?

apps/api/test/setup-functional-tests.ts
```diff
@@ -34,6 +36,7 @@ export function clearCache(keyOrPrefix?: string) {
 }
 
 container.register(RAW_APP_CONFIG, { useValue: process.env });
+registerLocalProviderAttributesSchemaGitHubHttpService();
```
baktun14: same thing here

apps/api/src/utils/map/provider.ts
```diff
@@ -125,12 +128,62 @@ function getStringAttribute(key: keyof ProviderAttributesSchema, attrMap: Map<st
...
 }
 
+function getLocationRegion(attrMap: Map<string, string>, schema: ProviderAttributesSchema): string | null {
+  const locationRegion = getStringAttribute("location-region", attrMap, schema);
+  if (locationRegion) {
+    return locationRegion;
+  }
+
+  // Legacy providers may still have `region` on-chain; console expects `location-region`.
+  return attrMap.get("region") || null;
```
baktun14: This is going to make it harder to spot providers that haven't set the good region attribute that we expect them to set

apps/api/swagger/openapi.json
```diff
@@ -2,7 +2,7 @@
   "openapi": "3.0.0",
   "servers": [
     {
-      "url": "https://console-api-sandbox.akash.network"
+      "url": "http://localhost:3080"
```
baktun14: any reason why this is now localhost?

apps/provider-console/src/components/become-provider/ProviderAttributes.tsx
```diff
@@ -81,24 +81,24 @@ const createGpuAttributes = (gpuConfigs: GpuConfig[] | undefined) => {
...
     return [
       {
-        key: "unknown-attributes",
+        key: "hardware-gpu-capability",
         value: "true",
         customKey: `capabilities/gpu/vendor/${vendor}/model/${model}`
       },
       {
-        key: "unknown-attributes",
+        key: "hardware-gpu-capability",
```
jigar-arc10, line 89: @Zblocker64 So the key and value onchain is not changed correct?

key is capabilities/gpu/vendor/${vendor}/model/${model} and value is true?

apps/provider-console/src/components/become-provider/ProviderAttributes.tsx
```diff
@@ -81,24 +81,24 @@ const createGpuAttributes = (gpuConfigs: GpuConfig[] | undefined) => {
 
     return [
       {
-        key: "unknown-attributes",
+        key: "hardware-gpu-capability",
```
jigar-arc10, line 84: **[Medium] GPU capability paths are now hidden in the UI.**

These generated GPU rows use `key: "hardware-gpu-capability"` with the real `capabilities/gpu/vendor/.../model/...` path stored in `customKey`. But the `customKey` input is rendered only when the row key is `"unknown-attributes"` (L264), so for these rows the path never renders. The user sees N identical `hardware-gpu-capability = true` rows with no way to see or edit which capability — a regression vs. the prior `unknown-attributes` behavior where the path was visible.

**AI prompt to fix:**
> In `apps/provider-console/src/components/become-provider/ProviderAttributes.tsx`, the `customKey` input is rendered only when the row key equals `"unknown-attributes"` (around line 264). GPU capability rows now use the key `"hardware-gpu-capability"` and store their real on-chain path in `customKey`, so their path is currently hidden. Update that conditional so it also renders the `customKey` input for `"hardware-gpu-capability"` — e.g. `["unknown-attributes", "hardware-gpu-capability"].includes(form.watch(`attributes.${index}.key`))`. Keep the existing `unknown-attributes` behavior unchanged. Do not modify unrelated lines.

apps/provider-console/src/components/become-provider/ProviderAttributes.tsx
```diff
@@ -122,6 +122,27 @@ const providerFormSchema = z.object({
...
+function toOnChainAttributeKey(attr: ProviderFormAttribute): string {
+  if (attr.key === "unknown-attributes") {
+    return attr.customKey?.trim() || "";
+  }
+
+  if (attr.key === "hardware-gpu-capability" && attr.customKey?.trim()) {
+    return attr.customKey.trim();
+  }
+
+  return attr.key || "";
```
jigar-arc10, line 136: **[Medium / blocking] A manually-added `hardware-gpu-capability` row writes a dead on-chain key.**

`hardware-gpu-capability` is selectable in the key dropdown (L225), but no `customKey` input renders for it (L264), so `customKey` stays empty and this fallback (`return attr.key`) returns the literal `"hardware-gpu-capability"`. The API resolves GPU capabilities only via the `capabilities/gpu/vendor/.../model/...` paths, so the provider pays a tx to write an attribute Console can never read.

Suggested fix — collapse the two wrapper branches and drop unresolved rows before submit:
```ts
function toOnChainAttributeKey(attr: ProviderFormAttribute): string {
  if (attr.key === "unknown-attributes" || attr.key === "hardware-gpu-capability") {
    return attr.customKey?.trim() || "";
  }
  return attr.key || "";
}

function toOnChainAttributes(attributes: ProviderFormValues["attributes"]): Array<{ key: string; value: string }> {
  return attributes
    .map(attr => ({ key: toOnChainAttributeKey(attr), value: attr.value }))
    .filter(attr => attr.key !== "");
}
```

**AI prompt to fix:**
> In `apps/provider-console/src/components/become-provider/ProviderAttributes.tsx`, `toOnChainAttributeKey` falls through to `return attr.key` for a `hardware-gpu-capability` row whose `customKey` is empty, emitting the literal invalid on-chain key `"hardware-gpu-capability"`. Merge the `hardware-gpu-capability` branch into the `unknown-attributes` branch so both resolve to `attr.customKey?.trim() || ""`, and update `toOnChainAttributes` to `.filter(attr => attr.key !== "")` so rows whose key can't be resolved are never submitted. Add a unit spec for `toOnChainAttributes` covering: a known key passes through, an `unknown-attributes` row uses `customKey`, a `hardware-gpu-capability` row with a path uses `customKey`, and a `hardware-gpu-capability` row with empty `customKey` is dropped.

apps/provider-console/src/components/become-provider/ProviderAttributes.tsx
```diff
@@ -122,6 +122,27 @@ const providerFormSchema = z.object({
...
+}
+
+function toOnChainAttributes(attributes: ProviderFormValues["attributes"]): Array<{ key: string; value: string }> {
+  return attributes.map(attr => ({
+    key: toOnChainAttributeKey(attr),
+    value: attr.value
+  }));
+}
+
 export const ProviderAttributes: React.FunctionComponent<ProviderAttributesProps> = ({ onComplete, existingAttributes, editMode }) => {
```
jigar-arc10, line 146: **[Minor] Edit-mode round-trip is asymmetric for GPU capabilities.**

The `existingAttributes.map(...)` mapping just below (≈L156–161) sends any key not in `attributeKeys` to `"unknown-attributes"`. On reload, on-chain `capabilities/gpu/...` keys aren't in `attributeKeys`, so they map back to `unknown-attributes` rather than `hardware-gpu-capability` — the distinct treatment doesn't survive a save/reopen. If `hardware-gpu-capability` is meant to be a first-class representation, reverse-map GPU paths on load.

**AI prompt to fix:**
> In `apps/provider-console/src/components/become-provider/ProviderAttributes.tsx`, the `existingAttributes.map(...)` default-values mapping (around line 156–161) sends any key not in `attributeKeys` to `"unknown-attributes"`, so on-chain GPU capability keys (`capabilities/gpu/vendor/.../model/...`) don't round-trip back to `"hardware-gpu-capability"`. Add a branch: if `attr.key` matches `/^capabilities\/gpu\//`, map it to `{ key: "hardware-gpu-capability", value: attr.value, customKey: attr.key }`; otherwise keep the existing known/unknown logic. Keep this consistent with the render gate and `toOnChainAttributeKey` so the create and edit flows show GPU capabilities identically.

jigar-arc10 (commented):
Three linked findings on the new `hardware-gpu-capability` handling (inline). Root cause: it's a `customKey`-backed key, but the render + round-trip logic still special-cases only `unknown-attributes`. The key rename, `customKey?.trim()` logic, and GPU key ordering are all correct — these are about the surrounding plumbing.

apps/provider-console/src/components/become-provider/ProviderAttributes.tsx
Zblocker64: correct

PR #3311, opened 2026-06-13
fix(indexer): remove legacy monitored value monitors

baktun14:
## Why

The indexer's `DeploymentBalanceMonitor` has been throwing non-stop in production — Sentry **CLOUDMOS-SYNCER-5C**, 5180+ events since 2026-05-08 — whenever it can't fetch a deployment's escrow balance.

That monitor and its sibling `AddressBalanceMonitor` are leftovers from the old Blockspy/Cloudmos system (the indexer README flagged them "Blockspy specific" and "not used in the deploy tool"). Nothing in the monorepo reads the `monitoredValue` table they write to, so the whole feature is dead weight that only generates error noise.

Fixes CLOUDMOS-SYNCER-5C

## What

Removes the legacy MonitoredValue feature end to end:

- Deleted both monitors (`apps/indexer/src/monitors/`) and unregistered their scheduled tasks in `index.ts`.
- Removed the `MonitoredValue.sync()` call (`buildDatabase.ts`) and the shared Sequelize model (`packages/database`), including its entry in `baseModels` — also consumed by `apps/api`, which never queries the table (verified, type-checks clean).
- Cleaned the table from the Drizzle `schema.ts` and removed the two task rows from the indexer README.

**⚠️ Production DB migration:** `0009_drop_monitored_value.sql` runs `DROP TABLE IF EXISTS "monitoredValue"`. No code in this PR reads or writes the table anymore, so the drop is non-blocking and safe. Historical migrations and `meta/` snapshots are intentionally left intact as immutable history.

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed obsolete balance-monitoring tasks from the scheduler list.
  * Cleaned up the indexer by dropping the legacy monitored-value data storage.
* **Chores**
  * Updated documentation to match the current scheduled task list.
  * Removed unused database schema references related to monitored values.


stalniy (approved):
🔥

PR #3334, opened 2026-06-18
fix: make BME settlement polling abortable

miloquinn:
## Summary
- Stop `waitForLedgerRecordsSettlement` immediately when its abort signal fires during the poll interval.
- Clear the pending timeout and return `false` instead of waiting for the full interval before observing cancellation.
- Add a regression test covering abort during the sleep between polling attempts.

## Testing
- `npm test --workspace packages/http-sdk -- --run src/bme/bme-http.service.spec.ts`
- `npm run validate:types --workspace packages/http-sdk`
- `npm run lint --workspace packages/http-sdk -- src/bme/bme-http.service.ts src/bme/bme-http.service.spec.ts` (passes with existing warnings elsewhere in `packages/http-sdk`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ledger settlement polling to support proper cancellation during polling intervals. Operations can now be aborted without waiting for the current interval to complete, providing better responsiveness when cancelling ledger verification processes.

* **Tests**
  * Added unit tests for ledger settlement polling cancellation behavior with abort signal verification to ensure correct behavior under interruption scenarios.

### DIFF
--- a/packages/http-sdk/src/bme/bme-http.service.spec.ts
+++ b/packages/http-sdk/src/bme/bme-http.service.spec.ts
@@ -1,0 +1,48 @@
+import { describe, expect, it, vi } from "vitest";
+
+import type { HttpClient } from "../utils/httpClient";
+import { BmeHttpService } from "./bme-http.service";
+
+describe(BmeHttpService.name, () => {
+  describe("waitForLedgerRecordsSettlement", () => {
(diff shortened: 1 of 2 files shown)


baktun14:
Hi, why do you need this? Bme is not visible to users anymore. Self custody is now on https://github.com/akash-network/console-air

PR #3387, opened 2026-07-04
feat(sdl): auto-match confidential compute deployments to capable providers

baktun14:
## Why

Part of CON-543.

When a tenant enables Confidential Compute (TEE) on a deployment, it must only ever be offered to providers that can actually run it. Today the per-service `params.tee` choice only lands in the provider manifest (`TEEParams`) — it never becomes an on-chain group placement requirement, so **any** provider can bid on a confidential workload and a won bid can't be honored as confidential.

`@akashnetwork/chain-sdk`'s `generateManifest` builds `group_spec.requirements.attributes` exclusively from `profiles.placement.<name>.attributes`; it does not derive anything from `params.tee`. So the builder must inject the matching requirement itself.

## What

Auto-manage a `tee/type` provider-matching requirement derived from the tenant's TEE choice, fully round-tripped and hidden from the manual attribute editor.

- **Export** (`sdlGenerator.ts`): derive each placement's TEE type from its services' `params.tee` (CON-449 guarantees one shared type per placement) and merge `tee/type: cpu | cpu-gpu` into the placement attributes — mirroring the existing `location-region` handling.
- **Import** (`sdlImport.ts`): drop `tee/type` in `hydratePlacement` (same lift as `location-region`) so it never populates the hand-editable attribute list and can't duplicate on re-export. The value is re-derived from `params.tee` (which already round-trips) on every export.

Single injection point covers both consumers: once `tee/type` is on the group spec, on-chain bidding and the marketplace bid-screening request (`buildPlacementScreeningRequest`) both pick it up automatically. Non-TEE deployments are unchanged.

**Acceptance criteria**
- [x] Enabling Confidential Compute auto-restricts the deployment to capable providers, no manual action.
- [x] CPU and CPU+GPU deployments round-trip (import → edit → re-export) losslessly, ending with exactly one requirement, no duplicates.
- [x] Re-importing an SDL that already carries the requirement does not add a second/conflicting one.
- [x] The requirement is not exposed in the manual attribute editor and cannot be hand-edited.
- [ ] Consistent in managed Console **and Console Air** — this repo covers managed Console; Console Air is a separate repo (`akash-network/console-air`) with its own copy of these utils, so the same pattern must be mirrored there as a follow-up.

**Verification**
- New unit tests in `sdlGenerator.spec.ts` / `sdlImport.spec.ts` cover injection (cpu/cpu-gpu), no-injection when off, merge with region + other attributes, single attribute for multi-service placements, import-drop, and no-duplication on re-import. Full affected SDL suite: 75 passing.
- Confirmed end-to-end against the real `generateManifest`: a generated TEE SDL yields `group requirements.attributes: [{"key":"tee/type","value":"cpu-gpu"}]`.

**Note (to confirm with provider/AEP-83):** matching key is `tee/type` with value `cpu`/`cpu-gpu` — the contract the console read side (`confidentialCompute.ts`) already uses. Only the constant changes if the provider expects a different key.

## Summary by CodeRabbit

* **New Features**
  * SDL generation now automatically adds the required confidential compute placement attribute when a service requests a TEE type.
  * Import and export flows now preserve this placement requirement without exposing it as a manually editable field.

* **Bug Fixes**
  * Prevented duplicate TEE placement requirements when multiple services share the same placement.
  * Improved round-tripping so existing placement attributes remain intact while derived TEE requirements are handled consistently.

### DIFF
--- a/apps/deploy-web/src/utils/sdl/sdlGenerator.spec.ts
+++ b/apps/deploy-web/src/utils/sdl/sdlGenerator.spec.ts
@@ -64,6 +64,46 @@ describe("sdlGenerator", () => {
       expect(parsedAny.profiles.placement["dcloud"].attributes?.["location-region"]).toBeUndefined();
     });
 
+    it.each([["cpu"], ["cpu-gpu"]] as const)("injects the tee/type placement requirement from a service's params.tee (%s)", tee => {
+      const result = generateSdl(buildFormValues(buildLogCollectorService({ title: "web", image: "nginx:latest", params: { tee } })));
+      const parsed = yaml.load(result) as { profiles: { placement: Record<string, { attributes?: Record<string, string> }> } };
+
+      expect(parsed.profiles.placement["dcloud"].attributes).toMatchObject({ "tee/type": tee });
+    });
+
+    it("does not inject a tee/type placement requirement when no service requests confidential compute", () => {
+      const result = generateSdl(buildFormValues(buildLogCollectorService({ title: "web", image: "nginx:latest" })));
+      const parsed = yaml.load(result) as { profiles: { placement: Record<string, { attributes?: Record<string, string> }> } };
+
+      expect(parsed.profiles.placement["dcloud"].attributes?.["tee/type"]).toBeUndefined();
+    });
+
+    it("merges the tee/type requirement alongside location-region and other declared attributes", () => {
+      const formValues = buildFormValues(buildLogCollectorService({ title: "web", image: "nginx:latest", params: { tee: "cpu-gpu" } }));
+      formValues.placements[0].region = "us-west";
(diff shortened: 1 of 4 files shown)


baktun14:
Closing as redundant. chain-sdk 1.0.0-alpha.38 (already pinned in this repo) auto-injects the `tee/type` provider-matching requirement into group requirements from `service.params.tee` in `generateManifest` (lines 89-92), which both on-chain bidding and bid-screening already consume. This PR was written against a stale local node_modules (alpha.36) that predates that derivation. Worse, alpha.38's guard doesn't dedupe against `profiles.placement.*.attributes`, so this change would produce a duplicate `tee/type` requirement. CON-543 is effectively delivered by the dependency.

PR #3400, opened 2026-07-09
feat(onboarding): gate the app behind onboarding and route deploys through configure

ygrishajev:
## Why

With the onboarding redesign we want a single, enforced first-run path: a brand-new user is guided through onboarding and their first deployment, and can't wander into the rest of the app until they've actually deployed something. Before this PR the redesign was split across two flags and had no real gate — a not-onboarded user could reach any page, and the old wallet-based `OnboardingRedirectEffect` only covered part of it. New-deployment entry points also bypassed the new configure screen, and on a fresh trial the deployment could be created before the trial wallet was ready to sign it.

Fixes CON-317

## What

**Flag consolidation**
- Merged `console_onboarding_redesign` + `bid_screening` into a single `onboarding_redesign_v1` flag across the FE flag type and the API `FeatureFlags` (including bid-screening's gate).

**Onboarding gate (`RequireOnboarding`)**
- A blocking wrapper in `_app` — nothing renders until the user's onboarding state is known (no flash of gated content).
- "Onboarded" = the user has ever created a lease (read client-side via `useAllLeases`); a user whose deployments are all closed still counts.
- Not-onboarded users are confined to the onboarding + first-deploy path — the onboarding picker, `/new-deployment/configure*`, and a single deployment's detail (`/deployments/{dseq}`) — and are bounced to `/onboarding` from everywhere else (including the `/deployments` list).
- Onboarded users are free everywhere and are sent into the console if they land on `/onboarding`.
- Flag-off preserves the legacy wallet-based behavior; public pages are never gated.
- Removed the legacy `OnboardingRedirectEffect` and `useIsOnboarded` in favor of the single gate.

**Deploy-entry routing**
- The classic `/new-deployment` deployment-type/template picker stays the deploy entry: a bare **Deploy** opens the picker; a chosen template or an `edit-deployment` intent opens the configure screen; git, redeploy, and VM (`deploy-linux`) intents keep the classic builder. Wired via `useNewDeploymentUrl` (in-app links) and `RedirectMappableBuilderToConfigure` (inbound `/new-deployment?...` links).

**Deployment creation waits for the trial wallet**
- The start-trial request now survives navigation (a stable mutation key + a global in-flight check in `useEnsureTrialStarted`), so navigating onboarding → configure no longer fires duplicate start-trial calls.
- Create-deployment is gated in one place (`useCreateDeployment`): it holds the broadcast until the trial wallet can sign, instead of firing prematurely and failing. Both the manual configure flow and the auto-deploy flow funnel through it.
- The gate no longer waits on the leases query for allow-list pages, so the trial wallet arriving mid-deploy doesn't remount an in-progress configure/auto flow (which previously stalled the deploy and forced a re-click).

**Tests**
- New e2e: `onboarding-gate` (the routing/gate contract) and `onboarding-journey` (new-user template auto-deploy + bring-your-own-image). Registration goes through passwordless email OTP; existing-user login uses email + password.
- Skip strategy revisited: legacy classic-deploy specs skip under the flag; the flag-agnostic SDL-builder and VM (`deploy-linux`) specs run under both; `deploy-from-a-template` and `managed-wallet-alerts` are migrated to the redesign flow.
- Unit coverage for `RequireOnboarding`, `useNewDeploymentUrl`, `useCreateDeployment`, `useEnsureTrialStarted`, and `RedirectMappableBuilderToConfigure`.

## Summary by CodeRabbit

* **New Features**
  * Added onboarding redesign V1 flag support to drive the new onboarding and “mappable to configure” redirect behavior.
  * Improved trial retry handling: retries now trigger before re-requesting quotes and starting deployment creation.
  * Enhanced “Deploy” link routing using a centralized destination resolver for consistent navigation.
* **Bug Fixes**
  * Fixed onboarding gating and redirects for logged-in and public pages.
  * Updated bid-screening behavior to return an empty provider list when onboarding redesign V1 is disabled.
* **Tests**
  * Added/updated unit and UI coverage for onboarding gating and trial retry flows.


apps/deploy-web/src/components/new-deployment/RedirectMappableBuilderToConfigure/RedirectMappableBuilderToConfigure.tsx
```diff
@@ -0,0 +1,47 @@
...
+export const DEPENDENCIES = { useFlag, useRouter, UrlService };
+
+type Props = { children: ReactNode; dependencies?: typeof DEPENDENCIES };
+
+/**
+ * Client-side redirect for inbound/external links to the classic builder. When the redesign flag is on and
+ * the `/new-deployment` query is a mappable intent (template or blank editor, not git/redeploy), it replaces
+ * to the configure screen. Intentionally client-only — no SSR (the team is phasing getServerSideProps out).
+ *
+ * The CI/CD template is a git/remote-deploy intent that `NewDeploymentContainer` opens as such (keyed on
```
baktun14, line 21: this option is currently disabled

apps/deploy-web/src/types/feature-flags.ts
```diff
@@ -9,7 +9,6 @@ export type FeatureFlag =
   | "console_embedded_login"
   | "console_auth_passwordless"
   | "console_auth_password_escape_hatch"
-  | "console_onboarding_redesign"
-  | "bid_screening"
```
baktun14, line 13: @stalniy I think we can enable bid screening for everyone

baktun14 (approved):

apps/deploy-web/src/types/feature-flags.ts
stalniy: we need to remove it  then

apps/deploy-web/src/components/new-deployment/RedirectMappableBuilderToConfigure/RedirectMappableBuilderToConfigure.tsx
ygrishajev: this is not a blocking thing

PR #3437, opened 2026-07-15
fix(billing): ensure Stripe customer exists before coupon redemption

baktun14:
## Why

Redeeming a coupon failed on any account that had never opened the "Purchase credits" tab. A Stripe customer is created lazily — the only place `customers.create` runs is the race-safe get-or-create `StripeService.getStripeCustomerId()`, and its only caller was `createSetupIntent` (the add-payment-method flow). Brand-new accounts had `stripeCustomerId = null`, and the coupon path hard-asserted the id, returning a 500 "Payment account not properly configured. Please contact support." with no credits added.

Fixes CON-676

## What

**Lazy-ensure the Stripe customer in the coupon path**
- `stripe.service.ts` — `applyCouponOrPromotionCode` ensures the customer via the existing race-safe `getStripeCustomerId()` only after the coupon is known to be redeemable (`valid` / `percent_off` / `amount_off` checks), immediately before invoice creation — so invalid or unsupported coupon attempts never provision a customer. Removed the now-redundant `assert`.
- `getStripeCustomerId()` now passes a Stripe idempotency key (`create-customer:${user.id}`), so concurrent provisioning (e.g. two simultaneous billing actions) can never create duplicate/orphaned customers. The DB reconciliation (conditional update on `stripeCustomerId IS NULL` + reload on lost race) is retained.
- `stripe.controller.ts` — dropped the `applyCoupon` pre-gate `assert(currentUser.stripeCustomerId, 500, ...)`. Kept `couponId`/`userId` asserts. `confirmPayment`/`removePaymentMethod` unchanged (they legitimately require a prior payment method).

### Tests
- Service specs: brand-new account (`stripeCustomerId: null`) redeems a coupon — asserts `customers.create` runs with the idempotency key, the user row is updated, and the invoice/item/`coupon_claim` use the newly-created customer id.
- Service specs: rejection paths do **not** provision a customer (unknown promotion code, unsupported percentage coupon).

### Not included
- **No eager provisioning at registration** — an earlier revision preloaded the customer from `registerUser` behind a `CustomerProvisioner` port; dropped per review since the lazy ensure is the correctness guarantee and the preload was a best-effort latency optimization that added a `user → billing` abstraction. If first-billing latency ever matters, it belongs as a `UserRegistered` event + job handler.
- No backfill for existing null-customer accounts (would require a Stripe call per historical user). They're covered by the lazy-ensure path on their next billing action.

## Summary by CodeRabbit

* **New Features**
  * New accounts can apply eligible fixed-amount coupons without an existing billing profile.
  * Billing profiles are created automatically when needed during coupon redemption.
  * Coupon redemption now prevents duplicate billing profiles during concurrent requests.

* **Bug Fixes**
  * Invalid or unsupported coupons continue to be rejected before creating a billing profile.


apps/api/src/billing/controllers/stripe/stripe.controller.spec.ts
```diff
@@ -273,6 +273,27 @@ describe(StripeController.name, () => {
...
+        transactionStatus: "pending"
+      });
+
+      const result = await controller.applyCoupon({
+        couponId: faker.string.alphanumeric(10),
+        userId: user.id
+      });
+
+      expect(stripe.applyCoupon).toHaveBeenCalledWith(user, expect.any(String));
+      expect(result.data).toMatchObject({ coupon: mockCoupon, amountAdded: 10, transactionId, transactionStatus: "pending" });
```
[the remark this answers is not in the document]
baktun14: Done in 2c57a7761 — the generated id is stored in a `couponId` variable and asserted exactly (`toHaveBeenCalledWith(user, couponId)`).

apps/api/src/billing/services/stripe/stripe.service.ts
```diff
@@ -329,6 +329,11 @@ export class StripeService extends Stripe {
     transactionId: string;
     transactionStatus: StripeTransactionOutput["status"];
   }> {
+    // Ensure the user has a Stripe customer before redeeming. Brand-new accounts may not have
+    // one yet since it is created lazily by the add-payment-method flow (see getStripeCustomerId).
+    const stripeCustomerId = await this.getStripeCustomerId(currentUser);
+    const payingUser = { ...currentUser, stripeCustomerId };
```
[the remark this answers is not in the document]
baktun14: Done in 2c57a7761 — `getStripeCustomerId` now runs inside `applyCouponOrPromotionCode` after the validity/percent_off/amount_off checks and immediately before invoice creation, so invalid or unsupported coupon attempts no longer provision a Stripe customer for brand-new accounts. Added service specs covering the no-customer-on-rejection paths (unknown code + unsupported percentage coupon).

apps/api/src/app/services/user-registered/user-registered.handler.spec.ts
```diff
@@ -0,0 +1,53 @@
+import { describe, expect, it, vi } from "vitest";
+import { mock } from "vitest-mock-extended";
+
+import type { StripeService } from "@src/billing/services/stripe/stripe.service";
+import type { LoggerService } from "@src/core/providers/logging.provider";
+import type { UserRepository } from "@src/user/repositories";
+import { UserRegisteredHandler } from "./user-registered.handler";
+
+import { createUser } from "@test/seeders/user.seeder";
```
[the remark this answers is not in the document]
baktun14: Not applicable — `user-registered.handler.spec.ts` (and the event/handler/job-queue approach) was removed in this branch; registration provisioning now goes through the `CustomerProvisioner` service, so this spec no longer exists.

apps/api/src/user/services/user/user.service.ts
```diff
@@ -57,6 +59,12 @@ export class UserService {
 
     if (wasInserted) {
       this.analyticsService.track(user.id, "account_created", { category: "user" });
+      // Provision a billing customer up front so billing actions (e.g. coupon redemption) never
+      // fail on a brand-new account. Best-effort: a provider outage must not fail registration —
+      // the billing layer lazily ensures the customer on the next billing action as a fallback.
+      await this.customerProvisioner.provisionCustomer(user).catch(error => {
+        this.logger.error({ event: "FAILED_TO_PROVISION_CUSTOMER", id: user.id, error });
+      });
```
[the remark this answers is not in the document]
baktun14: Done in 5c041d82d — provisioning is now dispatched fire-and-forget (`void ...provisionCustomer(user).catch(...)`), so registration no longer blocks on (or fails because of) Stripe. Verified safe: no request-scoped transaction wraps `registerUser` and the pg client is an app-wide pool disposed only on shutdown, so the background write completes normally; the lazy billing-path fallback remains the correctness guarantee. Kept the direct `CustomerProvisioner` port rather than reintroducing event/job dispatch (which this branch intentionally moved away from). Added a spec asserting registration returns even when provisioning never settles.

apps/api/src/billing/services/stripe/stripe.service.ts
```diff
@@ -405,15 +405,18 @@ export class StripeService extends Stripe {
...
     }
 
-    assert(currentUser.stripeCustomerId, 500, "Payment account not properly configured. Please contact support.");
-
     const amountToAdd = coupon.amount_off; // amount_off is already in cents
 
+    // Ensure the user has a Stripe customer only once the coupon is known to be redeemable. Brand-new
+    // accounts may not have one yet since it is created lazily by the add-payment-method flow (see
+    // getStripeCustomerId); provisioning it earlier would create customers for invalid/unsupported coupons.
+    const stripeCustomerId = await this.getStripeCustomerId(currentUser);
```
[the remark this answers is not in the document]
baktun14: **Idempotency (Major):** done in 5c041d82d — `customers.create` now passes `{ idempotencyKey: \`create-customer:${user.id}\` }`, so concurrent provisioning (eager registration + lazy billing) can't create duplicate/orphaned customers; the DB reconciliation is retained and the spec asserts the key.

**Promotion-code eligibility (Minor):** not adding a preflight. Customer-scoped / first-transaction / minimum-amount / redemption-limit restrictions are only resolved by Stripe when the discount is applied to the invoice, and creating that invoice itself requires a customer — so provisioning cannot be deferred past that point. The invoice-creation failure path already voids the invoice, and with the idempotency fix the only residual effect is a single legitimate customer the lazy fallback would create anyway. `coupon.valid`/`percent_off`/`amount_off` are already validated before provisioning.

apps/api/src/billing/controllers/stripe/stripe.controller.ts
```diff
@@ -147,7 +147,8 @@ export class StripeController {
   }> {
     const { currentUser } = this.authService;
 
-    assert(currentUser.stripeCustomerId, 500, "Payment account not properly configured. Please contact support.");
+    // The Stripe customer is ensured lazily inside stripe.applyCoupon (getStripeCustomerId),
+    // so brand-new accounts without a customer can still redeem — no pre-gate assert here.
```
ygrishajev: nit: this looks redundant, no logic added

apps/api/src/billing/controllers/stripe/stripe.controller.spec.ts
```diff
@@ -273,6 +273,28 @@ describe(StripeController.name, () => {
         }
       });
     });
+
+    it("redeems a coupon on a brand-new account without a Stripe customer", async () => {
```
ygrishajev: nit: this looks redundant, this is a unit test testing a controller, however this one exercises a mocked dep out of the controller concern

apps/api/src/user/services/user/user.service.ts
```diff
@@ -57,6 +59,13 @@ export class UserService {
 
     if (wasInserted) {
       this.analyticsService.track(user.id, "account_created", { category: "user" });
+      // Provision a billing customer up front so billing actions (e.g. coupon redemption) never
+      // fail on a brand-new account. Fire-and-forget best-effort: registration must not block on
+      // (or fail because of) a slow/unavailable Stripe — the billing layer lazily ensures the
+      // customer on the next billing action as a fallback.
+      void this.customerProvisioner.provisionCustomer(user).catch(error => {
```
ygrishajev: This eager provisioning doesn't fix the bug — `applyCoupon` already calls `getStripeCustomerId`, which creates the customer on demand, so correctness holds without it. As-is it's fire-and-forget preload: a best-effort latency optimization with no guarantee it even runs.

I'd drop the whole eager path — this `registerUser` change, the `CustomerProvisioner` port + token, `StripeCustomerProvisioner`, and the provider registration (plus their specs). That removes the extra abstraction and the `user -> billing` dependency, and leaves the real fix (lazy ensure + idempotency key) standing on its own.

If we later want the preload to cut first-billing latency, it belongs as a `UserRegistered` domain event + app-layer `JobHandler` (matching `first-purchase-bonus-granted` et al.) — not inlined into `registerUser` behind a bespoke port. But I don't think this PR needs it.

apps/api/src/billing/services/stripe/stripe.service.ts
```diff
@@ -729,13 +732,18 @@ export class StripeService extends Stripe {
...
+    // lazy billing paths) can never create duplicate/orphaned customers before the DB update wins.
+    const customer = await this.customers.create(
+      {
+        email: user.email ?? undefined,
+        name: user.username ?? undefined,
+        metadata: {
+          userId: user.id
+        }
+      },
+      { idempotencyKey: `create-customer:${user.id}` }
```
stalniy, line 745: **question:** I'm curious what happens in case of race condition. what `customer` will contain? does stripe return id of existing customer with the same parameters?

stalniy: @coderabbitai help to clarify my question

apps/api/src/user/services/user/user.service.ts
baktun14: Done in ef8e55562 — dropped the whole eager path: the `registerUser` change, `CustomerProvisioner` port + token, `StripeCustomerProvisioner`, the provider registration, and their specs. The lazy ensure in `applyCouponOrPromotionCode` + the idempotency key stand on their own; PR description updated to match.

apps/api/src/billing/controllers/stripe/stripe.controller.ts
baktun14: Removed in ef8e55562.

apps/api/src/billing/controllers/stripe/stripe.controller.spec.ts
baktun14: Removed in ef8e55562 (test + the `setup` param it needed). The real no-customer redemption behavior is covered in `stripe.service.spec.ts`.

apps/api/src/billing/services/stripe/stripe.service.ts
baktun14: What CodeRabbit said matches the implementation intent: with the idempotency key, Stripe treats the second concurrent `customers.create` as a replay and returns the *same* customer (same `id`) — no duplicate is created. Then the conditional DB update (`stripeCustomerId: null` guard) lets only one writer win; the loser reloads the row and returns the persisted id. Note the race surface shrank in ef8e55562: eager registration provisioning is gone, so this now only guards concurrent billing actions — the key + reconciliation still cover that.

stalniy (approved):

PR #3472, opened 2026-07-20
refactor: final batch of indexer migration to strict types

stalniy:
## Why

Closes CON-243


apps/indexer/src/providers/statusEndpointHandlers/rest.ts
stalniy, line 1: @baktun14 we probably don't need this anymore, right? this implementation was serving old providers before protobuf provider era

ygrishajev (approved):

PR #3498, opened 2026-07-21
refactor(billing): make StripeService hold an injected Stripe client instead of extending it

ygrishajev:
## Why

`StripeService` **extends** the Stripe SDK — it literally *is* a `Stripe` instance. So its public surface is the entire SDK, it can never be a real abstraction, and it can't be unit-tested in isolation (its own tests have to reach into `service.paymentIntents`, `service.charges`, … the inherited SDK resources).

The prior PR (#3497) routed every external caller through named methods, so nothing outside the service touches the raw SDK anymore. This PR takes the next step: make the SDK an **injected dependency** instead of a base class. That encapsulates it (only our named operations are public) and makes the service properly testable — its spec now injects and mocks the client directly. It's also the prerequisite for the layered `StripeService` split that follows.

Closes CON-720 · Part of CON-718. Stacked on #3497 — review/merge after it.

## What

- Adds a `STRIPE_CLIENT` DI provider (a factory that builds the `Stripe` client from the billing config) and injects it into `StripeService`, replacing the internal `super(...)` construction.
- `StripeService` no longer `extends Stripe`; every internal SDK call moves from `this.<resource>` to `this.stripe.<resource>`.
- The service's own spec injects a real `Stripe` client and mocks *it* (via `vi.spyOn(stripe.<resource>, …)`) instead of reaching through the service instance. Same assertions, same coverage — only the mock target changed.

## Verification

- Full **billing unit suite green — 351 tests / 26 files** (confirms dropping `extends Stripe` caused no fallout in the many `mock<StripeService>()` consumers).
- `stripe.service` spec preserves every test (proved 0 removed vs base); `stripe.controller` spec unaffected.
- `tsc`: no new errors in the changed files (pre-existing apps/api baseline unchanged).
- Prettier clean. Lint (ESLint v9 flat config) and the webhook integration test run in CI.

## Summary by CodeRabbit

* **New Features**
  * Added a centralized Stripe client for billing operations, making Stripe access consistent across customers, payments, invoices, and coupons.
* **Bug Fixes**
  * Improved Stripe payment flows, including replay/resume handling, concurrency/in-progress mapping, and more reliable 3DS processing.
  * Enhanced coupon application by creating the discounted invoice line item before finalizing the invoice and improving rollback behavior on failures.
* **Tests**
  * Updated Stripe-related tests to validate interactions with Stripe SDK primitives directly.

### DIFF
--- a/apps/api/src/billing/providers/index.ts
+++ b/apps/api/src/billing/providers/index.ts
@@ -1,3 +1,5 @@
 import "./config.provider";
+import "./stripe-client.provider";
 
 export * from "./config.provider";
+export * from "./stripe-client.provider";
--- a/apps/api/src/billing/providers/stripe-client.provider.ts
+++ b/apps/api/src/billing/providers/stripe-client.provider.ts
@@ -1,0 +1,19 @@
+import Stripe from "stripe";
+import type { InjectionToken } from "tsyringe";
+import { container, inject, instancePerContainerCachingFactory } from "tsyringe";
+
+import { BillingConfigService } from "@src/billing/services/billing-config/billing-config.service";
+
+export const STRIPE_CLIENT: InjectionToken<Stripe> = Symbol("STRIPE_CLIENT");
+
+container.register(STRIPE_CLIENT, {
+  useFactory: instancePerContainerCachingFactory(
+    c =>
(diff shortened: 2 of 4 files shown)


ygrishajev:
@coderabbitai review the PR

baktun14 (approved):

PR #3500, opened 2026-07-21
fix(billing): show all transaction types and refunds in billing history

baktun14:
## Why

Fixes CON-674.

The customer-facing **Billing → History** table only showed credit-card purchases. It was built exclusively from Stripe **charges**, so:

- **Coupon claims** and **manual credits** never appeared (they are created as Stripe invoices, not charges).
- **Refunds** never appeared (a refund reduces `amount_refunded` but leaves the charge `succeeded`).

Our internal `stripe_transactions` table already records every type and all refund state, so the fix sources the history from it instead. This matches what the internal admin console shows: card purchases, coupon claims, and manual credits together, with refunds surfaced on the original row.

## What

**API (`apps/api`)**
- `getCustomerTransactions` now reads `StripeTransactionRepository.findByUserId` + `countByUserId` (keyed on `userId`) instead of `charges.list`. Returns `{ transactions, totalCount, hasMore }` with offset pagination.
- Flat DTO: adds `type`, `amountRefunded`, `cardBrand`, `cardLast4`, `stripeInvoiceId`; drops the nested `paymentMethod`/`metadata`.
- Query param `startingAfter`/`endingBefore` → `offset`.
- `countByUserId` gains an optional `{ startDate, endDate }` filter.
- CSV export gains `Type`, `Refunded`, and `Invoice ID` columns.

**http-sdk**
- `Charge` → `BillingTransaction`; response is `{ transactions, totalCount, hasMore }`; params switch cursor → `offset`.

**deploy-web**
- History table columns: **Date**, **Type** (badge with `Visa **** 4242` subtext for card payments), **Description**, **Amount** (credits show green `+`, keeps the bonus subtext, adds a `refunded` subtext when `amountRefunded > 0`), **Status** (adds a `Refunded` badge), **Receipt**.
- `BillingContainer` drops the cursor bookkeeping in favor of `offset = pageIndex * pageSize` (net simplification).

### Needs verification before merge
**Historical completeness.** The old endpoint listed *every* Stripe charge; this DB-sourced list shows only rows in `stripe_transactions`. Any card payment predating that table (or never backfilled) would drop out of history. Please confirm against prod that historical `payment_intent` rows exist for older customers; if there is a gap, a one-time backfill of legacy charges is needed.

### Notes
- The generated `console-api-types` were intentionally **not** regenerated here: a local `sdk:gen` surfaced large pre-existing drift (format changes + `notifications` files unrelated to this change). No typed consumer reads the generated `listStripeTransactions` type (deploy-web uses the hand-written http-sdk), so this endpoint's stale generated types cause no build or runtime issue. A dedicated regen pass should follow separately.

## Summary by CodeRabbit

* **New Features**
  * Billing history now displays payment, coupon, and manual-credit transaction types.
  * Added refunded amounts, card details, descriptions, invoice IDs, receipt links, and first-purchase bonuses.
  * Transaction exports now include expanded billing details.

* **Improvements**
  * Replaced cursor-based pagination with offset-based pagination.
  * Added total transaction counts and date-range filtering.
  * Billing history now uses clearer status, amount, and transaction-type presentation.

* **Bug Fixes**
  * Missing card or receipt details now display appropriately without broken links.

### DIFF
--- a/apps/api/src/billing/controllers/stripe/stripe.controller.ts
+++ b/apps/api/src/billing/controllers/stripe/stripe.controller.ts
@@ -205,18 +205,17 @@ export class StripeController {
   @Protected([{ action: "read", subject: "StripePayment" }])
   async getCustomerTransactions(options?: {
     limit?: number;
-    startingAfter?: string;
-    endingBefore?: string;
+    offset?: number;
     startDate?: string;
     endDate?: string;
-  }): Promise<{ data: { transactions: Transaction[]; hasMore: boolean; nextPage: string | null; prevPage: string | null } }> {
+  }): Promise<{ data: { transactions: Transaction[]; totalCount: number; hasMore: boolean } }> {
     const { currentUser } = this.authService;
 
     if (!currentUser.stripeCustomerId) {
-      return { data: { transactions: [], hasMore: false, nextPage: null, prevPage: null } };
+      return { data: { transactions: [], totalCount: 0, hasMore: false } };
     }
 
-    const response = await this.stripe.getCustomerTransactions(currentUser.stripeCustomerId, options);
+    const response = await this.stripe.getCustomerTransactions(currentUser.id, options);
     return { data: response };
   }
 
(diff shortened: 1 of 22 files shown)


baktun14:
Superseded. This was too large and collided with the StripeService split (CON-718). Split into two stacked PRs, rebased onto the new `TransactionReportingService`, with the actionable review comments folded in:

- #3508 — API/data half (source history from internal transactions)
- #3509 — UI half (stacked on #3508)

Closing in favor of those.

PR #3504, opened 2026-07-22
refactor(billing): give the payment-intent charge lifecycle a single owner

ygrishajev:
## Why

`StripeService` is still a god object. After reporting was carved out, the next self-contained concern living inside it is the **payment-intent charge lifecycle** for user-initiated flows: creating a PaymentIntent, the idempotency-key replay/resume state machine, the settled-mid-request guard, and polling a transaction to a terminal status. That cluster owns the `stripe_transactions` records and has one job — take a payment and record it — unrelated to coupons, customers, or payment-method management.

This PR moves that cluster into a dedicated `StripeTransactionService`. `StripeService` sheds ~370 lines plus its `cockatiel` / `http-errors` / idempotency-error-message imports; the charge lifecycle gets a single-responsibility, independently testable home. It's behavior-preserving — **no change to charging or idempotency** — and the next step in the layered `StripeService` split (following the merged reporting extraction).

While moving it, one bit of circular logic is retired (see below), but the observable behavior is identical.

Closes CON-722 · Part of CON-718.

## What

- Adds `StripeTransactionService` (injected `STRIPE_CLIENT` + `StripeTransactionRepository` + `TimerService` + logger, mirroring the reporting service). It owns `createPaymentIntent` (keyless create / `findOrCreateByIdempotencyKey` / charge / replay short-circuit / resume-from-recorded-intent / settled-winner) and `resolveTransaction`, plus their private helpers — all moved verbatim from `StripeService`.
- **Retires the idempotency-key prefix sniffing.** Previously the service inferred reject-vs-tolerate from `idempotencyKey.startsWith("topup_")` — circular, since the controller *sets* that prefix. The caller now passes an explicit `onAmountMismatch: "reject" | "tolerate"` policy: `StripeController.confirmPayment` passes `"reject"`, the wallet-reload job passes `"tolerate"`. Same reject/tolerate outcomes as before.
- **Idempotency key string is unchanged.** `TOP_UP_IDEMPOTENCY_KEY_PREFIX` stays and the controller still builds the identical `topup_${userId}_${attemptKey}` string, so no in-flight key changes across deploy — only its *semantic* role (being sniffed) is gone. `STRIPE_CURRENCY` is now exported for reuse.
- `StripeController` and `WalletBalanceReloadCheckHandler` now depend on `StripeTransactionService` for charging; every other endpoint still goes through `StripeService`.
- `StripeService` drops the moved members and the now-dead imports.
- Tests: the `createPaymentIntent` and `resolveTransaction` describes move into a new `stripe-transaction.service` spec; the controller and reload-handler specs are rerouted to the new service; the amount-mismatch tests now assert the explicit `onAmountMismatch` policy instead of the key prefix.

## Verification

- **Billing unit suite green — 357 tests / 28 files** (the moved charge/resolve tests now live in the new spec, not lost).
- `tsc`: no new errors in the changed files (pre-existing apps/api baseline unchanged).
- Prettier clean. Lint (ESLint flat config) and integration/functional tests run in CI.

## Summary by CodeRabbit

- **New Features**
  - Introduced a centralized transaction orchestration layer for Stripe PaymentIntents, including idempotency replay protection and controlled behavior for amount mismatches.
  - Updated payment confirmation, coupon application, and wallet reload flows to use the new transaction flow, including required wait/resolution behavior.

- **Bug Fixes**
  - Prevents duplicate charges on retried requests with reused idempotency keys.
  - Improves correctness for in-progress/required-action outcomes and aligns results with webhook-settled states.

- **Tests**
  - Added a dedicated test suite for the new transaction service and refreshed related controller/handler/webhook tests.


ygrishajev:
@coderabbitai review

ygrishajev:
@coderabbitai i need you to specifically verify that the core of PR only moved the code from stripe service to a new one with only change in logic around idempotency amount mismatch handling, nothing else

baktun14 (approved):

PR #3506, opened 2026-07-23
refactor(billing): make StripeTransactionService the single writer for the webhook path

ygrishajev:
## Why

`stripe_transactions` was written by **two** owners: `StripeTransactionService` on the request path (previous slice) and `StripeWebhookService` on the webhook path — the concurrency class this refactor set out to close. This PR routes the webhook's transaction writes (and the coupled wallet mutation) through `StripeTransactionService`, so the table now has a **single writer** across both paths.

Behavior-preserving: no response, status-mapping, idempotency, or concurrency semantics change. The webhook keeps its role as the Stripe-event dispatcher; only the `stripe_transactions` writes move down into the L2 aggregate. Next step in the layered `StripeService` split (follows the request-path slice).

Part of CON-718, Closes CON-723

## What

- `StripeTransactionService` gains the webhook-settlement methods, each owning the `@WithTransaction` "lock the row → update it → refill/reduce the wallet" atomicity that used to live in the webhook:
  - `settleSucceededTransaction` — the former `updateTransactionAndTopUp` core (locked read, already-succeeded no-op, first-purchase bonus, row update, wallet top-up); returns the bonus so the dispatcher can publish the post-commit event.
  - `markPaymentIntentFailed` / `markPaymentIntentCanceled`.
  - `applyRefund` — the refund core (locked read, refund idempotency + delta, status guard, bonus clawback, row update, wallet reduce).
- It picks up `RefillService` + `FirstPurchaseBonusService` as dependencies (the collaborators that moved with the atomic core).
- `StripeWebhookService` stays the L3 dispatcher: verify signature, route the event, extract Stripe-shape primitives, resolve the user, fetch charge details, and publish the deliberately post-commit `FirstPurchaseBonusGranted` event. It performs **no** direct `stripe_transactions` writes and no longer injects `RefillService` / `FirstPurchaseBonusService`.
- **Stripe API calls stay outside the transaction** — `constructWebhookEvent` and `retrieveCharge` run in the dispatcher, before the L2 transaction opens. Only DB + wallet work happens inside `@WithTransaction`.
- `payment_method.attached` / `detached` are untouched (they belong to the upcoming `PaymentMethodService` slice).

## Verification

- **Billing unit suite green — 357 tests / 28 files.**
- `tsc`: no new errors in the changed files; the reworked `stripe-webhook.service.integration.ts` construction (a real `StripeTransactionService` wired with the same mocks, passed into the webhook) type-checks.
- Prettier clean.
- **Run before merge (need the test DB):** `npm run test:integration` (the `handleChargeRefunded` suite — full/partial/clawback/duplicate-delivery idempotency — plus the settle/failed/canceled paths, all now exercised through the real L2) and `npm run test:functional`. These require `npm run test:ci-setup` and run in CI.

### DIFF
--- a/apps/api/src/billing/services/stripe-transaction/stripe-transaction.service.spec.ts
+++ b/apps/api/src/billing/services/stripe-transaction/stripe-transaction.service.spec.ts
@@ -5,6 +5,8 @@ import { describe, expect, it, vi } from "vitest";
 import { mock } from "vitest-mock-extended";
 
 import type { StripeTransactionRepository } from "@src/billing/repositories";
+import type { FirstPurchaseBonusService } from "@src/billing/services/first-purchase-bonus/first-purchase-bonus.service";
+import type { RefillService } from "@src/billing/services/refill/refill.service";
 import { IDEMPOTENCY_KEY_MISMATCH_ERROR_MESSAGE, PAYMENT_IN_PROGRESS_ERROR_MESSAGE } from "@src/billing/services/stripe-error/stripe-error.service";
 import type { TimerService } from "@src/core/services/timer/timer.service";
 import { StripeTransactionService } from "./stripe-transaction.service";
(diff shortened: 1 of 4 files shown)


ygrishajev:
@coderabbitai review the last commit in this PR. The 1st is going to be merged as a part of another that is already reviewed

baktun14 (approved):

PR #3508, opened 2026-07-23
fix(billing): source billing history from internal transactions

baktun14:
## Why

Part of CON-674 (backend half; the UI change is in the stacked follow-up PR).

The customer-facing **Billing → History** table only showed credit-card purchases. It was built from Stripe **charges**, so:

- **Coupon claims** and **manual credits** never appeared (they are Stripe invoices, not charges).
- **Refunds** never appeared (a refund reduces `amount_refunded` but leaves the charge `succeeded`).

Our internal `stripe_transactions` table already records every type and all refund state, so this sources the history from it instead, matching what the internal admin console shows.

This PR is the API/data half of what used to be #3500, rebased onto Yaro's `TransactionReportingService` extraction (#3499). It splits the old oversized PR in two; the deploy-web UI + http-sdk contract land in the stacked PR on top of this branch.

## What

`apps/api` only. No migrations.

- `TransactionReportingService.getCustomerTransactions` now reads `StripeTransactionRepository.findByUserId` + `countByUserId` (keyed on `userId`) instead of `stripe.charges.list`. Returns `{ transactions, totalCount, hasMore }` with offset pagination. The service no longer depends on the injected Stripe client.
- Flat transaction DTO: adds `type`, `amountRefunded`, `cardBrand`, `cardLast4`, `stripeInvoiceId`; drops the nested `paymentMethod`/`metadata`.
- `countByUserId` gains an optional `{ startDate, endDate }` filter.
- Controller retrieves by `currentUser.id` and drops the obsolete `stripeCustomerId` prerequisite, so coupon/manual-credit-only users are no longer excluded.
- Query param `startingAfter`/`endingBefore` -> `offset`, validated as a non-negative integer at parse time.
- CSV export gains `Type`, `Refunded`, and `Invoice ID` columns and yields a generic sanitized placeholder on a fetch error instead of the raw exception message.

### Needs verification before merge

**Historical completeness.** The old endpoint listed every Stripe charge; this DB-sourced list shows only rows in `stripe_transactions`. Any card payment predating that table (or never backfilled) would drop out of history. Please confirm against prod that historical `payment_intent` rows exist for older customers; if there is a gap, a one-time backfill of legacy charges is needed.

## Summary by CodeRabbit

* **New Features**
  * Switched billing transaction pagination to offset-based paging across API, client, and UI, including `totalCount` and date-range filtering.
  * Updated transaction details everywhere (new type badges, refunded amounts, card brand/last4, and invoice ID).
  * Refreshed transaction CSV export schema with new columns and streaming improvements.

* **Bug Fixes**
  * Standardized API response and query shapes so listing and exporting stay consistent.
  * Improved CSV output robustness (sanitized descriptions, safer handling of missing card details, and cleaned-up error messaging).

### DIFF
--- a/apps/api/src/billing/controllers/stripe/stripe.controller.ts
+++ b/apps/api/src/billing/controllers/stripe/stripe.controller.ts
@@ -210,18 +210,13 @@ export class StripeController {
   @Protected([{ action: "read", subject: "StripePayment" }])
   async getCustomerTransactions(options?: {
     limit?: number;
-    startingAfter?: string;
-    endingBefore?: string;
+    offset?: number;
     startDate?: string;
     endDate?: string;
-  }): Promise<{ data: { transactions: Transaction[]; hasMore: boolean; nextPage: string | null; prevPage: string | null } }> {
+  }): Promise<{ data: { transactions: Transaction[]; totalCount: number; hasMore: boolean } }> {
     const { currentUser } = this.authService;
 
-    if (!currentUser.stripeCustomerId) {
-      return { data: { transactions: [], hasMore: false, nextPage: null, prevPage: null } };
-    }
-
-    const response = await this.transactionReporting.getCustomerTransactions(currentUser.stripeCustomerId, options);
+    const response = await this.transactionReporting.getCustomerTransactions(currentUser.id, options);
     return { data: response };
   }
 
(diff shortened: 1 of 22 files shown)


ygrishajev (approved):

baktun14:
## Addressed CodeRabbit feedback

Pushed in `7a34b1862`.

**Fixed:**
- **Bound `limit` at runtime** (`stripe.schema.ts`) — `limit` is now validated as `int().min(1).max(100)`; it previously only carried `.openapi()` docs, so clients could pass negative/fractional/huge page sizes straight into the query.
- **DB-side transaction count** (`stripe-transaction.repository.ts`) — `countByUserId` now runs `count(*)` instead of loading every matching row, mirroring `user-wallet.repository.ts`.
- **Seeder `created` in Unix seconds** (`payment.ts`) — was emitting milliseconds, which the `* 1000` rendering in `BillingView` turned into a far-future date.
- **Fallback for unrecognized transaction types** (`BillingView.tsx`) — default badge class + capitalized label, matching the existing status-badge fallback, so untyped API drift no longer renders a blank badge.

**Kept as-is:** the CSV export still calls `getCustomerTransactions` per batch, so the count runs once per page — but it is now a cheap indexed `count(*)`, so the "load every matching row" cost the comment flagged is gone. Extracting a count-free page fetch would churn the existing generator tests for negligible gain, so the shared method stays.

Validation: `tsc` + lint (api & deploy-web), the transaction-reporting unit spec (13), the BillingView/BillingContainer specs (28), and the stripe-transaction integration test (18) all pass.

PR #3520, opened 2026-07-24
fix(billing): stamp wallet activation only after trial provisioning succeeds

baktun14:
## Why

Related to CON-744

After deploying #3513, every trial started in a ~2h prod window ended up with `activated_at = NULL` even though provisioning fully succeeded on-chain (fee + deposit grants landed, users were actively deploying). Verified against prod data and public LCD state.

Root cause: `activated_at` doubled as both the in-flight claim marker and the completion marker.

1. A trial start stamped `activated_at` up front and rolled it back on failure. The rollback was unconditional, so a slow failing attempt could clear a stamp it no longer owned.
2. The idempotency fast-path read the mid-flight stamp as "trial started" and returned 200 while the claiming attempt could still fail and unstamp.

Net effect: provisioned, actively-deploying wallets that are invisible to `GET /v1/wallets`, skipped by the refill cron (`findDrainingWallets`), and ignored by the trial fingerprint check. Affected prod rows were repaired manually with a one-off `UPDATE`.

## What

- New nullable `user_wallets.activation_claimed_at` column (migration 0034, additive) guards in-flight provisioning:
  - `claimActivation` sets it conditionally; a claim older than 5 minutes is considered abandoned (crashed/hung attempt) and can be taken over.
  - `releaseActivationClaim` clears only the claim identified by its own timestamp, so a stale attempt cannot clobber a newer claim.
- `activated_at` is now written **only on success**, atomically with the granted allowances — there is no rollback to get wrong, and the fast-path/`GET /v1/wallets`/refill cron/fingerprint check can only ever observe completed activations.
- `RefillService` uses the new `markActivated` (direct stamp, no claim cycle) since funding-based activation has no post-claim provisioning step to guard.

No API contract changes. The migration is a single nullable-column `ALTER TABLE` — no backfill, no locks beyond a brief ACCESS EXCLUSIVE.

## Summary by CodeRabbit

* **New Features**
  * Added a new activation claim timestamp to track in-progress activation separately from completed activation.
  * Introduced stale activation claim takeover so older, abandoned claims can be recovered.

* **Bug Fixes**
  * Prevented concurrent activation attempts from overwriting each other’s claim/activation state.
  * Updated activation/release behavior so only the correct activation claim is released, enabling safe retries after failures.
  * Ensured already-activated wallets remain unchanged during activation flows.

### DIFF
--- a/apps/api/drizzle/0034_natural_piledriver.sql
+++ b/apps/api/drizzle/0034_natural_piledriver.sql
@@ -1,0 +1,1 @@
+ALTER TABLE "user_wallets" ADD COLUMN "activation_claimed_at" timestamp with time zone;
--- a/apps/api/drizzle/meta/0034_snapshot.json
+++ b/apps/api/drizzle/meta/0034_snapshot.json
@@ -1,0 +1,1367 @@
+{
+  "id": "ac14bdd7-96b1-4a0e-9244-688b051999de",
+  "prevId": "0e3866c6-bb02-4bb0-8241-6e6ac040f4f7",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "public.user_wallets": {
+      "name": "user_wallets",
+      "schema": "",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "serial",
+          "primaryKey": true,
+          "notNull": true
+        },
+        "user_id": {
+          "name": "user_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "address": {
+          "name": "address",
+          "type": "varchar",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "deployment_allowance": {
+          "name": "deployment_allowance",
+          "type": "numeric(20, 2)",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "'0.00'"
+        },
+        "fee_allowance": {
+          "name": "fee_allowance",
+          "type": "numeric(20, 2)",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "'0.00'"
+        },
+        "trial": {
+          "name": "trial",
+          "type": "boolean",
+          "primaryKey": false,
+          "notNull": false,
+          "default": true
+        },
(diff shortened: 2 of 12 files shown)


baktun14:
## Addressed CodeRabbit review

Pushed `9c90aaf` addressing the review feedback.

**✅ Fixed — best-effort claim release (`wallet-initializer.service.ts`)**
A failing `releaseActivationClaim` would propagate and mask the original provisioning error. Wrapped it in a `.catch` that logs `TRIAL_CLAIM_RELEASE_ERROR` (via `createOtelLogger`, matching `RefillService`) and swallows the release error, so the caller/logs keep the real root cause. The claim is already covered by the 5-min stale-takeover, so release is safe to treat as best-effort. Added a regression test (`surfaces the original provisioning error even when releasing the claim fails`).

**✅ Fixed — tightened vacuous assertion (`wallet-initializer.service.spec.ts`)**
Replaced `not.toHaveBeenCalledWith(wallet.id, { activatedAt: null })` with `not.toHaveBeenCalled()`. In the failure path `updateById` isn't called at all, so this is a stronger, non-vacuous assertion.

**⏭️ Deferred — DB-side claim timestamp (`user-wallet.repository.ts`)**
Not applied, by design. `releaseActivationClaim` releases a claim by **exact** match `eq(activationClaimedAt, claimedAt)`. `now()` produces microsecond-precision `timestamptz`, but the value round-trips back through a millisecond-precision JS `Date`, so an exact-match release would silently stop matching its own claim. The `new Date()` stamp is load-bearing for that exact-match, not merely stylistic. As the review notes, with a 5-min window clock skew is "unlikely to bite in practice… a hardening nit rather than a defect," so the correctness tradeoff isn't worth it here.

_Note: the CodeRabbit ESLint "unused eslint-disable directive" tool-run warning on `user-wallet.schema.ts` is a CodeRabbit tooling misconfiguration — our own `npm run lint` confirms the `import-x/no-cycle` directive is required (real cycle via `Users`), so it's left in place._

`npm run test:unit` (wallet-initializer), `npm run lint --quiet`, and `tsc --noEmit` (no new errors) all pass locally.

PR #3542, opened 2026-07-30
fix(deployment): await trial-wallet activation before creating a deployment

ygrishajev:
## Why

A managed (trial) wallet gets its address at registration, but can only *spend* once its on-chain
grants are provisioned. That provisioning ("start trial") was **driven from the client**, kicked off
when the user landed on onboarding. Anything that interrupted the browser between landing and the
first spend — a reload, a navigation, a closed tab — abandoned the in-flight start-trial, so the
action raced ahead of an un-activated wallet and **failed terminally**. This wasn't only deploys:
top-ups and coupon redemptions hit the same un-provisioned wallet the same way. A manual retry
usually "fixed" it, because by then the grants had landed.

A spending action should never have to assume a client-side side-effect completed. Activation is now
the server's responsibility, and every managed-wallet endpoint waits for it.

## What

Trial-wallet activation moves off the client and behind a **server-side background job**, and every
managed-wallet spend/credit **awaits activation via a retriable `409 wallet_provisioning`** instead of
failing. No happy-path UX change; the action just occasionally runs a little longer while grants land.

```mermaid
flowchart TD
    subgraph server["Server-side — activation runs on its own, survives client reloads"]
        direction TB
        A[User registers or verifies email] --> B{Email verified?}
        B -- no --> Z[No activation yet]
        B -- yes --> C[Enqueue ActivateTrial job]
        C --> D[Grant on-chain trial limits, then set activatedAt]
        D -- transient failure: retry with backoff --> D
        D -- success --> F([Wallet activated])
    end

    subgraph client["Client — deploy / top-up / coupon are managed-wallet actions"]
        direction TB
        G[User acts] --> I{Wallet activated?}
        I -- yes --> J([Action proceeds])
        I -- no --> K[API returns 409 wallet_provisioning<br/>and re-enqueues activation]
        K --> L{Retries left?}
        L -- yes: wait backoff --> G
        L -- no --> M[Show 'try again later or contact support']
    end

    F -. activation unblocks the gate .-> I
```

**Server-side activation (apps/api)**

- Activation is enqueued as an idempotent `ActivateTrial` job the moment a user's email is verified —
  on registration, on email-verification sync, and on the passwordless code-verify path. The job grants
  the on-chain limits then stamps `activatedAt`; it retries with backoff, so a transient chain/queue
  hiccup self-heals without any client involvement.
- `activatedAt` is set **only after the grants land** (same write as the allowances), so it doubles as
  the spend gate's "ready" signal — an action arriving mid-provisioning can't slip past onto a bare
  chain/funding error. No activation claim is needed: the queue is `policy: singleton` (one job per user)
  and the grants are idempotent, so a retry just re-converges.

**Shared spend/credit guard**

- A reusable `TrialActivationJobService.assertActivated` best-effort re-enqueues activation and throws the
  retriable `409 wallet_provisioning` when the wallet isn't activated. It's applied to **create-deployment**
  (managed signer), **top-up confirm** and **coupon apply** (`StripeController`). A failed re-enqueue is
  logged but the 409 is still returned, so the next client retry drives it.

**Client retry + gate removal (apps/deploy-web)**

- One shared `walletProvisioningRetry` retries the `wallet_provisioning` 409 with backoff; wired into the
  create-deployment, top-up (`confirmPayment`), and coupon (`applyCoupon`) mutations. The predicate matches
  both client error shapes (openapi-sdk `ApiError` and the http-sdk `AxiosError`).
- The client-side wallet-ready gate is **removed** from the deploy flow, the top-up form, and the coupon
  form — they act immediately and let the server 409 + retry handle provisioning. `useEnsureTrialStarted`
  is reduced to a pure reader; `useCreateDeployment` is gone (inlined into the flow). The now-dead
  `isWalletReady` prop is removed all the way up the Add-Credits chain.
- On an exhausted retry budget the flow surfaces a clear "your account is still being set up — try again
  shortly or contact support" message.

**Stale-cache self-heal**

- The managed-wallet query no longer sits on a stale no-address snapshot: it **polls until the address
  lands** (server-side provisioning can lag the first fetch), so post-deploy pages, the onboarding gate,
  and the balance stop requiring a manual reload.
- A successful deploy also invalidates the onboarding lease-existence cache, so a client-side nav home
  isn't bounced to `/onboarding`. The top-up/coupon flows keep the existing balance-poll so funds are
  confirmed before the user proceeds to deploy.

**Backward compatibility**

- `POST /v1/start-trial` is kept but reduced to *ensure the wallet + enqueue activation* (returning the
  wallet), so an older deploy-web still works if the API ships first in a staged rollout. New clients
  don't call it.

**Observability (OpenTelemetry, not analytics)**

- The activation job records its outcome — a completions counter tagged by status and, on failure, a
  classified reason (transient vs. terminal blocks like unverified email or a duplicate-fingerprint
  block), job duration, and registration→activation latency. Failures also log `TRIAL_ACTIVATION_JOB_FAILED`.
  Activation is expected to always succeed, so any failure is a real signal worth a dashboard alert.

**Notes**

- The email-verified and duplicate-fingerprint checks are kept: password auth (`?auth=password`) still
  exists, so the server must not assume every caller arrives already verified.
- Follow-up (out of scope): retire the orphaned `connectManagedWallet` / `ConnectManagedWalletButton`
  cluster that pointed at the old client-create endpoint.

## Summary by CodeRabbit

* **New Features**
  * Trial activation is now scheduled automatically after email verification, registration, and wallet creation.
  * Payments, coupons, and deployments retry automatically while wallet provisioning completes.
  * Wallet status updates automatically during provisioning.
  * Deployment and funding flows no longer require manual trial activation or retry actions.

* **Bug Fixes**
  * Improved payment polling and handling of wallet-provisioning delays.
  * Added clearer error reporting and monitoring for trial activation failures.


apps/api/src/billing/services/wallet-initializer/wallet-initializer.service.ts
```diff
@@ -43,31 +34,39 @@ export class WalletInitializerService {
...
+   * allowances. Concurrency is handled upstream — the ActivateTrial queue is `policy: singleton`, so at most one
+   * job per user is ever active — and the grants are idempotent (revoke-then-regrant / authz overwrite), so a
+   * crashed job just re-runs on retry and re-converges to the same limits. No claim or rollback needed.
+   */
   async initializeAndGrantTrialLimits(userId: string): Promise<UserWalletPublicOutput> {
-    const userWallet = await this.#ensureWalletVia(this.userWalletRepository.accessibleBy(this.authService.ability, "create"), userId);
+    const user = await this.userRepository.findById(userId);
+    assert(user, 404, "User Not Found");
+    assert(user.emailVerified, 400, "Email not verified");
+    await this.#assertNoDuplicateFingerprint(user);
```
baktun14, line 52: I don't think we need this anymore

baktun14 (approved):

PR #3552, opened 2026-08-04
fix(deployment): respect auto top-up setting on lease-start funding

ygrishajev:
## Why

On lease creation we auto-deposit ~48h of escrow to keep a managed deployment alive — but we did this
**regardless of the deployment's `autoTopUpEnabled` setting**. A user who disabled auto top-up to manage
deposits themselves still got our deposit racing with theirs, causing a `402 Insufficient balance`.

## What

`fundOnLeaseStarted` now skips the deposit when the deployment setting is explicitly disabled. A missing
setting still funds, so auto top-up stays enabled by default.

Covered by unit tests (disabled → skip, enabled → fund).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed deployment funding to respect auto-top-up settings when disabled

* **Tests**
  * Added test coverage for auto-top-up settings behavior


stalniy (approved):

baktun14:
The plan is to remove deployment top up an escrow account management entirely to users

baktun14 (approved):

PR #3603, opened 2026-08-17
feat(indexer): akash bme event handlers with exact ledger records

baktun14:
## Why

Closes CON-813

The old indexer treats BME as a special case spliced into the sync insert phase: `chainSync` scans EndBlocker events inline into a `bme_raw_event` staging table, and a separate indexer later derives ledger records from those rows plus six cumulative `DOUBLE` columns on the block table fed by `parseFloat`. That means float math on money and a staging table the new pipeline has no reason to keep.

## What

BME becomes a normal domain layer on the new pipeline, with the same deriver + writer shape as the other handlers.

- Three new `akash` tables, all exact numerics (`numeric(38,0)` amounts, `numeric(38,18)` prices, where the old `DECIMAL(20,10)` truncated the chain's 18-decimal Dec):
  - `bme_ledger_records`, keyed by the full on-chain `LedgerRecordID`. `record_height` is the record's creation height and `height` the block whose EndBlocker executed it, so pending records that execute later keep both.
  - `bme_status_changes`, keyed by `(height, ordinal)`.
  - `bme_canceled_records`, parsed into typed columns. The old indexer kept cancellations only as raw staging rows. `cancel_reason` is text on purpose: sandbox already emits `minimum_mint`, which the SDK enum does not know.
- `deriveBmeChanges` scans successful-tx events then block events in chain order. A malformed event is logged (`BME_EVENT_PARSE_FAILED`) and skipped rather than halting the block, and it still consumes its ordinal so a later parser fix replays with stable keys. Unset proto fields reach the wire as the literal JSON `null` (observed on sandbox) and are treated as absent.
- `BmeWriter` is plain conflict-ignoring appends inside the batch transaction; replaying a block is a no-op.
- `EventVaultFunded` is deliberately not captured: the vault's balance history is already exact in `cosmos.balance_changes` (the reason classifier maps the vault address to `bme`), and the reconcile CLI checks it against the chain.
- Not ported from the old indexer: the raw staging table, the synthesized `VaultFundedTransfer`/`MigrationMinted` events (the balance ledger records those movements from the generic coin events), oracle price parsing and the one-time denom row mutations (the resulting gap for pre-upgrade deployments is tracked in CON-842), and the cumulative `DOUBLE` block columns (computable by query from the ledger; the API layer is CON-817).

The migration only creates new tables and one enum, so it cannot block a production database.

### Verification on sandbox

- **AC1 (ledger parity):** backfilled every sandbox height carrying BME events (146 ranges) and compared each stored row field by field against the raw `block_results` events: 149 executed records, 106 status changes and 315 canceled records, 570 field sets, 0 mismatches.
- **AC2 (vault in balance ledger):** vault movements appear in `cosmos.balance_changes` with `bme`/`mint` reasons and cross-check the ledger records exactly (collateral inflow equals `remint_credit_accrued`, payout equals minted minus spread, cancel refunds net to zero).
- **AC3 (no staging):** no raw table exists in the schema, and the deriver spec pins that unhandled events store nothing.

## Summary by CodeRabbit

- **New Features**
  - Added support for tracking Akash BME (burn-mint-equilibrium) ledger activity.
  - Records mint status changes, executed ledger transactions, and canceled operations.
  - Captures related token amounts, prices, accounts, block details, and cancellation information.
  - Integrates BME event processing into block indexing for reliable historical querying.

- **Bug Fixes**
  - Improved handling of malformed or incomplete BME events so valid events continue processing.
  - Standardized account resolution across Akash indexing workflows.

### DIFF
--- a/apps/chain-indexer/drizzle/0008_good_galactus.sql
+++ b/apps/chain-indexer/drizzle/0008_good_galactus.sql
@@ -1,0 +1,58 @@
+CREATE TYPE "akash"."bme_mint_status" AS ENUM('mint_status_unspecified', 'mint_status_healthy', 'mint_status_warning', 'mint_status_halt_cr', 'mint_status_halt_oracle');--> statement-breakpoint
+CREATE TABLE "akash"."bme_canceled_records" (
+	"denom" text NOT NULL,
+	"to_denom" text NOT NULL,
+	"source" text NOT NULL,
+	"record_height" bigint NOT NULL,
+	"sequence" bigint NOT NULL,
+	"height" bigint NOT NULL,
+	"tx_index" integer,
+	"cancel_reason" text NOT NULL,
+	"owner_account_id" integer NOT NULL,
+	"to_account_id" integer NOT NULL,
+	"coins_to_burn_denom" text,
+	"coins_to_burn_amount" numeric(38, 0),
+	"denom_to_mint" text NOT NULL,
+	CONSTRAINT "bme_canceled_records_record_height_sequence_denom_to_denom_source_pk" PRIMARY KEY("record_height","sequence","denom","to_denom","source")
+);
+--> statement-breakpoint
+CREATE TABLE "akash"."bme_ledger_records" (
+	"denom" text NOT NULL,
+	"to_denom" text NOT NULL,
+	"source" text NOT NULL,
+	"record_height" bigint NOT NULL,
+	"sequence" bigint NOT NULL,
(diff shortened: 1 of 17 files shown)


baktun14:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

baktun14:
@coderabbitai review

PR #3629, opened 2026-08-20
feat(deployment): refuse sealed secrets before any KMS call when intake is off or oversized

stalniy:
## Why

Without ordering the checks correctly, a malformed or hostile request costs a billable key operation. And a new intake path needs a kill switch before it has any callers, not after.

Closes CON-850

## What

Two refusals decided at the entrance of `SdlSecretsUnsealerService.open()`, before anything is spent:

- **Flag gate** — `sdl_secrets_sealed_intake` off → `403 SDL_SECRETS_SEALED_INTAKE_DISABLED`.
- **Size bound** — `SDL_SECRETS_MAX_SEAL_BYTES` (16Kb) exceeded → `413 SDL_SECRETS_SEAL_TOO_LARGE`. Measured with `Buffer.byteLength`, not `.length`, so multi-byte padding cannot slip a payload past the cap.

Both precede `#splitSeal`, so neither can reach KMS. Tests assert `asymmetricDecrypt` was never called, plus a realistic 20-secret payload still opens so the cap does not bite legitimate traffic.

### Decisions worth flagging

- **The gate is in the service, not on a route.** There is no sealed-intake route to gate yet, and `open()` is the only place every present and future caller passes through. Consequence: with the flag off nothing user-visible changes in this phase — the gate is dormant until PRD-0003 wires a caller.
- **Fail-closed by default.** `sdl_secrets_sealed_intake` does not exist in Unleash yet; an unknown flag resolves `false`, so merging before the ops step is safe. Creating the flag is a prerequisite for PRD-0003, not for this PR.
- **16Kb derivation** is in the constant's JSDoc: a wrapped CEK is at most 512 bytes and header/IV/tag are a few hundred more, so the cap is effectively a payload bound with several times the headroom a real secret set needs.
- **`GET /v1/sdl-secrets-context` is deliberately *not* gated.** A public key is not sensitive and clients need it discoverable; the refusal belongs at the point of acceptance.

### Known limitation on test level

The PRD asks for the flag-off case as a functional test. That is not achievable: `FEATURE_FLAGS_ENABLE_ALL=true` in `env/.env.functional.test`, and `isEnabled()` returns `true` on that config before consulting Unleash. Turning it off would break every other flagged path and would then demand Unleash credentials. So this is a unit test with `mock<FeatureFlagsService>()`; the mock is flag-specific, so checking the wrong flag name fails the happy path rather than passing silently.

Stacked on #3628.

### DIFF
--- a/apps/api/src/core/services/feature-flags/feature-flags.ts
+++ b/apps/api/src/core/services/feature-flags/feature-flags.ts
@@ -3,7 +3,8 @@ export const FeatureFlags = {
   NOTIFICATIONS_ALERT_UPDATE: "notifications_general_alerts_update",
   AUTO_CREDIT_RELOAD: "auto_credit_reload",
   AUTO_RELOAD_FIXED_THRESHOLD: "auto_reload_fixed_threshold",
-  TRIAL_FINGERPRINT_CHECK: "trial_fingerprint_check"
+  TRIAL_FINGERPRINT_CHECK: "trial_fingerprint_check",
+  SDL_SECRETS_SEALED_INTAKE: "sdl_secrets_sealed_intake"
 } as const;
 
 export type FeatureFlagValue = (typeof FeatureFlags)[keyof typeof FeatureFlags];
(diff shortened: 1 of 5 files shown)


stalniy:
close, since it has no real value. KMS price is per key size and amount of calls (ie, tokens spent) not based on payload size

PR #3654, opened 2026-08-24
feat(deployment): switch existing deployments to always-on automatic funding

baktun14:
## Why

Part of CON-734

With always-on deployment funding, new deployments default to funded (#3636 moves that default into the repository). Deployments whose owner turned the old per-deployment toggle off are still stuck with funding off, and once the toggle is gone from the UI (#3651) there is no way for anyone to change that. This is the one-time switch back on.

## What

A data-only migration:

```sql
UPDATE "deployment_settings" SET "auto_top_up_enabled" = true, "updated_at" = now()
WHERE "auto_top_up_enabled" = false AND "closed" = false;
```

- Closed deployments keep their value: every funding query filters on `closed = false`, so flipping them would rewrite rows nothing reads.
- Runtime-limited deployments behave the same before and after. The closer honours `runtime_ends_at` regardless of the funding flag (#3630), so a limited deployment is funded up to its deadline and then stops.
- Row locks only, no table rewrite. `lock_timeout = '3s'` follows 0041, so pod boot fails fast and retries instead of queueing behind a lock held by the top-up cron.

The integration test replays the actual migration file against seeded rows: an open opt-out row flips on, a closed row keeps its opt-out, and an already-on row is not rewritten (its `updated_at` stays put).

## Merge gate

#3636 and #3651 have merged and this PR now targets `main`. It must not merge until `auto_reload_fixed_threshold` is at 100%: while the toggle is still visible, a user who watches "off" flip to "on" can simply turn it off again, and the API still lets them. The rollout order is documented on CON-734.

### DIFF
--- a/apps/api/drizzle/0042_backfill_auto_top_up_always_on.sql
+++ b/apps/api/drizzle/0042_backfill_auto_top_up_always_on.sql
@@ -1,0 +1,7 @@
+-- One-time backfill for always-on deployment funding (CON-734): rows where the owner turned the
+-- old per-deployment toggle off are switched back on. Runtime-limited deployments are unaffected
+-- in behaviour: funding stops at their deadline regardless of this flag. Closed deployments are
+-- left as they are, since every funding query filters on closed = false.
+SET lock_timeout = '3s';--> statement-breakpoint
+UPDATE "deployment_settings" SET "auto_top_up_enabled" = true, "updated_at" = now() WHERE "auto_top_up_enabled" = false AND "closed" = false;--> statement-breakpoint
+RESET lock_timeout;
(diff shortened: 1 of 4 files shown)


baktun14:
Closing this. The backfill moved out of the repo into a local one-off script.

Two reasons. A data-only migration is re-run by every environment forever, matching nothing after the first time. And `0042` is now taken on main by `0042_warn_before_runtime_limit` (#3652), so this needed renumbering before it could merge anyway.

The flip now lives in `apps/api/scripts/backfill-auto-top-up-always-on.mts`, read-only unless you pass `--apply`. The constraint this PR was waiting on still holds, it just applies to the run now instead of the merge: don't run it until `auto_reload_fixed_threshold` is at 100%.

Behaviour is the same as what was reviewed here. Same WHERE clause, so closed rows keep their opt-out, and it still sets `lock_timeout = '3s'` to fail fast rather than queue behind the top-up sweep.

Verified locally against `console-users`: 49 rows flipped, count back to 0, then restored.

Rollout notes are on CON-734.

PR #3657, opened 2026-08-25
fix(deployment): backfill missing settings rows so existing deployments keep being funded

baktun14:
## Why

Part of CON-895

The eager-create fix in #3656 only covers deployments from now on (plus any older deployment that starts a new lease). A deployment created outside the web app before that fix, with a lease already running, still has no `deployment_settings` row and stays invisible to the funding sweep until it drains and closes. CON-895 also asks for the affected population to be counted before release, since backfilling widens the sweep and increases both credit spend and the draw on the shared deposit authorization from CON-512.

## What

Adds a one-off `backfill-deployment-settings` CLI command (same shape as `top-up-deployments` and `close-expired-deployments`). It pages through managed wallets, reads each wallet's open-lease dseqs from the chain database, diffs them against the wallet's settings rows, and creates the missing rows on the repository defaults (auto top-up enabled). This has to be code rather than a SQL migration because leases and `deployment_settings` live in different databases.

With `--dry-run` it only counts and logs, which is the measurement step:

1. Run the dry run against prod and post the counts on CON-895.
2. Sanity-check the extra funding draw against the deposit authorization headroom.
3. Run it for real; the next hourly sweep picks the deployments up.

A concurrent row creation (detail page lazy create or lease-start funding) loses gracefully: the unique violation is treated as the backfill's goal already reached.

## Stacked on

#3656 (base branch `fix/deployment-fund-deployments-created-outside-web-app`)

### DIFF
--- a/apps/api/src/app/console.ts
+++ b/apps/api/src/app/console.ts
@@ -61,6 +61,16 @@ program
   });
 
 program
+  .command("backfill-deployment-settings")
+  .description("Create missing deployment settings rows for open deployments so the funding sweep can see them")
+  .option("-d, --dry-run", "Only count and log the missing rows without creating them", false)
+  .action(async (options, command) => {
+    await executeCliHandler(command.name(), async () => {
+      return container.resolve(TopUpDeploymentsController).backfillDeploymentSettings(options);
+    });
+  });
+
+program
   .command("close-expired-deployments")
   .description("Close deployments that reached their runtime limit")
   .option("-d, --dry-run", "Log which deployments would be closed without broadcasting", false)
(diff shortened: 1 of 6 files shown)


baktun14:
Closing this. The backfill moved out of the repo into a local one-off script.

The command runs once and then never again, so putting it in `console.ts` leaves a permanent entry nobody will use, plus a `LeaseRepository.findOpenDseqsByOwner` with no other caller.

It now lives in `apps/api/scripts/backfill-deployment-settings.mts`, read-only unless you pass `--apply`. Same cross-DB diff, but in four queries total rather than two per wallet, since a local script has no reason to page. `ON CONFLICT (dseq, user_id) DO NOTHING` covers the race the `isUniqueViolation` catch was there for.

Verified against a scratch chain DB: 3 missing rows found and created with `auto_top_up_enabled = true`, already-known dseqs skipped, closed leases and unmanaged owners ignored, re-run is a no-op.

#3656 is unaffected and still targets main.

The dry-run count and the deposit-authorization check against CON-512 are tracked on CON-895.

PR #3673, opened 2026-08-26
feat(deployment): read deployment settings no deployment may back, by page

stalniy:
## Why

A deployment's settings row records what it is — its stripped SDL and the manifest version it commits — written just before the create tx is broadcast. A broadcast that never lands therefore leaves a record on a dseq the chain has never heard of, and **nothing in the codebase can currently find those records**:

- the stale-deployment cleanup (`StaleManagedDeploymentsCleanerService`) starts from on-chain deployments and closes the lease-less ones; a dseq that was never broadcast is not among them
- the funding sweep looks each remembered dseq up by lease and skips whatever it finds nothing for — on every pass, forever (`draining-deployment.service.ts:121-123`)

Left alone they only accumulate. This PR adds the read those rows need. **The cleanup that consumes it is the stacked follow-up below.**

## What

`DeploymentSettingRepository.findUnbackedDeploymentSettings` — one page of deployment settings rows old enough that the create tx each was written for has either reached the chain or never will.

> **Note for reviewers: this method has no caller until slice 2.** It is not dead code, it is the first half of a stack. Its behaviour is fully exercised by the integration tests in this PR, which call it directly against real Postgres.

**Pages rather than caps.** Whether a row is an orphan or a deployment that simply has not been indexed yet is a question only the chain can answer, so this narrows the field and the caller confirms each row before deleting anything. A record the caller never examines would always be among the oldest of its cohort, and because pages come newest first, a cap on how many rows a pass could take would leave the remainder under a waterline that only ever rises — unreachable for good, and worst during the broadcast outage that produces orphans in bulk.

**A keyset, not an offset**, because the caller deletes as it walks and an offset would shift the rows underneath it.

**The cursor's timestamp travels as text**, for the same reason `runtimeEndsAtMarker` already does in this file. `created_at` is `now()` at microsecond precision and a JS `Date` holds only milliseconds, so a cursor round-tripped through one is the stored value truncated *down* — and the next page, asking for rows strictly older than it, would silently drop every sibling written in the same millisecond, on this page and on every later one. The comparison is written as a single row value `(created_at, id) < (marker::timestamp, id::uuid)` so it cannot drift out of step with the `ORDER BY` beside it.

**The predicate.** `closed = false` is only a cheap head start, not the protection for closed deployments — that belongs to the caller, which asks the chain for every state. `sdl is not null` skips rows a settings read created lazily, which remember no definition at all and carry only the choices their owner made.

### Tests

Three integration tests against real Postgres cover page size, newest-first ordering, and cursor exhaustion. A fourth seeds rows that **share a timestamp** and rows that **differ only in microseconds**, walks to exhaustion, and asserts every row appears exactly once.

Each of these mutations fails that suite, confirmed separately:

| Mutation | Result |
|---|---|
| `DESC` → `ASC` | 3 tests fail |
| delete `.limit(pageSize)` | 1 fails: `expected […(5)] to deeply equal ['90004','90003','90002']` |
| drop the `id` tiebreaker from the cursor | 5 of 8 rows returned |
| drop `desc(id)` from `ORDER BY` | 6 of 8 rows returned |
| truncate the marker to milliseconds | 7 of 8 rows returned |

The tie group in that fixture is deliberately **wider than a page**. Postgres promises no order among rows with equal sort keys, so a fixture whose ties fit inside one page could agree with an id-ordered walk by luck and pass against a broken `ORDER BY`. Four tied rows against a page size of two put a boundary inside the group. That constraint is recorded in a JSDoc on the fixture — resize one and you have to resize the other.

## Stack

This is **slice 1 of 2**.

- **Slice 1 (this PR)** — the paged read. `size: M`.
- **Slice 2** — `OrphanedDeploymentSettingsCleanerService`, its config, its CLI subcommand and its cron scheduling, plus the gate decision on closed-deployment rows. `size: L`. Linked below once opened.

**Slice 2 should merge after this one.**

## Checks

In `apps/api`, on this branch alone: `npm test` → `215 passed (215)` files / `2303 passed (2303)` tests, `npm run lint -- --quiet` → clean, `npx tsc --noEmit` → 41 errors, byte-identical to the pre-existing baseline on `main` (all in unrelated files).

## Naming

The method reads deployment settings rows that no on-chain deployment may back, so it is `findUnbackedDeploymentSettings` returning `UnbackedDeploymentSetting`. "Unbacked" is kept deliberately — unlike "definition" it carries the actual meaning, and it is the term the spec uses. "Candidate" is dropped from the type because the method name already implies it of everything returned, but survives in local variable names where it reads better.

### DIFF
--- a/apps/api/src/deployment/repositories/deployment-setting/deployment-setting.repository.integration.ts
+++ b/apps/api/src/deployment/repositories/deployment-setting/deployment-setting.repository.integration.ts
@@ -93,12 +93,12 @@ describe(DeploymentSettingRepository.name, () => {
     });
   });
 
-  describe("findUnbackedDefinitionCandidates", () => {
+  describe("findUnbackedDeploymentSettings", () => {
     it("returns one page of the newest candidates, not the oldest and not all of them", async () => {
       const { deploymentSettingRepository, rememberDefinitions } = await setup();
       const dseqs = await rememberDefinitions(["oldest", "older", "middle", "newer", "newest"]);
 
-      const page = await deploymentSettingRepository.findUnbackedDefinitionCandidates({ graceHours: 1, pageSize: 3 });
+      const page = await deploymentSettingRepository.findUnbackedDeploymentSettings({ graceHours: 1, pageSize: 3 });
 
       expect(page.map(candidate => candidate.dseq)).toEqual([dseqs.newest, dseqs.newer, dseqs.middle]);
     });
@@ -107,8 +107,8 @@ describe(DeploymentSettingRepository.name, () => {
       const { deploymentSettingRepository, rememberDefinitions } = await setup();
       const dseqs = await rememberDefinitions(["oldest", "older", "middle", "newer", "newest"]);
 
-      const first = await deploymentSettingRepository.findUnbackedDefinitionCandidates({ graceHours: 1, pageSize: 3 });
-      const second = await deploymentSettingRepository.findUnbackedDefinitionCandidates({ graceHours: 1, pageSize: 3, olderThan: first[2] });
+      const first = await deploymentSettingRepository.findUnbackedDeploymentSettings({ graceHours: 1, pageSize: 3 });
+      const second = await deploymentSettingRepository.findUnbackedDeploymentSettings({ graceHours: 1, pageSize: 3, olderThan: first[2] });
 
       expect(second.map(candidate => candidate.dseq)).toEqual([dseqs.older, dseqs.oldest]);
     });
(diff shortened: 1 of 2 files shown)


stalniy:
Stacked follow-up that consumes this read: https://github.com/akash-network/console/pull/3674 — merge this one first.

stalniy:
Superseded along with #3674 by an outbox-pattern implementation; the keyset candidate query it added is not needed once compensation is per-create rather than a scan. Replacement PR to follow.

PR #3674, opened 2026-08-26
feat(deployment): clean up deployment settings no deployment backs

stalniy:
## Why

The console records what a deployment is just before it broadcasts the create tx, so a broadcast that never lands leaves a `deployment_settings` row on a dseq the chain has never heard of. **Nothing could reach those rows.** The stale-deployment cleanup starts from on-chain deployments and closes the lease-less ones, and a dseq that was never broadcast is not among them. The funding sweep looks each remembered dseq up by lease and skips whatever it finds nothing for, on every pass, forever (`draining-deployment.service.ts:121-123`). Left alone they only accumulate.

This is the cleanup that removes them, and nothing else.

## What

`OrphanedDeploymentSettingsCleanerService`, reachable as `console cleanup-orphaned-deployment-settings`, scheduled daily at 03:17 by a `jobs:` entry in the Helm values.

Named for what it does to what it operates on, alongside `StaleManagedDeploymentsCleanerService` and `ExpiredDeploymentsCloserService`. The previous `OrphanedDefinitionsSweeper` said neither: "definition" named the SDL a row happens to carry rather than the row, and "sweep" is not a verb this codebase uses for deleting.

### The CronJob object name is shorter than the command, on purpose

The chart renders the object name as `console-api-<chain>-<name>`, and the CronJob controller caps that at **52 characters** because it appends `-<timestamp>` to derive Job names. The prefix is 20 characters, so the 36-character command name would render **56** — 4 over. The object name is therefore `cleanup-orphaned-settings`, rendering `console-api-mainnet-cleanup-orphaned-settings` and `console-api-sandbox-cleanup-orphaned-settings` at **45** characters, verified with `helm template` against the real chart. The `command:` arg keeps the full subcommand and is the field that must match `console.ts`; the values files say so, because the mismatch looks like a bug and "fixing" it would break the deploy.

### Closed-deployment records are KEPT — and that is a decision, not a miss

The originating spec contradicted itself. Its "What to build" said closed-deployment records should stay "because it can be used for re-deployment"; its acceptance criterion #2 said "a record for a closed deployment is removed automatically". This was raised at the spec gate and **a human decided the prose is normative: closed records are kept, and AC#2 is rewritten to "a record for a closed deployment is KEPT."** Please don't read it as an unimplemented requirement.

The reasoning: a closed deployment is exactly the one its owner redeploys from, and the stripped SDL is the only copy the console holds.

### The listing is deliberately unfiltered by state, and that is load-bearing

`closed` on the settings row is only ever written by `expired-deployments-closer.service.ts:92` and `draining-deployment.service.ts:140` — and neither covers a deployment its owner closed through `DELETE /v1/deployments/{dseq}`. So the **common** closed deployment is still `closed = false`, *is* a cleanup candidate, and survives only because the chain listing reports closed deployments too.

Narrowing the request to `{ owner, state: "active" }` reads like an obvious optimisation and would delete the only stored copy of the SDL for every closed deployment of every user. The test fixtures therefore answer like the chain does — unfiltered lists every state, a state filter narrows to it — and **three tests fail on that one token.**

### Fails closed throughout

An owner whose chain lookup does not answer is skipped whole rather than guessed at, and the run reports itself failed so the next one retries. A failed delete logs `DELETE_FAILED` rather than `CHAIN_LOOKUP_FAILED`, because the dry-run bake period is read entirely through these logs. Deleting is irreversible and a definition cannot be rebuilt from anything else, so every uncertainty resolves to keeping the row.

The candidate read and the delete are **never** routed through `accessibleBy`, and must not be: this runs from the CLI, whose principal carries an empty ability, and `DrizzleAbility` calls `throwUnlessCan` in a field initializer, so a scoped repository would throw `ForbiddenError` before any SQL ran and the cleanup would never delete anything. The integration suite pins that by running the real CLI entrypoint inside the execution context `console.ts` builds.

### Rollout

`--dry-run` on prod-mainnet for a bake period (logs `ORPHANED_DEPLOYMENT_SETTINGS_WOULD_DELETE` and deletes nothing), live on staging-sandbox. Same precedent as `close-expired-deployments`.

## Demo

Five settings rows for one owner, **page size 2** so it has to page, and the orphan is deliberately the oldest row — on the last page. `111` is the orphan (never on chain), `333` is the common closed shape (`closed = false` locally, closed on chain), `444` is already marked closed locally so the SQL guard excludes it before the chain is asked, which is why `examined` reads 4 and not 5.

```
SEEDED - five deployment settings rows for one owner, newest first
   dseq | on chain | local closed flag | settings row
  ------+----------+-------------------+------------------------
    555 | active   | false             | kept (sdl still stored)
    444 | closed   | true              | kept (sdl still stored)
    333 | closed   | false             | kept (sdl still stored)
    222 | active   | false             | kept (sdl still stored)
    111 | NEVER    | false             | kept (sdl still stored)

  $ console cleanup-orphaned-deployment-settings --dry-run
      log ORPHANED_DEPLOYMENT_SETTINGS_CLEANUP_START
      log ORPHANED_DEPLOYMENT_SETTINGS_WOULD_DELETE dseqs=["111"]
      log ORPHANED_DEPLOYMENT_SETTINGS_CLEANUP_END pages=2 examined=4 complete=true

AFTER --dry-run (ok=true) - nothing deleted     [table unchanged]
  $ console cleanup-orphaned-deployment-settings
      log ORPHANED_DEPLOYMENT_SETTINGS_CLEANUP_START
      log ORPHANED_DEPLOYMENT_SETTINGS_DELETED dseqs=["111"]
      log ORPHANED_DEPLOYMENT_SETTINGS_CLEANUP_END pages=2 examined=4 complete=true

AFTER live run (ok=true) - only the row no deployment backs is gone
   dseq | on chain | local closed flag | settings row
  ------+----------+-------------------+------------------------
    555 | active   | false             | kept (sdl still stored)
    444 | closed   | true              | kept (sdl still stored)
    333 | closed   | false             | kept (sdl still stored)
    222 | active   | false             | kept (sdl still stored)
    111 | NEVER    | -                 | DELETED
```

`pages=2` is the point. Under a record cap, a run reading 2 records would have taken the two newest and stopped — and `111`, the oldest, would have fallen below a waterline that only ever rises and been unreachable **forever**. That failure would have landed hardest during exactly the incident this feature exists for: a broadcast outage produces orphans in bulk, and the same event that mass-produces them is the one that blows a record cap.

## Known limitation

**The time budget is checked between pages, not between owners.** Worst-case wall clock is therefore the budget plus one page, which against a throttling chain node can carry a run past the job's `activeDeadlineSeconds`. Being killed there costs the closing log and nothing else: each owner's delete is its own committed statement, so a run that ends mid-page leaves the same state as one that ends between pages, and the next run resumes from the newest record either way. Documented in the service JSDoc rather than papered over.

## Known risk: a wrongly-deleted row loses the runtime limit too

Worth stating plainly, because review kept framing this as "losing the redeploy artifact" and that understates it. `#delete` removes the whole `deployment_settings` row, which is also where the runtime limit lives — so `runtime_limit_hours`, `runtime_ends_at`, `last_funded_at` and `runtime_ending_notified_for` go with the SDL.

The runtime limit is the part that is a behaviour change rather than a lost convenience. Verified against the code: `startRuntimeCountdown` only anchors `where isNotNull(runtime_limit_hours)` and has just two callers, both routed through it, so nothing re-anchors a deadline once the limit is gone; and `findExpiredRuntimeDeployments` requires `isNotNull(runtime_ends_at)`, so the closer never closes the deployment at its limit. The deployment would run on unbounded funding with its cap silently unenforced.

One clarification on the sequence: funding does **not** resume immediately. Every funding query reads `FROM deployment_settings`, so with no row the deployment is invisible to the funding sweep, the closer and the notifier alike. It resumes only once a settings read recreates the row via `findOrCreateByUserIdAndDseq` → `createWithDefaults`, which applies `AUTO_TOP_UP_ENABLED_BY_DEFAULT = true` with the limit columns null. So: unfunded and uncapped → (settings read) → funded and uncapped.

**Probability.** It requires a create whose broadcast is still pending past the grace period — the signer round trip plus a block, against a default grace of one hour, so two orders of magnitude of headroom. An owner whose chain lookup fails is skipped rather than guessed at. And nothing is at risk during the bake: prod-mainnet ships `--dry-run` and deletes nothing. `ORPHANED_DEPLOYMENT_SETTINGS_CLEANUP_GRACE_IN_H` buys further margin for free — raising it only ever delays a deletion.

If reviewers judge this too sharp an edge, the cheap mitigation is to null the two definition columns instead of deleting the row, leaving the owner's runtime choices intact. That is a different feature from what was specified, so it belongs in a decision rather than in this PR.

## Deferred, deliberately

**1. No per-request HTTP timeout on the chain listing.** Adding one means changing the shared `CHAIN_API_HTTP_CLIENT`, which backs nine services — out of scope for this change. Wall clock is bounded by `activeDeadlineSeconds` and `concurrencyPolicy: Forbid` instead, and the residual risk is a **skipped owner, never a lost record**, because the cleanup fails closed.

**2. Marking `closed = true` from chain state.** The cleanup already learns each deployment's state and throws it away, so the population of "closed on chain but never marked locally" is permanently undeletable by the cleanup and permanently re-examined by it. Marking it would converge the candidate set on live-plus-recent. Deferred because it is a **WRITE and would have to be dry-run gated** — a write that ignored the flag would break the bake-period guarantee on the very first prod run. And the honest fix may not live in the cleanup at all: `DeploymentWriterService.close()` not marking the row is the actual gap, and the cleanup marking it is repairing a symptom it happens to be positioned to see.

## Stack

This is **slice 2 of 2**, stacked on slice 1.

- **Slice 1** — the paged read this consumes: https://github.com/akash-network/console/pull/3673
- **Slice 2 (this PR)** — the cleaner, its config, its CLI subcommand and its scheduling.

**Merge slice 1 first.** This branch is cut from it, so its diff will collapse to just these files once slice 1 lands.

## Checks

In `apps/api`, on this branch: `npm test` → `217 passed (217)` files / `2334 passed (2334)` tests, `npm run lint -- --quiet` → clean, `npx tsc --noEmit` → 41 errors, byte-identical to the pre-existing baseline on `main` (all in unrelated files).

### DIFF
--- a/.helm/console-api-prod-mainnet-values.yaml
+++ b/.helm/console-api-prod-mainnet-values.yaml
@@ -62,9 +62,10 @@ jobs:
   # Drop the flag to start deleting for real once those logs look right. Deleting a remembered
   # definition cannot be undone, so this one bakes before it is trusted.
   # Daily, and offset from the other deployment_settings sweeps so no two of them share a minute.
-  # Object name kept short on purpose: the chart renders it as console-api-<chain>-<name> and the
-  # CronJob controller caps that at 52 chars, which the CLI command name would exceed.
-  - name: sweep-orphaned-definitions
+  # The object name is deliberately shorter than the subcommand: the chart renders it as
+  # {{ Chart.Name }}-{{ .Values.chain }}-<name> and the CronJob name is capped at 52 chars including that
+  # prefix. The `command:` arg below is the one that must match console.ts — do not "fix" the mismatch.
+  - name: cleanup-orphaned-settings
     schedule: "17 3 * * *" # 03:17 every day
     # Backstop for the cleanup's own time budget, which is set lower so it stops and logs before this fires.
     # Forbid keeps a slow run from being joined by the next night's.
--- a/.helm/console-api-prod-mainnet-values.yaml
+++ b/.helm/console-api-prod-mainnet-values.yaml
@@ -58,6 +58,24 @@ jobs:
       - ./dist/console.js
       - notify-expiring-deployments
       - --dry-run
+  # Ships in dry-run for a bake period: --dry-run only logs ORPHANED_DEFINITION_WOULD_SWEEP.
+  # Drop the flag to start deleting for real once those logs look right. Deleting a remembered
+  # definition cannot be undone, so this one bakes before it is trusted.
+  # Daily, and offset from the other deployment_settings sweeps so no two of them share a minute.
+  - name: sweep-orphaned-definitions
+    schedule: "17 3 * * *" # 03:17 every day
+    # One chain listing per owner, run one after another, so a slow or rate-limiting node stretches the run
+    # rather than failing it. Forbid keeps a slow run from being joined by the next night's, and the deadline
+    # ends it instead: the sweep is capped and idempotent, so a run cut short is one that made progress.
+    concurrencyPolicy: Forbid
+    activeDeadlineSeconds: 1800
+    command:
+      - node
+      - --require
+      - ./dist/instrumentation.js
+      - ./dist/console.js
+      - sweep-orphaned-definitions
+      - --dry-run
   - name: mint-act
     schedule: "*/10 * * * *" # every 10 minutes
     concurrencyPolicy: Forbid
--- a/.helm/console-api-prod-mainnet-values.yaml
+++ b/.helm/console-api-prod-mainnet-values.yaml
@@ -58,15 +58,16 @@ jobs:
       - ./dist/console.js
       - notify-expiring-deployments
       - --dry-run
-  # Ships in dry-run for a bake period: --dry-run only logs ORPHANED_DEFINITION_WOULD_SWEEP.
+  # Ships in dry-run for a bake period: --dry-run only logs ORPHANED_DEPLOYMENT_SETTINGS_WOULD_DELETE.
   # Drop the flag to start deleting for real once those logs look right. Deleting a remembered
   # definition cannot be undone, so this one bakes before it is trusted.
   # Daily, and offset from the other deployment_settings sweeps so no two of them share a minute.
+  # Object name kept short on purpose: the chart renders it as console-api-<chain>-<name> and the
+  # CronJob controller caps that at 52 chars, which the CLI command name would exceed.
   - name: sweep-orphaned-definitions
     schedule: "17 3 * * *" # 03:17 every day
-    # One chain listing per owner, run one after another, so a slow or rate-limiting node stretches the run
-    # rather than failing it. Forbid keeps a slow run from being joined by the next night's, and the deadline
-    # ends it instead: the sweep is capped and idempotent, so a run cut short is one that made progress.
+    # Backstop for the cleanup's own time budget, which is set lower so it stops and logs before this fires.
+    # Forbid keeps a slow run from being joined by the next night's.
     concurrencyPolicy: Forbid
     activeDeadlineSeconds: 1800
     command:
@@ -74,7 +75,7 @@ jobs:
       - --require
       - ./dist/instrumentation.js
       - ./dist/console.js
-      - sweep-orphaned-definitions
+      - cleanup-orphaned-deployment-settings
       - --dry-run
   - name: mint-act
     schedule: "*/10 * * * *" # every 10 minutes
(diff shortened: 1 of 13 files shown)


stalniy:
Superseded by an outbox-pattern implementation. Instead of a nightly cron scanning `deployment_settings` and querying the chain per candidate, the deployment setting and a pg-boss compensation job are written in the same transaction; a successful broadcast cancels the job atomically, so the compensation only fires for creates that failed or crashed. That removes the scan entirely — work becomes proportional to the create rate rather than the table size — and with it the paging, time budget and `CLEANUP_CAPPED` starvation window. The chain-existence check is retained as the sole delete authority, since a compensation can also fire when the broadcast succeeded but the process died before cancelling it. Replacement PR to follow.

PR #3744, opened 2026-08-31
fix(template): stream github archives instead of aborting large downloads at 60s

stalniy:
## Why

Closes CON-924

The `Publish Akash Templates` workflow silently published a gutted template gallery. In [attempt 1 of run 33407172476](https://github.com/akash-network/console/actions/runs/33407172476/job/99537495643) the `awesome-akash` archive download was aborted 60.006s after it started:

```
15:14:14.069  GET_TEMPLATES  repository: awesome-akash
15:15:14.075  FETCH_TEMPLATES_FROM_README_FAILED
              error: "TimeoutError: The operation was aborted due to timeout (code: 23)
                  at new DOMException (node:internal/per_context/domexception:76:18)
                  at Timeout._onTimeout (node:internal/abort_controller:154:9)"
15:15:14.113  TEMPLATE_GALLERY_CACHE_BUILD_STATS
              categoryCount: 1, categoryNames: ["Blockchain"], totalTemplates: 62
```

That stack is a byte-for-byte match for `AbortSignal.timeout()` firing — `node:internal/abort_controller:154` is its timer callback and `code: 23` is `DOMException.TIMEOUT_ERR`. It is our own `AbortSignal.timeout(60_000)` in `GitHubArchiveService`, not an undici-internal timeout: those surface through global `fetch` as `TypeError: fetch failed` with a `.cause`, never as a bare `TimeoutError`.

Two defects combined:

1. **The 60s budget covered the whole transfer, not just stalls.** The signal passed to `fetch` also governs `response.arrayBuffer()`. The `awesome-akash` tarball is 67 MB and `cosmos/chain-registry` is another 66 MB downloaded concurrently, so the runner needed a sustained >= 1.1 MB/s or a perfectly healthy download was killed mid-flight. Buffering the whole archive before extracting made it worse: one archive's CPU-heavy extraction blocked the event loop while the other's clock kept running.

2. **The failure was swallowed and the broken result was published.** `fetchTemplatesFromReadme` catches everything and returns `[]`, so the step exited 0 and wrangler deployed 62 templates instead of 395. `getTemplatesFromRepo` also persisted that `[]` into the commit-SHA-keyed cache file, which in the long-running API would pin the empty result for that SHA.

## What

**`github-archive.service.ts`** — the response body now streams into gunzip/tar instead of being buffered via `arrayBuffer()`, and the 60s whole-operation deadline became a 30s **stall** timer that resets on every chunk. A large-but-progressing download now succeeds; only a genuinely dead connection aborts, and the abort reason names the URL and the stall window instead of the opaque `TimeoutError`. Streaming also drops peak memory by ~134 MB and lets extraction overlap the download, so the two archives no longer starve each other.

Downloads are also retried through a `cockatiel` retry policy (exponential backoff with jitter, 3 attempts total), matching how the rest of the repo handles transient transport failures. `handleWhen` excludes 4xx responses, since a missing archive won't appear on a retry.

**`template-gallery.service.ts`** — a repository that yielded zero categories is now an error instead of something cached and published.

> [!IMPORTANT]
> This is a behavioural change worth a look: `awesome-akash` failing will now **fail the workflow** rather than deploy a partial gallery, leaving the previous Cloudflare Pages deployment in place. Per-template-source resilience is untouched, so a single broken template still only skips itself.

`fetchTemplatesFromReadme` still returns `[]` on failure (an existing test asserts it). The new gallery-level guard means that no longer reaches production, but happy to move the propagation into the fetcher instead if reviewers prefer.

### Verification

Retry coverage is pinned for every shape this can fail in, since the point of the change is that a repeat of CON-924 recovers on its own:

- a mid-stream body error after partial data (`controller.error(new TypeError("terminated"))`)
- a transport-level `TypeError: fetch failed` with a `.cause`
- a bare `TimeoutError` `DOMException`, i.e. the exact error from the incident
- retries exhausted, surfacing the last error
- a 404, asserted to make exactly one attempt
- a stalled download aborting, plus one proving a download slower than the stall timeout still completes

90 tests pass across `src/template`. `eslint --quiet` clean; `tsc --noEmit` clean for `src/template` apart from a pre-existing `tar.Header` typing error that also fails on `main`. Ran `npm run build:akash-templates` against the real repos: **30 categories, 395 templates in ~9.5s**.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GitHub archive downloads with automatic retries for transient failures.
  * Added protection against stalled, timed-out, truncated, or interrupted downloads.
  * Reduced memory usage by processing archives as they download.
  * Prevented empty template galleries from being published or cached.

### DIFF
--- a/apps/api/src/template/services/github-archive/github-archive.service.spec.ts
+++ b/apps/api/src/template/services/github-archive/github-archive.service.spec.ts
@@ -1,12 +1,17 @@
 import { gzipSync } from "node:zlib";
 import tar from "tar";
-import { describe, expect, it, vi } from "vitest";
+import { afterEach, describe, expect, it, vi } from "vitest";
 import { mock } from "vitest-mock-extended";
 
 import type { CreateLogger } from "@src/core";
 import { GitHubArchiveService } from "./github-archive.service";
 
 describe(GitHubArchiveService.name, () => {
+  afterEach(() => {
+    vi.useRealTimers();
+    vi.restoreAllMocks();
+  });
+
   describe("getArchive with fileFilter", () => {
     it("only stores content for files matching the filter", async () => {
       const { service, installArchive } = setup();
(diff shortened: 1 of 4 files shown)


stalniy:
Addressed in cb8704ed2.

Applied the cancel, but not the suggested `completed`-flag form — worth recording why, because verifying the finding turned up a wrinkle.

The concern is valid. `releaseLock()` does not cancel, and I confirmed the reader's `cancel()` does reach the response body's source `cancel` hook, so the body really was being left open.

Two details changed the shape of the fix:

1. **The `completed` flag isn't needed.** `cancel()` on an already-closed stream resolves rather than rejecting, so an unconditional cancel leaves the success path untouched. It *does* reject on an errored stream (with the stored error), which is the only reason the `.catch` is there.

2. **The guard would have been unreachable in the scenario as described.** If the body has gone silent, the generator is suspended at `await reader.read()`, so `generator.return()` cannot run the `finally` at all — `pipeline` never settles, and no cleanup code in that block executes. I reproduced this in isolation: with a single chunk followed by silence, the pipeline hangs until something errors the body externally (for us, the 30s stall timer, after which it settles with the original `Z_DATA_ERROR` rather than the abort reason).

Where the fix *does* pay off is the realistic case for a ~70 MB archive: the body is still delivering, so the generator is parked at `yield`, `pipeline` destroys it, and the `finally` runs immediately — cancelling the body with no timer involved. Verified in isolation, then pinned by a test that asserts the source's `cancel` hook fires when extraction fails, which fails without the change and passes with it.

91 tests pass across `src/template`.

stalniy:
Addressed all three findings from the latest round.

**🟡 `ArchiveNotAvailableError` covered the whole 4xx range** (claude, [github-archive.service.ts:126](https://github.com/akash-network/console/blob/main/apps/api/src/template/services/github-archive/github-archive.service.ts#L126)) — fixed in 2c55a5df5.

This one I'd rate above "optional": it was a regression I introduced in this PR, not a pre-existing gap. `handleWhen` excluded the whole class from retry, so a 429 secondary rate limit — plausible precisely because we pull two ~70 MB archives concurrently — would fail the publish on the first attempt, and with the new empty-gallery guard that now fails the whole workflow. Narrowed to `404` and `410`, the only statuses meaning "will never appear"; 429, 403 and 5xx now fall through to the retry policy. Covered by tests for 429 and 502 retrying and 404 still making exactly one attempt.

**🟣 Cache-read path trusts a stale empty gallery** — fixed in ffcf7f572.

Correct, and it undercut the PR's own claim: the guard only refused to *write* `[]`, while the read branch returned any file already on disk for that sha without looking at it. So the exact CON-924 artefact — a `[]` written before this lands — would keep being served until upstream HEAD moved. An empty cached gallery is now treated as a miss, so it refetches and overwrites. Verified the test fails without the guard.

**🔵 Stall test never reached `response.body`** (coderabbit) — fixed in 788e14244.

Right, and worth more than its label: the mock left `fetch()` itself pending, so the assertion never exercised `reader.read()` — the path the stall timer actually guards during a large download. Added a test driving a body that delivers one chunk then goes silent, modelling how undici errors the body on abort. Confirmed it fails when the stall timeout is disabled, so it pins the behaviour rather than the shape.

95 tests pass across `src/template`; lint clean; `tsc` clean apart from a pre-existing `tar.Header` typing error that also fails on `main`. Re-ran the real publish script: 30 categories, 395 templates.

On the ruled-out backpressure concern in the extended reasoning — agreed, and thanks for checking it. `reader.read()` resolves against the fetch stream's own buffer, so slow gunzip/tar downstream doesn't starve `restartStallTimer`; the resets track the socket, not extraction pacing.

PR #3817, opened 2026-09-05
fix(config): ignore non relative return paths on the maintenance redirect

### LINKED ISSUE #2564: Valdiate that the return url param is a relative path
## Why

We currently don't validate that the url to return to is relative and there's risk of phishing attack by putting an absolute path

## What

Strip or ignore absolute paths with domain name

cnYui:
## Why

Fixes #2564

`getReturnPath` takes the user-controlled `return` query param, decodes it, and hands the raw value to `NextResponse.redirect(new URL(returnPath, request.url), 307)`. The surrounding `try/catch` only guards `decodeURIComponent` throwing on malformed input — nothing checks that the value is a relative path.

So whenever `MAINTENANCE_MODE` is not `"true"` (i.e. normal production), `/maintenance?return=<absolute url>` issues a 307 off the console origin. Verified against production:

```console
$ curl -sSI "https://console.akash.network/maintenance?return=https%3A%2F%2Fexample.com%2Fphish"
HTTP/1.1 307 Temporary Redirect
location: https://example.com/phish

$ curl -sSI "https://console.akash.network/maintenance?return=%2F%2Fexample.com%2Fphish"
HTTP/1.1 307 Temporary Redirect
location: https://example.com/phish

$ curl -sSI "https://console.akash.network/maintenance?return=%2Fdeployments"
HTTP/1.1 307 Temporary Redirect
location: /deployments
```

`stats.akash.network` returns exactly the same three responses — `apps/stats-web/src/middleware.ts` carries an identical `getReturnPath`.

## What

`getReturnPath` now resolves the decoded value against `request.url` and keeps it only when it stays on the request's own origin, falling back to `"/"` otherwise. It returns the resolved `pathname + search + hash` instead of the raw string, so the value the caller feeds back into `new URL(returnPath, request.url)` can no longer carry an origin at all.

Applied identically to `apps/deploy-web/src/middleware.ts` and `apps/stats-web/src/middleware.ts`. The `try/catch` and both call sites are untouched.

Going through the URL parser rather than string prefix checks is deliberate: `searchParams.get()` already percent-decodes once before `decodeURIComponent` runs again, so a prefix test has to account for double encoding (`?return=%252F%252Fexample.com`), and `/\example.com` is folded to `//example.com` by WHATWG URL and escapes the origin as well.

Nothing changes for the only value the maintenance branch itself writes — `request.nextUrl.pathname + request.nextUrl.search`, which always starts with a single `/`.

### Tests

Extended `apps/deploy-web/src/middleware.spec.ts` and added a mirror `apps/stats-web/src/middleware.spec.ts` (stats-web had no middleware spec). Four cases each: relative path preserved, absolute URL ignored, protocol-relative ignored, backslash-prefixed ignored.

With the middleware change reverted but the specs kept, the three attack cases fail in both apps:

```
FAIL  src/middleware.spec.ts > middleware > ignores an absolute return url when leaving the maintenance page
AssertionError: expected 'https://evil.example/phish' to be 'http://localhost/'
FAIL  src/middleware.spec.ts > middleware > ignores a protocol relative return url when leaving the maintenance page
AssertionError: expected 'http://evil.example/phish' to be 'http://localhost/'
FAIL  src/middleware.spec.ts > middleware > ignores a backslash prefixed return url when leaving the maintenance page
AssertionError: expected 'http://evil.example/phish' to be 'http://localhost/'
```

With the fix in place (running the `test:unit` scripts' vitest invocation directly, since I am on Windows and the `NODE_ENV=test ...` prefix is not valid in cmd.exe):

```
$ cd apps/deploy-web && NODE_ENV=test DEPLOYMENT_ENV=staging vitest run src/middleware.spec.ts
Test Files  1 passed (1)
     Tests  7 passed (7)

$ cd apps/stats-web && NODE_ENV=test vitest run
Test Files  6 passed (6)
     Tests  56 passed (56)
```

`eslint` and `prettier --check` are clean on all four files (lint-staged also ran them on commit), and `tsc --noEmit` passes in `apps/stats-web`. `tsc --noEmit` in `apps/deploy-web` already reports 94 errors on `main`, all in unrelated `*.spec.tsx` / `tests/` files and none in anything touched here.

### On not reusing `getValidInternalReturnToUrl`

`apps/deploy-web/src/utils/getValidInternalReturnToUrl` guards the client-side `returnTo` flow against the same class of bug, but it does not fit this call site: it is `window`-dependent and middleware has no `window` (without one it rejects every absolute URL, including same-origin ones), it lives in deploy-web so stats-web cannot import it, and its relative branch (`startsWith("/") && !startsWith("//")`) returns `/\example.com/phish` unchanged, which `new URL(value, request.url)` then resolves to `http://example.com/phish`.

That leaves the guard duplicated across the two middlewares, which already duplicate `getReturnPath` and `setContentSecurityPolicyHeaders` verbatim. Happy to extract a shared helper into `packages/` if you would rather have one home for both — I kept the change local so it stays reviewable as a fix.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved maintenance-page redirects to accept only same-site return URLs.
  - Invalid, absolute, protocol-relative, and backslash-prefixed destinations now redirect to the site root.
  - Valid return destinations preserve their path, query parameters, and URL fragments.

- **Tests**
  - Added coverage for valid and rejected return URL scenarios across deployment and statistics pages.


baktun14:
Nice thanks, can you make sure your commits are signed please?

cnYui:
Done — signed the commit and force-pushed. The branch head is now `5ff9e1d`, which GitHub reports as `Verified` (SSH signature). No content changed, only the signature was added; CI has re-triggered on the new head. Let me know if anything else is needed.

baktun14:
> Done — signed the commit and force-pushed. The branch head is now `5ff9e1d`, which GitHub reports as `Verified` (SSH signature). No content changed, only the signature was added; CI has re-triggered on the new head. Let me know if anything else is needed.

please resolve the conflicts as well

cnYui:
Done — conflicts resolved and pushed (branch head `29fcf57`, GitHub reports `Verified`, SSH-signed).

I merged the latest `main` into the branch. The only real conflict was in `apps/deploy-web/src/middleware.spec.ts`, where `main` and this PR each appended new `it(...)` blocks at the same spot; I kept both test sets. `middleware.ts` auto-merged cleanly, and the `getReturnPath` same-origin validation from this PR is intact alongside the new PWA-asset/maintenance handling from `main`.

Verified locally (`npm ci` on the merged tree):
- `apps/deploy-web` → `vitest run src/middleware.spec.ts` → 17 passed
- `apps/stats-web` → `vitest run src/middleware.spec.ts` → 4 passed
- `eslint src/middleware.ts src/middleware.spec.ts` → clean

The PR now shows `MERGEABLE`. Let me know if anything else is needed.

cnYui:
Good catch by the automated review — this was a real bypass. A same-origin *absolute* return like `return=http://<host>//evil.example/phish` passes the origin check, but its pathname is `//evil.example/phish`; `getReturnPath` returned that string and the caller re-parsed it with `new URL(returnPath, request.url)`, where the leading `//` makes it protocol-relative and escapes to `evil.example` (CWE-601).

Fixed in `52ec89a`: `getReturnPath` now returns the validated same-origin `URL` object and it's passed directly to `NextResponse.redirect`, so the pathname is never re-parsed. Added a regression test for `http://localhost//evil.example/phish` asserting the redirect host stays `localhost`.

Verified locally (apps/deploy-web):
- `vitest run src/middleware.spec.ts` → 18 passed (incl. the new case)
- `eslint src/middleware.ts src/middleware.spec.ts` → clean
- `tsc --noEmit` → no new errors

Commit is SSH-signed / GitHub-Verified.

PR #3840, opened 2026-09-07
test(deployment): cover recording a deployment setting through a worker

stalniy:
## Why

Part of CON-952.

`RecordDeploymentSettingHandler` carries this comment on `handle`:

> Unscoped because job execution runs under an empty ability, which a scoped read would throw on before any SQL ran.

Nothing held it to that, and its unit spec cannot: a mocked repository consults no ability and writes no row. If someone "tidied" that call into an `accessibleBy` read, every existing test would stay green and the job would die in production after its retries, silently leaving the funding sweep with no row to find.

## What

Runs the handler through a real worker against a real database, and pins the behaviour a mock cannot get wrong on the repository's behalf.

The second test is the one worth reading: `createDefaultIfMissing` is an `onConflictDoNothing` insert, so a late job must **not** re-enable auto top-up on a deployment whose owner turned it off. Verified by mutation — swapping `onConflictDoNothing` for `onConflictDoUpdate` fails it:

```
× leaves the auto top up choice a user already made alone
AssertionError: expected { …(15) } to match object { autoTopUpEnabled: false, …(1) }
```

Adds `createDseq` to the deployment-setting seeder, since the dseq is needed before the row exists (it keys the job).

### Verification

- `npm run test:integration` — this file passes, and the full integration suite is green
- `npm run lint -- --quiet` — clean
- `npx tsc --noEmit` — 42 errors, unchanged from the `main` baseline (this app has never typechecked clean)

### DIFF
--- a/apps/api/src/deployment/services/record-deployment-setting/record-deployment-setting.handler.integration.ts
+++ b/apps/api/src/deployment/services/record-deployment-setting/record-deployment-setting.handler.integration.ts
@@ -1,0 +1,63 @@
+import { eq } from "drizzle-orm";
+import { container } from "tsyringe";
+import { describe, expect, it } from "vitest";
+
+import type { ApiPgDatabase } from "@src/core";
+import { JOB_NAME, POSTGRES_DB, resolveTable } from "@src/core";
+import { UserRepository } from "@src/user/repositories";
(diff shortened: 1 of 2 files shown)


stalniy:
@coderabbitai review

PR #3841, opened 2026-09-07
test(deployment): cover releasing funding claims for a reverted tx

stalniy:
## Why

Part of CON-952.

`releaseFundingClaim` is a compare-and-set: it clears `last_funded_at` only where the column still equals the exact marker the claim wrote. That marker is produced as `last_funded_at::text` from a **timezone-less** `timestamp` column, makes a round trip through pg-boss JSON, and comes back as `${claim.claimedAt}::timestamp`.

A mocked repository cannot get that comparison wrong on the handler's behalf, so the guard was untested — **drop it from the WHERE clause and every existing test stays green**, while in production a claim another pass had already re-taken would be cleared, letting a second deposit through.

## What

Covers the handler through a real worker against a real database, for all four outcomes: reverted releases, landed holds, not-seen holds, and a re-taken claim holds.

It goes through the real machinery rather than hand-rolling it — `ReconcileManagedTxJobService.schedule()` for the enqueue (so the singleton key and the delay are the production ones), `makeJobDue` to bring the delayed job forward, and a real `claimForFunding` to produce the marker in its true format.

Verified by mutation — dropping the marker from the WHERE clause fails exactly the test meant to catch it:

```
× keeps a claim another pass has since re-taken
AssertionError: expected null to deeply equal 2026-09-07T22:24:39.689Z
```

`TxPresenceService.findTx` is spied at the tsyringe boundary rather than nocked: its own spec already covers the SDK decoding, and what is untested here is the database write.

### Note for whoever writes the next handler test

The first version of this failed with a two-hour offset, and the assertion was at fault, not the code: `claimedAt` is a timezone-less string, so `new Date(claimedAt)` parses it as **local** time while the driver reads the column back as UTC. Assert against the value read from the database, not a reparse of the marker. Several remaining handlers compare timestamp markers the same way.

### Verification

- `npm run test:integration` — this file passes, full integration suite green
- `npm run lint -- --quiet` — clean
- `npx tsc --noEmit` — 42 errors, unchanged from the `main` baseline

### DIFF
--- a/apps/api/src/deployment/services/reconcile-managed-tx/reconcile-managed-tx.handler.integration.ts
+++ b/apps/api/src/deployment/services/reconcile-managed-tx/reconcile-managed-tx.handler.integration.ts
@@ -1,0 +1,106 @@
+import { faker } from "@faker-js/faker";
+import { eq } from "drizzle-orm";
+import { container } from "tsyringe";
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+import { type LandedTx, TxPresenceService } from "@src/chain/services/tx-presence/tx-presence.service";
+import type { ApiPgDatabase } from "@src/core";
+import { JOB_NAME, POSTGRES_DB, resolveTable } from "@src/core";
+import { DeploymentSettingRepository } from "@src/deployment/repositories/deployment-setting/deployment-setting.repository";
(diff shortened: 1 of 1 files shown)


stalniy:
@coderabbitai review

PR #3842, opened 2026-09-07
docs(dx): require integration coverage for job handlers

stalniy:
## Why

Part of CON-952.

A unit spec cannot cover a pg-boss job handler. The worker installs a fake system user and an **empty** CASL ability, so the two ways a job fails in production are both invisible to mocked repositories: refused by the ability layer (dies after its retries, no user-facing error), or writing zero rows while reporting success.

This is not hypothetical drift. While CON-952 was in progress, #3819 landed an 18th handler (`probe-trial-deployment`) with a unit spec only. Writing the rule down is what stops the backfill from being permanent.

## What

Adds a *Background Job Handlers* section to the `console-tests` skill, and marks integration coverage as required for job handlers in the test-level table.

The four mechanics it states, each of which cost real debugging to learn:

- run the handler through the real worker, never `handler.handle(payload)` alone — calling `handle` directly tests an execution environment that does not exist in production
- assert on end state, never on `expect(repo.method).toHaveBeenCalled()` — a write that matched zero rows satisfies the spy and fails the row read
- wait with `expectJobCompleted`, so a refused job reports its cause instead of timing out opaquely
- fake collaborators at the tsyringe boundary or with `nock`, because `test:ci-setup` starts only the database and the mock OAuth2 server — anything reaching notifications, provider-proxy, the tx signer, or KMS dies on `ECONNREFUSED`

It also records the one fair exception: a handler that touches no table and makes no ability-scoped call cannot exhibit either failure mode, so a test there asserts a spy was called after a queue round trip, which tests the harness rather than the handler. Better to say so than to write it.

Depends on #3839 (merged), which added the harness this references.

### DIFF
--- a/.claude/skills/console-tests/SKILL.md
+++ b/.claude/skills/console-tests/SKILL.md
@@ -14,7 +14,7 @@ Choose the lowest-effective test level:
 | Level | When to use in this repo | Mocking strategy |
 |-------|-------------------------|-----------------|
 | **Unit** (default) | Components, hooks, services, utilities | All deps mocked via DI (`mock<T>()`) — never `vi.mock()` |
-| **Integration** | Services with heavy DB logic or Repository patterns | Real DB fixtures; mock only 3rd-party calls |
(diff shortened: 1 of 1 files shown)


stalniy:
@coderabbitai review

PR #3895, opened 2026-09-09
fix(deployment): stop funding deployments the chain has left unbilled past the arrears limit

baktun14:
## Why

The hourly funding sweep has been reporting the same deployment as draining and short of balance every hour since Aug 30. On chain that deployment still holds its original 0.5 ACT escrow with nothing ever transferred to the provider. The escrow module's `EndBlocker` is a no-op, so an account only settles on a deposit, a withdraw or a close, and this lease's provider never withdrew. Its predicted close height is 15 days in the past and it will read `active` until something touches the account.

Had the owner added credits, either funding path would have deposited into it, and the next settlement would have taken two weeks of back rent for a lease no provider has been billing. The sweep should not pay that.

## What

- New `AUTO_TOP_UP_MAX_ARREARS_IN_H` (default 72). A draining deployment whose predicted close is further in the past than this is left out of the fundable set by both the cron and the event-driven path, and recorded as `TOP_UP_DEPLOYMENT_OVERDUE_ON_CHAIN` / `FUND_DRAINING_DEPLOYMENT_OVERDUE_ON_CHAIN`, with `auto_top_up_deployments_overdue_on_chain_total` and `fund_draining_deployments_skips_total{reason="overdue_on_chain"}` counting them.
- Deployments in arrears inside the limit are still funded as before, so a deployment whose escrow ran out between sweeps is still rescued.
- Overdue deployments stay in the owner's active set, so the weekly coverage behind the credits-low email is unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added a configurable 72-hour limit for deployment arrears.
  * Deployments overdue beyond this limit are excluded from automatic funding while remaining active.
  * Added monitoring and warning logs for deployments skipped due to on-chain overdue status.

* **Bug Fixes**
  * Prevented automatic top-ups for deployments that have exceeded the configured arrears window.

* **Tests**
  * Added coverage for default settings, validation, overdue deployments, and monitoring behavior.

### DIFF
--- a/apps/api/src/deployment/config/env.config.spec.ts
+++ b/apps/api/src/deployment/config/env.config.spec.ts
@@ -73,6 +73,18 @@ describe("deployment envSchema", () => {
     });
   });
 
+  describe("AUTO_TOP_UP_MAX_ARREARS_IN_H", () => {
+    it("defaults to three days", () => {
+      const result = envSchema.safeParse(setup());
+
+      expect(result.success && result.data.AUTO_TOP_UP_MAX_ARREARS_IN_H).toBe(72);
+    });
+
+    it("rejects a limit that would leave no room for a provider to bill", () => {
+      expect(envSchema.safeParse(setup({ AUTO_TOP_UP_MAX_ARREARS_IN_H: 0 })).success).toBe(false);
+    });
+  });
+
   describe("AUTO_TOP_UP_TARGET_RUNWAY_IN_H", () => {
     it("accepts the default target runway and look-ahead window", () => {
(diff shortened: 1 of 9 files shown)


baktun14:
Closing: the unwithdrawn lease is a provider-side withdrawal problem, not a phantom lease. The workload may well be running, so the back rent is owed and refusing to fund it would get the deployment closed as soon as the provider withdraws. The sweep's existing behavior (fund deployments in arrears, floor at the current height) is correct here; the hourly insufficient-balance line for a $0 wallet is expected.

PR #3896, opened 2026-09-09
fix(billing): send the credits-low email at most once per wallet per cooldown

baktun14:
## Why

One paid wallet received ten credits-low emails between Sep 2 and Sep 8. Its owner hovers around the threshold: they close deployments, which drops the weekly cost under the balance and unlatches the notice after the 30 minute recovery window, then redeploy, which puts the balance under the weekly cost again and sends a fresh email 30 minutes later. Every email was "correct" under the latch and confirm windows from CON-896, and the wallet still got one every 12 to 24 hours.

The latch and the two confirm windows protect against misreads, not against a wallet that genuinely crosses the threshold several times a day. This adds the missing bound: one credits-low email per wallet per cooldown, whatever the coverage does in between.

## What

- New `CREDITS_LOW_RESEND_COOLDOWN_H` (default 168). The confirmed-recovery clear of `credits_low_notified_at` now also requires the stamp to be older than the cooldown, checked in the same SQL update as the recovery window. While the latch holds, a wallet that drops low again is skipped as `already_notified` exactly as today, so no second email can go out inside the cooldown.
- No schema change. The immediate clear for a wallet with nothing left on auto top-up is unchanged, so a wallet that closes everything and comes back weeks later still gets its first warning.
- Latch, recovery and low confirm windows are otherwise unchanged.

### DIFF
--- a/apps/api/src/billing/config/env.config.ts
+++ b/apps/api/src/billing/config/env.config.ts
@@ -51,6 +51,8 @@ export const envSchema = z.object({
   CREDITS_LOW_RECOVERY_CONFIRM_WINDOW_MIN: z.number({ coerce: true }).min(0).default(30),
   /** How long credits must keep reading low before the email goes out, so a single misread cannot send one. */
   CREDITS_LOW_CONFIRM_WINDOW_MIN: z.number({ coerce: true }).min(0).default(30),
+  /** The notified latch holds at least this long before a recovery can clear it, so closing deployments and redeploying cannot re-arm the email within hours. */
+  CREDITS_LOW_RESEND_COOLDOWN_H: z.number({ coerce: true }).min(0).default(168),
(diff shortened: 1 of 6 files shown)


baktun14:
Re the immediate-unlatch branch bypassing the cooldown: left as is on purpose.

Holding the latch there would trade one extra email for a lost first warning. A wallet with nothing left on auto top-up gets no further checks, so a latch held through the cooldown has nothing to clear it. When that user comes back weeks later and deploys with a low balance, the first check reads low and latched, skips as `already_notified`, and the warning never goes out until they happen to read sufficient for 30 minutes. That is the wallet the email exists for.

The re-arm pattern this PR fixes is the other one: all nine unlatches for the flapping wallet were `sufficient_balance` recoveries with deployments still running, none were `zero_cost`. If the close-everything-and-redeploy variant shows up in the logs, the right fix is a separate signal for "recovered while unobserved", not a held latch.

PR #3902, opened 2026-09-10
fix(deployment): stream the trial probe's process listing and bound what it logs

baktun14:
## Why

Follow-ups to #3901 from watching its first probes in production. Two SSH boxes on different providers returned an empty process listing: the loop wrote through `head`, whose buffering turned a slow or blocked listing into nothing at all before the idle timeout cut the shell. The clean-verdict log line also carried the contents of the config files the probe read, which for an honest trial user can be the app's own keys. The remaining commits bound the logged excerpt by bytes rather than characters, decode native IPv6 socket peers, and only restart a deployment's probe on update when that moves it earlier.

## What

- `--procs` prints each process as it goes instead of piping through `head`, reads `/proc/<pid>/stat` with the shell builtin, takes `comm` from it, and wraps every cmdline read in `timeout 2` where the image has `timeout` (Debian and Alpine both do).
- The `TRIAL_WORKLOAD_PROBED` log line keeps the `== path` headers of the `--files` and `--recent-conf` sections and drops the file bodies; stored detection rows are unchanged.
- The logged excerpt is cut at 4 KiB of UTF-8 without splitting a character.
- `--net` decodes native IPv6 peers as the kernel encodes them and keeps loopback v6 out of the v4 range.
- `restartForUpdatedDeployment` leaves a pending probe alone when it is already due sooner than the restart would schedule.

## Summary by CodeRabbit

- **New Features**
  - Trial workload probes now automatically restart after successful deployment updates when appropriate.
  - Workload probes provide broader system insights, including process, network, executable, and configuration data.
  - Probe evidence is limited to 4,096 UTF-8 bytes and retains file names without exposing file contents.

- **Bug Fixes**
  - Improved text sanitization removes unsafe control characters while preserving tabs and line breaks.
  - Probe scheduling avoids unnecessary restarts when an existing probe is already due soon.
  - Deployment updates remain successful even if probe restarting fails.


baktun14:
Superseded: this branch still carried the three commits already squashed into main with #3901, which is where the conflicts came from. The one new commit is re-based on main in a fresh branch, see the replacement PR.

PR #3904, opened 2026-09-10
feat(deployment): name every deployment the api creates

stalniy:
## Why

Part of CON-953 — https://linear.app/ovrclk/issue/CON-953/store-a-deployments-name-on-the-deployment-not-in-the-browser

A deployment's name lives only in the browser's localStorage today, so it is lost on a device switch or a storage clear, invisible to every other client of the console API, and unavailable to anything the server renders.

This is **slice 1 of 3**: it puts the name in the database and makes sure every deployment the API creates ends up with one. Returning it from the read endpoints is slice 2; renaming through `PATCH` is slice 3.

## What

- `POST /v1/deployments` and `PUT /v1/deployments/{dseq}` accept an optional `name`.
- Created with no name, a deployment is named after the services its SDL declares — de-duplicated across placements, joined with `+`, so a one-service SDL is named `web`.
- Omitting `name` on `PUT` leaves the stored name alone. That is a deliberate exception to this endpoint's wholesale-replacement convention, and the same rule `runtimeLimitHours` already relies on: `undefined` drops the column out of drizzle's `SET` clause.
- `name` is trimmed before it is measured and bounded at 256 characters, so `"  api  "` is stored as `"api"` while `""`, `"   "` and 257 characters each answer 400.
- Migration `0053_deployment_name` adds `name text` to `deployment_settings`, nullable, with no backfill — deployments created before this ship keep no name rather than having one invented for them.

The column is `text` rather than `varchar(256)` on purpose. The 256 bound applies to what a *caller* may supply and lives in the request schema; a name the console *derives* is bounded by nothing the caller controls, so it is clamped in code rather than rejected by Postgres mid-create.

### Example

Creating a deployment without a name, adapted from `test/functional/deployments.spec.ts` with its fixture helpers inlined:

```ts
const response = await app.request("/v1/deployments", {
  method: "POST",
  body: JSON.stringify({ data: { sdl: twoServiceSdl } }),
  headers: new Headers({ "Content-Type": "application/json", "x-api-key": userApiKeySecret })
});

expect(response.status).toBe(201);
const { data } = (await response.json()) as { data: { dseq: string } };

expect(await deploymentSettingRepository.findOneBy({ userId: user.id, dseq: data.dseq })).toMatchObject({ name: "db+web" });
```

Supplying `{ data: { sdl, name: "  checkout stack  " } }` instead stores `"checkout stack"` and derives nothing.

## Two decisions that want a human

**1. A multi-service name comes out alphabetical, not in document order.** `two-service-sdl.yml` declares `web` before `db`, and the deployment is named `db+web`. The issue asks for the service names "joined with `+`" and prints `web+db` as its example, but it also states the order is the order of services inside `manifest.groups` — and the chain SDK sorts services with `compareStrings` before emitting the manifest, so that pair can never come out `web+db`. The explicit rule was followed over the example. This is now pinned by a functional test, so flipping it is a one-line change plus a decision about rows already created.

**2. The name is stripped from every deployment-settings response.** It is exposed only by `GET /v1/deployments` and `GET /v1/deployments/{dseq}` (slice 2), and deliberately not by `/v1/deployment-settings` or `/v2/deployment-settings`.

`DeploymentSettingService.withEstimatedTopUpAmount` assembles that response *opt-out*: it destructures the columns that must not be published out of the row and spreads the rest, and its return type is an `Omit<…>` built the same way. Adding any column to `deployment_settings` therefore publishes it on four shipped routes by default. Without the exclusion this PR adds, `name` joined those routes as a field `DeploymentSettingResponseSchema` does not declare — present in the JSON, absent from `openapi.json` and the published docs. It is stripped to keep a documented response documented, and the issue names only the two deployment reads as returning the name.

The trade-off to weigh: a client rendering console state from a settings endpoint gets every stored field except the name, and must also call `GET /v1/deployments/{dseq}` once slice 2 lands. If those endpoints should carry it, the change is additive — declare `name` on `DeploymentSettingResponseSchema` and regenerate the docs — and does not alter what this PR stores.

## Not in this slice

- Reading the name back from `GET /v1/deployments` and `GET /v1/deployments/{dseq}` — slice 2.
- Renaming through `PATCH /v1/deployments/{dseq}`, including a name-only patch of a deployment with no recorded SDL — slice 3.
- `deploy-web` changes, which CON-953 puts out of scope.

## Tests

- **Unit** — `deployment-name.spec.ts` covers the derivation: one service, several, a service under two placements, an over-long join and a manifest with no services. `deployment-writer.service.spec.ts` covers `input.name ?? derived` precedence. `deployment-setting.service.spec.ts` pins the settings-response exclusion, so the next column added to that table cannot leak silently.
- **Integration** — `deployment-setting.repository.integration.ts` covers the omit-keeps-it rule against the real column, because that guarantee lives in whether drizzle emits the column in the `SET` clause and a mocked cursor cannot get it wrong on the code's behalf.
- **Functional** — `deployments.spec.ts` covers create and PUT end to end over HTTP: the derived name, a supplied name, trimming, the 256 boundary and the 400s.

Checks on this branch:

- `npm test` — passed (exit 0)
- `npm run lint -- --quiet` — passed (exit 0)
- `npx tsc --noEmit` — passed (exit 0)

Known non-blocking review follow-ups, left for a later slice rather than widening this diff: no assertion yet that a services-only `PATCH` cannot erase a stored name (the guarantee currently rests on `name` being absent from `replaceDefinitionIfVersionMatches`'s `SET` clause); `MAX_DEPLOYMENT_NAME_LENGTH` sits in the derivation util rather than in `config/sdl.config.ts` with the module's other request bounds; and two pairs of tests duplicate a setup that could be folded together.

## Demo

# CON-953 slice 1 — naming a deployment at the moment it is created

*2026-09-10T19:40:26Z by Showboat 0.6.1*

A deployment's name has lived in the browser's localStorage until now, so it was lost on a device switch and invisible to every other client of the console API. This slice puts the name in the database and makes sure every deployment the API creates ends up with one. Reading the name back from the deployment endpoints is slice 2; renaming through PATCH is slice 3.

Start with the column. The migration adds it nullable and backfills nothing, which is what leaves deployments created before this change without a name rather than inventing one for them.

```bash
cat apps/api/drizzle/0053_deployment_name.sql
echo
echo "UPDATE / SET statements in the migration: $(grep -ciE "update|set " apps/api/drizzle/0053_deployment_name.sql || true)"
```

```output
ALTER TABLE "deployment_settings" ADD COLUMN "name" text;
UPDATE / SET statements in the migration: 0
```

Next, what a deployment gets called when the caller supplies no name. The API derives it from the manifest it already builds for the chain: the service names, de-duplicated, joined with `+`. Below, two real SDL files go through the real manifest generator and the real derivation the create path calls.

```bash
cd apps/api
cat > .demo-derive.ts <<'TS'
import { readFileSync } from "node:fs";
import { generateManifest, yaml } from "@akashnetwork/chain-sdk";
import type { SDLInput } from "@akashnetwork/chain-sdk";
import { deriveDeploymentName } from "./src/deployment/utils/deployment-name/deployment-name";

for (const file of ["hello-world-sdl.yml", "two-service-sdl.yml"]) {
  const parsed = yaml.raw<SDLInput>(readFileSync(`test/mocks/${file}`, "utf8"));
  const manifest = generateManifest(parsed);
  if (!manifest.ok) throw new Error("sdl did not generate a manifest");
  const services = manifest.value.groups.flatMap(g => g.services.map(s => s.name));
  console.log(file);
  console.log(`  services in manifest.groups : ${JSON.stringify(services)}`);
  console.log(`  deployment is named         : ${deriveDeploymentName(manifest.value.groups)}`);
}
TS
npx tsx .demo-derive.ts 2>&1 | grep -v "npm warn"
rm -f .demo-derive.ts
```

```output
hello-world-sdl.yml
  services in manifest.groups : ["web"]
  deployment is named         : web
two-service-sdl.yml
  services in manifest.groups : ["db","web"]
  deployment is named         : db+web
```

Worth pausing on `db+web`. The issue asks for the service names "joined with `+`" and prints `web+db` as its example, but it also states the order is the order of services inside `manifest.groups` — and the chain SDK sorts services alphabetically while building the manifest, so that pair can only ever come out `db+web`. The explicit rule was followed rather than the example; a reviewer should confirm that is the intended reading.

Now what a caller may supply. The name is optional, trimmed before it is measured, and bounded at 256 characters — a bound that lives in the request schema, not in the database. Below, the real create and update request schemas judge a series of names.

```bash
cd apps/api
cat > .demo-schema.ts <<'TS'
import "@hono/zod-openapi";
import { CreateDeploymentRequestSchema, UpdateDeploymentRequestSchema } from "./src/deployment/http-schemas/deployment.schema";

const attempts: Array<[string, string]> = [
  ["a plain name", "staging api"],
  ["padded with spaces", "   checkout stack   "],
  ["nothing but spaces", "   "],
  ["empty", ""],
  ["256 characters", "n".repeat(256)],
  ["257 characters", "n".repeat(257)]
];

for (const [label, name] of attempts) {
  const create = CreateDeploymentRequestSchema.safeParse({ data: { sdl: "version: 2.0", name } });
  const update = UpdateDeploymentRequestSchema.safeParse({ data: { sdl: "version: 2.0", name } });
  const stored = create.success
    ? (create.data.data.name!.length > 40 ? `stored, ${create.data.data.name!.length} characters` : `stored as ${JSON.stringify(create.data.data.name)}`)
    : "refused (400)";
  console.log(`POST ${label.padEnd(19)} -> ${stored.padEnd(28)} | PUT -> ${update.success ? "accepted" : "refused (400)"}`);
}

const omitted = CreateDeploymentRequestSchema.safeParse({ data: { sdl: "version: 2.0" } });
console.log(`POST ${"no name at all".padEnd(19)} -> ${omitted.success ? "accepted, the api derives one" : "refused (400)"}`);
TS
npx tsx .demo-schema.ts 2>&1 | grep -v "npm warn"
rm -f .demo-schema.ts
```

```output
POST a plain name        -> stored as "staging api"      | PUT -> accepted
POST padded with spaces  -> stored as "checkout stack"   | PUT -> accepted
POST nothing but spaces  -> refused (400)                | PUT -> refused (400)
POST empty               -> refused (400)                | PUT -> refused (400)
POST 256 characters      -> stored, 256 characters       | PUT -> accepted
POST 257 characters      -> refused (400)                | PUT -> refused (400)
POST no name at all      -> accepted, the api derives one
```

Then the database itself. This creates a scratch database, runs the real migrations against it, and drives the real repository — no mocks. It matters because the guarantee that "updating a deployment without naming it leaves the name alone" is not a branch in the code: it is whether drizzle puts the column in the `SET` clause at all. The emitted SQL is shown alongside each result.

```bash
cd apps/api
cat > .demo-db.ts <<'TS'
import "reflect-metadata";
import dotenv from "dotenv";
import dotenvExpand from "dotenv-expand";
import { RAW_APP_CONFIG } from "@src/core/providers/raw-app-config.provider";

dotenvExpand.expand(dotenv.config({ path: "env/.env.functional.test" }));
container.register(RAW_APP_CONFIG, { useValue: process.env });

async function main() {
  const { TestDatabaseService } = await import("./test/services/test-database.service");
  const database = new TestDatabaseService("demo.spec.ts");
  await database.setup();

  const { DeploymentSettingRepository } = await import("@src/deployment/repositories/deployment-setting/deployment-setting.repository");
  const { UserRepository } = await import("@src/user/repositories");
  const { TxService } = await import("@src/core/services/tx/tx.service");
  const { POSTGRES_DB, resolveTable } = await import("@src/core");

  const pg = container.resolve<any>(POSTGRES_DB);
  const txManager = new TxService(pg);
  const settings = new DeploymentSettingRepository(pg, resolveTable("DeploymentSettings") as any, txManager);
  const users = new UserRepository(pg, resolveTable("Users") as any, txManager);
  const user = await users.create({ userId: "demo-user" });

  const [column] = await pg.execute(
    `select data_type, is_nullable, character_maximum_length from information_schema.columns
     where table_name = 'deployment_settings' and column_name = 'name'`
  );
  console.log(`the column: name ${column.data_type}, nullable=${column.is_nullable}, length limit in the database=${column.character_maximum_length ?? "none"}`);

  console.log("");
  console.log("a deployment created with a name, then updated with a new sdl and no name:");
  await settings.upsertDefinition({ userId: user.id, dseq: "1001", sdl: "version: 2.0", manifestVersion: "BAUG", name: "web" });
  const created = await settings.findOneBy({ userId: user.id, dseq: "1001" });
  console.log(`  after create : sdl=${JSON.stringify(created?.sdl)} name=${JSON.stringify(created?.name)}`);

  await settings.upsertDefinition({ userId: user.id, dseq: "1001", sdl: "version: 2.1", manifestVersion: "BQYH" });
  const updated = await settings.findOneBy({ userId: user.id, dseq: "1001" });
  console.log(`  after update : sdl=${JSON.stringify(updated?.sdl)} name=${JSON.stringify(updated?.name)}  <- the rename-free update left the name alone`);

  console.log("");
  console.log("a deployment recorded the way every deployment before this change was recorded, with no name:");
  await settings.upsertDefinition({ userId: user.id, dseq: "1002", sdl: "version: 2.0", manifestVersion: "BAUG" });
  const legacy = await settings.findOneBy({ userId: user.id, dseq: "1002" });
  console.log(`  reads back   : name=${JSON.stringify(legacy?.name)}`);

  await database.teardown();
  process.exit(0);
}

main();
TS
npx tsx .demo-db.ts 2>&1 \
 | grep -vE "npm warn|Setting up test databases|Dropping test databases|Failed to find Response internal state key" \
 | grep -vE "^\{$|^\}$|^  (severity_local|severity|code|message|file|line|routine):" \
 | sed -E "s/.*(on conflict[^-]*) --.*/      emitted SQL: \1/" \
 | grep -vE "^\[|^    context:|^    orm:"
rm -f .demo-db.ts
```

```output
the column: name text, nullable=YES, length limit in the database=none

a deployment created with a name, then updated with a new sdl and no name:
      emitted SQL: on conflict ("dseq","user_id") do update set "sdl" = $7, "name" = $8, "manifest_version" = $9, "updated_at" = now() returning "id"
  after create : sdl="version: 2.0" name="web"
      emitted SQL: on conflict ("dseq","user_id") do update set "sdl" = $6, "manifest_version" = $7, "updated_at" = now() returning "id"
  after update : sdl="version: 2.1" name="web"  <- the rename-free update left the name alone

a deployment recorded the way every deployment before this change was recorded, with no name:
      emitted SQL: on conflict ("dseq","user_id") do update set "sdl" = $6, "manifest_version" = $7, "updated_at" = now() returning "id"
  reads back   : name=null
```

Finally, which endpoints carry the name. This reads the generated OpenAPI document that ships with the API, so it is the contract a client actually sees rather than a claim about it.

```bash
cd apps/api && python3 -c '
import json
doc = json.load(open("swagger/openapi.json"))

def props(schema):
    if not isinstance(schema, dict):
        return []
    data = schema.get("properties", {}).get("data", {})
    if data.get("type") == "array":
        data = data.get("items", {})
    return list(data.get("properties", {}).keys())

rows = []
for path, ops in sorted(doc["paths"].items()):
    if "deployment" not in path:
        continue
    for verb, op in sorted(ops.items()):
        body = op.get("requestBody", {}).get("content", {}).get("application/json", {}).get("schema", {})
        ok = op.get("responses", {}).get("200") or op.get("responses", {}).get("201") or {}
        resp = ok.get("content", {}).get("application/json", {}).get("schema", {})
        if op.get("requestBody") or resp:
            rows.append(("%-6s %s" % (verb.upper(), path), "name" in props(body), "name" in props(resp)))

print("%-58s %14s %14s" % ("endpoint", "accepts a name", "returns a name"))
for label, accepts, returns in rows:
    print("%-58s %14s %14s" % (label, "yes" if accepts else "-", "yes" if returns else "-"))
'
```

```output
endpoint                                                   accepts a name returns a name
GET    /akash/deployment/{version}/deployments/info                     -              -
GET    /akash/deployment/{version}/deployments/list                     -              -
GET    /v1/addresses/{address}/deployments/{skip}/{limit}               -              -
GET    /v1/deployment-alerts/{dseq}                                     -              -
POST   /v1/deployment-alerts/{dseq}                                     -              -
GET    /v1/deployment-funding-config                                    -              -
POST   /v1/deployment-settings                                          -              -
GET    /v1/deployment-settings/{userId}/{dseq}                          -              -
PATCH  /v1/deployment-settings/{userId}/{dseq}                          -              -
GET    /v1/deployment/{owner}/{dseq}                                    -              -
GET    /v1/deployments                                                  -              -
POST   /v1/deployments                                                yes              -
DELETE /v1/deployments/{dseq}                                           -              -
GET    /v1/deployments/{dseq}                                           -              -
PATCH  /v1/deployments/{dseq}                                           -              -
PUT    /v1/deployments/{dseq}                                         yes              -
POST   /v1/deposit-deployment                                           -              -
GET    /v1/providers/{provider}/deployments/{skip}/{limit}              -              -
POST   /v2/deployment-settings                                          -              -
GET    /v2/deployment-settings/{dseq}                                   -              -
PATCH  /v2/deployment-settings/{dseq}                                   -              -
```

Two things to read off that table. The `yes` column is this slice: `POST /v1/deployments` and `PUT /v1/deployments/{dseq}` take a name. Nothing returns one yet — the two GET reads gain it in slice 2, which is why the create results above were read from the database rather than from a response body.

The rest of the table is a decision that wants a human. The `/v1/deployment-settings` and `/v2/deployment-settings` endpoints are named after a deployment's settings and the name is stored on that very row, yet they neither accept nor return it. That is deliberate: the settings response is assembled by dropping the fields it must not publish and spreading the rest, so the new column joined four shipped routes as a field the response schema never declared — present in the JSON, absent from the docs above. It is stripped to keep a documented response documented, and the issue names only the two deployment reads as returning the name. If those endpoints should carry it, the change is to declare it on `DeploymentSettingResponseSchema` and regenerate the docs.

## Summary by CodeRabbit

* **New Features**
  * Deployment names can now be provided when creating or updating deployments.
  * Names are trimmed, must be non-empty, and may contain up to 256 characters.
  * Deployments without an explicit name automatically receive one based on their services.
  * Existing names are preserved when updates omit the name.
  * Deployments can be renamed or assigned a name after creation.
* **Documentation**
  * Deployment API specifications now document the optional name field and its validation rules.

### DIFF
--- a/apps/api/src/deployment/utils/deployment-name/deployment-name.ts
+++ b/apps/api/src/deployment/utils/deployment-name/deployment-name.ts
@@ -1,0 +1,21 @@
+import type { Manifest } from "@akashnetwork/chain-sdk";
+
+/** Bounds the name in the request schema alone, the column being unsized, so a name the console derives for a caller who supplied none is shortened rather than refusing a create that would succeed today. */
+export const MAX_DEPLOYMENT_NAME_LENGTH = 256;
+
+const SERVICE_NAME_SEPARATOR = "+";
+
+export function deriveDeploymentName(groups: Manifest): string | undefined {
+  const serviceNames = new Set(
+    groups
+      .flatMap(group => group.services)
+      .map(service => service.name.trim())
+      .filter(serviceName => serviceName.length > 0)
+  );
+
+  if (serviceNames.size === 0) {
+    return undefined;
+  }
+
+  return [...serviceNames].join(SERVICE_NAME_SEPARATOR).slice(0, MAX_DEPLOYMENT_NAME_LENGTH);
+}
--- a/apps/api/src/deployment/utils/deployment-name/deployment-name.ts
+++ b/apps/api/src/deployment/utils/deployment-name/deployment-name.ts
@@ -7,7 +7,7 @@ const SERVICE_NAME_SEPARATOR = "+";
 
 export function deriveDeploymentName(groups: Manifest): string | undefined {
   const serviceNames = new Set(
-    groups
+    Iterator.from(groups)
       .flatMap(group => group.services)
       .map(service => service.name.trim())
       .filter(serviceName => serviceName.length > 0)
--- a/apps/api/drizzle/0053_deployment_name.sql
+++ b/apps/api/drizzle/0053_deployment_name.sql
@@ -1,0 +1,1 @@
+ALTER TABLE "deployment_settings" ADD COLUMN "name" text;
--- a/apps/api/drizzle/meta/0053_snapshot.json
+++ b/apps/api/drizzle/meta/0053_snapshot.json
@@ -1,0 +1,1756 @@
+{
+  "id": "ac23eb43-aae2-4886-bc9e-754d7776ba29",
+  "prevId": "5bd1a611-9b71-43ba-9ac6-260581ff5218",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "public.user_wallets": {
+      "name": "user_wallets",
+      "schema": "",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "serial",
+          "primaryKey": true,
+          "notNull": true
+        },
+        "user_id": {
+          "name": "user_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "address": {
+          "name": "address",
+          "type": "varchar",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "deployment_allowance": {
+          "name": "deployment_allowance",
+          "type": "numeric(20, 2)",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "'0.00'"
+        },
+        "fee_allowance": {
+          "name": "fee_allowance",
+          "type": "numeric(20, 2)",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "'0.00'"
+        },
+        "trial": {
+          "name": "trial",
+          "type": "boolean",
+          "primaryKey": false,
+          "notNull": false,
+          "default": true
+        },
+        "activated_at": {
+          "name": "activated_at",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "credits_low_notified_at": {
+          "name": "credits_low_notified_at",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "credits_sufficient_since": {
+          "name": "credits_sufficient_since",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "credits_low_since": {
+          "name": "credits_low_since",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "abuse_locked_at": {
+          "name": "abuse_locked_at",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "abuse_locked_reason": {
+          "name": "abuse_locked_reason",
+          "type": "varchar(64)",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "created_at": {
+          "name": "created_at",
+          "type": "timestamp",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "now()"
+        },
+        "updated_at": {
+          "name": "updated_at",
+          "type": "timestamp",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "now()"
+        }
+      },
+      "indexes": {},
+      "foreignKeys": {
+        "user_wallets_user_id_userSetting_id_fk": {
+          "name": "user_wallets_user_id_userSetting_id_fk",
+          "tableFrom": "user_wallets",
+          "tableTo": "userSetting",
+          "columnsFrom": [
+            "user_id"
+          ],
+          "columnsTo": [
+            "id"
+          ],
+          "onDelete": "cascade",
+          "onUpdate": "no action"
+        }
+      },
+      "compositePrimaryKeys": {},
+      "uniqueConstraints": {
+        "user_wallets_user_id_unique": {
+          "name": "user_wallets_user_id_unique",
+          "nullsNotDistinct": false,
+          "columns": [
+            "user_id"
+          ]
+        },
+        "user_wallets_address_unique": {
+          "name": "user_wallets_address_unique",
+          "nullsNotDistinct": false,
+          "columns": [
+            "address"
+          ]
+        }
+      },
+      "policies": {},
+      "checkConstraints": {},
+      "isRLSEnabled": false
+    },
+    "public.payment_methods": {
+      "name": "payment_methods",
+      "schema": "",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "uuid",
+          "primaryKey": true,
+          "notNull": true,
+          "default": "uuid_generate_v4()"
+        },
+        "user_id": {
+          "name": "user_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "fingerprint": {
+          "name": "fingerprint",
+          "type": "varchar(255)",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "payment_method_id": {
+          "name": "payment_method_id",
+          "type": "varchar(255)",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "is_validated": {
+          "name": "is_validated",
+          "type": "boolean",
+          "primaryKey": false,
+          "notNull": true,
+          "default": false
+        },
+        "is_default": {
+          "name": "is_default",
+          "type": "boolean",
+          "primaryKey": false,
+          "notNull": true,
+          "default": false
+        },
+        "created_at": {
+          "name": "created_at",
+          "type": "timestamp",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "now()"
+        },
+        "updated_at": {
+          "name": "updated_at",
+          "type": "timestamp",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "now()"
+        }
+      },
+      "indexes": {
+        "payment_methods_fingerprint_payment_method_id_unique": {
+          "name": "payment_methods_fingerprint_payment_method_id_unique",
+          "columns": [
+            {
+              "expression": "fingerprint",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            },
+            {
+              "expression": "payment_method_id",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": true,
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        },
+        "payment_methods_user_id_is_default_unique": {
+          "name": "payment_methods_user_id_is_default_unique",
+          "columns": [
+            {
+              "expression": "user_id",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            },
+            {
+              "expression": "is_default",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": true,
+          "where": "\"payment_methods\".\"is_default\" = true",
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        },
+        "payment_methods_fingerprint_idx": {
+          "name": "payment_methods_fingerprint_idx",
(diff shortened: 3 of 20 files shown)


apps/api/src/deployment/utils/deployment-name/deployment-name.ts
```diff
@@ -0,0 +1,21 @@
...
+
+const SERVICE_NAME_SEPARATOR = "+";
+
+export function deriveDeploymentName(groups: Manifest): string | undefined {
+  const serviceNames = new Set(
+    groups
+      .flatMap(group => group.services)
+      .map(service => service.name.trim())
+      .filter(serviceName => serviceName.length > 0)
+  );
```
stalniy, line 14: @coderabbitai it's possible to use latest lazy generators here to optimize the logic, right?

stalniy: but it will do 1 iteration instead of 3, right?

PR #3925, opened 2026-09-11
refactor(deployment): fingerprint stored secrets before and after a rotation

stalniy:
## Why

Part of CON-876.

The upcoming key rotation re-wraps every user's data encryption key, and it must never read, open or write a single stored secret. "We didn't touch them" has to be evidence, not a claim — on a live system where users are writing their own secrets throughout the run.

This is **slice 1 of 2**: the instrument that produces that evidence. The rotation itself, its CLI command and its preflight land in the next PR, stacked on this one. No CLI or HTTP surface is added here.

## What

`StoredSecretFingerprintService` measures every stored secret without opening one, and reconciles the measurement afterwards per row.

```ts
const before = await storedSecretFingerprintService.take();

// ... the rotation re-wraps data keys, touching nothing in deployment_settings ...

const reconciliation = await storedSecretFingerprintService.reconcile(before);

if (reconciliation.unexplained.length > 0) {
  // the run failed: these deployments changed with no writer behind them
}
```

`take()` sweeps `deployment_settings` where `sealed_secrets is not null`, keyset-paged on `id`, and returns a sha256 over id, byte length and token in id order, plus the row count and byte total. Under live traffic the two digests legitimately differ, so `reconcile()` decides **per row** instead of by digest:

| what happened to a row | verdict |
| --- | --- |
| token changed, `updated_at` moved | the owner's own write — `ownerRewritten` |
| token appeared, stamped at or after the run opened | `created` |
| token gone, row written since, or row deleted outright | `removed` |
| token changed, appeared or vanished with **no writer behind it** | `unexplained`, named by deployment id |

Both writers of `sealed_secrets` stamp `updated_at` in the same statement, which is what makes that discriminator work. A row that no longer exists stays tolerated: nothing it left behind can tell a tamper from an owner deleting their deployment.

Two details worth a reviewer's eye:

- **Keyset, not offset.** Offset paging over a predicate that shifts under live traffic silently skips rows, and a digest from a sweep that skipped rows proves nothing.
- **Markers, not `Date`s.** `updated_at` travels as fixed-width `to_char(..., 'YYYY-MM-DD"T"HH24:MI:SS.US')` text. A `Date` truncates microseconds to milliseconds, so two writes inside one millisecond would compare equal and a healthy owner write would be reported as a tamper. The fixed width also makes the text order chronologically, which the "stamped at or after the run opened" check needs.

No schema change and no migration — this reads columns that already exist.

### Tests

Integration is the primary level here: every property is about a real database and what survives a concurrent writer, which a mock cannot get wrong on the code's behalf. The unit spec covers the classification table with a mocked sweep.

```ts
it("reports a token rewritten without its updated_at moving as a re-seal", async () => {
  const { service, storeSecrets, writeSecretsWithoutTouchingUpdatedAt } = await setup();
  const { id } = await storeSecrets(newSealedToken());
  await storeSecrets(newSealedToken());

  const before = await service.take();
  await writeSecretsWithoutTouchingUpdatedAt(id, newSealedToken());
  const reconciliation = await service.reconcile(before);

  expect(reconciliation).toMatchObject({ ownerRewritten: 0, unexplained: [id] });
});
```

Checks on `apps/api`: `npm test`, `npm run lint -- --quiet` and `npx tsc --noEmit` all pass.

One unrelated fix rides along, because it was making the suite red one run in ten: `setupDeploymentListMock` in `test/functional/deployments.spec.ts` drew each dseq from `faker.string.numeric()`, which returns a **single digit**, so two listed deployments shared a dseq about 10% of the time and the unnamed one inherited the other's name. The dseqs are now numbered from one base. The flake predates this branch (it arrived with #3905); no assertion changed.

## Demo

# Fingerprinting stored secrets before and after a rotation

*2026-09-11T15:11:56Z by Showboat 0.6.1*

A key rotation is about to re-wrap every user's data encryption key. It must never touch a user's **stored secrets** — and it has to be able to prove that afterwards, on a live system where users are writing their own secrets the whole time.

This slice builds the instrument that proves it. There is no CLI command yet (that lands with the rotation itself), so this demo drives the service directly against a real PostgreSQL database, the same way the rotation will.

Everything below runs against a throwaway database built from the project's own migrations.

```bash
cd apps/api && ../../node_modules/.bin/tsx ../../.work/con-876-re-wrap-data-keys-onto/demo.ts prepare
```

```output
database con876_fingerprint_demo created and migrated
```

The fleet is three deployments belonging to one user: **dseq 1001** and **dseq 1002** each hold a sealed secrets token, **dseq 1003** holds none. Every scenario below re-seeds that same fleet, so the numbers are comparable across scenarios.

First: what does the instrument actually read? Here is every statement one fingerprint pass issues, with the page size turned down to one row so the paging is visible.

```bash
cd apps/api && ../../node_modules/.bin/tsx ../../.work/con-876-re-wrap-data-keys-onto/demo.ts trace
```

```output
every statement one fingerprint pass issued, in order:
  1. select to_char(now()::timestamp, $1) as marker
  2. select "id", "sealed_secrets", to_char("updated_at", $1) from "deployment_settings" where "deployment_settings"."sealed_secrets" is not null order by "deployment_settings"."id" asc limit $2
  3. select "id", "sealed_secrets", to_char("updated_at", $1) from "deployment_settings" where ("deployment_settings"."sealed_secrets" is not null and "deployment_settings"."id" > $2) order by "deployment_settings"."id" asc limit $3
  4. select "id", "sealed_secrets", to_char("updated_at", $1) from "deployment_settings" where ("deployment_settings"."sealed_secrets" is not null and "deployment_settings"."id" > $2) order by "deployment_settings"."id" asc limit $3
```

Three things a reader can check for themselves in that output:

- it selects **three columns** — the id, the sealed token, and the row's write timestamp. It never decrypts anything, and it never asks for a data key;
- it pages by `id > $2` rather than by offset, so rows cannot be skipped when live traffic shifts the table under the sweep;
- it opens with `to_char(now()::timestamp, $1)`, the reading a write landing right then would leave. That is the clock every later judgement is made against.

Now the scenarios. The first is the boring one: take a fingerprint, do nothing, reconcile.

```bash
cd apps/api && ../../node_modules/.bin/tsx ../../.work/con-876-re-wrap-data-keys-onto/demo.ts quiet
```

```output
what happened during the run : nothing at all

fingerprint before : digest 2164adfd2689689eaf1f10291e5b7b6aacd041dc5578e3ac443039c0e1567aa2
                     2 rows, 172 bytes of sealed secrets
fingerprint after  : digest 2164adfd2689689eaf1f10291e5b7b6aacd041dc5578e3ac443039c0e1567aa2
                     2 rows, 172 bytes of sealed secrets

explained by a writer : 0 rewritten, 0 created, 0 removed
unexplained           : none

VERDICT: clean run, the rotation may report success
```

Identical digest, identical row count, identical byte total. That is the easy case.

The hard case is the one a naive digest comparison gets wrong: a **real user**, updating their own deployment while the rotation runs. Their token legitimately changes, so the fleet digest legitimately moves.

```bash
cd apps/api && ../../node_modules/.bin/tsx ../../.work/con-876-re-wrap-data-keys-onto/demo.ts owner-write
```

```output
what happened during the run : the owner of dseq 1001 updated their deployment, re-sealing their own secrets

fingerprint before : digest 2164adfd2689689eaf1f10291e5b7b6aacd041dc5578e3ac443039c0e1567aa2
                     2 rows, 172 bytes of sealed secrets
fingerprint after  : digest ee6823c8f910cfef23dcbf05bb0b82bcecae5a2a31159af97134324ca8fe3ad7
                     2 rows, 176 bytes of sealed secrets

explained by a writer : 1 rewritten, 0 created, 0 removed
unexplained           : none

VERDICT: clean run, the rotation may report success
```

The digest moved and the run is still clean: the write went through the application, which stamps `updated_at` in the same statement, so the changed token has a writer behind it and is counted rather than reported.

Now the same shaped change, made **behind the application's back** — the case the whole instrument exists for. A token is replaced by raw SQL that leaves `updated_at` alone.

```bash
cd apps/api && ../../node_modules/.bin/tsx ../../.work/con-876-re-wrap-data-keys-onto/demo.ts reseal
```

```output
what happened during the run : something replaced the token of dseq 1001 behind the application's back

fingerprint before : digest 2164adfd2689689eaf1f10291e5b7b6aacd041dc5578e3ac443039c0e1567aa2
                     2 rows, 172 bytes of sealed secrets
fingerprint after  : digest 6d5538176e48216226e7d593ef662567b51708818301abe1b4d5d9af5364f930
                     2 rows, 175 bytes of sealed secrets

explained by a writer : 0 rewritten, 0 created, 0 removed
unexplained           : 11111111-1111-4111-8111-111111111111 (dseq 1001)

VERDICT: the rotation must fail loudly and name these deployments
```

Named, not just counted. The rotation gets the deployment id to put in its failure.

The next two are the cases a code review caught, and they are worth watching together because **they end in exactly the same fingerprint**. First, a token erased by raw SQL.

```bash
cd apps/api && ../../node_modules/.bin/tsx ../../.work/con-876-re-wrap-data-keys-onto/demo.ts clear
```

```output
what happened during the run : something erased the token of dseq 1001 behind the application's back

fingerprint before : digest 2164adfd2689689eaf1f10291e5b7b6aacd041dc5578e3ac443039c0e1567aa2
                     2 rows, 172 bytes of sealed secrets
fingerprint after  : digest 8d913963ea14d12a3c03558265d10f3fc00898da57780f3d78152c4f5f027ea9
                     1 rows, 86 bytes of sealed secrets

explained by a writer : 0 rewritten, 0 created, 0 removed
unexplained           : 11111111-1111-4111-8111-111111111111 (dseq 1001)

VERDICT: the rotation must fail loudly and name these deployments
```

Now a user simply deleting that deployment — an ordinary thing to do at any moment, including during a rotation.

```bash
cd apps/api && ../../node_modules/.bin/tsx ../../.work/con-876-re-wrap-data-keys-onto/demo.ts owner-deletes
```

```output
what happened during the run : the owner deleted deployment dseq 1001 outright

fingerprint before : digest 2164adfd2689689eaf1f10291e5b7b6aacd041dc5578e3ac443039c0e1567aa2
                     2 rows, 172 bytes of sealed secrets
fingerprint after  : digest 8d913963ea14d12a3c03558265d10f3fc00898da57780f3d78152c4f5f027ea9
                     1 rows, 86 bytes of sealed secrets

explained by a writer : 0 rewritten, 0 created, 1 removed
unexplained           : none

VERDICT: clean run, the rotation may report success
```

Same digest `8d913963…`, same one row, same 86 bytes — and opposite verdicts. Nothing in the fingerprint itself separates these two; the instrument separates them by going back and asking the row that lost its token whether anything wrote it. The erased row was still sitting there with its write timestamp untouched. The deleted row was gone, and a deletion is the one case nothing left behind can explain, so it stays tolerated.

Last one, the mirror image: a token **planted** on a deployment that had none. Before the review round this looked like an ordinary new deployment and passed quietly.

```bash
cd apps/api && ../../node_modules/.bin/tsx ../../.work/con-876-re-wrap-data-keys-onto/demo.ts plant
```

```output
what happened during the run : something planted a token on dseq 1003, which had none

fingerprint before : digest 2164adfd2689689eaf1f10291e5b7b6aacd041dc5578e3ac443039c0e1567aa2
                     2 rows, 172 bytes of sealed secrets
fingerprint after  : digest c96f82ddfa26706b69b4a72b1b439ccd0b29988ca6de9393f456c623d8c00dda
                     3 rows, 261 bytes of sealed secrets

explained by a writer : 0 rewritten, 0 created, 0 removed
unexplained           : 33333333-3333-4333-8333-333333333333 (dseq 1003)

VERDICT: the rotation must fail loudly and name these deployments
```

Caught, because dseq 1003 carries a write timestamp from before the run opened — no writer of this run put that token there.

**What a person sees across the six scenarios:** every change a real user can make while the rotation runs is tolerated and counted, and every change nothing explains is named by deployment id. The next slice does the actual re-wrapping and turns a non-empty `unexplained` list into the run's failure.

To re-run this document: `uvx showboat verify .work/con-876-re-wrap-data-keys-onto/demo-1.md` from the repository root, with the project's test PostgreSQL running.

## Summary by CodeRabbit

* **New Features**
  * Added stored-secret fingerprinting to track encrypted secret records, including counts, sizes, update markers, and per-record tokens.
  * Added reconciliation to identify newly created, removed, rewritten, or otherwise changed stored secrets.
  * Added efficient paginated processing for large collections of sealed secrets.
* **Bug Fixes**
  * Improved deployment test data generation by using sequential deployment identifiers for more consistent results.

### DIFF
--- a/apps/api/src/deployment/repositories/deployment-setting/deployment-setting.repository.ts
+++ b/apps/api/src/deployment/repositories/deployment-setting/deployment-setting.repository.ts
@@ -48,6 +48,13 @@ export type OpenDeployment = {
   createdAt: Date;
 };
 
+export type SealedSecret = {
+  id: string;
+  sealedSecrets: string;
+  /** The write timestamp as stored, in text, so a write landing in the same millisecond as the previous one still reads as a change a `Date` would flatten. */
+  updatedAtMarker: string | null;
+};
+
 export type LiveTrialDeployment = {
   userId: string;
   dseq: string;
@@ -168,6 +175,32 @@ export class DeploymentSettingRepository extends BaseRepository<Table, Deploymen
       }
 
       yield batch as OpenDeployment[];
+
+      if (batch.length < batchSize) {
+        return;
+      }
+
+      cursor = batch[batch.length - 1].id;
+    }
+  }
+
+  /** Keyset-paged on `id` rather than offset-paged, because a sweep whose offsets shift under live traffic skips rows and its digest then proves nothing. */
+  async *findSealedSecretsIteratively({ batchSize }: { batchSize: number }): AsyncGenerator<SealedSecret[]> {
+    let cursor: string | undefined;
+
+    while (true) {
+      const batch = await this.pg
+        .select({ id: this.table.id, sealedSecrets: this.table.sealedSecrets, updatedAtMarker: sql<string | null>`${this.table.updatedAt}::text` })
+        .from(this.table)
+        .where(and(isNotNull(this.table.sealedSecrets), ...(cursor ? [gt(this.table.id, cursor)] : [])))
+        .orderBy(asc(this.table.id))
+        .limit(batchSize);
+
+      if (!batch.length) {
+        return;
+      }
+
+      yield batch as SealedSecret[];
 
       if (batch.length < batchSize) {
         return;
--- a/apps/api/src/deployment/repositories/deployment-setting/deployment-setting.repository.ts
+++ b/apps/api/src/deployment/repositories/deployment-setting/deployment-setting.repository.ts
@@ -51,7 +51,7 @@ export type OpenDeployment = {
 export type SealedSecret = {
   id: string;
   sealedSecrets: string;
-  /** The write timestamp as stored, in text, so a write landing in the same millisecond as the previous one still reads as a change a `Date` would flatten. */
+  /** The write timestamp as stored, fixed-width and microsecond-exact, so two readings compare and order as text without a `Date` flattening them to milliseconds. */
   updatedAtMarker: string | null;
 };
 
@@ -83,6 +83,8 @@ export type AutoTopUpDeployment = {
  * initialised managed wallet, so there is no owner for whom auto top-up cannot work.
  */
 const AUTO_TOP_UP_ENABLED_BY_DEFAULT = true;
+
+const UPDATED_AT_MARKER_FORMAT = 'YYYY-MM-DD"T"HH24:MI:SS.US';
 
 @singleton()
 export class DeploymentSettingRepository extends BaseRepository<Table, DeploymentSettingsInput, DeploymentSettingsOutput> {
@@ -190,7 +192,7 @@ export class DeploymentSettingRepository extends BaseRepository<Table, Deploymen
 
     while (true) {
       const batch = await this.pg
-        .select({ id: this.table.id, sealedSecrets: this.table.sealedSecrets, updatedAtMarker: sql<string | null>`${this.table.updatedAt}::text` })
+        .select({ id: this.table.id, sealedSecrets: this.table.sealedSecrets, updatedAtMarker: this.#updatedAtMarker() })
         .from(this.table)
         .where(and(isNotNull(this.table.sealedSecrets), ...(cursor ? [gt(this.table.id, cursor)] : [])))
         .orderBy(asc(this.table.id))
@@ -208,6 +210,28 @@ export class DeploymentSettingRepository extends BaseRepository<Table, Deploymen
 
       cursor = batch[batch.length - 1].id;
     }
+  }
+
+  /** The reading a write happening now would leave, so a row carrying a stamp below it carries one no writer of this run produced. */
+  async readCurrentUpdatedAtMarker(): Promise<string> {
+    const [row] = await this.pg.execute<{ marker: string }>(sql`select to_char(now()::timestamp, ${UPDATED_AT_MARKER_FORMAT}) as marker`);
+
+    return row.marker;
+  }
+
+  /** Answers for rows the sealed-secrets sweep cannot reach, so a row that lost its token can still be asked whether anything wrote it; an id absent from the result no longer exists. */
+  async findUpdatedAtMarkers(ids: string[]): Promise<Map<string, string | null>> {
+    if (ids.length === 0) {
+      return new Map();
+    }
+
+    const rows = await this.pg.select({ id: this.table.id, updatedAtMarker: this.#updatedAtMarker() }).from(this.table).where(inArray(this.table.id, ids));
+
+    return new Map(rows.map(row => [row.id, row.updatedAtMarker]));
+  }
+
+  #updatedAtMarker() {
+    return sql<string | null>`to_char(${this.table.updatedAt}, ${UPDATED_AT_MARKER_FORMAT})`;
   }
 
   async #findAutoTopUpDeployments(address?: string): Promise<AutoTopUpDeployment[]> {
--- a/apps/api/src/deployment/services/stored-secret-fingerprint/stored-secret-fingerprint.service.integration.ts
+++ b/apps/api/src/deployment/services/stored-secret-fingerprint/stored-secret-fingerprint.service.integration.ts
@@ -39,7 +39,7 @@ describe(StoredSecretFingerprintService.name, () => {
     const reconciliation = await service.reconcile(await service.take({ batchSize: 1 }), { batchSize: 1 });
 
     expect(reconciliation.after).toEqual(reconciliation.before);
-    expect(reconciliation).toMatchObject({ ownerRewritten: 0, created: 0, removed: 0, reSealed: [] });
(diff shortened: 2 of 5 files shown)


baktun14:
Closing this as superseded. The `rewrap-data-keys` command that landed on `main` the same day under CON-875 (#3911, #3912, #3913, #3919) already covers what this stack set out to do: the before/after fingerprint with per-row drift classification (an owner's own write tolerated, a token changed without a write named as corrupted), the enabled-version preflight, per-batch transactions, resumability by selection alone, a dry run, and stepping over rows nothing can open. Rebasing this branch would mean carrying two implementations of the same command.

Two things this stack had that `main` does not are being ported in a small follow-up under CON-876: guarding the re-wrap write on the `wrapped_by_kid` it opened, and refusing a `--target-version` that differs from the one the console is configured to wrap under, along with the concurrent-read integration test.

Serhii has left the company, so I am closing this on his behalf.

PR #3956, opened 2026-09-15
feat(deployment): list closed deployments newest first, with their groups

baktun14:
## Why

Part of CON-962.

`GET /v1/deployments` could only answer with the active deployments a wallet holds, in the chain's oldest-first order. A client wanting the archive, or the newest page first, had to query the chain itself — which is what deploy-web's deployments list does today, and why it cannot show anything the console knows.

Two things the chain returned were also being dropped on the way out. The service rebuilt each item without the `groups` the chain sent alongside it, so a caller could not tell how much cpu, memory or gpu a listed deployment asks for. And `DeploymentLeaseSchema` has no `reclamation`, so zod strips the block a provider sets when it flags a lease for reclamation — from this endpoint and from every other response built on that schema.

Stacked on #3955; merge that one first.

## What

`state` (`active` by default, or `closed`) and `reverse` reach the chain query. A listed deployment carries its `groups` as the chain describes them, reusing the http-sdk's own shape so the generated type stays in step. A lease carries its optional `reclamation` block.

All four are additive. A caller sending neither parameter gets the active deployments oldest first, exactly as before.

Tests: unit cases for the forwarded state and order, including down the database fallback, and for the returned groups; functional cases for the closed-and-reversed query reaching the chain, for the groups on the way back, and for a state the chain does not hold. Docs snapshot and generated SDK regenerated.

## Summary by CodeRabbit

- **New Features**
  - Deployment listings now support active/closed filtering and reverse ordering.
  - Pagination includes configurable page sizes, cursor-based continuation, and `hasMore` indicators.
  - Deployment results now include on-chain resource groups and deployment names.
  - Lease-related responses may include provider reclamation details, including deadlines and closure reasons.

- **Bug Fixes**
  - Invalid deployment listing filters and pagination values now return structured `400` errors.

- **Documentation**
  - Updated API documentation and client types to reflect the expanded deployment listing and lease response fields.


baktun14:
On the `size: L`: 492 of the 726 counted lines are the regenerated functional docs snapshot, which the labeler does not yet recognise as a generated artifact the way it does the OpenAPI spec and the SDK types. Hand-written, this is 234 lines. #3962 adds the snapshot to that exclusion list.

PR #4023, opened 2026-09-23
fix(deployment): look up a few deployments directly instead of paging the owner's history

baktun14:
## Why

Fixes CON-995

To decide whether a managed deployment needs funding, we asked the chain for every lease and deployment the owner ever had, then kept the few being checked. For an owner with about 50,000 past deployments that came to roughly 87 MB of JSON and 40+ seconds per check, and initial funding runs the check for every deployment created. When two of these overlapped on the background jobs pod, garbage collection stalled the event loop long enough for pg-boss housekeeping to miss its 3 second database connect timeout, which paged the "Background job failure could not be recorded" alert.

## What

- `DrainingDeploymentRpcService` now reads up to 10 deployments one at a time: `/deployments/info` plus a lease list filtered by dseq, both already in the http-sdk. Larger sets keep paging the owner's history, because two requests per deployment would outnumber the pages for most owners.
- A deployment the chain reports as not found reads as absent, as it would when missing from a page. Any other error still throws, so `DrainingDeploymentService` keeps falling back to the database.
- Initial funding always checks a single deployment, so it takes the direct path. So do the auto top-up sweep and the weekly coverage check for owners with 10 or fewer auto top-up deployments.
- The top-up integration test's chain stubs now answer both the paged and the per-deployment queries. Its scenarios pass either way, which I checked by forcing the limit to 0.

Against mainnet, the two direct reads for one of the affected deployments returned about 2 KB in about 0.6 s each.


baktun14:
Neither point needs a change.

Failure handling hasn't changed: the old paging loop had no per-page catch either, so one failed page already rejected the whole call. Both callers still fall back to the database on any RPC error (`draining-deployment.service.ts:598-606` and `613-621`).

On fallback frequency, the funding path reads a single dseq, so a small owner still costs one lease request and one deployment request, the same as one page of each before. Only a sweep over 2 to 10 dseqs sends more requests (two per dseq), and a failed one takes the same database fallback a failed page did.

PR #4024, opened 2026-09-23
fix(managed-wallet): resend a broadcast the node never answered

baktun14:
## Why

Fixes CON-996

When the chain connection dropped in the middle of a broadcast, tx-signer couldn't tell whether a node had the transaction. It sent the signed bytes once, polled until the transaction's deadline passed, and reported it as not included (502). A deployment create failed exactly this way on 2026-09-23: `rpc-proxy-private` got a connection reset from a Cloudflare-fronted RPC node during `broadcast_tx_sync` and answered 500 `could not proxy request`, and the transaction never reached the chain. The proxy logs about a dozen of these upstream resets a day, so any managed transaction can hit this now and then.

## What

- While no node has answered the broadcast, each poll now sends the same signed bytes again. Resending is safe because the chain executes a signed transaction at most once: a node that already holds it answers "tx already exists in cache", and a copy that already landed is refused.
- Resends stop at the first answer, whether the node accepts or refuses. A refusal doesn't prove the first copy was lost, so polling still decides the outcome, and the not-included and unknown outcomes are reported exactly as before.
- The transport still never replays `broadcast_tx_*` on its own. The resend lives in `SigningClientService`, which knows the bytes are the same signed transaction.
- Separate commit: the mutation gate's observability ignorer didn't recognize `this.#logger`, which Babel parses as a `PrivateName`, so arguments to log calls on private-field loggers counted as surviving mutants. It now reads the private name, which moves this change from 72.7% to 94.1% on the gate.


baktun14:
Neither point needs a change.

The out-of-gas retry still sees both signals it acts on: the first broadcast's `BroadcastTxError`, rethrown by `#broadcast` (`signing-client.service.ts:211`), and a committed out-of-gas tx, which polling returns (line 247). A refusal on a resend only shows that one node refused, not that the first broadcast missed, so it's logged as `SIGN_AND_BROADCAST_RESEND_REFUSED` and polling still decides.

On poll cadence, a failed `getTx` already ends the poll (line 261), so a full outage plays out the same way. When queries work but broadcasts hang, a resend can stretch a cycle by up to `RPC_REQUEST_TIMEOUT_MS`, but the exit check runs after `getTx` (line 251), so the last lookup still comes after the tx expires.

PR #4030, opened 2026-09-23
feat(deployment): list unavailable gpu models below the ones providers can fill

baktun14:
## Why

Closes CON-1002

Since #4007 the GPU model list offers only the models an online provider has free capacity for. Users can't see which catalog models nobody can fill right now, and nothing says how many providers could take the models they can pick.

## What

- The model list shows the available models first, each labelled with the number of online providers that have free capacity for it, then the rest of the vendor's catalog grayed out under "Unavailable". Search finds those too.
- A model locked for a trial keeps its lock next to its count. The unlock button only appears for models a provider offers, since unlocking a model nobody offers would not produce a bid.
- A pinned model that nobody offers now sits under "Unavailable". Its memory and interface still show, and the trigger now shows its catalog display name (for example "A100") instead of the raw SDL value.
- Memory, interface and the vendor step are unchanged, and the vendor step still hides while NVIDIA is the only available vendor.
- When availability is absent or empty, the whole catalog stays selectable with no counts, as before.

### Verification

- deploy-web: full unit suite green (396 files, 4671 tests), lint clean, no new tsc errors against the main baseline
- Checked in the browser in light and dark mode, against a local API with counts added to its placement options answer


baktun14:
The dropped provider count the Claude review flagged in `withPinnedModel` is fixed in 74ac5e360: the pinned entry now spreads the offered model before merging in the pinned values, so a picked model keeps its "N providers" hint. GpuCard.spec ("keeps the provider count on a model once it is picked") and a `withPinnedGpu` case in gpu.spec cover it.

